PR #6, opened 2017-07-21
Add reference OSL implementations

jstone-lucasfilm:


lgritz:
Jonathan, how should we handle future changes on the OSL side? Not that we're expecting things to change much, but it really has just barely settled down.

One strategy would put the initiative on us (that is, people modifying OpenShadingLanguage project), whenever we think we've changed something on our side that would affect the .osl files generated, to submit a MaterialX project PR to propagate the changes.

Another strategy is just to let the OSL accumulate changes and have your side initiate a single big copy once per "release" (once you start really marking releases).

Other strategies?

dbsmythe:
I've already posted my thoughts on this to the Google Group, but your “Another strategy” outline is what I was thinking: it allows work on the two projects to proceed independently, and when future revisions to the MaterialX spec or codebase necessitate updates to the reference OSL shaders now included in the MaterialX GitHub, we can bake out a new set of .osl shaders implementing the intentions of the spec and any new or enhanced standard nodes pulling in updates from the OpenShadingLanguage project.

jstone-lucasfilm:
For updates on the OSL side that are triggered by new versions of the MaterialX specification, I think Doug's suggestion is right on, and this will allow each release of MaterialX to contain a matching set of reference shaders.

For updates to OSL shaders *within* a version of the MaterialX specification, which might be required as fixes and clarity improvements are made after the initial launch of a version, I can see an advantage in Larry's first suggestion, where the OSL team creates pull requests as they recognize that useful updates should be propagated to the MaterialX repository as well.

Does that seem like a reasonable distinction to make?

lgritz:
Yes. You guys do a pull when you are ready to branch/release MaterialX.

We'll do a "push" (that is, submit a PR to the MaterialX project, initiated from the OSL developer side) if we think something has changed on our side that is important for you to pick up between MX versions (for example, if we fix a bug).

In either case, I suggest that the README file on the MaterialX side document the SHA-1 hash that designates exactly which commit of OSL was the last place we synced the projects. Otherwise, we're always going to be wondering if anything has diverged and unsure exactly when we synced, only to be resolved by checking out both projects and diffing all the files.

jstone-lucasfilm:
Good idea, and we'll add that data to the README.

dbsmythe:
Yes, good ideas all around, agreed with all of them.

-Doug

PR #10, opened 2017-08-05
Improvements to MaterialX names

jstone-lucasfilm:
See CHANGELOG.md for the full list of API changes.

Also:

- Move Element-related unit tests into their own module.
- Minor fixes to Python code examples.


source/MaterialXCore/Element.h
```diff
@@ -94,9 +94,13 @@ class Element : public enable_shared_from_this<Element>
...
     /// @{
 
-    /// Return the element's name string.  The name of a MaterialX element
-    /// must be unique among all elements at the same scope, and cannot be
-    /// modified after the element is created.
+    /// Set the element's name string.  The name of a MaterialX element must be
+    /// unique among all elements at the same scope.
+    /// @throws Exception if an element at the same scope already possesses the
+    ///    given name.
+    void setName(const string& name);
```
niklasharrysson: You could add to the doc string that references within the document will not be updated automatically by this method, so if an element is renamed its references might need to be updated as well.

niklasharrysson:
LGTM

source/MaterialXCore/Element.h
jstone-lucasfilm: That's a good suggestion, though I'd recommend waiting until we have a corresponding helper method (e.g. Document::updateNameReferences) that we can point the developer to.

PR #11, opened 2017-08-08
Improved topological sort by reversing the traversal order.

niklasharrysson:
Improved topological sort by reversing the traversal order:
  - Using a more intuitive top-down traversal
  - Resulting node list is now in correct task order and does not need to be revered by shader generators
  - Updated unit tests to follow the new order


source/MaterialXCore/Node.cpp
```diff
@@ -252,14 +252,28 @@ vector<ElementPtr> NodeGraph::topologicalSort() const
...
+        }
+        else
+        {
+            for (InputPtr input : child->getChildrenOfType<Input>())
+            {
+                connectionCount += int(!input->getNodeName().empty());
+            }
+        }
+
+        inDegree[child] = connectionCount;
```
niklasharrysson: That's a good suggestion. We must only count true connections in the graph, so checking for NULL_EDGE is required in this case.

jstone-lucasfilm: Since an Edge can be implicitly cast to a boolean, here's one way this logic could be written:

```
int connectionCount = 0;
for (int i = 0; i < child->getUpstreamEdgeCount(); i++)
{
    if (child->getUpstreamEdge(i))
    {
        connectionCount++;
    }
}
inDegree[child] = connectionCount;
```

source/MaterialXCore/Node.cpp
```diff
@@ -252,14 +252,21 @@ vector<ElementPtr> NodeGraph::topologicalSort() const
 
     const vector<ElementPtr>& children = getChildren();
 
-    // Calculate in-degrees for all nodes, and enqueue those with degree 0.
-    std::unordered_map<ElementPtr, size_t> inDegree(children.size());
+    // Calculate in-degrees for all children.
+    std::unordered_map<ElementPtr, int> inDegree(children.size());
```
jstone-lucasfilm: To my mind, the original `size_t` type in this map seemed more clear, since an in-degree can only be zero or positive.  Would it be reasonable to keep this as size_t, and then to adjust the decrement logic below to prevent negative values from being generated?

source/MaterialXCore/Node.cpp
```diff
@@ -252,14 +252,21 @@ vector<ElementPtr> NodeGraph::topologicalSort() const
...
     std::deque<ElementPtr> childQueue;
     for (ElementPtr child : children)
     {
-        NodePtr node = child->asA<Node>();
-        inDegree[child] = node ? node->getDownstreamPorts().size() : 0;
-        if (inDegree[child] == 0)
+        int connectionCount = 0;
+        for (size_t i = 0; i < child->getUpstreamEdgeCount(); ++i)
+        {
+            connectionCount += int(child->getUpstreamEdge(MaterialPtr(), i) != NULL_EDGE);
```
jstone-lucasfilm: Perhaps we can take advantage of the implicit cast to boolean here?

```
if (child->getUpstreamEdge(MaterialPtr(), i))
{
    connectionCount++;
}
```

source/MaterialXCore/Node.cpp
niklasharrysson: Sure I can change that. The fix to handle negative values was just an extra precaution I added.. In theory it should never happen and would actually be an error in the graph (a node reached by downstream traversal should not have a zero in-degree). Would it be better to throw an exception if that happens?

jstone-lucasfilm: Since this situation could only be caused by a logic error in the library (rather than a data error in a MaterialX document), my sense is that it should be handled by our unit tests, rather than being an Exception that NodeGraph::topologicalSort might throw.

What would you think about the following, which prevents the logic error case (however unlikely) from generating an underflow?

```
if (inDegree[downstreamElem] > 1)
{
    inDegree[downstreamElem]--;
}
else
{
    inDegree[downStreamElem] = 0;
    childQueue.push_back(downstreamElem);
}
```

niklasharrysson: OK, let's go with that :)

PR #21, opened 2017-09-16
Updates for Travis and Clang 3.9

jstone-lucasfilm:


.travis.yml
```diff
@@ -17,6 +18,6 @@ script:
   - mkdir build
   - cd build
   - cmake -DMATERIALX_BUILD_PYTHON=ON ..
-  - cmake --build . --target install -- -j4
+  - sudo cmake --build . --target install -- -j4
```
lgritz: Unless you truly require install in the write-protected /usr area, you ought to be able to use `-DCMAKE_INSTALL_PREFIX=$HOME/someplace/you/own`, avoid the sudo, and thereby use the containerized Travis instances that are more plentiful (less likely to have a backlog) and have quicker startup time.

jstone-lucasfilm:
I suspect you're right, and there's likely to be a more elegant solution here.

In this situation, the process performing the installation is Python's "distutils.setup" module (which we use to install MaterialX Python in preparation for unit testing), and its default target is the "site-packages" or "dist-packages" folder of the active Python installation.

In the previous Trusty environment, this target folder was considered writable (as it is on Appveyor, for example), but the newly-updated Trusty environment considers it to be write-protected.

One potential solution would be to point "distutils.setup" to a specific local folder, and to add this same folder to the PYTHONPATH in preparation for Python unit tests, though I'd hoped to avoid making any assumptions about the underlying folder structure in Trusty environments.

If you have any additional suggestions, they would definitely be appreciated!

jstone-lucasfilm:
The solution I'm currently leaning towards is to add a MATERIALX_INSTALL_PYTHON option to the root CMake file, which would control whether the "distutils.setup" step is executed.

In Travis builds, this option would be set to OFF, and PYTHONPATH would be manually set to "${CMAKE_BINARY_DIR}/installed/python" in order to make MaterialX Python accessible to unit tests.

PR #43, opened 2017-10-30
Document management | Add in duplicate Element skip on XML reads

bernardkwok:
- Replaces the old boolean argument of reading XIncludes with a new structure.
- The options includes the Xincludes option + a new option to skip duplicates on read.
- The option structure is now available on all "read" methods consistently now.

### DIFF
--- a/source/MaterialXFormat/XmlIo.h
+++ b/source/MaterialXFormat/XmlIo.h
@@ -29,6 +29,32 @@ struct XmlReadOptions
         skipDuplicates(false) {}
 
     virtual ~XmlReadOptions() {}
+    
+    /// Set option for whether XInclude references will be read from disk and included in the document.
+    /// @param value Option value
+    void setReadXincludes(bool value)
+    {
+        readXincludes = value;
+    }
+
+    /// Get option for whether XInclude references will be read from disk and included in the document.
+    bool getReadXincludes() const
+    {
+        return readXincludes;
+    }
+
+    /// Set option for whether to skip reading in Elements with duplicate names.
+    /// @param value Option value
+    void setSkipDuplicates(bool value)
+    {
+        skipDuplicates = value;
+    }
+
+    /// Get option for whether to skip reading in Elements with duplicate names.
+    bool getSkipDuplicates() const
+    {
+        return skipDuplicates;
+    }
 
     /// If true, XInclude references will be read from disk and included in the document. Defaults to true.
     bool readXincludes;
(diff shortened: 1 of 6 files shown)


source/PyMaterialX/PyXmlIo.cpp
```diff
@@ -16,7 +16,7 @@ namespace mx = MaterialX;
 void bindPyXmlIo(py::module& mod)
 {
     mod.def("readFromXmlFileBase", &mx::readFromXmlFile,
-        py::arg("doc"), py::arg("filename"), py::arg("searchPath") = mx::EMPTY_STRING, py::arg("readXIncludes") = true);
+        py::arg("doc"), py::arg("filename"), py::arg("searchPath") = mx::EMPTY_STRING, py::arg("readingOptions") = static_cast<mx::XmlReadOptions*>(nullptr));
```
jstone-lucasfilm: Maybe this needs to be a cast to a (const XmlReadOptions*)?  PyBind11 has very strict requirements about how default arguments are declared, and the missing "const" might cause this to break.

http://pybind11.readthedocs.io/en/stable/advanced/functions.html#default-arguments-revisited

source/PyMaterialX/PyDefinition.cpp
```diff
@@ -49,4 +51,7 @@ void bindPyDefinition(py::module& mod)
         .def("hasLanguage", &mx::Implementation::hasLanguage)
         .def("getLanguage", &mx::Implementation::getLanguage)
         .def_readonly_static("CATEGORY", &mx::Implementation::CATEGORY);
+
+    py::class_<mx::XmlReadOptions>(mod, "XmlReadOptions")
```
bernardkwok: Silly mistake. Was missing the definition of this.

jstone-lucasfilm: How will a Python user access the individual members of this structure?  I believe we need to make a decision about the Python API, either by exposing the underlying member variables with "def_readwrite", or by adding and binding appropriate "set" methods:

http://pybind11.readthedocs.io/en/stable/classes.html

In this case, it seems reasonable to use the "def_readwrite" approach, which mirrors the way that these variables are accessed in C++.

Also, I would recommend moving this binding to PyXmlIo.cpp, where the read methods that use this class are bound.

source/MaterialXFormat/XmlIo.h
```diff
@@ -19,38 +19,61 @@ namespace MaterialX
...
+    bool readXincludes;
+    /// If true, Will skip reading in Elements with duplicate names. Defaults to false.
+    bool skipDuplicates;
+};
+
 /// Read a document as XML from the given character buffer.
 /// @param doc The document into which data is read.
 /// @param buffer The character buffer from which data is read.
+/// @param readOptions Options to use during reading. Default is null which indicates to use the 
+/// default values as specified in the XmlReadOptions structure.
```
jstone-lucasfilm: Formatting suggestion:  In the codebase so far, we've been starting the second and following lines of a "param" description with four spaces, to make the text more readable.

PR #65, opened 2018-01-12
Autodesk update

bernardkwok:
Element: 
- New copy options for Element copy: CopyOptions 
Document:
- usage of CopyOptions for library import .
Observer: 
- Add interface to be able to disable / enable notifications per observed document. 
- Add in scoped disabling class.
Element: 
- geometry identifier token substitution (equivalence to file identifier substitution).
Material: edge traversal utilities.


jstone-lucasfilm:
Looks good!

PR #85, opened 2018-04-04
Move XML write predicate check.

niklasharrysson:
Move XML write predicate check, before xinclude writes, to let the predicate control xincludes.


jstone-lucasfilm:
Looks good, thanks!

PR #91, opened 2018-04-13
Template build fix

bernardkwok:


jstone-lucasfilm:
Looks good, and thanks for the fix!

PR #124, opened 2018-06-21
Value class float formatting

niklasharrysson:
This change list adds methods to the Value class to control how float values are formatted when converted to strings. It enables support for selecting the format as "fixed", "scientific" or "default" as well as setting the precision. 
Our use-case for this is when doing code generation targeting a shading language compiler that doesn't support scientific notation.

### DIFF
--- a/source/MaterialXCore/Value.cpp
+++ b/source/MaterialXCore/Value.cpp
@@ -15,7 +15,7 @@ namespace MaterialX
 {
 
 Value::CreatorMap Value::_creatorMap;
-int Value::_floatFormat = 0;
+Value::FloatFormat Value::_floatFormat = Value::FloatFormatDefault;
 int Value::_floatPrecision = 6;
 
 namespace {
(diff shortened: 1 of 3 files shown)


source/MaterialXCore/Value.cpp
```diff
@@ -8,12 +8,15 @@
 #include <MaterialXCore/Util.h>
 
 #include <sstream>
+#include <iomanip>
```
jstone-lucasfilm: For consistency, I'd recommend alphabetizing this block of includes.

source/MaterialXCore/Value.cpp
```diff
@@ -232,6 +244,20 @@ template<class T> T Value::asA() const
     return typedVal->getData();
 }
 
+Value::ScopedFloatFormatting::ScopedFloatFormatting(FloatFormat format, int precision)
+    : _format(Value::getFloatFormat())
```
jstone-lucasfilm: For consistency, I'd place the colon and comma at line endings rather than line beginnings.

source/MaterialXCore/Value.h
```diff
@@ -64,6 +73,44 @@ class Value
...
+    }
+
+    /// RAII class for scoped setting of float formatting.
+    /// Flags are reset when the object goes out of scope.
+    class ScopedFloatFormatting
+    {
+      public:
+        ScopedFloatFormatting(FloatFormat format, int precision = 6);
+        ~ScopedFloatFormatting();
+      private:
```
jstone-lucasfilm: For consistency, I'd put a newline between the public and private sections.

PR #127, opened 2018-07-09
Fix potential name conflict in type aliases

schmidka:
This PR consists of two commits. It is intended to address https://github.com/materialx/MaterialX/issues/126.

5f08ca1: Adds forward declarations for all classes that are used in type aliases. I did not add any comments explaining why the forward declarations are needed, since it seemed like the headers only comment for doxygen documentation.

~2a1ab03: Adds doxygen comments for a few public type aliases that did not previously have them, for consistency. The comments are the same as are used for those same type aliases in other public headers.~

There are no changes to how the code functions, the ABI or API (for anyone not running into the name conflict, anyhow).

### DIFF
--- a/source/MaterialXCore/Definition.h
+++ b/source/MaterialXCore/Definition.h
@@ -32,24 +32,24 @@ class Member;
 class ShaderRef;
 
 /// A shared pointer to a NodeDef
-using NodeDefPtr = shared_ptr<class NodeDef>;
+using NodeDefPtr = shared_ptr<NodeDef>;
 /// A shared pointer to a const NodeDef
-using ConstNodeDefPtr = shared_ptr<const class NodeDef>;
+using ConstNodeDefPtr = shared_ptr<const NodeDef>;
 
 /// A shared pointer to an Implementation
-using ImplementationPtr = shared_ptr<class Implementation>;
+using ImplementationPtr = shared_ptr<Implementation>;
 /// A shared pointer to a const Implementation
-using ConstImplementationPtr = shared_ptr<const class Implementation>;
+using ConstImplementationPtr = shared_ptr<const Implementation>;
 
 /// A shared pointer to a TypeDef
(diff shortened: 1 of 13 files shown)


source/MaterialXCore/Definition.h
```diff
@@ -25,6 +25,12 @@ extern const string GEOMETRIC_NODE_CATEGORY;
...
 extern const string CONDITIONAL_NODE_CATEGORY;
 
+class NodeDef;
+class Implementation;
+class TypeDef;
+class Member;
+class ShaderRef;
+
 /// A shared pointer to a NodeDef
 using NodeDefPtr = shared_ptr<class NodeDef>;
```
jstone-lucasfilm: Now that we're forward-declaring classes before defining type aliases, I would recommend that we remove the "class" keywords from the type aliases themselves, as they're no longer needed.

source/MaterialXCore/Traversal.h
```diff
@@ -14,8 +14,15 @@
 namespace MaterialX
 {
 
+class Element;
+class Material;
+
+/// A shared pointer to an Element
```
jstone-lucasfilm: Although I see the consistency argument in providing Doxygen comments for all type aliases, these are really forward declarations of the "official" type aliases for ElementPtr and MaterialPtr in Element.h and Material.h.  From that perspective, it may be better to omit these duplicate Doxygen comments, leaving the versions in Element.h and Material.h as authoritative.

source/MaterialXCore/Definition.h
schmidka: Updated.

source/MaterialXCore/Traversal.h
schmidka: I see. I didn't realize they were the only duplicates, excepting the class-specific aliases. I dropped the commit adding the duplicate doxygen from the PR.

schmidka:
PR updated following comments.

PR #184, opened 2019-03-01
Adding cross platform get/set env var functions

willmuto-lucasfilm:
This change adds cross platform functions for getting and setting environment variables. It also re-adds the unit test for reading a MaterialX document using the environment variable search path.

### DIFF
--- a/source/MaterialXFormat/Environ.cpp
+++ b/source/MaterialXFormat/Environ.cpp
@@ -1,0 +1,59 @@
+//
+// TM & (c) 2019 Lucasfilm Entertainment Company Ltd. and Lucasfilm Ltd.
+// All rights reserved.  See LICENSE.txt for license.
+//
+
+#include <MaterialXFormat/Environ.h>
+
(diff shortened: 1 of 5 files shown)


source/MaterialXFormat/Environ.cpp
```diff
@@ -0,0 +1,59 @@
...
+    DWORD size = GetEnvironmentVariable(name.c_str(), nullptr, 0);
+
+    if (size != 0 && size != ERROR_ENVVAR_NOT_FOUND) 
+    {
+        std::unique_ptr<char[]> buffer(new char[size]);
+        GetEnvironmentVariable(name.c_str(), buffer.get(), size);
+        return string(buffer.get());
+    }
+
+    return EMPTY_STRING;
```
jstone-lucasfilm: This return of EMPTY_STRING can now be moved outside of the two platform-specific paths, in order to make the code slightly more streamlined.

willmuto-lucasfilm: Will do

PR #187, opened 2019-03-08
ShaderX to MaterialX merge

bernardkwok:
- initial PR for ShaderX merge


source/MaterialXGenShader/Factory.h
```diff
@@ -0,0 +1,79 @@
...
+    {
+        CreatorMap& map = creatorMap();
+        auto it = map.find(typeName);
+        return (it != map.end() ? it->second() : nullptr);
+    }
+
+private:
+    static CreatorMap& creatorMap()
+    {
+        static CreatorMap s_creatorMap;
```
jstone-lucasfilm: I'm a bit concerned about defining a static variable within an inline method, which effectively hides the translation unit in which the variable will be instantiated.

It may be better to make this variable a static member of the class (i.e. following the pattern of Element::_creatorMap and Value::_creatorMap), and require a specific translation unit to instantiate this for each unique class leveraging Factory.

CMakeLists.txt
```diff
@@ -50,8 +60,25 @@ if(MSVC)
     if(MATERIALX_WARNINGS_AS_ERRORS)
         add_compile_options(/WX)
     endif()
+
+    # Use /Z7 to avoid having pdbs
```
jstone-lucasfilm: This modification may be better left as Autodesk-specific, since different users will have different needs for symbolic debugging, and I'd be interested in your team's thoughts on the advantages of this approach.

meshula: Outside perspective: we prefer /Zi on our builds so this would conflict. If it really needs to be here, it should probably be guarded with an option that defaults OFF.

source/MaterialXGenShader/Factory.h
meshula: Further to that point, if you intend to ultimately have MaterialX be buildable as a shared library, life will be a lot less confusing if the static is not in a header file.

meshula:
Overall it's very exciting to see ShaderX begin to land!

README.md
```diff
@@ -26,13 +26,14 @@ The Python bindings for MaterialX are based on [PyBind11](https://github.com/pyb
...
-                and graph traversal, while the MaterialXFormat module supports
-                XML serialization.
-    python    - Support modules for MaterialX Python.
+- [documents](documents) - The MaterialX specification, developer guide, example and test suite files.
+- [source](source) - Cross-platform C++ libraries for MaterialX with Python bindings.
+- [python](python) - Support modules for MaterialX Python.
+
+### Usage
+
+Use of this code is subject to the terms of the Autodesk license agreement provided at the time of installation or download, or which otherwise accompanies this software in either electronic or hard copy form.
```
jstone-lucasfilm: This language was probably not intended to be merged to MaterialX master, though it makes perfect sense within the Autodesk fork.

CMakeLists.txt
```diff
@@ -35,13 +36,22 @@ mark_as_advanced(MATERIALX_PYTHON_EXECUTABLE)
 mark_as_advanced(MATERIALX_PYTHON_OCIO_DIR)
 mark_as_advanced(MATERIALX_PYTHON_PYBIND11_DIR)
 
+# Shader generation options
+option(MATERIALX_BUILD_GEN_OSL "Build the OSL shader generator back-end." ON)
```
jstone-lucasfilm: For discussion:  Since the new MaterialXGenGlsl and MaterialXGenOsl modules are lightweight and introduce no external library dependencies, perhaps it would be better to remove these two options, and to build these new modules in all configurations.  Developers can still choose which modules they want to link into their applications, so they're not compelled to include any more code than they need.

CMakeLists.txt
bernardkwok: We needed this for creating proper pdbs for release-with-debug builds we can make this optional based on whether that configuration is set. Would this work?

README.md
bernardkwok: Yes. We'll remove this.

CMakeLists.txt
bernardkwok: We can always build these as long as it's agreeable that the Python package always includes these 2 sub-modules.

source/MaterialXGenShader/Factory.h
niklasharrysson: Good catch, I'll make the change.

CMakeLists.txt
bernardkwok: This has been removed.

README.md
bernardkwok: This has been removed.

CMakeLists.txt
bernardkwok: We'll change the build rules so both are built by default with no options.

CMakeLists.txt
jstone-lucasfilm: Thanks Bernard!

source/MaterialXGenShader/Factory.h
niklasharrysson: As it turned out the Factory class methods doesn't need to be static at all, was a left-over from an older use-case. So I've made them all non-static now.

documents/DeveloperGuide/ShaderGeneration.md
```diff
@@ -0,0 +1,427 @@
+# Shader Generation
```
jstone-lucasfilm: This seems excellent, I'm really glad to see a developer document for shader generation!

source/README.md
```diff
@@ -0,0 +1,59 @@
...
+-   Basic "default" non-LUT color management transforms for OSL and
+    GLSL:
+    -   lin_rec709, gamma18, gamma22, gamm24, acescg, srgb_texture
+-   Basic "light" definitions and implementations for GLSL:
+    -   point, directional, spotlight, IBL
+
+## Library Tree Structure
+- Refer to documentation in the [Libraries](../documents/Libraries) folder.
+
+## Unsupported definitions for Shader Generation
```
jstone-lucasfilm: Some of the details in this document seem a bit too low-level for a README at source root.  Perhaps we should restrict this document to a brief overview of the source modules (MaterialXCore, MaterialXFormat, MaterialXShaderGen, etc), and point the developer to other documents (e.g. ShaderGeneration.md, Libraries/README.md) for further details.

bernardkwok: I wonder if it's better to move the details of this into the DeveloperGuide/ShaderGeneratoin.md file.
How does that sound?

jstone-lucasfilm: That sounds wise to me, and some of this information may already be there.

PR #211, opened 2019-05-10
Cmake: use relative paths for source dir

Da-Krunch:
Using directly ${CMAKE_SOURCE_DIR} can cause problems when materialx is embedded as a submodule. In general that can result in fragile setups.
That can be easily fixed using relative paths, in this case: ${CMAKE_CURRENT_SOURCE_DIR}/../..


jstone-lucasfilm:
Looks good, and thanks for the fix!

PR #213, opened 2019-05-13
Add a few nonzero parameter default values.

Da-Krunch:
Adding default values to a few parameters that should have non-zero values if unset.


jstone-lucasfilm:
Looks good, and thanks for the fixes!

PR #215, opened 2019-05-14
Render and Viewer updates

bernardkwok:
- Add in C++ version of matrix / vector transform math.
- Add in new shader ball and default base and ball shaders for Viewer.
- Add in xinclude enhancement include search for local path of parent by default. Add in test "library" with on dependent library definition and a material "preset" which references the library definition by relative path.
- Minor fix to env placement to perform proper flip in V (1-V) and usage of 90 degree world matrix rotation vs a 2D uv rotation which causes wrapping within triangles.
- Minor fixes to test suite to properly catch exceptions due to read errors such as missing include files.
- Updates to new vector/matrix math restrictions in the spec.
- Restore approximate burley diffuse implementation for OSL with oren_nayer.
- Approximate sheen bsdf implementation for OSL.


source/MaterialXView/Viewer.cpp
```diff
@@ -208,9 +208,20 @@ Viewer::Viewer(const mx::StringVec& libraryFolders,
     const mx::MeshList& meshes = _envGeometryHandler->getMeshes();
     if (!meshes.empty())
     {
+        // Invert u
+        mx::MeshPtr mesh = meshes[0];
+        mx::MeshStreamPtr stream = mesh->getStream(mx::MeshStream::TEXCOORD_ATTRIBUTE, 0);
+        mx::MeshFloatBuffer &buffer = stream->getData();
+        size_t stride = stream->getStride();
+        for (size_t i = 0; i < buffer.size() / stride; i++)
```
bernardkwok: Instead of trying to perform 1-V in the shader it is done here. Load time is a bit slower but avoid re-computation at shader time. Also did not seem easy to set up an invert on V only for texcoord.

jstone-lucasfilm: If we wanted to keep this as a shader computation, one option would be to use a scalar "invert" node (with the channels attribute selecting the U-coordinate) followed by a vector2 "combine" node (with the channels attribute selecting the V-coordinate).  Let me know how this approach sounds to you.

source/MaterialXCore/Types.cpp
```diff
@@ -85,6 +85,41 @@ template <> Matrix33 MatrixN<Matrix33, float, 3>::getAdjugate() const
...
+    return Vector3(
+      _arr[0][0] * rhs[0] + _arr[0][1] * rhs[1] + _arr[0][2] * rhs[2],
+      _arr[1][0] * rhs[0] + _arr[1][1] * rhs[1] + _arr[1][2] * rhs[2],
+      _arr[2][0] * rhs[0] + _arr[2][1] * rhs[1] + _arr[2][2] * rhs[2]
+    );
+}
+
+Vector2 Matrix33::transformPoint(const Vector2& rhs) const
+{
+    Vector3 rhsVector3(rhs[0], rhs[1], 1.0f);
```
jstone-lucasfilm: As a naming convention for this variable (as well as similar variables below), what would you think of "rhs3"?  That seems less verbose without reducing clarity.

source/MaterialXCore/Types.cpp
```diff
@@ -185,10 +220,47 @@ template <> Matrix44 MatrixN<Matrix44, float, 4>::getAdjugate() const
...
+}
+
+Vector3 Matrix44::transformVector(const Vector3& rhs) const
+{
+    Vector4 rhsVector4(rhs[0], rhs[1], rhs[2], 0.0f);
+    rhsVector4 = transform(rhsVector4);
+    return Vector3(rhsVector4[0], rhsVector4[1], rhsVector4[2]);
+}
+
+Vector4 Matrix44::transformNormal(const Vector4& rhs) const
```
jstone-lucasfilm: What would be the proposed use case for this Vector4 overload of transformNormal?  Perhaps we can omit this and provide only the more common Vector3 version?

source/MaterialXCore/Types.cpp
```diff
@@ -185,10 +220,47 @@ template <> Matrix44 MatrixN<Matrix44, float, 4>::getAdjugate() const
         _arr[0][0]*_arr[2][1]*_arr[1][2] - _arr[1][0]*_arr[0][1]*_arr[2][2] - _arr[2][0]*_arr[1][1]*_arr[0][2]);
 }
 
+
+Vector4 Matrix44::transform(const Vector4& rhs) const
```
jstone-lucasfilm: For these pure matrix/vector multiplication methods, I would recommend the original name "multiply", with the name "transform" being reserved for the specific point/vector/normal cases.

source/MaterialXCore/Types.cpp
feldstj: It was convenient for something I did in Mesh.cpp with it. Plus, the MaterialX spec's transformnormal supports Vector4s so I was trying to be consistent.

jstone-lucasfilm: Got it, and thanks for those additional details.  This seems fine to me.

jstone-lucasfilm: Got it, and thanks for those additional details.  This seems fine to me.

jstone-lucasfilm: It looks like the vector4 version of transformnormal in the specification was an error, so I'm re-opening this thread.  Given that additional context, I would recommend that we remove this overload of transformNormal in the code, and replace it with the equivalent math in Mesh.cpp.

jstone-lucasfilm:
Looks great!

PR #218, opened 2019-05-18
Updated the amount parameter to an input in rotate nodes

laserallan:


jstone-lucasfilm:
Thanks for the fix, David!

PR #240, opened 2019-06-19
Updates from Lucasfilm development

jstone-lucasfilm:


libraries/pbrlib/pbrlib_defs.mtlx
```diff
@@ -301,7 +301,7 @@
 
   <!--
     Node: <complex_ior>
-    A node for converting scientific/complex IOR to the artistic IOR used by the conductorbsdf node.
+    A node for converting scientific/complex IOR to the artistic IOR used by the conductor_brdf node.
   -->
   <nodedef name="ND_complex_ior" node="complex_ior" type="multioutput" nodegroup="pbr"
            doc="A node for converting scientific/complex IOR to the artistic IOR used by the conductorbsdf node.">
```
niklasharrysson: Thanks for updating the node name above. The same change can be made in the doc string here.

jstone-lucasfilm: Ah, good point -- I'll make that update as well

PR #241, opened 2019-06-21
ADSK dev/ merge to master

bernardkwok:
- Fixes for implicit type conversions found in GLSL noise implementation which can cause shader compile failures.
- Fix to Viewer for UI updates for left,righ arrow keys.
- Fix order so that AO will appear after transparent objects in Viewer.


jstone-lucasfilm:
Looks good, thanks Bernard!

PR #245, opened 2019-06-26
Stdlib defs/NG updates for 1.36.3/1.37

dbsmythe:
Updated stdlib_defs: conforms to 1.37 spec other than having child output elems for nodedefs, and kept geomattrvalue as well as adding geompropvalue.
Updated stdlib_ng: new transform* nodegraph implementations, minor naming fixes.
Updated stdlib TestSuite files to conform to new stdlib_defs.


dbsmythe:
Closing so we can instead push to a shared stdlib_updates branch.

PR #251, opened 2019-06-28
ADSK dev to ILM master merge

bernardkwok:
- Movement <normalmap> from pbr library to stdlib as a new math node.
- Addition default bitangent typedef
- Addition of MaterialX module utility
- Update to web page generation utility to examine unit tests
- Fix to set transparency option before creating advanced options in UI for viewer
- Remove invalid implicit casts in GLSL noise implementation.
- Minor fix for error reporting for OSL validation driver

### DIFF
--- a/cmake/modules/FindMaterialX.cmake
+++ b/cmake/modules/FindMaterialX.cmake
@@ -1,0 +1,229 @@
+#
+# TM & (c) 2017 Lucasfilm Entertainment Company Ltd. and Lucasfilm Ltd.
+# All rights reserved.  See LICENSE.txt for license.
+#
+# Helper CMake module to Find MaterialX include dirs, libraries and document libraries
+#
+# Example Usage
+#  - Build and Install MaterialX 
+#  - register the path to this cmake module using CMAKE_MODULE_PATH
+#  - use find_package(MaterialX REQUIRED) to locate core components
+#  - additonal components can be found using shaderx, bxdf and pytton
+#     e.g:  find_package(MaterialX REQUIRED COMPONENTS shaderx)
+#           will find inc, lib, corelib and generators
+#
+# Variables defined:
+# MATERIALX_FOUND            True if headers and requested libraries were found
+# MATERIALX_BASE_DIR         MaterialX root installation directory
+# MATERIALX_INCLUDE_DIRS     MaterialX include directory
+# MATERIALX_LIB_DIRS         MaterialX lib directory
+# MATERIALX_CORE_LIBS        MaterialX Core libraries and ShaderX i.e. Core, Format & GenShader
+# MATERIALX_GENERATOR_LIBS   MaterialX Generator libraries i.e GenGlsl, GenOsl
+# MATERIALX_RENDER_LIBS      MaterialX Render libraries i.e RenderGlsl, RenderOsl
+# MATERIALX_STDLIB_DIR       Path to the MaterialX standard library directory
+# MATERIALX_PBRLIB_DIR       Path to the MaterialX pbr library directory
+# MATERIALX_BXDFLIB_DIR      Path to the Surface shaders library directory e.g. standard surface
+# MATERIALX_PYTHON_DIR       Path to MaterialX Python library
+# MATERIALX_RESOURCES_DIR    Path to MaterialX Resources (sample data, mtlx etc)
+#
+
+# make core variables required
+set (MATERIALX_REQUIRED_VARS
+        MATERIALX_INCLUDE_DIRS
+        MATERIALX_LIB_DIRS
+        MATERIALX_STDLIB_DIR
+        MATERIALX_CORE_LIBS
+    )
+
+# make shaderx a required if requested
+if ("shaderx" IN_LIST MaterialX_FIND_COMPONENTS)
+ list (APPEND MATERIALX_REQUIRED_VARS
+        MATERIALX_GENERATOR_LIBS
+        MATERIALX_RENDER_LIBS)
+endif()
+
+# make bxdf surfacce shaders required if requested
+if ("bxdf" IN_LIST MaterialX_FIND_COMPONENTS)
+ list (APPEND MATERIALX_REQUIRED_VARS 
+        MATERIALX_PBRLIB_DIR
+        MATERIALX_BXDFLIB_DIR)
+endif()
+
+
+if ("python" IN_LIST MaterialX_FIND_COMPONENTS)
+ list (APPEND MATERIALX_REQUIRED_VARS
+        MATERIALX_PYTHON_DIR
+        MATERIALX_RESOURCES_DIR)
+endif()
+
+
+
+# Locate MaterialX base directory based on ENV var
+find_path(MATERIALX_BASE_DIR
+    NAMES
+        include/MaterialXCore/Library.h
+    HINTS
+        "${MATERIALX_ROOT}"
+        "$ENV{MATERIALX_ROOT}"
+    )
+
+find_path(MATERIALX_INCLUDE_DIRS 
+    MaterialXCore/Library.h
+    HINTS
+        "${MATERIALX_ROOT}"
+        "$ENV{MATERIALX_ROOT}"
+        "${MATERIALX_BASE_DIR}"
+    PATH_SUFFIXES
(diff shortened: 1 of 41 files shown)


cmake/modules/FindMaterialX.cmake
```diff
@@ -0,0 +1,229 @@
...
+# make core variables required
+set (MATERIALX_REQUIRED_VARS
+        MATERIALX_INCLUDE_DIRS
+        MATERIALX_LIB_DIRS
+        MATERIALX_STDLIB_DIR
+        MATERIALX_CORE_LIBS
+    )
+
+# make shaderx a required if requested
+if ("shaderx" IN_LIST MaterialX_FIND_COMPONENTS)
```
jstone-lucasfilm: Now that the ShaderX functionality is merged into MaterialX, I would recommend that we refer to these features by their unique functionality (e.g. shader code generation, PBR nodes).  Would "shadergen" be a reasonable string to check for here?

libraries/stdlib/genglsl/mx_normal_map.glsl
```diff
@@ -1,11 +1,11 @@
-void mx_normalmap(vec3 value, int tangent_space, vec3 N, vec3 T, out vec3 result)
+void mx_normal_map(vec3 value, int map_space, float normal_scale, vec3 N, vec3 T,  out vec3 result)
```
jstone-lucasfilm: This is somewhat of an edge case, but historically we've omitted underscores for node names in the standard library (e.g. tiledimage), adding them only to nodes in new libraries such as pbrlib and bxdf.  Perhaps it would make sense to omit the underscore here, since this node now belongs to the standard library?

cmake/modules/FindMaterialX.cmake
bernardkwok: "shadergen" sounds good as that is the prefix used for all generation modules.

libraries/stdlib/genglsl/mx_normal_map.glsl
bernardkwok: This is the implementation, but I assume your referring to the to the nodedef name? We could keep the <normalmap>
name as it would also be backwards compatible. I think it was Doug who suggested that this be renamed. Whatever name you decide on we case use.

dbsmythe: I think I went along with the name "normal_map" as suggested by Niklas when it was added to the pbrlib.  If it's going to be moved to stdlib, I think it should be called "**normalmap**" and go into the "convolution2d" nodegroup next to "heighttonormal".  As an aside, the nodegroup name "convolution2d" is maybe not the best choice at this point, since neither heighttonormal or normalmap actually perform convolution, and the one node that does ("blur") is not implemented anywhere that I know of.

jstone-lucasfilm: I believe heighttonormal does perform a simple convolution, since it needs to detect gradients in the height field.  So it’s just normalmap that would be the odd man out.  Perhaps it’s well placed as a texture node?

dbsmythe: Except that as currently implemented, "normalmap" doesn't actually read from an image file- it instead has an "in" input that I would expect would be connected to an image node.  Here's what's currently in the pbrlib_defs.mtlx file (though I am noticing now that "tangent_space" should probably be a boolean not an integer):
     
  <nodedef name="ND_normalmap" node="normalmap" type="vector3" nodegroup="pbr">
    <input name="in" type="vector3" value="0.5, 0.5, 1.0"/>
    <parameter name="tangent_space" type="integer" value="1"/>
    <input name="normal" type="vector3" defaultgeomprop="Nworld"/>
    <input name="tangent" type="vector3" defaultgeomprop="Tworld"/>
  </nodedef>

dbsmythe: >      
>   <nodedef name="ND_normalmap" node="normalmap" type="vector3" nodegroup="pbr">
>     <input name="in" type="vector3" value="0.5, 0.5, 1.0"/>
>     <parameter name="tangent_space" type="integer" value="1"/>
>     <input name="normal" type="vector3" defaultgeomprop="Nworld"/>
>     <input name="tangent" type="vector3" defaultgeomprop="Tworld"/>
>   </nodedef>

jstone-lucasfilm:
That’s correct, and it’s just performing a transformation on the results of a texture sample, presumably using the geometric normal as another input.  I’m open to other thoughts on categories for this node, but I’d recommend steering clear of convolution, for the reasons you note above.

libraries/stdlib/genglsl/mx_normal_map.glsl
dbsmythe: Well for some reason I can't paste the text or a file in here, so please forgive the lack of proper formatting:
nodedef name="ND_normalmap" node="normalmap" type="vector3"
input name="in" type="vector3" value="0.5, 0.5, 1.0"
parameter name="tangent_space" type="integer" value="1"
input name="normal" type="vector3" defaultgeomprop="Nworld"
input name="tangent" type="vector3" defaulteomprop="Tworld"

dbsmythe: Maybe it should be a "math" node, same as transformpoint/normal/vector?

jstone-lucasfilm: That works for me, and I’d be interested in thoughts from Bernard.

bernardkwok: Yes. it is simply a vector transform so "math" sounds good. I'll also keep the name as "normalmap" and move it after the place2d "math" definition as that seems to be closest image related item.

dbsmythe: Bernard- would you mind sending me a quick description of "normalmap" that I can put into the actual Spec document?

dbsmythe: Did you also switch "tangentspace" to be a boolean?  Would be helpful to get a one-line description of each input and parameter for the spec as well.
There was also talk of adding a "heightmap" node- anything on that?

bernardkwok: Let me know if this is sufficient:
--------------------------------
The <normalmap> node is a math node that allows for a normal vector defined in one space to be transformed into another space. Inputs:
- "in" : input normal vector
- "space" : the input space for the normal. Possible options are "object" for object space and "tangent" for tangent space.
- "scale" : the normal can be scaled (offset) by a given multiplier.
- "normal" and "tangent". Two vectors from which a bitangent vector is computed to form the basis used for transformation. By default the vectors are in world space, meaning the output vector will be in world space.
```
  <nodedef name="ND_normalmap" node="normalmap" type="vector3" nodegroup="math">
    <input name="in" type="vector3" value="0.5, 0.5, 1.0"/> 
    <parameter name="space" type="string" value="tangent" enum="object,tangent"/>
    <input name="scale" type="float" value="1.0"/>
    <input name="normal" type="vector3" defaultgeomprop="Nworld"/>
    <input name="tangent" type="vector3" defaultgeomprop="Tworld"/>
  </nodedef>
```
Below is an example of this utility being used to read in a image containing tangent space normals and outputting world space normals scaled by a factor of 1.5.
```
  <image name="normal_image" type="vector3" colorspace="Raw">
    <parameter name="file" type="filename" value="normal_image.png" />
  </image>
  <normalmap name="surface_normals" type="vector3">
    <input name="in" type="vector3" nodename="normal_image" />
    <input name="scale" type="float" value="1.5" />
  </normalmap>
```
--------------------------------

We did have the discussion about add in a <heightmap> node which would take file image as input, but have decided that the existing <heighttonormal> is sufficient though we may want to add another kernel option.

cmake/modules/FindMaterialX.cmake
jstone-lucasfilm: I think this is the one remaining change we need before merging, unless we'd like to include additional improvements in this PR.

bernardkwok: Forgot to fix this. Update added now.

PR #259, opened 2019-07-08
Adding official hires MaterialX logo images

dbsmythe:


jstone-lucasfilm:
Fantastic, it will be great to have a higher-quality logo available!  Two thoughts:

- Is there any additional content being provided by the pre-composited versions with black and white backgrounds?  If they're only here for convenience, then I would recommend omitting these and providing only the alpha version (as we currently do with the logo in documents/Images).  These files will be downloaded with every clone of the MaterialX repository, so it's valuable to keep the size of the repository as small as possible.
- Perhaps the new alpha logo should simply replace the version already in documents/Images, rather than being placed in a new folder?  That will reduce confusion for developers looking for the "official" logo that they should use.

dbsmythe:
The logo image that's currently in the distribution's "Images" folder is very low-res: 256x256 I believe.
I would be fine with replacing it with something higher-res, though the current "with alpha" version may be too large to be the only version (it's 3500x2700, or more than 10x larger in pixels).
If there was only going to be one logo image file in the GitHub dist, I think it should be a 2048x1580 with alpha (which doesn't currently exist), and I could put the other versions on the website since I suspect that not all image readers can handle PNGs with alpha, and having already-over-white/black versions could be useful.

jstone-lucasfilm:
Sounds good!  An alpha version of the logo at 2048x1580 resolution seems like a good default for most needs.

dbsmythe:
Closing this PR, will send a new one with just the 2048x1580-with-alpha version.  This will *supplement* the current 256x256 version, just in case anything depends on it staying that size or aspect ratio.

PR #262, opened 2019-07-10
Feature for colorspace transformation based on colorspace specified in shaderref

mjyip-lucasfilm:


source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -478,6 +484,12 @@ ShaderGraphPtr ShaderGraph::create(const ShaderGraph* parent, const string& name
...
                     inputSocket->setValue(bindInputValue);
+
+                    input->setIsBindInput(true);
+                    ColorManagementSystemPtr colorManagementSystem = context.getShaderGenerator().getColorManagementSystem();
+                    const string& targetColorSpace = context.getOptions().targetColorSpaceOverride.empty() ?
+                        element->getDocument()->getActiveColorSpace() : context.getOptions().targetColorSpaceOverride;
+                    graph->populateInputColorTransformMap(colorManagementSystem, graph->_nodeMap[newNodeName], bindInput, targetColorSpace);
                 }
                 inputSocket->setPath(bindInput->getNamePath());
                 input->setPath(inputSocket->getPath());
```
bernardkwok: At line 486, you'll note that a check for a connection is done. If there is no connection then you can add in a transform. At line 498 it's adding in the link your probably seeing. This is to connect this to to an "output socket" which is of type ShaderGraphOutputSocket.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -251,7 +251,7 @@ void ShaderGraph::addDefaultGeomNode(ShaderInput* input, const GeomPropDef& geom
 void ShaderGraph::addColorTransformNode(ShaderInput* input, const ColorSpaceTransform& transform, GenContext& context)
 {
     ColorManagementSystemPtr colorManagementSystem = context.getShaderGenerator().getColorManagementSystem();
-    if (!colorManagementSystem || input->getConnection())
+    if (!input->getIsBindInput() && (!colorManagementSystem || input->getConnection()))
```
bernardkwok: Looking later on the input is connected as an "output socket" ShaderGraphOutputSocket if unconnected. Believe you can just check to see if the connection is of this type. Then you would not require to tag the input as a "bind input".

source/MaterialXGenShader/ShaderGraph.cpp
bernardkwok: Above at line 448, is the check for BindParam. In this case it's connected to a ShaderGraphInputSocket which could be  as a test guard as well. Both basically mean your not connected to an internal node.

source/MaterialXGenShader/ShaderNode.h
```diff
@@ -144,9 +144,23 @@ class ShaderInput : public ShaderPort
...
+    /// Set isBindParam value
+    void setIsBindParam(bool isBindParam) { _isBindParam = isBindParam; }
+
+    /// Get isBindParam value
+    const bool& getIsBindParam() const { return _isBindParam; }
+
   protected:
     ShaderOutput* _connection;
     string _channels;
+    bool _isBindInput;
```
bernardkwok: There there is no difference in logic between these to, I think we can just keep isBindInput and remove isBindParam. Then update the comments to say that it's either a <bindinput> or <bindparam> on a <shaderref>..

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -251,13 +251,12 @@ void ShaderGraph::addDefaultGeomNode(ShaderInput* input, const GeomPropDef& geom
...
     {
-        // Ignore inputs with connections as they are not 
+        // Ignore inputs (that are not bind inputs or bind params with connections as they are not
```
jstone-lucasfilm: Just missing a closing parenthesis here!  I might also recommend phrasing this comment "Ignore unbound inputs with connections, as they are not allowed to have colorspaces specified."  That covers both the bindinput and bindparam cases without the need for parentheses.

source/MaterialXGenShader/ShaderNode.h
```diff
@@ -144,9 +144,16 @@ class ShaderInput : public ShaderPort
     /// Get optional channels value
     const string& getChannels() const { return _channels; }
 
+    /// Set isBindInput value. If true, either a BindInput or BindParam on a shaderref
+    void setIsBindInput(bool isBindInput) { _isBindInput = isBindInput; }
+
+    /// Get isBindInput value
+    const bool& getIsBindInput() const { return _isBindInput;  }
```
jstone-lucasfilm: I would recommend returning this boolean result by value, rather than by const reference.  Returning by const reference provides benefits for non-trivial types like std::string, but for a boolean it's clearer and equally efficient to return by value.

source/MaterialXGenShader/ShaderNode.h
```diff
@@ -144,9 +144,16 @@ class ShaderInput : public ShaderPort
...
+
   protected:
     ShaderOutput* _connection;
     string _channels;
+    bool _isBindInput;
```
bernardkwok: I found a more generic way to cache this. 

On the `ShaderPort` class there is an `_flags` data member. We can reuse this by adding in a new flag, say  BIND_INPUT. Note that there is already an EMITTED flag so that value cannot be reused.

The condition test would then be something like `input.getFlags() & ShaderInput:BIND_INPUT` and setting is `input.setFlags(input.getFlags() | ShaderInput::BIND_INPUT)`.

source/MaterialXGenShader/ShaderNode.h
```diff
@@ -144,6 +144,9 @@ class ShaderInput : public ShaderPort
     /// Get optional channels value
     const string& getChannels() const { return _channels; }
 
+    /// Flag set on input shader port.
+    static const unsigned int BIND_INPUT = 1 << 1;
```
jstone-lucasfilm: To make it easier for readers to track the list of supported flags, I would recommend moving this definition just below the definition of ShaderPort::EMITTED:

```
    /// Flags set on shader ports.
    static const unsigned int EMITTED = 1 << 0;
    static const unsigned int BIND_INPUT = 1 << 1;
```

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -450,6 +455,12 @@ ShaderGraphPtr ShaderGraph::create(const ShaderGraph* parent, const string& name
                 if (bindParamValue)
                 {
                     inputSocket->setValue(bindParamValue);
+
+                    input->setFlags(input->getFlags() | input->BIND_INPUT);
```
jstone-lucasfilm: Since there are already accessor methods for ShaderPort:setEmitted() and ShaderPort::isEmitted(), I would recommend adding corresponding methods for ShaderPort::setBindInput() and ShaderPort::isBindInput(), so that clients don't need to use bit-wise operators.

PR #265, opened 2019-07-14
Adding sample python code scripts

dbsmythe:
Adding mxvalidate.py (now with camelCase function names), writelooks.py, writenodegraphs.py and a README.md describing them to a new Scripts folder under "python".

### DIFF
--- a/python/Scripts/mxvalidate.py
+++ b/python/Scripts/mxvalidate.py
@@ -1,0 +1,394 @@
+#!/usr/bin/env python
+'''
+Verify that a specified file is a valid MaterialX document.
+
+Doug Smythe
+'''
+
+import sys, os, string
+import MaterialX as mx
+
+
+def Usage():
+    print "mxvalidate.py: verify that a specified file is a valid MaterialX document."
+    print "Usage:  mxvalidate.py [options] <file.mtlx>"
+    print "    -h[elp]         Print usage information"
+    print "    -v[erbose]      Print summary of elements found in the document"
+    print "    -vv[eryverbose] Also print material bindings, nodedef interfaces and nodegraph trees"
+    print "    -i[ncludedefs]  Automatically XInclude stdlib_defs"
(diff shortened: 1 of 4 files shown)


python/Scripts/mxvalidate.py
```diff
@@ -0,0 +1,394 @@
...
+    #   "MaterialX.ExceptionFileMissing" if the file is not found, or
+    #   "MaterialX.ExceptionParseError" if the document is found but not readable.
+    try:
+	mx.readFromXmlFile(doc, filename)
+    except mx.ExceptionFileMissing as err:
+	#TODO: library doesn't correctly set missing filename if it's an XIncluded file (it's blank)
+	print err
+	return
+    except mx.ExceptionParseError as err:
+	m = err.split('(')[1][:-1]
```
jstone-lucasfilm: There's some potential for fragility here, as the text format of the parse error message isn't guaranteed to remain fixed in future versions of MaterialX (e.g. it may later contain additional information including multiple string literals).  For future robustness, I would recommend just printing the error message, as is done in the ExceptionFileMissing case.

python/Scripts/mxvalidate.py
```diff
@@ -0,0 +1,394 @@
+#!/usr/bin/env python
+'''
+Verify that a specified file is a valid MaterialX document.
+
+Doug Smythe
```
jstone-lucasfilm: For the MaterialX repository, we've dropped attribution lines for each script and source file, since GitHub can track authorship at a finer granularity (e.g. multiple authors contributing to different sections of a single file).

python/Scripts/mxvalidate.py
dbsmythe: done

python/Scripts/mxvalidate.py
dbsmythe: done

PR #266, opened 2019-07-14
Update stdlib_defs.mtlx with vectorN math operator definitions

dbsmythe:
Add vectorN definitions for sqrt, ln, exp and trig functions.


dbsmythe:
This is failing the unit tests, looks like because there's a check for missing functions for any definitions.  Which is good, but which means we'll need Niklas or someone to manage this merge.

dbsmythe:
Closing because Bernard has submitted a more complete PR.

PR #270, opened 2019-07-22
Merges from dev to ILM master

bernardkwok:
General:
- Fixes for auto non-const usage in iterators

ShaderGen:
- Sanitize all generated identifiers during code generation. 
- UDIM transform support (for tiled atlases) + UDIM computation utilities
- Fix utility for transparency detection to test for bindinputs which set for nodedef inputs which are marked as interfaces (vs assuming any interface is tranparent).

File Management:
- Library path read / include cleanup.
- Add additional skips for read/import since read can have includes, and includes can still conflict.

Test:
- Add external library path and test root options to unit test framework (_options.mtlx)

UI:
- Add in `uiadvanced` option support for Property Editor in viewer.

### DIFF
--- a/source/MaterialXCore/Element.cpp
+++ b/source/MaterialXCore/Element.cpp
@@ -32,6 +32,7 @@ const string ValueElement::UI_NAME_ATTRIBUTE = "uiname";
 const string ValueElement::UI_FOLDER_ATTRIBUTE = "uifolder";
 const string ValueElement::UI_MIN_ATTRIBUTE = "uimin";
 const string ValueElement::UI_MAX_ATTRIBUTE = "uimax";
+const string ValueElement::UI_ADVANCED_ATTRIBUTE = "uiadvanced";
 
 Element::CreatorMap Element::_creatorMap;
 
--- a/source/MaterialXFormat/XmlIo.cpp
+++ b/source/MaterialXFormat/XmlIo.cpp
@@ -76,7 +76,6 @@ void elementFromXml(const xml_node& xmlNode, ElementPtr elem, const XmlReadOptio
 void elementToXml(ConstElementPtr elem, xml_node& xmlNode, const XmlWriteOptions* writeOptions)
 {
     bool writeXIncludeEnable = writeOptions ? writeOptions->writeXIncludeEnable : true;
-    const StringVec& ignoredXIncludes = writeOptions ? writeOptions->ignoredXIncludes : StringVec();
     ElementPredicate elementPredicate = writeOptions ? writeOptions->elementPredicate : nullptr;
 
     // Store attributes in XML.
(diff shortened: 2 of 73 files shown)


source/MaterialXFormat/XmlIo.cpp
```diff
@@ -106,6 +107,11 @@ void elementToXml(ConstElementPtr elem, xml_node& xmlNode, const XmlWriteOptions
             {
                 if (!writtenSourceFiles.count(sourceUri))
                 {
+                    // Check to see if we have flagged this XInclude to be ignored, if so skip it.
```
bernardkwok: @jstone-lucasfilm , currently we've added this in so that we can import libraries implicitly and allow for xincludes which are explicit library imports. When we write a file we want to be able to not write out implicit ones but still write out explicit ones. 

Let us know what you think of this. This could be an xinclude "write" predicate instead.

source/MaterialXGenShader/Nodes/HwImageNode.cpp
```diff
@@ -0,0 +1,72 @@
...
+void HwImageNode::addInputs(ShaderNode& node, GenContext&) const
+{
+    // Add additional scale and offset inputs to match implementation arguments
+    ShaderInput* input = node.addInput(UV_SCALE, Type::VECTOR2);
+    input->setValue(Value::createValue<Vector2>(Vector2(1.0f, 1.0f)));
+    input = node.addInput(UV_OFFSET, Type::VECTOR2);
+    input->setValue(Value::createValue<Vector2>(Vector2(0.0f, 0.0f)));
+}
+
+void HwImageNode::setValues(const Node& node, ShaderNode& shaderNode, GenContext& context) const
```
bernardkwok: We added custom source code handling to allow insertion of a transform before lookup. This is currentlu used for UDIM transforms from the original space to the normalized space if we decide to use a texture atlas.

source/MaterialXFormat/XmlIo.cpp
jstone-lucasfilm: Couldn't this functionality be handled by the existing write predicate for _elements_?  If you specified a write predicate that returned false for any element with a specific source URI, then the XInclude associated with that source URI should also be omitted from the written document.

source/MaterialXGenShader/Nodes/HwImageNode.cpp
jstone-lucasfilm: That sounds like a good addition!

source/MaterialXFormat/XmlIo.cpp
bernardkwok: This is true. We can add in the element predicate or in our case add to our existing element predicate. I'll revert this modification.

source/MaterialXCore/Element.cpp
```diff
@@ -32,6 +32,7 @@ const string ValueElement::UI_NAME_ATTRIBUTE = "uiname";
 const string ValueElement::UI_FOLDER_ATTRIBUTE = "uifolder";
 const string ValueElement::UI_MIN_ATTRIBUTE = "uimin";
 const string ValueElement::UI_MAX_ATTRIBUTE = "uimax";
+const string ValueElement::UI_ADVANCED_ATTRIBUTE = "uiadvanced";
```
bernardkwok: Exposed and added to viewer Property Editor.

source/MaterialXView/Viewer.cpp
```diff
@@ -78,13 +78,18 @@ mx::DocumentPtr loadLibraries(const mx::StringVec& libraryFolders, const mx::Fil
         mx::FilePath path = searchPath.find(libraryFolder);
         mx::FilePathVec filenames = path.getFilesInDirectory("mtlx");
 
+        mx::CopyOptions copyOptions;
```
bernardkwok: For the Substance files we still need this as reads can have xincludes which conflict, and imports can still also conflict.

PR #274, opened 2019-08-07
Fix for OSL implementation of mx_roughness_dual.

bernardkwok:
Fix for OSL implementation of mx_roughness_dual. Avoid error if the function attempts to overwrite the input parameter reference when that parameter is a shader input.


jstone-lucasfilm:
Thanks for the fix!

PR #276, opened 2019-08-09
Reload standard libraries

mjyip-lucasfilm:
Added hotkey to reload standard libraries. Also added a label that displays the current shading model (if it exists).

### DIFF
--- a/source/MaterialXView/Viewer.cpp
+++ b/source/MaterialXView/Viewer.cpp
@@ -1592,8 +1592,7 @@ void Viewer::updateDisplayedProperties()
 {
     _propertyEditor.updateContents(this);
     mx::TypedElementPtr elem = getSelectedMaterial() ? getSelectedMaterial()->getElement() : nullptr;
-    const std::string& node = (elem && !elem->getAttribute("node").empty()) ? elem->getAttribute("node"): "";
-    if (!node.empty())
+    if (elem && !elem->getAttribute("node").empty())
     {
         _shaderLabel->setVisible(true);
         _shaderTextBox->setVisible(true);
--- a/source/MaterialXView/Viewer.cpp
+++ b/source/MaterialXView/Viewer.cpp
@@ -1596,7 +1596,7 @@ void Viewer::updateDisplayedProperties()
     {
         _shaderLabel->setVisible(true);
         _shaderTextBox->setVisible(true);
-        _shaderTextBox->setValue(node);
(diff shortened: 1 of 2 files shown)


source/MaterialXView/Viewer.cpp
```diff
@@ -810,6 +806,17 @@ void Viewer::loadDocument(const mx::FilePath& filename, mx::DocumentPtr librarie
         for (const auto& renderablePath : renderablePaths)
         {
             mx::ElementPtr elem = doc->getDescendant(renderablePath);
+            if (!elem->getAttribute("node").empty())
```
jstone-lucasfilm: There are other situations in which we'll need to update the "shading model" display for the current material, e.g. when the user changes the selected material or selected geometry.  To keep everything in sync, I would recommend extending the existing Viewer::updatePropertyEditor method, making it general enough to include your new functionality as well.  One potential name for the new, more comprehensive method would be Viewer::updateDisplayedProperties.

mjyip-lucasfilm: Sounds good! At the moment, each time a new material is chosen, loadDocument is called and then loadDocument calls updateDisplayedProperties. Do you suggest I move the updateDisplayedProperties outside of loadDocument?

jstone-lucasfilm: My thinking is that we would replace updatePropertyEditor with the more general updateDisplayedProperties, and then call the new method from all of the old locations where we used to call updatePropertyEditor.  That would include loadDocument, but also other locations where the user is switching between materials or geometries.

source/MaterialXView/Viewer.cpp
```diff
@@ -1568,7 +1588,21 @@ void Viewer::computeCameraMatrices(mx::Matrix44& world,
     world *= mx::Matrix44::createTranslation(_modelTranslation).getTranspose();
 }
 
-void Viewer::updatePropertyEditor()
+void Viewer::updateDisplayedProperties()
 {
     _propertyEditor.updateContents(this);
+    mx::TypedElementPtr elem = getSelectedMaterial() ? getSelectedMaterial()->getElement() : nullptr;
+    const std::string& node = (elem && !elem->getAttribute("node").empty()) ? elem->getAttribute("node"): "";
```
jstone-lucasfilm: One minor note here:  If the return value of `elem->getAttribute("node")` is empty, then we can use that empty string as the value of this variable, saving us a bit of extra logic to explicitly check for the empty string and substitute `""`.

PR #283, opened 2019-08-19
Improvements to FilePath and FileSearchPath

jstone-lucasfilm:
- Make FilePath construction independent of the current platform, allowing FilePath to be used with stored strings.
- Add support for iteration over FileSearchPath objects.
- Update the interface of readFromXmlFile and writeToXmlFile to leverage FilePath and FileSearchPath.
- Extend Python bindings for FilePath and FileSearchPath.

### DIFF
--- a/source/MaterialXFormat/XmlIo.cpp
+++ b/source/MaterialXFormat/XmlIo.cpp
@@ -5,12 +5,9 @@
 
 #include <MaterialXFormat/XmlIo.h>
 
-#include <MaterialXFormat/File.h>
-
 #include <MaterialXFormat/PugiXML/pugixml.hpp>
 
 #include <MaterialXCore/Types.h>
-#include <MaterialXCore/Util.h>
 
 #include <cstring>
 #include <fstream>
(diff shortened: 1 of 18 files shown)


source/MaterialXFormat/XmlIo.cpp
```diff
@@ -120,36 +117,35 @@ void elementToXml(ConstElementPtr elem, xml_node& xmlNode, const XmlWriteOptions
     }
 }
 
-void xmlDocumentFromFile(xml_document& xmlDoc, string filename, const string& searchPath)
+void xmlDocumentFromFile(xml_document& xmlDoc, FilePath filename, FileSearchPath searchPath)
 {
-    FileSearchPath fileSearchPath = FileSearchPath(searchPath);
-    fileSearchPath.append(getEnvironmentPath());
+    searchPath.append(getEnvironmentPath());
```
bernardkwok: I believe append() does not check to see if it already exists so the side-effect is that calling this function multiple times with the same input searchPath could keep on adding the same environment path.
Seems this method should continue to create a local variable and append to that instead.

source/MaterialXView/Viewer.cpp
```diff
@@ -67,25 +67,22 @@ void writeTextFile(const std::string& text, const std::string& filePath)
...
         readOptions.skipConflictingElements = true;
-        for (const std::string& filename : filenames)
+
+        mx::FilePath libraryPath = searchPath.find(libraryFolder);
+        for (const mx::FilePath& filename : libraryPath.getFilesInDirectory("mtlx"))
         {
-            mx::FilePath file = path / filename;
             mx::DocumentPtr libDoc = mx::createDocument();
-            mx::readFromXmlFile(libDoc, file, mx::EMPTY_STRING, &readOptions);
-            libDoc->setSourceUri(file);
```
bernardkwok: I assume we don't need this anymore since import will add in a source URI?

source/MaterialXFormat/XmlIo.cpp
jstone-lucasfilm: Although we could be more clear about this, the searchPath variable is _already_ a local copy, because it's now passed by value instead of by reference.

source/MaterialXView/Viewer.cpp
jstone-lucasfilm: Yes, exactly.  It should be very rare that we need to explicitly set the source URI, and in this case the call to importLibrary() will handle that for us.

source/MaterialXFormat/XmlIo.cpp
bernardkwok: Ah. I missed this. I was assuming it was still by reference. Please ignore the comment then.

PR #284, opened 2019-08-20
Added python bindings for TypeDesc

laserallan:
Added missing base class for DefaultColorManagementSystem

### DIFF
--- a/source/MaterialXGenShader/TypeDesc.cpp
+++ b/source/MaterialXGenShader/TypeDesc.cpp
@@ -47,7 +47,7 @@ const TypeDesc* TypeDesc::registerType(const string& name, unsigned char basetyp
         throw ExceptionShaderGenError("A type with name '" + name + "' is already registered");
     }
 
-    TypeDescPtr ptr = TypeDescPtr(new TypeDesc(name, basetype, semantic, size, editable, channelMapping));
+	TypeDescPtr ptr = TypeDescPtr(new TypeDesc(name, basetype, semantic, size, editable, channelMapping), TypeDesc::destroy);
     map[name] = ptr;
 
     return ptr.get();
--- a/source/MaterialXGenShader/TypeDesc.cpp
+++ b/source/MaterialXGenShader/TypeDesc.cpp
@@ -11,7 +11,7 @@ namespace MaterialX
 
 namespace
 {
-    using TypeDescPtr = std::shared_ptr<TypeDesc>;
+    using TypeDescPtr = std::unique_ptr<TypeDesc>;
     using TypeDescMap = std::unordered_map<string, TypeDescPtr>;
 
     // Internal storage of the type descriptor pointers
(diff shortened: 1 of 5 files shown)


source/MaterialXGenShader/TypeDesc.cpp
```diff
@@ -47,7 +47,7 @@ const TypeDesc* TypeDesc::registerType(const string& name, unsigned char basetyp
         throw ExceptionShaderGenError("A type with name '" + name + "' is already registered");
     }
 
-    TypeDescPtr ptr = TypeDescPtr(new TypeDesc(name, basetype, semantic, size, editable, channelMapping));
+    TypeDescPtr ptr = TypeDescPtr(new TypeDesc(name, basetype, semantic, size, editable, channelMapping), TypeDesc::destroy);
```
jstone-lucasfilm: Given the current behavior of TypeDesc::registerType, I see the logic behind implementing Python bindings in this way, but I'm wondering if perhaps it would be better to pass type descriptions consistently by TypeDescPtr.  Adding @bernardkwok for his thoughts as well.

jstone-lucasfilm: To keep type comparisons as efficient as possible, another option would be to define TypeDescPtr as a _raw pointer_, removing smart pointers entirely from the management of TypeDesc object lifetimes.  This seems more in keeping with the spirit of the current design, where all TypeDesc objects are allocated during static initialization, and are deleted during static destruction.

source/MaterialXGenShader/TypeDesc.cpp
```diff
@@ -47,10 +47,11 @@ const TypeDesc* TypeDesc::registerType(const string& name, unsigned char basetyp
         throw ExceptionShaderGenError("A type with name '" + name + "' is already registered");
     }
 
-    TypeDescPtr ptr = TypeDescPtr(new TypeDesc(name, basetype, semantic, size, editable, channelMapping));
-    map[name] = ptr;
+    std::unique_ptr<TypeDesc> uniquePtr(new TypeDesc(name, basetype, semantic, size, editable, channelMapping));
+    TypeDesc* rawPtr = uniquePtr.get();
+    map[name] = std::move(uniquePtr);
```
niklasharrysson: Thanks for fixing this! It clearly should have been a unique_ptr from the start.

For longer term I think we could improve on encapsulating the ownership of TypeDesc instances better. Letting TypeDesc be a lightweight carrier of the pointer without exposing it, similar to string interning. To avoid the risk of API users taking ownership of them. But still keeping type compares efficient.

laserallan: Looks like an improvement to me.

jstone-lucasfilm: Thanks to David for putting this changelist together!

Niklas — Yes, we could consider an object like observer_ptr (https://en.cppreference.com/w/cpp/experimental/observer_ptr) for the future, which would represent all non-owning, client-facing pointers.  For now, though, I think returning a raw pointer to a const object is a fine solution.

PR #290, opened 2019-09-03
Write relative XInclude paths as Posix

laserallan:


niklasharrysson:
@jstone-lucasfilm What du you think about choosing a fixed separator convention for the file format? Instead of an option to choose. So for example make the posix convention always be used when writing to file, and let the API handle conversion to the platform's convention when reading paths from file?

To my mind this would be more consistent, and then the files we produce will always look the same on all platforms. It will also be consistent with how geometry path separators are defined, which is always using the posix convention.

jstone-lucasfilm:
@niklasharrysson That would work if all include paths were relative, but this new flag will also affect the writing of absolute path prefixes like "D:\LocalPath" and "\\NetworkPath", so it's not simply a path separator selection.  I see your point, though, about writing consistent Posix paths when possible, and David made the same point in an off-line thread.

As an alternative, what would you think about writing *relative* paths consistently as Posix in XInclude references, but writing *absolute* paths consistently in the native format (which may be either Posix or Windows)?  This would maintain consistency for the common case of relative XInclude paths, but would never convert an absolute path to a confusing form like "/D:/LocalPath" to maintain that consistency.

dbsmythe:
Personally, I would vote to never allow full file paths in filenames precisely because the root part of a file’s final location is highly site- and operating system-dependent.  Even if the operating system is the same, the actual location will differ from one site or even project to another.  We should IMHO strongly encourage using MATERIALX_SEARCH_PATH (or whatever the Windows equivalent is) to access files from library/etc (non local project) locations.

jstone-lucasfilm:
@dbsmythe That's a great ideal to strive for, but even our own material libraries have historically broken that rule, and it seems impractical to enforce this for other studios.

laserallan:
I’m fine with that. My original concern is validating a written materialx
file against a reference file in a test that works across platforms. This
would also apply to any build process using file contents (or hashes of
file contents). With going through a filepath, suddenly windows and Linux
and windows produces different results for the same input which is not good
imo.

jstone-lucasfilm:
@laserallan If that approach works for you, then I would recommend removing the new flag, and simply checking for filePath.isAbsolute() before writing out the XInclude reference.  Relative paths can always be written using Posix conventions, but absolute paths must be written natively.

niklasharrysson:
I agree we should strongly encourage the use of relative paths and using search paths to find the includes. But like Jonathan wrote I don't think we can ban absolute paths completely.

However in the rare cases where you do need absolute paths I don't think paths like "/D:/LocalPath/MyFile.mtlx" are too confusing. It's the correct syntax for an URI on Windows, omitting the "file://"" prefix. The specification states we support URI's already for the filename data type, so maybe we could use the same for includes?

jstone-lucasfilm:
@niklasharrysson That's a good point, though I think we would lose the distinction between network and absolute paths on Windows.  If the Windows-native path is "\\NetworkPath" and we store it in Posix as "/NetworkPath", then it will be incorrectly reconstructed by a Windows-native importing application as "NetworkPath".  To preserve the full set of Windows distinctions, I can't think of any way around writing absolute paths in native format.

niklasharrysson:
Good point, we would need the full URI syntax then, which might feel to verbose. Anyway, I'm totally fine with using the native syntax for absolute paths, since you have already limited the portability in that case.

laserallan:
@jstone-lucasfilm : Just to make sure I understand you correctly. I should change the pull request to check for relative paths and always force them to be posix so we only use windows paths for absolute paths on windows platforms? 
It would make the code more sensible since the entire process can happen before the file is included rather than now where I need to convert it back to a FilePath to do the conversion when writing out the file.

jstone-lucasfilm:
@laserallen Yes, I’m imagining that the XInclude-writing code would check isAbsolute, then write Posix (i.e. universal) format for relative paths, and Native format for absolute paths.  For other file access situations, it’s still valuable for FilePath to assume native formatting, but for XIncludes this pattern seems like a good step forward.  Eventually, the ideal would be for FilePath to support a UriFormat in addition to Posix and Windows, but that’s not critical right now.

dbsmythe:
What would you all think about “highly recommending” that absolute file paths should be split into a relative path (which would be posix syntax) plus a fileprefix to add the OS-specific component like “/D:/“ or whatever?

-Doug

jstone-lucasfilm:
It's a good recommendation, but might be beyond the scope of MaterialX itself, and in the domain of best practices for DCC vendors in the future.  Without naming names, I'm familiar with DCC tools that don't support global filepaths for asset dependencies, so all asset references are on absolute local or network paths.

laserallan:
Changed the fix to just make sure any relative path is written as a posix path

PR #293, opened 2019-09-13
Initial implementation of texture baking

mjyip-lucasfilm:


source/MaterialXGenShader/GenOptions.h
```diff
@@ -93,6 +93,12 @@ class GenOptions
     /// compress a set of UDIMs into a single normalized image for
     /// hardware rendering.
     bool hwNormalizeUdimTexCoords;
+
+    /// Variables for texture space renderings
+    bool textureSpaceRender;
+    std::string textureSpaceInput;
```
jstone-lucasfilm: Here I think it would be better to store the original InputPtr, which gives us access to both the input name and input type string.

source/MaterialXView/Viewer.h
```diff
@@ -136,6 +137,11 @@ class Viewer : public ng::Screen
     bool _translationActive;
     ng::Vector2i _translationStart;
 
+    // TextureBaker information
+    mx::TextureBaker _textureBaker;
```
jstone-lucasfilm: Since you've already built the infrastructure to create and store a TextureBakerPtr, I would use that data type here instead of the raw object instance.

source/MaterialXView/Viewer.cpp
```diff
@@ -1108,6 +1113,21 @@ bool Viewer::keyboardEvent(int key, int scancode, int action, int modifiers)
...
     }
 
+    // Bake textures and create MTLX file for material
+    if (key == GLFW_KEY_B && action == GLFW_PRESS)
+    {
+        MaterialPtr material = getSelectedMaterial();
+        mx::DocumentPtr doc = material ? material->getDocument() : nullptr;
+        mx::TypedElementPtr elem = material ? material->getElement() : nullptr;
+        const std::string udim = material ? material->getUdim() : "";
+        _textureBaker.bakeAllInputTextures(_textureDimensions, _textureFormat, _searchPath, elem, _genContext, udim);
```
jstone-lucasfilm: Here I think it's important to give the user a chance to select the location to which their MaterialX document and textures will be written, so that we're not creating additional files inside their material library.  I would recommend calling the same NanoGUI method used in createSaveMaterialsInterface, giving the user a dialog that lets them choose the location of the saved MaterialX file, and then saving the baked textures to the same destination folder.

source/MaterialXRenderGlsl/GlslValidator.h
```diff
@@ -152,6 +162,9 @@ class GlslValidator : public ShaderValidator
     /// Flag to indicate if validator has been initialized properly.
     bool _initialized;
 
+    /// Data type being rendered out to texture
+    std::string _type = "color3";
```
jstone-lucasfilm: This data should be knowable from the input type string (or better yet, InputPtr) that was stored in the GenOptions instance by the client.

source/MaterialXView/Material.cpp
```diff
@@ -288,16 +288,16 @@ void Material::bindViewInformation(const mx::Matrix44& world, const mx::Matrix44
     mx::Matrix44 invTransWorld = world.getInverse().getTranspose();
 
     // Bind view properties.
-    _glShader->setUniform(mx::HW::WORLD_MATRIX, ng::Matrix4f(world.getTranspose().data()));
-    _glShader->setUniform(mx::HW::VIEW_PROJECTION_MATRIX, ng::Matrix4f(viewProj.getTranspose().data()));
+    _glShader->setUniform(mx::HW::WORLD_MATRIX, ng::Matrix4f(world.getTranspose().data()), _enableWarnings);
```
jstone-lucasfilm: Instead of an enable warnings flag, it may be better just to check for the presence of each uniform before we set it.  You can see that pattern below for the WORLD_INVERSE_TRANSPOSE_MATRIX uniform, which is not always present when this method is called.

source/MaterialXRenderGlsl/GlslValidator.cpp
```diff
@@ -313,6 +315,66 @@ void GlslValidator::validateCreation(const StageMap& stages)
     _program->build();
 }
 
+void GlslValidator::renderScreenSpaceQuad(GenContext& context)
+{
+    bindTarget(true);
+    _type = context.getOptions().textureSpaceInputType;
+    (_type == "color3" || _type == "color4")? glEnable(GL_FRAMEBUFFER_SRGB) : glDisable(GL_FRAMEBUFFER_SRGB);
```
bernardkwok: I'm not sure I understand why a color3 vs a color4 determines whether to perform gamma correction?
Perhaps this is for normal map generation. In which case I would allow a user setting to turn this on or off then no knowledge of this textureSpaceInputType is required.

source/MaterialXRenderGlsl/GlslValidator.cpp
```diff
@@ -313,6 +315,66 @@ void GlslValidator::validateCreation(const StageMap& stages)
...
+    glViewport(0, 0, _frameBufferWidth, _frameBufferHeight);
+
+    _program->bind();
+
+    // load and generate the texture
+    unsigned int VAO;
+    glGenVertexArrays(1, &VAO);
+    glBindVertexArray(VAO);
+
+    _program->bindTextures(_imageHandler);
```
bernardkwok: Could load in a quad to render with but this works as well.

source/MaterialXRenderGlsl/GlslValidator.h
```diff
@@ -39,7 +40,7 @@ class GlslValidator : public ShaderValidator
 {
   public:
     /// Create a GLSL validator instance
-    static GlslValidatorPtr create();
+    static GlslValidatorPtr create(unsigned int dim = 1024);
```
bernardkwok: Not sure why the default has changed from 512 to 1024? This will slow down unit tests.

source/MaterialXView/Material.cpp
bernardkwok: Agreed.

jstone-lucasfilm:
As a next step, I'd like @niklasharrysson and @bernardkwok to look more closely at this proposed change, to make sure any modifications to GLSL shader generation are aligned with the design of the system.  In particular, feel free to suggest alternative approaches in situations where the "textureSpaceRender" flag is applied during code generation, as we'd like this behavior to be aligned with future use cases as well.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -143,8 +143,19 @@ void ShaderGraph::addUpstreamDependencies(const Element& root, ConstMaterialPtr
             ShaderInput* input = rootNode->getInput(connectingElement->getName());
             if (input)
             {
-                input->breakConnection();
-                input->makeConnection(output);
+                if (context.getOptions().textureSpaceRender)
```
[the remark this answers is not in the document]
niklasharrysson: I agree, it's unfortunate if we need to have code branches in the general graph construction depending on rendering target. But I also don't understand why this is needed? Why does the shadergen graph need to be connected differently if the target is texture baking? Should it not be the same shader generated, but just rendering to another "output device".

source/MaterialXGenGlsl/GlslShaderGenerator.cpp
```diff
@@ -645,7 +645,12 @@ void GlslShaderGenerator::emitVariableDeclaration(const ShaderPort* variable, co
     }
     else
     {
-        string str = qualifier.empty() ? EMPTY_STRING : qualifier + " ";
+        string str = "";
+        if (context.getOptions().textureSpaceRender && stage.getName() == "vertex" && qualifier == "in")
```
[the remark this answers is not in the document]
niklasharrysson: I agree with Bernard. And supporting layout location for uniforms seems useful in general.

One option would be to let the location index be a signed integer, with -1 default, and then emit layout location if the index is positive. Then the texture baking specific handling here could be moved out from this general method.

Another option would be to _always_ emit layout locations for GLSL, but I'm not sure if that has implications elsewhere in other shader generators.

A third option would be to sub-class GlslShaderGenerator and add a derived shader generator specific for texture baking. This might also improve some of the other places where texture baking has been added in, to avoid feature creep into the general code. But again I done fully understand why the shader needs to be different for texture rendering compared to frame rendering?

source/MaterialXGenShader/GenOptions.cpp
```diff
@@ -14,7 +14,8 @@ GenOptions::GenOptions() :
     hwTransparency(false),
     hwSpecularEnvironmentMethod(SPECULAR_ENVIRONMENT_FIS),
     hwMaxActiveLightSources(3),
-    hwNormalizeUdimTexCoords(false)
+    hwNormalizeUdimTexCoords(false),
+    textureSpaceRender(false)
```
niklasharrysson: If a new derived shader generator was used for texture baking we would not need an option for this.

source/MaterialXGenShader/GenOptions.h
```diff
@@ -93,6 +93,9 @@ class GenOptions
     /// compress a set of UDIMs into a single normalized image for
     /// hardware rendering.
     bool hwNormalizeUdimTexCoords;
+
+    /// Enable texture-space rendering for the given element.
+    bool textureSpaceRender;
```
niklasharrysson: If a new derived shader generator was used for texture baking we would not need an option for this.

source/MaterialXGenShader/ShaderGraph.cpp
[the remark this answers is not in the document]
niklasharrysson: If the graph needs to be different then an alternative that might work is to do any such target specific edits after the general graph is constructed. There is also the possibility to sub-class ShaderGraph into a TextureBakeGraph and override logic where needed.
I just find it unfortunate the we need to check all these texture bake braches for _all_ shaders generated in any language or target :)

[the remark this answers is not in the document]
jstone-lucasfilm: It’s possible that some/all of these branches for texture-space rendering are no longer needed, and I’ll test that in the next pass of edits.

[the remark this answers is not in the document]
jstone-lucasfilm: I've verified that the ShaderGraph branches were not needed, and they're now removed.

source/MaterialXGenShader/GenOptions.h
jstone-lucasfilm: As it turns out, this flag was the legacy of an earlier implementation, and is no longer needed for texture baking.  :)

source/MaterialXGenGlsl/GlslShaderGenerator.cpp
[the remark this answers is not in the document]
jstone-lucasfilm: Removed.

jstone-lucasfilm:
Thanks for the feedback, @niklasharrysson and @bernardkwok.  As it turns out, the texture-space rendering flag is no longer needed by the current implementation, and has been removed.  Take a second look when you have a chance, and leave additional notes as needed.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void TextureBaker::bakeGraphOutput(OutputPtr output, GenContext& context, const FilePath& outputFolder)
+{
+    if (!output)
+    {
+        return;
+    }
+
+    _rasterizer = GlslValidator::create(_frameBufferRes);
```
bernardkwok: This can still be initialized outside and passed in as an argument as it should not vary for each bake
As mentioned before, I'm just a bit worried about the overhead of recreating the window+gl context every time..

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    // Create document.
+    DocumentPtr bakedTextureDoc = createDocument();
+
+    // Create top-level elements.
+    NodeGraphPtr bakedNodeGraph = bakedTextureDoc->addNodeGraph("NG_baked");
+    MaterialPtr bakedMaterial = bakedTextureDoc->addMaterial("M_baked");
+    ShaderRefPtr bakedShaderRef = bakedMaterial->addShaderRef(shaderRef->getName() + "_baked", shaderRef->getAttribute("node"));
+    bakedNodeGraph->setColorSpace("srgb_texture");
```
bernardkwok: I'm not sure but as "srgb" output is an option shouldn't this also be set based on that option?

source/MaterialXRenderGlsl/TextureBaker.cpp
jstone-lucasfilm: Good point, and I've modified the logic so that the rasterizer is now initialized in the TextureBaker constructor.

source/MaterialXRenderGlsl/TextureBaker.cpp
jstone-lucasfilm: I've updated the naming conventions and documentation here for clarity, but the encodeSrgb option only applies to the output of a single texture-space render.  This "srgb_texture" colorspace is being applied to the entire document, with color outputs being interpreted as sRGB, and scalar/vector outputs remaining linear.

PR #300, opened 2019-10-04
Initial support for units

bernardkwok:
**Added**
- Initial support for units, including the Unit, UnitDef, and UnitTypeDef classes.
- Support for unit conversion in shader code generation.
- Utility to generate Markdown from `nodedef` definition.

**Changed**
- Standard surface `subsurface_color` changed type to color3 from vector3. `subsurface_scale` supported in code generation
- Search path argument for MaterialXView adds search paths versus replacing.


source/MaterialXGenShader/HwShaderGenerator.h
```diff
@@ -196,6 +196,7 @@ namespace HW
     extern const string USER_DATA_LIGHT_SHADERS;
 }
 
+/// Shader stage identifiers.
```
jstone-lucasfilm: This was removed in a previous change, in order to keep Doxygen from incorrectly presenting the Stage namespace as a class.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -861,7 +991,7 @@ void ShaderGraph::optimize(GenContext& context)
     {
         std::set<ShaderNode*> usedNodes;
 
-        // Traverse the graph to find nodes still in use
+        // Travers the graph to find nodes still in use
```
jstone-lucasfilm: Merge error?

source/MaterialXCore/UnitConverter.cpp
```diff
@@ -0,0 +1,208 @@
+//
+// TM & (c) 2019 Lucasfilm Entertainment Company Ltd. and Lucasfilm Ltd.
+// All rights reserved.  See LICENSE.txt for license.
+//
+
+#include <algorithm>
+#include <MaterialXCore/Util.h>
+#include <MaterialXCore/Value.h>
+#include <MaterialXCore/UnitConverter.h>
```
jstone-lucasfilm: The new source files UnitConverter.cpp and UnitConverter.h might be more at home in MaterialXGenShader, unless there's an important use case you're imagining for them in lower-level libraries.

source/MaterialXCore/UnitConverter.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+namespace MaterialX
+{
+const string DistanceUnitConverter::DISTANCE_UNIT = "distance";
+
+DistanceUnitConverter::DistanceUnitConverter(UnitTypeDefPtr unitTypeDef) :
```
jstone-lucasfilm: This class seems much more general than just distance units.  Would it be better named DefaultUnitConverter?

source/MaterialXRender/ShaderRenderer.h
```diff
@@ -3,11 +3,11 @@
 // All rights reserved.  See LICENSE.txt for license.
 //
 
-#ifndef MATERIALX_SHADERVALIDATOR_H
```
jstone-lucasfilm: Does this pull request need to be merged with the latest changes in MaterialX master?  I wouldn't expect to see references to shader validators here.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -44,7 +44,23 @@
...
+      <unit name="millimeter" scale="0.001"/>
+      <unit name="centimeter" scale="0.01"/>
+      <unit name="kilometer" scale="1000.0"/>
+      <unit name="inch" scale="0.0254"/>
+      <unit name="foot" scale="0.3048"/>
+      <unit name="yard" scale="0.9144"/>
+      <unit name="mile" scale="1609.344"/>
+   </unitdef>
+
+   <!-- ======================================================================== -->
```
jstone-lucasfilm: Minor whitespace error here

resources/Materials/Examples/StandardSurface/standard_surface_default.mtlx
```diff
@@ -21,7 +21,7 @@
       <bindinput name="transmission_extra_roughness" type="float" value="0"/>
       <bindinput name="subsurface" type="float" value="0"/>
       <bindinput name="subsurface_color" type="color3" value="1, 1, 1"/>
-      <bindinput name="subsurface_radius" type="vector3" value="1, 1, 1"/>
+      <bindinput name="subsurface_radius" type="color3" value="1, 1, 1"/>
```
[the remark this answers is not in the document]
bernardkwok: I believe the desire was to make this a color to match up more with the user presentation of this which as a color which could be scaled to derive the final distance. The pbr node internally is still a vector and a distance. Pinging @niklasharrysson for this.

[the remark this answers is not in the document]
jstone-lucasfilm: My main concern is that color inputs will be transformed between color spaces as if they were perceptual colors (e.g. radiance, reflectance), which may not behave as expected for physical distances like subsurface radius.

If this has already been discussed in detail, I don't want to slow down the process of integrating the change to MaterialX, but I wanted to double-check that this behavior is truly intended.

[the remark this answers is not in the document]
niklasharrysson: Yes this change was a request from the Arnold team who is responsible for the Standard Surface specification. Their motivation for changing this to a color3 was that the perceived effect of changing it is very "color like". Which also includes connecting it to color textures that should be effected by color management. So the radius/color input can be use to control the perceptual color of the SSS and _scale_ input can then be used to turn this into a physical distance for mean free path.

As Bernard pointed out we choose to keep the parameterization of the `subsurface_bsdf` node in pbrlib with a vector3 still, because here we don't use a color + scalar but a single vector for radius. We also figured it would make sense to keep the lower level nodes in pbrlib more strictly physical while higher level nodes like `standard_surface` can choose to add more artistic interpretations of this.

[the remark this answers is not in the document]
jstone-lucasfilm: Got it, and thanks for the detailed answer, @niklasharrysson.  That makes sense to me, where the mean free path is the combination of a physical scale and a color in a particular color space.  I'll mark this question as resolved.

source/MaterialXCore/Element.cpp
```diff
@@ -325,6 +327,23 @@ ConstElementPtr Element::getRoot() const
     return root;
 }
 
+const string& Element::getActiveColorSpace(bool returnClosest) const
```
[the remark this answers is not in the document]
bernardkwok: Currently elements from an imported document if this method is will used to find an elements target colorspace this will just be that elements color space if there is one and hence no transform will occur. 
There may be a higher level element which would be the local target. If the document's root is used but is empty then no transform will occur. This allows for the search for the target to find the higher level element to be used as the source which could include the colorspace set on the children of the root of the imported document.

source/MaterialXCore/UnitConverter.cpp
bernardkwok: We thought about this one for a bit and it can handle more than distance, but will only handle conversion where a scale is used. We didn't want to introduce an arbitrary offset just to make it a generic "scalar" unit converter. 

We could call it something else, but not sure what a good name would be, but it's not really a default as it would be a bit misleading. Any suggestions welcome.

source/MaterialXCore/UnitConverter.cpp
bernardkwok: We want to handle 2 cases, 1 for shadergen and 1 for non-shadergen. For software rendering, as with colorspace, the desire is to be able to be able to get pre-transformed values without adding shader code for this.

source/MaterialXGenShader/ShaderGraph.cpp
bernardkwok: Yes, will fix this.

source/MaterialXRender/ShaderRenderer.h
bernardkwok: I thought that I had done this already. Could be I didn't re-merge from master back to this branch.

resources/Materials/Examples/Units/distance_units.mtlx
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<materialx version="1.37" >
```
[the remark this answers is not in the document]
bernardkwok: Will do.

source/MaterialXGenShader/HwShaderGenerator.h
bernardkwok: I'll clean up.

resources/Materials/Examples/StandardSurface/standard_surface_default.mtlx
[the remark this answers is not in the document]
dbsmythe: While the perception of what subsurface_radius does may be color-like, it is most definitely not a color, and really should not be a color3 type for all the reasons Jonathan already mentioned.  Also, it is fairly common to multiply the subsurface_radius by a scalar, which makes total sense for a vector, but not as much so for a color.  The only reason to express subsurface_radius as a color3 I can think of is if that's the only 3-channel data type your UI supports, but in that case I would hope that the application's interface to MaterialX would already have a color3 UI -> MaterialX vector3 conversion under-the-hood, which should apply here as well.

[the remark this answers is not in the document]
dbsmythe: Thought about this more on the drive into work, and it's a bit more complex than that.. technically speaking, subsurface_radius *is* a distance (mean free path length for scattering or something like that), and is typically specified in millimeters, but it's the distance/path length *per color channel*, which is dependent on what your color primaries are.  So the questions is, if you have a subsurface_radius value relative to one set of primaries, is the conversion of that value to a different set of primaries actually using the same math as converting an actual color value?  My gut feeling is "no" but I'd defer to someone with more knowledge in this area.

source/MaterialXCore/UnitConverter.cpp
bernardkwok: As mentioned by Ashwin, there is no mentioned or usage of offset.  DistanceUnitConverter is now called DefaultUnitConverter with a comment that it only handles conversion involving scalar multiplication.

resources/Materials/Examples/StandardSurface/standard_surface_default.mtlx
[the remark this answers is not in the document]
niklasharrysson: @dbsmythe I agree this is complex, and I think there are good arguments for both using a color and using a vector for this radius. The change we did now was to make the MaterialX nodedef in line with the Arnold teams description in the Standard Surface whitepaper. The Arnold team had a long debate about this internally and finally settled for using a color in the whitepaper. So I think it's best if we use the same in the MaterialX nodedef and avoid deviating from the whitepaper. In fact, as discussed before, we are planning to move this nodedef from the MaterialX repo and into the Standard Surface whitepaper repo, so it will be owned by the Arnold team. And we will just reference this in as a submodule in MaterialX.

On the other hand for the <subsurface_bsdf> nodedef in the PRB library, which is an interface that we "own" in the MaterialX repo, we are keeping the vector for radius because in that case it's a mean free path input directly, without any scalar multiplier to make it a "distance" mean free path.

source/MaterialXCore/Element.cpp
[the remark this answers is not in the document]
jstone-lucasfilm: I can see where you might want to transform a color to the "next colorspace up" in the hierarchy, but it's not clear to me why you would select the "furthest but non-root" colorspace that this function computes.  I would imagine that would only handle cases where exactly one level of libraries was imported into a document, rather than handling the general, hierarchical case.

As a side note, I think it's reasonable for now to handle just the common case of a single "working" colorspace, with all source colors being transformed into this single space.  For the future, I agree with your goal of handling complex colorspace hierarchies, but I believe this will require additional research and design work.

resources/Materials/Examples/Units/distance_units.mtlx
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<materialx version="1.36" >
+
+<!-- scale -->
+<constant name="realscale" type="float">
+   <parameter name="value" type="float" value="1.25" unit="centimeter" unittype="distance"/>
+</constant>
+
+  <!-- scale -->
```
jstone-lucasfilm: Minor:  A few odd whitespace inconsistencies in this file.

source/MaterialXGenShader/UnitConverter.cpp
```diff
@@ -0,0 +1,208 @@
+//
+// TM & (c) 2019 Lucasfilm Entertainment Company Ltd. and Lucasfilm Ltd.
+// All rights reserved.  See LICENSE.txt for license.
+//
+
+#include <algorithm>
```
jstone-lucasfilm: Minor:  This standard library should already be included through MaterialXCore/Library.h.

source/MaterialXRender/ImageHandler.h
```diff
@@ -279,7 +279,7 @@ class ImageHandler
     /// Clear the contents of the image cache.
     /// deleteImage() will be called for each cache description to
     /// allow derived classes to clean up any associated resources.
-    void clearImageCache();
+    virtual void clearImageCache();
```
jstone-lucasfilm: Merge error?  This was removed in a previous change, since the method is never overridden, and it's called from the destructor of a polymorphic class.

source/MaterialXView/Main.cpp
```diff
@@ -50,7 +50,7 @@ int main(int argc, char* const argv[])
         }
         if (token == "--path" && !nextToken.empty())
         {
-            searchPath = mx::FileSearchPath(nextToken);
+            searchPath.append(mx::FileSearchPath(nextToken));
```
jstone-lucasfilm: This is a good improvement, though I would suggest that we can now remove the explicit cast to mx::FileSearchPath.  The FileSearchPath class should correctly handle "append" with a standard string input.

source/MaterialXCore/Element.cpp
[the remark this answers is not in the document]
bernardkwok: I was thinking about this and was leaning towards leaving this out. I'll revert this and the import colorspace changes as I'd like to work out the bigger picture of the fact that we'd like to differentiate between "import" which embeds one file into another and a "reference" which the dependent document keeps it's own properties relative to a local "root". This traversal is partially a "guess" to try and go as far up as possible to find this local "root".

source/MaterialXRender/ImageHandler.h
bernardkwok: It's required so that PyImageHandler can override it in a newly exposed method.

source/MaterialXView/Main.cpp
bernardkwok: Unfortunately, you can't use a string input argument since it's ambiguous as to which signature it resolves to (the one with FileSearchPath or the one with a FilePath).

source/MaterialXRender/ImageHandler.h
jstone-lucasfilm: I think that the change to PyImageHandler is an unintended merge artifact as well.  We removed that binding at the same time the “virtual” modifier was removed here.

bernardkwok: Reverted now.

resources/Materials/Examples/Units/distance_units.mtlx
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jstone-lucasfilm: These new materials should probably be moved to Materials/TestSuite, since they're more focused on testing MaterialX features, where Materials/Examples is more focused on user-facing material examples.

source/MaterialXView/Viewer.h
```diff
@@ -86,8 +87,9 @@ class Viewer : public ng::Screen
     void drawScene2D();
 
     void setupLights(mx::DocumentPtr doc);
+    void setupUnitConverter(mx::DocumentPtr doc);
     void loadDocument(const mx::FilePath& filename, mx::DocumentPtr libraries);
-    void reloadShaders();
+    void reloadShaders(bool forceCreation);
```
jstone-lucasfilm: What's the proposed use case for forceCreation=true?  At the moment, I'm only seeing this method called from one place in the code, and forceCreation is set to false.

source/PyMaterialX/PyMaterialXCore/PyElement.cpp
```diff
@@ -159,6 +160,25 @@ void bindPyElement(py::module& mod)
         .def("_getValue", &mx::ValueElement::getValue)
         .def("_getBoundValue", &mx::ValueElement::getBoundValue)
         .def("_getDefaultValue", &mx::ValueElement::getDefaultValue)
+        .def("setUnit", &mx::ValueElement::setUnit)
+        .def("hasUnit", &mx::ValueElement::hasUnit)
+        .def("getUnit", &mx::ValueElement::getUnit)
+        .def("setUnitType", &mx::ValueElement::setUnitType)
+        .def("hasUnitType", &mx::ValueElement::hasUnitType)
+        .def("getUnitType", &mx::ValueElement::getUnitType)
+        .def_readonly_static("VALUE", &mx::ValueElement::VALUE_ATTRIBUTE)
```
jstone-lucasfilm: To match the conventions of our existing Python bindings, these Python constants should have the "ATTRIBUTE" prefix, and constants like "INTERFACENAME" should have the same underscore as their C++ equivalents (e.g. "INTERFACE_NAME_ATTRIBUTE").

resources/Materials/Examples/StandardSurface/standard_surface_emission.mtlx
```diff
@@ -77,7 +77,7 @@
       <bindinput name="transmission_extra_roughness" type="float" value="0"/>
       <bindinput name="subsurface" type="float" value="0"/>
       <bindinput name="subsurface_color" type="color3" value="1, 1, 1"/>
-      <bindinput name="subsurface_radius" type="vector3" value="1, 1, 1"/>
+      <bindinput name="subsurface_radius" type="color3" value="1, 1, 1"/>
```
meshula: is this a typo?

dbsmythe:
@meshula I think you're unfortunately at a bit a disadvantage here, in that you haven't seen the companion changes to the MaterialX Specification document, which (I hope) explains everything [I will email you a preview copy of the updated sections].  The thought here is that units would be used very sparingly, only applied to things that really do want to express themselves in terms of a specific real-world size expressed in a specific unit.  So you could say that a scanned texture image used in a material library is 2.0 feet by 4.0 feet, or that the values in a texture used for displacement are in millimeters.  We would then support automatic conversion of these values from their specified unit to an application-defined "scene unit" for any given unittype (e.g. "distance") within the MaterialX library itself, at least at ShaderGen code generation time.  If a node or nodedef does not express a specific choice of unit (you need both the "what unit it's in" and the "what unit do you want"), then no conversion would take place.  We would not likely want to add units to quantities such as luminous intensities expressed in lux or candelas or anything else where there isn't a clear need for unit conversion.
I am hopeful that the revisions to the 1.37 Spec can go live very soon; we're just making sure that the Spec accurately describes what we really want, and using the development of the library code to verify this.

meshula:
Thanks for sending it over, I'll follow up with any questions in the email!

source/MaterialXCore/Element.cpp
```diff
@@ -584,6 +586,26 @@ ValuePtr ValueElement::getDefaultValue() const
     return ValuePtr();
 }
 
+const string& ValueElement::getDefaultUnit() const
```
jstone-lucasfilm: What would you think of naming this method "getActiveUnit"?  To my mind, it's closely analogous to "getActiveColorspace", which applies any applicable rules (e.g. scoping, inheritance) to return the final colorspace that applies to the given element.

bernardkwok: This is fine with me. I think the suggestion for getDefaultUnit() was from you originally.

source/MaterialXGenShader/UnitConverter.h
```diff
@@ -0,0 +1,218 @@
...
+
+/// A shared pointer to an UnitConverterRegistry
+using UnitConverterRegistryPtr = shared_ptr<UnitConverterRegistry>;
+/// A shared pointer to a const UnitConverterRegistry
+using ConstUnitConverterRegistryPtr = shared_ptr<const UnitConverterRegistry>;
+
+/// @class UnitConverterRegistry
+/// A registry of unit converters.
+///
+class UnitConverterRegistry
```
jstone-lucasfilm: What is the conceptual distinction between UnitConverterRegistry and UnitSystem?  Could the former be merged into the latter, making it responsible for managing the list of converters as well?

One related note is that we might consider the name UnitManagementSystem instead of UnitSystem, making a clearer parallel with ColorManagementSystem, and avoiding confusion with "unit systems" like Imperial Units and International Units.

source/MaterialXView/Viewer.cpp
```diff
@@ -694,6 +721,17 @@ void Viewer::createAdvancedSettings(Widget* parent)
...
         updateDisplayedProperties();
     });
+
+    // Units
+    {
+        new ng::Label(advancedPopup, "Unit Options");
+        // Unit selection widget
+        Widget* sampleGroup = new Widget(advancedPopup);
+        sampleGroup->setLayout(new ng::BoxLayout(ng::Orientation::Horizontal));
+        new ng::Label(sampleGroup, "Unit Space:");
```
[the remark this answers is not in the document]
bernardkwok: Yes. There is not use to have this ambiguous name.

source/MaterialXGenShader/UnitConverter.h
bernardkwok: The registry can be used to keep a 1:1 relation between unit types and converters have have this done outside of shadergen (e.g. for software renderers). The UnitSystem which is mostly a construct for insertion of shader code As you may recall the registry and converters used to be in core but it was moved to shadergen as it's only used here.. It should still be possible to have conversion support without shader generation.  Ping @ashwinbhat for this as well.

meshula:
Just a quick comment about my earlier question about how the units are intended to be used. The PR as it stands now addresses the things I was worried about, specifically the possibility of writing incorrect conversions such as attempting to convert units of distance with units of temperature or something.

source/MaterialXView/Viewer.cpp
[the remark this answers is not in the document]
jstone-lucasfilm: Ok, good to know, and this is minor cleanup that we can make in an upcoming pass.  I'll close this out for now.

source/MaterialXGenShader/UnitConverter.h
jstone-lucasfilm: That makes sense, and thanks for the clarification.

PR #301, opened 2019-10-06
Generalize shader validator classes

jstone-lucasfilm:
- Rename the ShaderValidator, GlslValidator, and OslValidator classes to ShaderRenderer, GlslRenderer, and OslRenderer, generalizing their role to rendering as well as validation.
- Rename class methods for consistency with both rendering and validation use cases.
- Update Python bindings and documentation.


jstone-lucasfilm:
I’d appreciate any thoughts you have on this change, @niklasharrysson and @bernardkwok.

PR #304, opened 2019-11-01
Adsk contrib/adsk to master explicit output

feldstj:


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 project(MaterialX)
 set(MATERIALX_MAJOR_VERSION 1)
-set(MATERIALX_MINOR_VERSION 36)
```
feldstj: Updated the MaterialX version, was needed for the upgrade path to work correctly.

libraries/bxdf/disney_brdf_2012.mtlx
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <materialx version="1.36" cms="ocio" colorspace="lin_rec709">
-  <nodedef name="ND_disney_brdf_2012_surface" type="surfaceshader" node="disney_brdf_2012">
```
feldstj: Tons of changes like this one which remove the type in favor of a child output called "out".

source/MaterialXCore/Definition.cpp
```diff
@@ -91,15 +91,7 @@ vector<ShaderRefPtr> NodeDef::getInstantiatingShaderRefs() const
...
-    validateRequire(hasType(), res, message, "Missing type");
-    if (isMultiOutputType())
-    {
-        validateRequire(getOutputCount() >= 2, res, message, "Multioutput nodedefs must have two or more output ports");
-    }
-    else
-    {
-        validateRequire(getOutputCount() == 0, res, message, "Only multioutput nodedefs support output ports");
-    }
+    validateRequire(!hasType(), res, message, "Nodedef should not have a type but an explicit output");
```
feldstj: NodeDefs should no longer have an explicit type attribute for validation.

source/MaterialXCore/Definition.h
```diff
@@ -88,6 +88,25 @@ class NodeDef : public InterfaceElement
         return getAttribute(NODE_ATTRIBUTE);
     }
 
+    /// Return the element's output type.
+    const string& getType() const override
```
feldstj: Overrides getType() to return the type of the child output, if there is more than one, multioutput is returned.

source/MaterialXCore/Document.cpp
```diff
@@ -601,6 +601,21 @@ void Document::upgradeVersion()
         minorVersion = 36;
     }
 
+    // Upgrade path for 1.37 (change types to child outputs)
+    for (NodeDefPtr nodeDef : getNodeDefs())
```
feldstj: This is the upgrade path for old documents which had NodeDefs with type attributes. Removes these attributes and creates a child output called "out".

source/MaterialXCore/Document.h
```diff
@@ -320,7 +320,10 @@ class Document : public GraphElement
                           const string& node = EMPTY_STRING)
     {
         NodeDefPtr child = addChild<NodeDef>(name);
-        child->setType(type);
+        if (!type.empty() && type != MULTI_OUTPUT_TYPE_STRING)
```
feldstj: Only add an output of the specified type called "out" if a user does not create a type-less NodeDef or a multioutput NodeDef.

source/MaterialXCore/Interface.cpp
```diff
@@ -13,6 +13,16 @@
 namespace MaterialX
 {
 
+namespace
+{
+    auto ValueElementLess = [](ValueElementPtr lhs, ValueElementPtr rhs)
```
feldstj: Value Element comparison routine needed by sets further down.

source/MaterialXCore/Interface.cpp
```diff
@@ -317,10 +327,17 @@ InputPtr InterfaceElement::getActiveInput(const string& name) const
 vector<InputPtr> InterfaceElement::getActiveInputs() const
 {
     vector<InputPtr> activeInputs;
+    ValueElementSet activeInputsSet(ValueElementLess);
```
feldstj: For outputs, inputs, and value elements in general, to get the activeOutputs, activeInputs, and activeValueElements, use a set to avoid duplicating inherited value elements. This was for example a parent NodeDef can have an output called "out" and the child NodeDef can have the same output called "out" and we wont end up with two outputs, but just one. This is also particularly useful for inputs as the input that is most derived will be used over the parent input allowing it to be more specialized.

source/MaterialXTest/GenShaderUtil.cpp
```diff
@@ -650,53 +650,59 @@ void ShaderGeneratorTester::validate(const mx::GenOptions& generateOptions, cons
...
             }
 
-            // Allow to skip nodedefs to test if specified
-            const std::string nodeDefName = nodeDef->getName();
-            if (_skipNodeDefs.count(nodeDefName))
+            if (!nodeDef)
             {
-                _logFile << ">> Skipped testing nodedef: " << nodeDefName << std::endl;
-                continue;
+                CHECK(nodeDef);
```
feldstj: This is a minor change. Just makes this code easier to debug by enabling users to place a breakpoint inside this if statement for when nodeDef is a nullptr. I found it useful to do so for debugging purposes so figured I'd contribute it back.

PR #314, opened 2019-11-18
Improvements to image functionality

jstone-lucasfilm:
- Rename the ImageDesc class to Image, and create a clearer separation between its interface and data.
- Move image creation logic to the Image class.
- Move image caching and fallback behavior to the base ImageHandler class.
- Add an initial Image::getTexelColor method, opening the door to future image processing and analysis functionality.

### DIFF
--- a/source/MaterialXRender/ImageHandler.cpp
+++ b/source/MaterialXRender/ImageHandler.cpp
@@ -19,43 +19,148 @@ const string VADDRESS_MODE_SUFFIX(IMAGE_PROPERTY_SEPARATOR + "vaddressmode");
 const string FILTER_TYPE_SUFFIX(IMAGE_PROPERTY_SEPARATOR + "filtertype");
 const string DEFAULT_COLOR_SUFFIX(IMAGE_PROPERTY_SEPARATOR + "default");
 
-string ImageDesc::BASETYPE_UINT8 = "UINT8";
-string ImageDesc::BASETYPE_HALF = "HALF";
-string ImageDesc::BASETYPE_FLOAT = "FLOAT";
+const string ImageLoader::BMP_EXTENSION = "bmp";
+const string ImageLoader::EXR_EXTENSION = "exr";
+const string ImageLoader::GIF_EXTENSION = "gif";
+const string ImageLoader::HDR_EXTENSION = "hdr";
+const string ImageLoader::JPG_EXTENSION = "jpg";
+const string ImageLoader::JPEG_EXTENSION = "jpeg";
+const string ImageLoader::PIC_EXTENSION = "pic";
+const string ImageLoader::PNG_EXTENSION = "png";
+const string ImageLoader::PSD_EXTENSION = "psd";
+const string ImageLoader::TGA_EXTENSION = "tga";
+const string ImageLoader::TIF_EXTENSION = "tif";
+const string ImageLoader::TIFF_EXTENSION = "tiff";
+const string ImageLoader::TX_EXTENSION = "tx";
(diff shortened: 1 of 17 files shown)


source/MaterialXRender/OiioImageLoader.cpp
```diff
@@ -78,82 +78,59 @@ bool OiioImageLoader::saveImage(const FilePath& filePath,
...
-            read = true;
-        }
+        image->setResourceBuffer(imageBuf);
+    }
+    else
+    {
+        image = nullptr;
     }
-
     imageInput->close();
```
lgritz: Are you leaking the imageInput here?

jstone-lucasfilm: Good catch, thanks Larry.  It looks like we're missing the call to ImageInput::destroy, and similarly the call to ImageOutput::destroy in the function above.

lgritz: Also, the fact that you're using a raw pointer means it will only work with OIIO 1.x. In 2.x it changed to a unique_ptr. The 1.x/2.x agnostic idiom is 

    auto imageInput = OIIO::ImageInput::open(filePath); // works for raw or unique_ptr
    ... all rest the same...
    OIIO::ImageInput::destroy(imageInput); // destroy is superfluous once min is >= 2.0

jstone-lucasfilm: Ah, thanks for the tip!  We’ll plan to update this to the version-agnostic idiom for future compatibility.

source/MaterialXRender/ImageHandler.cpp
```diff
@@ -111,83 +216,65 @@ bool ImageHandler::saveImage(const FilePath& filePath,
...
-}
 
-bool ImageHandler::createColorImage(const Color4& color, ImageDesc& desc)
-{
-    unsigned int bufferSize = desc.width * desc.height * desc.channelCount;
-    if (bufferSize < 1)
+    if (message && !filePath.isEmpty())
     {
-        return false;
+        *message = string("Error loading image: ") + filePath.asString();
```
bernardkwok: Minor: I'm just curious what more we get from this as it just says something failed. Would we want to extend this to get failure message from a given image loader?

source/MaterialXRender/ImageHandler.cpp
```diff
@@ -19,43 +19,148 @@ const string VADDRESS_MODE_SUFFIX(IMAGE_PROPERTY_SEPARATOR + "vaddressmode");
...
+            for (unsigned int c = 0; c < image->getChannelCount(); c++)
+            {
+                *pixel++ = color[c];
+            }
+        }
+    }
+    return image;
+}
+
+Color4 Image::getTexelColor(unsigned int x, unsigned int y) const
```
bernardkwok: This doesn't seem to be used anywhere. I'm just curious about where this is intended to be used?

source/MaterialXRender/ImageHandler.cpp
jstone-lucasfilm: For now, it’s just a way to recognize that the requested file failed to load, even though a valid image may have been returned using the constant-color fallback logic.  For the future, I like the idea of enhancing this error message with additional details from the loader.

source/MaterialXRender/OiioImageLoader.cpp
lgritz: ... or even past compatibility. OIIO 2.0 is already a full year old and 2.1 is about to be in release candidate stage. :-)

PR #365, opened 2020-02-21
Assume column vectors in createTranslation

jstone-lucasfilm:
This changelist updates Matrix33::createTranslation and Matrix44::createTranslation to assume column vectors, matching the prevailing convention in the MaterialX library.

Additional documentation and clarifications to Matrix33 and Matrix44 have been added where needed.

### DIFF
--- a/python/MaterialXTest/main.py
+++ b/python/MaterialXTest/main.py
@@ -96,26 +96,26 @@ class TestMaterialX(unittest.TestCase):
         # Translation and scale
         trans = mx.Matrix44.createTranslation(mx.Vector3(1, 2, 3))
         scale = mx.Matrix44.createScale(mx.Vector3(2))
-        self.assertTrue(trans == mx.Matrix44(1, 0, 0, 0,
-                                             0, 1, 0, 0,
-                                             0, 0, 1, 0,
-                                             1, 2, 3, 1))
+        self.assertTrue(trans == mx.Matrix44(1, 0, 0, 1,
+                                             0, 1, 0, 2,
+                                             0, 0, 1, 3,
+                                             0, 0, 0, 1))
         self.assertTrue(scale == mx.Matrix44(2, 0, 0, 0,
                                              0, 2, 0, 0,
                                              0, 0, 2, 0,
                                              0, 0, 0, 1))
 
         # Indexing operators
-        self.assertTrue(trans[3, 2] == 3)
-        trans[3, 2] = 4
-        self.assertTrue(trans[3, 2] == 4)
-        trans[3, 2] = 3
+        self.assertTrue(trans[2, 3] == 3)
+        trans[2, 3] = 4
+        self.assertTrue(trans[2, 3] == 4)
+        trans[2, 3] = 3
 
         # Matrix methods
-        self.assertTrue(trans.getTranspose() == mx.Matrix44(1, 0, 0, 1,
-                                                            0, 1, 0, 2,
-                                                            0, 0, 1, 3,
-                                                            0, 0, 0, 1))
+        self.assertTrue(trans.getTranspose() == mx.Matrix44(1, 0, 0, 0,
+                                                            0, 1, 0, 0,
+                                                            0, 0, 1, 0,
+                                                            1, 2, 3, 1))
         self.assertTrue(scale.getTranspose() == scale)
         self.assertTrue(trans.getDeterminant() == 1)
         self.assertTrue(scale.getDeterminant() == 8)
@@ -128,18 +128,18 @@ class TestMaterialX(unittest.TestCase):
         prod3 = trans * 2
         prod4 = trans
         prod4 *= scale
-        self.assertTrue(prod1 == mx.Matrix44(2, 0, 0, 0,
-                                             0, 2, 0, 0,
-                                             0, 0, 2, 0,
-                                             2, 4, 6, 1))
-        self.assertTrue(prod2 == mx.Matrix44(2, 0, 0, 0,
-                                             0, 2, 0, 0,
-                                             0, 0, 2, 0,
-                                             1, 2, 3, 1))
-        self.assertTrue(prod3 == mx.Matrix44(2, 0, 0, 0,
-                                             0, 2, 0, 0,
-                                             0, 0, 2, 0,
-                                             2, 4, 6, 2))
+        self.assertTrue(prod1 == mx.Matrix44(2, 0, 0, 1,
+                                             0, 2, 0, 2,
+                                             0, 0, 2, 3,
+                                             0, 0, 0, 1))
+        self.assertTrue(prod2 == mx.Matrix44(2, 0, 0, 2,
+                                             0, 2, 0, 4,
+                                             0, 0, 2, 6,
+                                             0, 0, 0, 1))
+        self.assertTrue(prod3 == mx.Matrix44(2, 0, 0, 2,
+                                             0, 2, 0, 4,
+                                             0, 0, 2, 6,
+                                             0, 0, 0, 2))
         self.assertTrue(prod4 == prod1)
 
         # Matrix division
--- a/source/MaterialXCore/Types.cpp
+++ b/source/MaterialXCore/Types.cpp
@@ -65,39 +65,39 @@ template <> Matrix33 MatrixN<Matrix33, float, 3>::getAdjugate() const
         _arr[0][0]*_arr[1][1] - _arr[1][0]*_arr[0][1]);
 }
 
-Vector3 Matrix33::multiply(const Vector3& rhs) const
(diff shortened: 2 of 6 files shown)


lgritz:
Sorry if I've been out of the loop on this, but doesn't switching to column vectors lose memory layout compatibility with Imath, OpenGL, RenderMan?

jstone-lucasfilm:
@lgritz MaterialX still uses the same row-major memory layout for matrices, but the convention has been for matrix-vector operations to assume column vectors (i.e. v' = Mv).  Previously, all MaterialX methods followed this convention _except_ for createTranslation, and this changelist brings that one odd method back in line.

lgritz:
I see. So you've had an incompatible memory layout (semantically) all along? I guess I missed that. Sorry for the bother.

jstone-lucasfilm:
@lgritz I believe OpenGL follows the same convention, with row-major memory layout and column vectors, but I will double check.

lgritz:
The notation is really tricky. Let me say it in a more explicit way.

    Matrix44 M = createTranslationMatrix(x, y, z);
    float* p = (float*)&M;

do the contents of p look like

    A:    [ 1 0 0 0  0 1 0 0  0 0 1 0  x y z 1 ]

or

    B:    [ 1 0 0 1  0 1 0 y  0 0 1 z  0 0 0 1 ]

 ???

RenderMan and OpenGL use memory order A. (Though in the *documentation*, when writing like math notation, OpenGL says it like B. But the memory layout is still A.)

lgritz:
I'm so sorry for hijacking this PR.

lgritz:
Imath/OpenEXR also uses A.

jstone-lucasfilm:
@lgritz This changelist would align MaterialX consistently with B, where createTranslation was the one method that previously didn't behave this way.  But I see your point about the behavior of RenderMan and OpenEXR, and I'm not against making a switch in the future.

jstone-lucasfilm:
@lgritz To be honest, though, it's only a small amount of code that would need to change if we were to switch MaterialX over to row vectors, instead of merging this pull request.  Do you believe that would be a more aligned approach with respect to projects at the ASWF?  If so, I'm open to making that change and updating any dependent projects.

dbsmythe:
If I had a vote, I’d go with A- I don’t recall ever seeing the B-style matrix layout in memory, and if A is what’s widely used in other open-source projects and popular renderers, I think we should go the same.

-Doug

bernardkwok:
- For @jstone-lucasfilm: Small note that the GlslRenderer code currentl calls createTranslation() so that transpose needs to be removed as well.
- MDL and metal also follows OpenGL convention so there is consistency there. 
- For HLSL we've just transposed at shader bind time.

jstone-lucasfilm:
@dbsmythe You definitely get a vote, and I'm starting to lean towards adopting row vectors (A) as our convention.  If we move forward with that approach, then all transform-generating methods _other_ than createTranslation would have their logic transposed, and matrices would concatenate in reverse order.

@bernardkwok Thanks for those additional notes, though following the line of the discussion above, it may be preferable for us to switch over to row vectors instead of column vectors.  Send any additional thoughts my way.

lgritz:
How have you managed to make two different matrix libraries at Lucasfilm, that use different memory layout conventions, without the users of those libraries coming after the developers in the hallways with pitchforks?

Dunno if this helps or is merely a cautionary tale, but what OpenGL did is they said that vectors were columns (departing from earlier Iris GL that clearly used row vectors), but that the matrices were stored in *column-major* order. This let them write all the documentation and books with column vector notation matching the one true way from physics and mathematics, but actually the data layout in memory was choice (A), matching what Iris GL and RenderMan and many other systems used.

The whole affair is a big mess. We borked this thing good in computer graphics 40 years ago.

Aside: Does anybody know off the top of their head which convention USD uses?

Meanwhile, in OpenEXR land, we have plans to split the Imath (matrix/vector, etc) stuff off into a separate project, thoroughly modernize it (C++11, constexpr, noexcept, header only, etc.), and then try to really push it as a commonly used vec/mat lingua franca and a lighter weight dependency for other projects to use. I think it would be great if, maybe at some upcoming OpenEXR TSC meeting, we could have reps from your project (and others) to really hash this out and, if we can't all settle on one matrix library (truth be told, I don't even care which one), at least try to agree for sure on data layouts so that we can go back and forth with simple pointer casting.

The reason I care so much is that there are totally plausible apps out there that need to simultaneously use OpenEXR, USD, MaterialX, OpenVDB, each one of which uses its own independent (essentially equivalent, yet different!) vector/matrix classes, and the poor app has to be doing all this pointless data shuffling between them. It's madness.

jstone-lucasfilm:
We’re looking forward to the modernized Imath project, which could become a drop-in replacement for the matrix implementations in MaterialX and other lightweight libraries.  In the meantime, I’ll plan to update MaterialX so that its transform methods assume row vectors, a choice which is relatively independent of its core functionality, and primarily affects its high-level rendering and viewer modules.

jstone-lucasfilm:
@lgritz Thanks for this suggestion, and the alignment of MaterialX matrices with row-vector conventions has been implemented as pull request #366.

One additional thought for future discussion:

As an alternative to basing OpenEXR, MaterialX, USD, and related libraries on a single set of vector/matrix classes, we might consider achieving interoperability by supporting implicit conversions between the vector/matrix representations across these libraries.  An example of this approach can be seen in the recent update to NanoGUI, which now contains its own (extremely minimal) vector/matrix library, with special constructors in the new classes to enable implicit conversions between vector/matrix instances in NanoGUI and Enoki:

https://github.com/mitsuba-renderer/nanogui/commit/6b82fb18ed599017028b042ebb5aa831d0f93454

lgritz:
CC @kdt3rd

I think our aim with Imath is not to force everybody to use one library for their internals (though it would be nice if we ended up with an Imath that was good enough for most packages to be happy to use it instead of making their own), but rather to try to use just one library for their **public interfaces**, with the right implicit conversions set up to have  zero-cost conversions between that and a different implementation they may prefer for their internals.

But maybe even that is not needed, if you have the right set of conversion conventions. I'm just thinking out loud here, but what if you have something like (excuse my not quite right C++ syntax):

```
// Common code that everybody uses:

// type trait for whether a class is canonically a V3f memory layout, supports
// operator[], members .{x,y,z}, and ctr from 3 floats
template<typename T> struct IsV3f : std::false_type {};
template<> struct IsV3f<Imath::V3f>  : std::true_type {}

#define IS_V3F(T) std::enable_if_t<IsV3f<T>(), int> = 0


// My library code:
// 

class MyV3f {
   ...
    // Construct from ANY other equivalent V3f
    template<typename OtherVec, IS_V3F(OtherVec)>
    MyV3f (const OtherVec& v) : x(v.x), y(v.y), z(v.z) {}

    // Convert TO any other V3f
    template<typename OtherVec, IS_V3F(OtherVec)>
    operator OtherVec () const { return OtherVec(this->x, this->y, this->z); }
};

// register it as a canonical V3f
template<> struct IsV3f<MyV3f> : std::true_type {}

// API with my library's native vec:
float my_func (const MyV3f& v);

// API that works with any other equivalent vec:
template<typename V, std::enable_if_t<IS_V3F(V)>
float my_func (const V& v) { return my_func(*(const MyV3f*)&v); }
```

Would that basically work? So you could have two libraries with their own vector classes, that don't know about each other at all (not even their names), but as long as they each "register" themselves with the IsV3f type trait, and provide templated conversion to/from any other V3f that promises the same mem layout... it'll all just work?

jstone-lucasfilm:
@lgritz From a high-level, that's the strategy I was thinking of, where type traits would be used to recognize compatibility between two math libraries that don't directly include each other

More specifically, I was imagining a slightly more generic implementation, inspired by the SFINAE constructors in NanoGUI and Enoki:

```
template <class T, size_t N>
class MyVector
{
    ...

    // Define these properties in any vector class to provide compatibility across the ecosystem.
    using VectorValue = T;
    static constexpr size_t VectorSize = N;

    // Convert from any compatible vector class.
    template <typename T,
              std::enable_if_t<T::VectorSize == VectorSize &&                         
              std::is_same_v<typename T::VectorValue, VectorValue>, int> = 0>
    MyVector(const T& otherVec)
    {
        for (size_t i = 0; i < VectorSize; i++)
        {
            _vec[i] = (VectorValue) otherVec[i];
        }
    }

    ...
}
```

This would be the minimal implementation required for a given vector class to be compatible with the entire ecosystem.  A vector class could then provide additional optimized paths as needed, where any particular method could provide a compatible variant that reinterprets the underlying bits of a const vector reference from another class (as in your example).

lgritz:
Cool, yeah, maybe this is the way forwrd.

PR #375, opened 2020-03-02
Syntax Example Updates

dbsmythe:
Updates to Syntax example files to better conform to 1.37 syntax and conventions

### DIFF
--- a/resources/Materials/Examples/Syntax/GeomInfos.mtlx
+++ b/resources/Materials/Examples/Syntax/GeomInfos.mtlx
@@ -4,22 +4,28 @@
   <collection name="col2" includegeom="/root/world/geo/g[4567]"/>
   <collection name="col3" includegeom="/root/world/geo/g[89]"/>
 
+  <geompropdef name="uv2" type="vector2" geomprop="texcoord" index="2"/>
+  <geompropdef name="objsize" type="float" unittype="distance" unit="meters"/>
+
   <geominfo name="gi1" collection="col1">
-    <geomprop name="diffalbedo" type="string" value="txt/albedo/albedo.1001.tif"/>
-    <geomprop name="dirtcolor" type="string" value="txt/dirt/dirt.1001.tif"/>
-    <geomprop name="areamask" type="string" value="txt/area/area.1001.tif"/>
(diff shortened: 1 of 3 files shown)


resources/Materials/Examples/Syntax/GeomInfos.mtlx
```diff
@@ -22,7 +22,7 @@
     <token name="noisemask" type="string" value="txt/noisemask/noisemask.1002.tif"/>
   </geominfo>
   <geominfo name="gi3" collection="col3">
-    <geomprop name="objsize" type="float" value="5.7" unit="foot"/>
+    <geomprop name="objsize" type="float" value="5.7" unittype="distance" unit="foot"/>
```
dbsmythe: This should not need a "unittype" specified, only the unit.  From the Spec (p.73): "A geomprop may also specify a unittype and unit if appropriate to indicate that the geometric property's value is in that unit; see the Units section above, although typically the <geompropdef> would define the unittype and unit, and a geomprop would only provide an overriding unit if the unit for its value differed from the geompropdef's defined default unit."

resources/Materials/Examples/Syntax/GeomInfos.mtlx
```diff
@@ -5,7 +5,7 @@
   <collection name="col3" includegeom="/root/world/geo/g[89]"/>
 
   <geompropdef name="uv2" type="vector2" geomprop="texcoord" index="2"/>
-  <geompropdef name="objsize" type="float" unittype="distance" unit="meters"/>
+  <geompropdef name="objsize" type="float" unittype="distance" unit="meter"/>
```
dbsmythe: The Spec still has its unit examples in the plural, e.g. "meters" and "feet".  I changed them to singular ("meter" and "foot") in the draft 1.38 doc, though that feels strange to me- I prefer the plural, but I guess that ship has sailed.  Note that the two units defined for "angle" (degrees and radians) are plural: it would be nice if these were consistent.

jstone-lucasfilm: Good points, and let’s amend this in an upcoming update to the data libraries.

PR #401, opened 2020-04-06
ADSK master to ILM master

bernardkwok:


CMakeLists.txt
```diff
@@ -96,51 +135,81 @@ add_subdirectory(source/MaterialXCore)
...
+if(MATERIALX_BUILD_TESTS)
+    add_subdirectory(source/MaterialXTest)
+endif()
+
 if(${CMAKE_VERSION} VERSION_GREATER "3.6.2")
     if(MATERIALX_BUILD_VIEWER)
         set_property(DIRECTORY PROPERTY VS_STARTUP_PROJECT MaterialXView)
     else()
-        set_property(DIRECTORY PROPERTY VS_STARTUP_PROJECT MaterialXTest)
+        if (NOT MATERIALX_DISABLE_BUILD_RENDER)
```
jstone-lucasfilm, line 189: Should this check instead be for MATERIALX_BUILD_TESTS?  I'm imagining that we'd want MaterialXTest to be the startup project in any solution where it's present without MaterialXView, but I may be missing a potential use case.

CMakeLists.txt
```diff
@@ -13,8 +13,16 @@ enable_testing()
 option(MATERIALX_BUILD_PYTHON "Build the MaterialX Python package from C++ bindings. Requires Python 2.7 or greater." OFF)
 option(MATERIALX_BUILD_VIEWER "Build the MaterialX Viewer." OFF)
 option(MATERIALX_BUILD_DOCS "Create HTML documentation using Doxygen. Requires that Doxygen be installed." OFF)
+
+option(MATERIALX_BUILD_GEN_GLSL "Build the GLSL shader generator back-end." ON)
+option(MATERIALX_BUILD_GEN_OSL "Build the OSL shader generator back-end." ON)
```
jstone-lucasfilm, line 18: @bernardkwok Are there situations in which you've found it valuable to disable shader generation in GLSL or OSL?  Because MaterialXGenGlsl and MaterialXGenOsl have no external library dependencies, I've been imagining that it's safe to build these modules in all configurations, but there may be use cases I'm not aware of.

source/MaterialXCore/Document.cpp
```diff
@@ -825,12 +872,168 @@ void Document::upgradeVersion()
...
+    if (majorVersion >= MATERIALX_MAJOR_VERSION &&
+        minorVersion >= MATERIALX_MINOR_VERSION)
+    {
+        setVersionString(makeVersionString(majorVersion, minorVersion));
+    }
+}
 
-    if (majorVersion == MATERIALX_MAJOR_VERSION &&
-        minorVersion == MATERIALX_MINOR_VERSION)
+bool Document::convertMaterialsToNodes(bool replaceNodes)
```
jstone-lucasfilm, line 896: This method might be better as a module-local function in the anonymous namespace, like updateChildSubclass above, since it's not part of the Document class interface that clients or subclasses would need to know about.

source/MaterialXView/Viewer.cpp
```diff
@@ -883,6 +901,8 @@ void Viewer::loadDocument(const mx::FilePath& filename, mx::DocumentPtr librarie
 {
     // Set up read options.
     mx::XmlReadOptions readOptions;
+    readOptions.desiredMajorVersion = 1;
+    readOptions.desiredMinorVersion = 38;
```
jstone-lucasfilm, line 905: We may not want to activate this feature yet for the MaterialX viewer, as it would cause documents saved from the viewer to become 1.38 documents.  For now, I think it's better for core MaterialX tools to maintain 1.37 standards when documents are loaded and saved, with only custom pipelines leveraging 1.38 features.

source/MaterialXCore/Document.cpp
```diff
@@ -315,6 +320,12 @@ void Document::upgradeVersion()
         }
         minorVersion = 23;
     }
+    if (majorVersion == desiredMajorVersion &&
+        minorVersion == desiredMinorVersion)
+    {
+        setVersionString(makeVersionString(majorVersion, minorVersion));
+        return;
+    }
```
jstone-lucasfilm, line 328: Since this feature really exists to allow clients to request _future_ MaterialX versions, maybe it would be better to explicitly add a "applyFutureUpgrades" flag instead of allowing the user to request arbitrary past versions.  This would remove a good deal of boilerplate code from the upgradeVersion() method, and would make it more clear which choice is relevant to the client.  By default, documents would upgrade to the current major and minor version, and the new flag would allow them to apply all future upgrades that are currently supported (e.g. converting material elements to nodes for 1.38).

CMakeLists.txt
bernardkwok: Hi @jstone-lucasfilm, This is part of make as many modules as possible options to accommodate web. The idea is both are never required in this situation and a new OpenGL ES generator will be added.

source/MaterialXView/Viewer.cpp
bernardkwok: Yes. That makes sense. I'll revert this.

source/MaterialXCore/Document.cpp
bernardkwok: Agreed. This does make the upgrade path simpler to manage as well, and it is not likely that a partial upgrade path would work due to other dependent changes.

CMakeLists.txt
jstone-lucasfilm: Thanks for the clarification, @bernardkwok, and that makes sense.

source/MaterialXCore/Document.cpp
```diff
@@ -826,10 +958,18 @@ void Document::upgradeVersion()
...
     }
 
-    if (majorVersion == MATERIALX_MAJOR_VERSION &&
-        minorVersion == MATERIALX_MINOR_VERSION)
+    // Apply latest updates targetted for the next release version
+    if (applyLatestUpdates)
+    {
+        // Convert material Elements to Nodes
+        convertMaterialsToNodes(getDocument());
+        minorVersion = 38;
```
jstone-lucasfilm, line 966: With this latest round of changes, we're effectively enabling support for material nodes in 1.37 documents, as an optional, forward-looking feature that pipelines and tools can enable.  Within that new context, maybe it makes more sense to leave the minor version unchanged when performing this upgrade, so that we're not removing the ability to distinguish between 1.37 and 1.38 documents in the future.

source/MaterialXCore/Interface.cpp
```diff
@@ -73,30 +74,63 @@ bool PortElement::validate(string* message) const
...
     {
-        validateRequire(connectedNode != nullptr, res, message, "Invalid port connection");
+        if (!connectedNode)
+        {
+            NodeGraphPtr nodeGraph = resolveRootNameReference<NodeGraph>(getNodeName());
+            if (!nodeGraph)
+            {
+                validateRequire(connectedNode != nullptr, res, message, "Invalid port connection");
+            }
+        }
```
jstone-lucasfilm, line 84: Could this proposed change perhaps be an error?  It looks like it's requiring that connectedNode be non-NULL after confirming that it's NULL.

source/MaterialXRenderGlsl/CMakeLists.txt
```diff
@@ -22,11 +18,18 @@ if(APPLE)
...
+        message("Error in building MaterialXRender: X11 was not found")
+    endif(NOT X11_FOUND)
     if(NOT X11_Xt_FOUND)
         message("Error in building MaterialXRender: Xt was not found")
-    endif()
+    endif(NOT X11_Xt_FOUND)
+    find_package(OpenGL REQUIRED)
+    if(NOT OPENGL_FOUND)
+        message("Error in building MaterialXRender: OpenGL was not found")
+    endif(NOT OPENGL_FOUND)
```
jstone-lucasfilm, line 32: This change may be an unintended reversion, since these calls were simplified in the master branch a while back.  In theory, we shouldn't need to check for `OPENGL_FOUND` or `X11_FOUND` after calling find_package on `OpenGL` and `X11` with the `REQUIRED` flag.

source/MaterialXFormat/Util.h
```diff
@@ -0,0 +1,52 @@
...
+
+namespace MaterialX
+{
+
+/// Removes the extension from the provided filename
+string removeExtension(const string& filename);
+
+/// Read the given file and return a string containing its contents; if the read is not
+/// successful, then the empty string is returned.
+string readFile(const string& filename);
```
jstone-lucasfilm, line 27: Now that these two functions are part of MaterialXFormat, we should probably change their arguments from std::string to FilePath for consistency with other functionality in the library.

source/MaterialXRenderGlsl/GlslRenderer.cpp
```diff
@@ -294,14 +317,28 @@ void GlslRenderer::save(const FilePath& filePath)
...
+
+    // Save using the handler.
+    if (!_imageHandler->saveImage(filePath, image, true))
+    {
+        errors.push_back("Failed to save to file:" + filePath.asString());
+        throw ExceptionShaderRenderError(errorType, errors);
+    }
+}
+
+ImagePtr GlslRenderer::saveImage()
```
jstone-lucasfilm, line 330: This new method is only called from one location (GlslRenderer::save), and it just wraps a call FrameBuffer::createColorImage with some additional error checking.  Maybe it would be better to remove this layer of indirection, and call FrameBuffer::createColorImage directly from GlslRenderer::save?

source/MaterialXCore/Document.cpp
bernardkwok: Isn't the opposite true? That if this is kept at 1.37 then we are unable to tell if the upgrade has occurred? i.e. it's not adding support but is a replacement. (I just found that it wasn't possible to actually remove material element class since it's required for upgrade purposes.).
BTW, The change for atan2() will be added in this forward-looking logic and will not be able to tell the difference unless we have a version bump.

When the next version comes out then this logic should be changed to be the same as for other version upgrades. (I need to add back in a check if this is 1.37 plus the feature enabling) before performing this. 

Currently we have dependents which are on 1.37 and we need to "downgrade" to convert material nodes back to material elements and hence check for a 1.38 version to revert back without first having to scan all the elements to find out if material nodes exist.

jstone-lucasfilm: This is a good question, and here is my current thinking:

For the upgrade from material elements to material nodes, the 1.37 codebase will support both options as equally valid (with this proposed changelist), so there is not a need to distinguish between them with versioning.  For the upgrade of the atan2 function signature, any single version of MaterialX is practically restricted to supporting one signature or the other, so that upcoming change will require a true version bump to distinguish between them.

When we ultimately update the latest MaterialX version from 1.37 to 1.38, we’ll remove support for material elements, and any such elements in 1.37 documents will be automatically upgraded to nodes, so there’s no situation (that I’m aware of) where our support for both options in the 1.37 codebase creates ambiguity.

Does that ring true to you, or are there cases I’m overlooking?

bernardkwok: Yes. Having the separation based on what is not backwards compatible (signature change) is a good minimum criteria for a version bump. I will remove any explicit version checking but keep "existence" testing (does a material node exist) to allow for proper conversion.

source/MaterialXCore/Interface.cpp
bernardkwok: Yes, This looks like a debugging change. The if (!connectedNode) should be removed.

source/MaterialXFormat/Util.h
bernardkwok: Yes, I think these we're written before the switch to FilePath and should be updated now.

source/MaterialXRenderGlsl/CMakeLists.txt
bernardkwok: Yes. This looks like a synchronization issue on our side. I'll revert this part.

source/MaterialXRenderGlsl/GlslRenderer.cpp
bernardkwok: I think this is a synchronization problem so I'll clean this up.

Issue #403, opened 2020-04-09
[Lucasfilm] Multiple Scattering Compensation In Generated GLSL


jstone-lucasfilm:
This task represents additional work to compensate for multiple scattering in MaterialX-generated GLSL, building upon recent improvements to energy conservation (#402), and using the following references as a guide:

https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf
https://blog.selfshadow.com/publications/turquin/ms_comp_final.pdf
https://blog.selfshadow.com/2018/05/13/multi-faceted-part-1/

jstone-lucasfilm:
This has been addressed in pull request #407

PR #404, opened 2020-04-13
Initial support for wedge rendering

jstone-lucasfilm:
- Add initial support for "wedge" rendering, where a single property is ramped between two values across successive screenshots, and the resulting image set is merged into a strip.  The new feature is activated by pressing the 'W' key in the MaterialX viewer, and settings for the render are hardcoded for now.
- Add hotkeys for adjusting camera zoom (+/-).
- Add utilities FilePath::addExtension and MaterialX::stringEndsWith.
- Refactor Viewer::drawContents for greater flexibility.
- Update changelog for recent work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,13 @@
 ## [1.37.1] - Development
 
 ### Added
+- Added initial support for wedge rendering in the viewer.
+- Added utility methods Backdrop\:\:setContainsElements and Backdrop\:\:getContainsElements.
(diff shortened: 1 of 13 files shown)


jstone-lucasfilm:
Here's an example of a wedge render, where the roughness property is interpolated between 0 and 1.  This feature should be useful in upcoming physically-based shading tasks such as #403:

MaterialXView_ChromeRoughnessWedge

Issue #359, opened 2020-02-07
Importing MaterialX.PyMaterialXGenOsl or MaterialX.PyMaterialXGenGlsl errors.


alister-chowdhury:
Attempting to import either of these modules results in an error to the tune of:
```python
>>> import MaterialX.PyMaterialXGenOsl
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: PyMaterialXGenShader.cpython-37m-x86_64-linux-gnu.so.1: cannot open shared object file: No such file or directory
```
In both python 2 and python 3.
It appears `MaterialX.PyMaterialXGenShader` needs to be imported first.

```python
>>> import MaterialX.PyMaterialXGenGlsl
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: PyMaterialXGenShader.so.1: cannot open shared object file: No such file or directory
>>>
>>> import MaterialX.PyMaterialXGenShader
>>> import MaterialX.PyMaterialXGenGlsl
>>> import MaterialX.PyMaterialXGenOsl
```

I'm not sure if these modules, are intended to act as an implementation detail (as none of these modules are imported in the `main.py` / `__init__.py` files).

Assuming they are expected to be used explicitly, a simple (and inelegant) solution, would be to simply do an import (in the working order) in `main.py`.

jstone-lucasfilm:
Hi Alister,

The intent is for these modules to be imported explicitly as needed, since applications are often interested in only a single shading language for generation and rendering.  I agree, though, that it's unexpected for PyMaterialXGenShader to be required before PyMaterialXGenGlsl or PyMaterialXGenOsl may be used.  My sense is that PyMaterialXGenShader should be imported automatically when either of the language-specific shader generators are imported.

Adding @niklasharrysson and @bernardkwok to the thread for their thoughts.

jstone-lucasfilm:
Thanks to @alister-chowdhury for bringing this up, and see pull request #411 for further discussion.  For now, I'm closing out the original issue, and we can add Python wrappers for common import sequences as needed in the future.

Issue #355, opened 2020-02-06
Ability to use a cmake installed version of pybind11.


alister-chowdhury:
Hello,

We'd quite like to be able to provide our own version of pybind11 (which is partially covered here #146).

Sadly, if pybind11 is made available by using a CMake install, they aren't explicit put in a `/tools` directory but instead will be placed in `/share/cmake/pybind11` etc.

as expected by:
[`list(APPEND CMAKE_MODULE_PATH "${MATERIALX_PYTHON_PYBIND11_DIR}/tools")`](https://github.com/materialx/MaterialX/blob/master/source/PyMaterialX/CMakeLists.txt#L26)

Using `find_package(pybind11)` is a pretty standard way of accessing the modules.

Would it be acceptable to change the logic such that it first searches for pybind11, before appending the path and including pybind11Tools?

jstone-lucasfilm:
This sounds like a great proposal for a change, Alister.  Would you be open to writing a pull request for this, as you have a working CMake install of PyBind11 to test against?  You would need to sign up as a contributor to MaterialX, but the approval process should be quick.

https://github.com/materialx/MaterialX/blob/master/CONTRIBUTING.md

alister-chowdhury:
Going to try and get the higher ups to approve it all, and will get back to you :)

jstone-lucasfilm:
Thanks to @alister-chowdhury for adding this support in #411, and I'm closing out the original issue.

PR #411, opened 2020-04-29
Python related fixes for #355 and #359

alister-chowdhury:
Python related fixes for #355 and #359

The logic introduced allows pybind11 to try and use CMake if no explicit path was provided.

Along with that, the python import errors were caused by the respective modules not importing their required dependencies and were instead linked against them (this actually seems to fix some of the FLTO related errors we were typically having).

I also found that STRIDE_2D, STRIDE_3D and DEFAULT_STRIDE would cause an error when importing `PyMaterialXRender` to the tune of:
`/build/inst/python/MaterialX/PyMaterialXRender.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZN9MaterialX10MeshStream9STRIDE_2DE`

The solution to this is to make them a read only property rather than linking to their pointers.


alister-chowdhury:
Apologies for the updates, we currently don't have a build system that deals with both py2 and py3 installs living alongside each other.

alister-chowdhury:
Em

```python
======================================================================
FAIL: test_ReadXml (__main__.TestMaterialX)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "main.py", line 465, in test_ReadXml
    self.assertTrue(doc.validate()[0], filename + ' is not valid.')
AssertionError: False is not true : CustomNode.mtlx is not valid.

```

I'm not so sure what could be causing this?
I didn't actually change any of the logic involved in reading documents (that I'm aware of).

jstone-lucasfilm:
Thanks for putting this changelist together, @alister-chowdhury!  If the cause of the MacOS validation error isn't clear, we can take a closer look on our side.

jstone-lucasfilm:
@alister-chowdhury, how does this latest version look to you?  I've addressed the cross-platform compilation issues, but before merging I'd like to validate that it fixes the original issues you ran into.

alister-chowdhury:
@jstone-lucasfilm 
So pybind11 is still found via cmake, however #359 is still a problem.

And importing `PyMaterialXRender` errors from `undefined symbol: _ZN9MaterialX10MeshStream9STRIDE_2DE` on import (because of the reverting of the changes in PyMesh.cpp).

jstone-lucasfilm:
@alister-chowdhury In my local environment, Windows 10 with Python 3.8, I'm now able to directly call `import MaterialX.PyMaterialXGenGlsl`, `import MaterialX.PyMaterialXGenOsl`, and `import MaterialX.PyMaterialXRender` without warnings or errors.

Would you be able to dig further into the errors that you're seeing locally, to see if there's a sequence of import calls that eliminates them?  My expectation was that the new calls to `py::module::import` would be sufficient to address them, but that doesn't appear to be the case.

jstone-lucasfilm:
@alister-chowdhury In case it helps to track down the issue you're seeing, I've verified that the master branch of MaterialX emits a Python error when I call `import MaterialX.PyMaterialXGenGlsl` without preparing it first with `import MaterialX.PyMaterialXGenShader`.  Running with this pull request, I get no errors when calling `import MaterialX.PyMaterialXGenGlsl` in a clean Python 3 session.

alister-chowdhury:
@jstone-lucasfilm I can confirm on linux / py3 I still get the error

PyMaterialXCore, PyMaterialXFormat and PyMaterialXGenShader import without complaint.

PyMaterialXGenGlsl, PyMaterialXGenOsl and PyMaterialXRender:
`ImportError: PyMaterialXGenShader.cpython-37m-x86_64-linux-gnu.so.1: cannot open shared object file: No such file or directory`

PyMaterialXRenderGlsl and PyMaterialXRenderOsl
`ImportError: PyMaterialXRender.cpython-37m-x86_64-linux-gnu.so.1: cannot open shared object file: No such file or directory`

When I order the imports as:
```python
import MaterialX.PyMaterialXCore
import MaterialX.PyMaterialXFormat
import MaterialX.PyMaterialXGenShader
import MaterialX.PyMaterialXGenGlsl
import MaterialX.PyMaterialXGenOsl
import MaterialX.PyMaterialXRender
import MaterialX.PyMaterialXRenderGlsl
import MaterialX.PyMaterialXRenderOsl
```
The only error is:
`ImportError: .../build/inst/python/MaterialX/PyMaterialXRender.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZN9MaterialX10MeshStream9STRIDE_2DE`
(From PyMaterialXRender).

I believe the "cannot open shared object file...." is coming from them being linked, and python/MaterialX/... is not going to be in LD_LIBRARY_PATH, so it errors (unless said .so has already been explicitly imported).

A especially inelegant solution would be to import them in the __init__.
I am not familiar with developing for OSX platforms, but I take it that they simply must be linked or strange problems propagate?

jstone-lucasfilm:
Thanks for those additional details, @alister-chowdhury, and those are very helpful in clarifying the picture.

It sounds like the `py::module::import` mechanism doesn't give us fully cross-platform behavior, so I would recommend that we remove these calls and simply require a manual import order for now.  We can follow up with the PyBind11 developers in the future for a more idiomatic way to accomplish this type of Python module inclusion.

Similarly, it sounds like it's not safe to bind constants such as STRIDE_2D, STRIDE_3D, and DEFAULT_STRIDE which are defined in shared headers.  I believe we can safely remove these Python bindings completely, as they're not likely to be required in Python scripts.

alister-chowdhury:
@jstone-lucasfilm would you like me to add imports ordered inside the \_\_init\_\_?

In regards to the STRIDE stuff, I originally opted to make it a read only property, so it acts as a const variable (from pythons point of view).

jstone-lucasfilm:
> Would you like me to add imports ordered inside the __init__?

Since each application will have a different favored shading language (GLSL, OSL, and soon MDL), it's probably overkill to have the core __init__.py file bring all shader generators into the Python environment.  One option for the future would be to have a wrapper for each supported language, but for now I think it's most flexible to make the client responsible for import order, and we can create high-level wrappers for the most common inclusion patterns as they become clear.

> In regards to the STRIDE stuff, I originally opted to make it a read only property, so it acts as a const variable (from pythons point of view).

That's an elegant solution, if access to those constants turns out to be useful.  But it's my sense that they were added just for the sake of completeness, and wouldn't be used in a practical Python script.  So for now, I'm thinking that we can take the simplest path and simply omit those Python bindings.

alister-chowdhury:
I think it could be possible to prevent import errors for most of the time, by simply importing PyMaterialXGenShader and PyMaterialXRender.
However, hopefully our build setup is an outlier and most people don't suffer from this :)

> But it's my sense that they were added just for the sake of completeness, and wouldn't be used in a practical Python script. So for now, I'm thinking that we can take the simplest path and simply omit those Python bindings.

I expect so too, I'd imagine when you reach that level, you're probably going to go into C++.

PR #417, opened 2020-05-03
Improvements to tangent frame generation

jstone-lucasfilm:
- Detect and skip duplicate vertices in TinyObjLoader::load, providing more accurate adjacency data for tangent frame generation.
- On hardware shading targets, generate bitangents in the vertex shader rather than storing them as an independent data stream.
- Refactor Mesh::generateTangents for improved robustness and clarity.

Supporting changes:

- Add a VectorN less-than operator, allowing vectors to be used as map keys.
- Add a MeshStream::getElement method to simplify geometry processing logic.


jstone-lucasfilm:
Here's a comparison of roughness wedges for an anisotropic metal material, before and after these changes.  Notice the slight vertical bands in the original version, which are fixed by the improved tangent frame:

RoughnessWedge_BrushedMetal_Original

RoughnessWedge_BrushedMetal_ImprovedTangentFrame

PR #427, opened 2020-05-15
Github actions support

ScatteredRay:
Some initial support for continuous integration support with Github Actions.

Has a builder for linux, macos, and windows with vs2017 and 2019.

https://github.com/ScatteredRay/MaterialX/actions/runs/105845563


jstone-lucasfilm:
Much appreciated, @ScatteredRay!  We've been wanting to leverage GitHub Actions for some time, and this looks like an important step forward.  I'd like @bernardkwok to take a closer look at the integration of render tests, to make sure this new work is aligned with Autodesk usage.

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+      working-directory: build
+    - name: build
+      run: cmake --build . --target install --config ${{env.config}}
+      working-directory: build
+    - uses: actions/upload-artifact@v2
+      with:
+        name: MaterialX_${{matrix.name}}
+        path: build/installed/
+    - name: unit tests
+      run: ctest -VV --output-on-failure --build-config ${{env.config}}
```
bernardkwok, line 61: Just a small difference I noted here that we run parallel tests -j # in case you want to add this in. Not sure if it makes a difference but shadergen and render tests we've found can time out waiting.

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+      if: ${{matrix.test_render == 'ON'}}
+      with:
+        name: MaterialX_${{matrix.name}}_test_images
+        path: build/source/MaterialXTest/**/*.png
+    - name: main tests
+      run: cmake -E chdir ../python/MaterialXTest ${{matrix.python}} main.py
+      working-directory: build
+    - name: genshader tests
+      run: cmake -E chdir ../python/MaterialXTest ${{matrix.python}} genshader.py
+      working-directory: build
```
bernardkwok, line 89: I assume all this is possible, but want to ask since I'm very familiar with this CI setup. 
We have the additional post set of zipping up builds and publishing back to github. I think this is on every build currently here but we do it based on tags to create release.

This and the previous comments are shown in our .travis file
https://github.com/autodesk-forks/MaterialX/blob/adsk_contrib/dev/.travis.yml

All can be looked at later when we try this out. Mostly a heads-up of our setup. Thanks.

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
+name: Build CI
+
+on:
+  push:
+    branches: [ master ]
```
bernardkwok, line 5: For a push vs a PR it would be nice if a different set of builds could be run to save time as pushes I assume can be less stringent / comprehensive.

PR #428, opened 2020-05-16
Fixes for tests_to_html

ScatteredRay:
* Fix for running under Python3
* Fix for building html even when there is a mismatch of items

### DIFF
--- a/python/MaterialXTest/tests_to_html.py
+++ b/python/MaterialXTest/tests_to_html.py
@@ -20,7 +20,7 @@ def createDiff(image1Path, image2Path, imageDiffPath):
         diff = ImageChops.difference(image1, image2)
         diff.save(imageDiffPath)
     except Exception:
-        print "Failed to create image diff between: " + image1Path + ", " + image2Path
+        print ("Failed to create image diff between: " + image1Path + ", " + image2Path)
 
 def main(args=None):
 
--- a/python/MaterialXTest/tests_to_html.py
+++ b/python/MaterialXTest/tests_to_html.py
@@ -4,6 +4,7 @@ import sys
 import os
 import datetime
 import argparse
+import itertools
 
 try:
     # Use pip to install Pillow and Image to enable image diffs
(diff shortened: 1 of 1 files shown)


python/MaterialXTest/tests_to_html.py
```diff
@@ -62,13 +63,12 @@ def main(args=None):
...
             print ("OSL files: " + str(oslFiles))
-            continue
-        elif len(glslFiles) > 0 and len(oslFiles) > 0:
+        if len(glslFiles) > 0 or len(oslFiles) > 0:
             fh.write("<h2>" + subdir + ":</h2><br>\n")
             fh.write("<table>\n")
-            for glslFile, oslFile in zip(glslFiles, oslFiles):
-                fullGlslPath = os.path.join(subdir, glslFile)
-                fullOslPath = os.path.join(subdir, oslFile)
+            for glslFile, oslFile in itertools.zip_longest(glslFiles, oslFiles):
```
ScatteredRay, line 69: This should probably be improved further to match based on file name.

jstone-lucasfilm:
This proposal looks good to me, but I'd like @bernardkwok at Autodesk to take a closer look before proceeding.

python/MaterialXTest/tests_to_html.py
bernardkwok: Yes it would be good to perform a folder scan like this:
```
def walk_results(inputRoot, nameStr, outputList, fullfilePaths):
    for subdir, _, files in os.walk(inputRoot):
        for curFile in files:
          if curFile.endswith(".png") and not "diff.png" in curFile:
            if nameStr in curFile:
                outputList.append(curFile)
                fullfilePaths[curFile] = os.path.join(subdir, curFile)
```
and add in explicit reference location and comparison location(s) e.g.
```
   parser.add_argument('-i', '--inputdir', dest='inputdir', action='store', help='Input directory', default=".")
    parser.add_argument('-r', '--refdir', dest='groundrefdir', action='store', help='Ground Truth directory', default=".")
```
but this can be looked at afterwards as we do want to add additional languages and this script is mostly "quick-and-dirty" way based on 2 hard-coded folders.

PR #441, opened 2020-06-01
Merge from ADSK to ILM

bernardkwok:
* MDL shadergen
* Misc other modifications.


libraries/bxdf/standard_surface.mtlx
```diff
@@ -124,7 +124,7 @@
     <rotate3d name="tangent_rotate" type="vector3">
       <input name="in" type="vector3" interfacename="tangent" />
       <input name="amount" type="float" nodename="tangent_rotate_degree" />
-      <parameter name="axis" type="vector3" interfacename="normal"/>
+      <input name="axis" type="vector3" interfacename="normal"/>
```
jstone-lucasfilm, line 127: @bernardkwok Wouldn't this signature change to "rotate3d" need to wait on the next major version upgrade to MaterialX?  That would allow us to robustly handle the original signature in 1.37 and the new signature in 1.38.

PR #445, opened 2020-06-30
Load standard libraries regardless of build folder location

mjyip-lucasfilm:
This changelist fixes a dependency on the build folder being located within the repository root folder.


jstone-lucasfilm:
Looks good to me!

PR #448, opened 2020-07-19
Python Texture Bake and Multiple Udims Texture Bake Support

mjyip-lucasfilm:


source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -4,187 +4,342 @@
 //
 
 #include <MaterialXRenderGlsl/TextureBaker.h>
+#include <MaterialXGenShader/Shader.h>
+#include <MaterialXGenShader/Util.h>
+#include <MaterialXRender/Util.h>
```
jstone-lucasfilm, line 9: For consistency with the codebase, these includes should be organized as:

```
#include <MaterialXRenderGlsl/TextureBaker.h> [the matching header for this source file]

#include <MaterialXRenderGlsl/GlslProgram.h> [other MaterialX dependencies order by their proximity to the MaterialXRenderGlsl library, then alphabetical within each library]
#include <MaterialXRender/Shader.h>
#include <MaterialXGenShader/Shader.h>
#include <MaterialXGenShader/Util.h>

#include <iostream> [standard library dependencies]
```

source/PyMaterialX/PyMaterialXRenderGlsl/PyTextureBaker.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+namespace py = pybind11;
+namespace mx = MaterialX;
+
+void bindPyTextureBaker(py::module& mod)
+{
+    py::class_<mx::TextureBaker, mx::GlslRenderer, mx::TextureBakerPtr>(mod, "TextureBaker")
+        .def_static("create", &mx::TextureBaker::create)
+        .def_static("bakeAndSave", &mx::TextureBaker::bakeAndSave);
+}
```
jstone-lucasfilm, line 19: Minor: There should be a final newline here.

python/Scripts/baketextures.py
```diff
@@ -0,0 +1,58 @@
...
+        if arg in ['-h', '-help', '--help']:
+            usage()
+            sys.exit(0)
+        elif arg in ['-hdr', '-HDR']:
+            hdr = True
+        elif (arg in ['-t', 'texRes', 'texres'] and i + 1 < len(sys.argv)):
+            texres = int(sys.argv[i + 1])
+            i += 1
+        else:
+            filename = arg
```
tmosley-lucasfilm, line 39: We should use the built-in `argparse` module for this. I believe for this inputs, that's would be close to this (though you don't need all my comments):

```
import argparse

parser = argparse.ArgumentParser(description="Bake out MaterialX sharderref inputs into textures.")

# Here's a boolean flag
parser.add_argument("--hdr", "--HDR", dest="hdr", type="store_true", help="Save images to hdr format.")

# This one takes the user input and converts it to an int, with a default value.
# `dest` in optional (default to flag name), but good to be explicit when you have multiple flag names
parser.add_argument("-t", "--texRes", "--texres", dest="tex_res", type=int, default=1024, help="Resolution at which to save textures.")

# If `parse_args` is called with no input, it uses `sys.argv`. If a help flag is passed, this automatically generates and prints usage
opts = parser.parse_args()

# Values are stored as attributes on this object.
# opts.hdr
# opts.tex_res
```
https://pymotw.com/3/argparse/

python/Scripts/baketextures.py
```diff
@@ -0,0 +1,58 @@
...
+Generate the baked mtlx file and baked textures programmatically.
+'''
+
+import sys, os, string
+import glfw
+import struct
+import MaterialX as mx
+from MaterialX import PyMaterialXRender as r
+from MaterialX import PyMaterialXRenderGlsl as rg
+from MaterialX import PyMaterialXGenShader as gs
```
tmosley-lucasfilm, line 12: You should keep these expanded to improve readability. `mx` might be ok, but even there `mtlx` would probably be better.

`from MaterialX import PyMaterialXRenderGlsl as mtlx_render`

python/Scripts/baketextures.py
```diff
@@ -0,0 +1,58 @@
...
+
+    doc = mx.createDocument()
+
+    try:
+        mx.readFromXmlFile(doc, filename)
+    except mx.ExceptionFileMissing as err:
+        print(err)
+
+    stdlib = mx.createDocument()
+    libraryFolder = list([mx.FilePath("../../libraries/stdlib"), mx.FilePath("../../libraries/pbrlib"), mx.FilePath("../../libraries/bxdf"), mx.FilePath("../../libraries/pbrlib/genglsl")])
```
tmosley-lucasfilm, line 49: Just some style notes here:

1. Try to keep line lengths under 80 characters. (PEP8)
2. Local variables should use `snake_case` instead of `camelCase` (PEP8)
3. The `list` call here is redundant. 

```
library_folder = [
    mx.FilePath("../../libraries/stdlib"),
    mx.FilePath("../../libraries/pbrlib"),
    ... ]
```

python/Scripts/baketextures.py
```diff
@@ -0,0 +1,58 @@
...
+            i += 1
+        else:
+            filename = arg
+
+    doc = mx.createDocument()
+
+    try:
+        mx.readFromXmlFile(doc, filename)
+    except mx.ExceptionFileMissing as err:
+        print(err)
```
tmosley-lucasfilm, line 46: To use a logger instead of print, up at the top you can do this:

```
import logging
logger = logging.getLogger('baketextures.py')
```

Then, anywhere else in this file just call:
```
logger.info("My info message!")
logger.debug("My debug message!")
```

python/Scripts/baketextures.py
```diff
@@ -0,0 +1,58 @@
...
+    stdlib = mx.createDocument()
+    libraryFolder = list([mx.FilePath("../../libraries/stdlib"), mx.FilePath("../../libraries/pbrlib"), mx.FilePath("../../libraries/bxdf"), mx.FilePath("../../libraries/pbrlib/genglsl")])
+    searchPath = mx.FileSearchPath(os.getcwd())
+    mx.loadLibraries(libraryFolder, searchPath, stdlib, set(), None)
+    doc.importLibrary(stdlib)
+
+    rg.TextureBaker.bakeAndSave(doc, filename, hdr, texres)
+
+if __name__ == '__main__':
+    glfw.init()
```
tmosley-lucasfilm, line 57: If you switch from `print` to a logger, also include in this block before `main()`:

```
# If this ever got bigger, you could add flags to easily control how verbose the log output is.
logging.basic_config(level=logging.INFO)
```

python/Scripts/baketextures.py
mjyip-lucasfilm: Amazing thank you!

python/Scripts/baketextures.py
jstone-lucasfilm: That's a good suggestion, and we might consider using "mx_render", "mx_render_glsl", and "mx_gen_shader" for the local names in this script.

mjyip-lucasfilm: You exactly read my mind! Thanks all :)

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -180,11 +197,149 @@ void TextureBaker::writeBakedDocument(NodePtr shader, const FilePath& filename)
     writeToXmlFile(bakedTextureDoc, filename);
 }
 
-FilePath TextureBaker::generateTextureFilename(OutputPtr output)
+// Helper function to generate mtlx filename
```
jstone-lucasfilm, line 200: For consistency, I would keep the high-level descriptions of these functions in the header, so that they're visible to clients without access to the source.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -16,14 +21,14 @@ TextureBaker::TextureBaker(unsigned int width, unsigned int height, Image::BaseT
 {
     // Assign a default extension for texture baking.
     _extension = (baseType == Image::BaseType::UINT8) ?
-                 ImageLoader::PNG_EXTENSION :
```
jstone-lucasfilm, line 19: Unintended whitespace change here?

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -180,11 +197,149 @@ void TextureBaker::writeBakedDocument(NodePtr shader, const FilePath& filename)
...
+            ShaderRefPtr shaderRef = elem->asA<ShaderRef>();
+            TypedElementPtr materialRef = (shaderRef ? shaderRef->getParent()->asA<TypedElement>() : nullptr);
+            materials.push_back(materialRef);
+        }
+        renderablePaths.push_back(renderableElem->getNamePath());
+    }
+    return renderablePaths;
+}
+
+// Helper function to generate texture filenames
```
jstone-lucasfilm, line 277: Since this is a purely module-local function, I would move it to an anonymous namespace at the top of the file.  Here's an example of this pattern in MaterialXCore:

https://github.com/materialx/MaterialX/blob/master/source/MaterialXCore/Document.cpp#L18

source/MaterialXRenderGlsl/TextureBaker.h
```diff
@@ -42,25 +44,28 @@ class TextureBaker : public GlslRenderer
     }
 
     /// Bake textures for all graph inputs of the given shader reference.
-    void bakeShaderInputs(ShaderRefPtr shaderRef, GenContext& context, const FilePath& outputFolder);
+    void bakeShaderInputs(ConstShaderRefPtr rf, GenContext& context, const FilePath& outputFolder, const string udim = "");
```
jstone-lucasfilm, line 47: Maybe just keep the name of this first argument as shaderRef for consistency with the rest?

python/Scripts/baketextures.py
```diff
@@ -0,0 +1,43 @@
...
+        mx.FilePath("../../libraries/pbrlib/genglsl")
+        ]
+    searchPath = mx.FileSearchPath(os.getcwd())
+    mx.loadLibraries(library_folder, searchPath, stdlib, set(), None)
+    doc.importLibrary(stdlib)
+
+    mx_render_glsl.TextureBaker.bakeAndSave(doc, opts.filename, opts.hdr, opts.tex_res)
+
+if __name__ == '__main__':
+    main()
```
jstone-lucasfilm, line 43: Should have a final newline here for consistency

source/MaterialXView/Viewer.cpp
```diff
@@ -1621,31 +1672,60 @@ void Viewer::bakeTextures()
...
+
+                try
+                {
+                    mx::TextureBakerPtr baker = mx::TextureBaker::create(_bakeTextureRes, _bakeTextureRes, baseType);
+                    baker->setImageHandler(imageHandler);
+                    baker->bakeShaderInputs(shaderRef, _genContext, _bakeFilename.getParentPath(), mat->getUdim());
+                    std::vector<std::string> udimSet = udimSetValue->asA<std::vector<std::string>>();
+                    if (mat->getUdim() == udimSet.back())
+                    {
+                        baker->writeBakedDocument(shaderRef, _bakeFilename, udimSetValue);
```
jstone-lucasfilm, line 1706: Would it be possible for the viewer to leverage bakeAllShaders as well, or are there different requirements for this version of the code?

mjyip-lucasfilm: The Viewer chooses the selected material and leverages the existing imagehandler presets in the viewer. BakeAllShaders just uses the explicitly stated material in the document and creates a basic default imagehandler from scratch. I would say BakeAllShaders is a little more naive compared to what is done in Viewer.cpp.

jstone-lucasfilm: Got it, and that all makes sense.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -180,11 +281,69 @@ void TextureBaker::writeBakedDocument(NodePtr shader, const FilePath& filename)
...
+    for (const auto& renderablePath : renderablePaths)
+    {
+        ElementPtr elem = doc->getDescendant(renderablePath);
+        TypedElementPtr typedElem = elem ? elem->asA<TypedElement>() : nullptr;
+        if (!typedElem)
+        {
+            continue;
+        }
+
+        ShaderRefPtr sr = elem->asA<ShaderRef>();
```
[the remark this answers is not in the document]
mjyip-lucasfilm: Added a check on line 306 to be extra sure.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -180,11 +282,69 @@ void TextureBaker::writeBakedDocument(NodePtr shader, const FilePath& filename)
...
+    FileSearchPath filename = initFileSearchPath();
+    GenContext genContext = initGenContext();
+    genContext.registerSourceCodeSearchPath(filename);
+    StringVec renderablePaths = getRenderablePaths(doc);
+    StringResolverPtr resolver = StringResolver::create();
+    for (const auto& renderablePath : renderablePaths)
+    {
+        ElementPtr elem = doc->getDescendant(renderablePath);
+        TypedElementPtr typedElem = elem ? elem->asA<TypedElement>() : nullptr;
+        if (!typedElem || elem->getCategory() != "shaderref")
```
jstone-lucasfilm, line 306: There's a more compact way to accomplish this that's more commonly used in MaterialX:

```
ShaderRefPtr sr = typedElem ? typedElem ->asA<ShaderRef>() : nullptr;
if (!sr)
{
    continue;
}
```

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -5,11 +5,100 @@
...
+
+#include <iostream>
 
 namespace MaterialX
 {
 
+namespace
+{
+
+// Helper function to initialize file search pat
```
jstone-lucasfilm, line 21: Should be "search path"

source/MaterialXView/Viewer.cpp
```diff
@@ -830,6 +845,42 @@ void Viewer::createAdvancedSettings(Widget* parent)
...
     }
+
+    ng::Label* textureLabel = new ng::Label(advancedPopup, "Texture Baking Options");
+    textureLabel->setFontSize(20);
+    textureLabel->setFont("sans-bold");
+
+    ng::CheckBox* bakeTexturesBox = new ng::CheckBox(advancedPopup, "Bake Textures");
+    bakeTexturesBox->setChecked(_bakeTextures);
+    bakeTexturesBox->setCallback([this](bool enable)
+        {
```
jstone-lucasfilm, line 856: For consistency, the braces for these lambda closures should be aligned with the function call, as in:

```
bakeTexturesBox->setCallback([this](bool enable)
{
    _bakeTextures = enable;
});
```

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -180,11 +282,69 @@ void TextureBaker::writeBakedDocument(NodePtr shader, const FilePath& filename)
...
+        FilePath out = generateOutMtlxFilename(file, sr->getName());
+        StbImageLoaderPtr stbimg = StbImageLoader::create();
+        ImageHandlerPtr imageHandler = GLTextureHandler::create(stbimg);
+
+        // Check for any udim set.
+        ValuePtr udimSetValue = doc->getGeomPropValue("udimset");
+
+        if (udimSetValue && udimSetValue->isA<StringVec>())
+        {
+            for (const string& udim : udimSetValue->asA<StringVec>())
```
bernardkwok, line 320: To avoid code repeat, a string array could be kept of all udim strings. If there are none then a single entry with a EMPTY_STRING could be used to pass as the udim.

source/MaterialXRenderGlsl/TextureBaker.h
```diff
@@ -42,25 +44,28 @@ class TextureBaker : public GlslRenderer
...
+    /// Bake material and its inputs to textures 
+    static void bakeAllShaders(DocumentPtr& doc, string file, bool HDR = false, int textureRes = 1024);
 
   protected:
     TextureBaker(unsigned int width, unsigned int height, Image::BaseType baseType);
 
-    // Generate a texture filename for the given graph output.
-    FilePath generateTextureFilename(OutputPtr output);
+    /// Generate a texture filename for the given graph output.
+    FilePath generateTextureFilename(OutputPtr output, const string& srName = "", const string& udim = "");
```
bernardkwok, line 68: Nitpick. EMPTY_STRING can be used here.

source/MaterialXView/Viewer.cpp
```diff
@@ -37,6 +38,10 @@ const int ALBEDO_TABLE_SIZE = 64;
 const int IRRADIANCE_MAP_WIDTH = 256;
 const int IRRADIANCE_MAP_HEIGHT = 128;
 
+const int MIN_TEXTURE_RES = 256;
+const int MAX_TEXTURE_RES = 4096;
```
bernardkwok, line 42: For later: To be more flexible the maximum here is for OpenGL. It could be better to get this from the baker class itself.
Then for instance a software baker can set a different max resolution.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -5,11 +5,100 @@
...
+#include <iostream>
 
 namespace MaterialX
 {
 
+namespace
+{
+
+// Helper function to initialize file search pat
+FileSearchPath initFileSearchPath()
```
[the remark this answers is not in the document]
jstone-lucasfilm: This is a good point, though I'd consider this a change that we need to make more broadly to our default search paths (e.g. in MaterialXView), so it's probably beyond the scope of the current changelist.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -180,11 +282,69 @@ void TextureBaker::writeBakedDocument(NodePtr shader, const FilePath& filename)
     writeToXmlFile(bakedTextureDoc, filename);
 }
 
-FilePath TextureBaker::generateTextureFilename(OutputPtr output)
+FilePath TextureBaker::generateTextureFilename(OutputPtr output, const string& srName, const string& udim)
```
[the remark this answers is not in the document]
mjyip-lucasfilm: I guess the reason I included the udim string was to be consistent about how I handle different udims. (e.g. specular_roughness_1001.hdr vs. specular_roughness_1002.hdr). Would you recommend that I just create a catch-all suffix term that I pass in as an argument?

source/MaterialXRenderGlsl/TextureBaker.cpp
mjyip-lucasfilm: Good idea! Thanks!

source/MaterialXRenderGlsl/TextureBaker.cpp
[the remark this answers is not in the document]
jstone-lucasfilm: Some form of UDIM distinction is needed here, and although it could be expressed as an integer instead of a string, the approach you've taken here seems fine.

PR #451, opened 2020-07-28
Normal map texture baking feature

mjyip-lucasfilm:


source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -124,6 +144,13 @@ void TextureBaker::bakeShaderInputs(ConstShaderRefPtr shaderRef, GenContext& con
         if (output)
         {
             FilePath filename = FilePath(outputFolder / generateTextureFilename(output, shaderRef->getName(), udim));
+
+            if (bindInput->getName() == "normal" && containsNormalMapNode(shaderRef))
```
jstone-lucasfilm, line 148: This is really close to the mark, though it would be even better if we could remove the dependency on the name "normal".  Would it be possible to scan each bindinput, regardless of name, and check the corresponding output for a connected <normalmap> node?

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -162,9 +194,6 @@ void TextureBaker::bakeGraphOutput(OutputPtr output, GenContext& context, const
 
     renderTextureSpace();
 
-    // TODO: Add support for graphs containing geometric nodes such as position and normal.
```
jstone-lucasfilm, line 165: While we're getting closer to support for "position" and "normal" nodes in pattern graphs, we're not quite there yet, so we should hold onto this comment.  :)

source/MaterialXRenderGlsl/TextureBaker.h
```diff
@@ -72,6 +72,7 @@ class TextureBaker : public GlslRenderer
     ShaderGeneratorPtr _generator;
     string _udim;
     string _extension;
+    std::set<string> _shadersWorldSpaceNormals;
```
jstone-lucasfilm, line 75: Very minor, but here we can use the existing StringSet alias.

source/MaterialXRenderGlsl/TextureBaker.cpp
mjyip-lucasfilm: Good point! I will try that.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -97,6 +97,14 @@ FilePath generateOutMtlxFilename(string file, string srName)
     return out;
 }
 
+// Helper function to check if shader requires normals to be transformed from tangent space to world space
+bool connectsToNormalMapNode(OutputPtr output)
+{
+    ElementPtr normalMapNode = (output) ? output->getParent()->getChild(output->getAttribute("nodename")) : nullptr;
```
jstone-lucasfilm, line 103: In order to avoid this hard-coded "nodename" string, we can use the built-in method Output::getNodeName, which is provided by the base PortElement class.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -124,6 +132,13 @@ void TextureBaker::bakeShaderInputs(ConstShaderRefPtr shaderRef, GenContext& con
         if (output)
         {
             FilePath filename = FilePath(outputFolder / generateTextureFilename(output, shaderRef->getName(), udim));
+
+            if (connectsToNormalMapNode(output))
+            {
+                ElementPtr normalMapNode = output->getParent()->getChild(output->getAttribute("nodename"));
+                output->setAttribute("nodename", normalMapNode->getChild("in")->getAttribute("nodename"));
```
jstone-lucasfilm, line 139: Here we can use built-in methods Output::setNodeName, Node::getInput, and Input::getNodeName, in order to avoid these hardcoded strings.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -142,6 +157,12 @@ void TextureBaker::bakeShaderInputs(NodePtr shader, GenContext& context, const F
         if (output)
         {
             FilePath filename = FilePath(outputFolder / generateTextureFilename(output, shader->getName(), udim));
+            if (connectsToNormalMapNode(output))
+            {
+                ElementPtr normalMapNode = output->getParent()->getChild(output->getAttribute("nodename"));
+                output->setAttribute("nodename", normalMapNode->getChild("in")->getAttribute("nodename"));
```
jstone-lucasfilm, line 163: Same as above.

PR #463, opened 2020-08-13
Shader translation

mjyip-lucasfilm:
- Added ShaderTranslator class
- Added python bindings for ShaderTranslator

Additional supporting changes:
- Moved TextureBaker::connectsToNormalMapNode() to MaterialXGenShader/Util.h


source/MaterialXGenShader/ShaderTranslator.h
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Returns set of all the available potential translations
+    StringSet getAvailableTranslations(string start)
+    {
+        return _shadingTranslations[start];
+    }
+
+  protected:
+    /// Constructor.
+    static ShaderTranslatorPtr create(ConstDocumentPtr doc)
```
jstone-lucasfilm, line 37: Should this static creation method be public and accessible via Python?

source/MaterialXGenShader/ShaderTranslator.h
```diff
@@ -0,0 +1,76 @@
+//
+// TM & (c) 2017 Lucasfilm Entertainment Company Ltd. and Lucasfilm Ltd.
```
jstone-lucasfilm, line 2: For newly created files like this one, let's go ahead and use "2020" as the copyright year.  Going forward, I think the best approach will be to mark each file with the year in which it was first added to the codebase, rather than requiring uniformity across files.

source/MaterialXGenShader/ShaderTranslator.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+void ShaderTranslator::connectToTranslationInputs(ShaderRefPtr shaderRef)
+{
+    for (BindInputPtr bindInput : shaderRef->getBindInputs())
+    {
+
+        OutputPtr output = bindInput->getConnectedOutput();
+        if (output)
+        {
+            NodeGraphPtr translationNg = output->getDocument()->getNodeGraph("NG_" + _translationNode->getCategory());
```
jstone-lucasfilm, line 57: For consistency with other MaterialX code, I would name this variable "translationGraph", and similarly name the NodeGraph class member "_graph" instead of "_ng".

source/MaterialXGenShader/Util.h
```diff
@@ -123,6 +123,10 @@ vector<Vector2> getUdimCoordinates(const StringVec& udimIdentifiers);
 /// 0..1 space.
 void getUdimScaleAndOffset(const vector<Vector2>& udimCoordinates, Vector2& scaleUV, Vector2& offsetUV);
 
+/// Checks if output connected to shaderreeef input requires normals to be transformed from tangent
```
jstone-lucasfilm, line 126: Shouldn't this method work for _any_ output of a NodeGraph?  If so, then we can omit the details here about the ShaderRef input (which will actually be replaced with a simpler shader node mechanism in MaterialX v1.38).

mjyip-lucasfilm: Yes you are right! I will change the comment (and the typo)

source/MaterialXGenShader/ShaderTranslator.h
mjyip-lucasfilm: It wouldn't hurt! I will go ahead and change that.

source/MaterialXGenShader/ShaderTranslator.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+        shaderRef->removeBindInput(bindInput->getName());
+    }
+}
+
+void ShaderTranslator::insertOutputNgUpstreamElems(OutputPtr translatedOutput, OutputPtr graphOutput)
+{
+    ElementPtr top = graphOutput->asA<Element>();
+    vector<ElementPtr> upstreamElements;
+    while (top)
```
jstone-lucasfilm, line 82: There's a more idiomatic way to describe this kind of graph traversal in MaterialX, using the Element::traverseGraph method:

https://github.com/materialx/MaterialX/blob/master/source/MaterialXTest/MaterialXCore/Traversal.cpp#L100

In addition to being fast and compact, it has the advantage of safely handling cycles and loops in the graph.

source/MaterialXGenShader/ShaderTranslator.cpp
```diff
@@ -0,0 +1,207 @@
...
+        top = top->getUpstreamElement();
+    }
+
+    // rebuilding upstream node dependencies from translation output to input
+    for (ElementPtr upstreamElem : upstreamElements)
+    {
+        if (upstreamElem == graphOutput->asA<Element>())
+        {
+            // connecting output to the upstream node (that will be copied over)
+            translatedOutput->setNodeName(upstreamElem->asA<Output>()->getNodeName());
```
jstone-lucasfilm, line 94: Perhaps this logic would be clearer using the built-in Output::setConnectedNode method?

source/MaterialXGenShader/ShaderTranslator.cpp
```diff
@@ -0,0 +1,207 @@
...
+bool ShaderTranslator::translateAllMaterials(DocumentPtr doc, string destShader)
+{
+    ShaderTranslatorPtr translator = ShaderTranslator::create(doc);
+    vector<TypedElementPtr> renderableShaderRefs;
+    std::unordered_set<OutputPtr> outputs;
+    findRenderableShaderRefs(doc, renderableShaderRefs, false, outputs);
+    for (TypedElementPtr elem : renderableShaderRefs)
+    {
+        ShaderRefPtr sr = elem ? elem->asA<ShaderRef>() : nullptr;
+        string startShader = sr->getNodeString();
```
jstone-lucasfilm, line 190: It might be clearer to name these variables "sourceShader" and "sourceName", making the common parallel with "destShader" above.

source/MaterialXGenShader/ShaderTranslator.h
```diff
@@ -0,0 +1,75 @@
...
+    /// Reads shading translation nodes from the document
+    void loadShadingTranslations();
+
+    /// Connects translation node inputs from the original shaderRef
+    void connectToTranslationInputs(ShaderRefPtr shaderRef);
+
+    /// Copies translation nodegraph upstream node dependencies over to the working nodegraph.
+    /// Used when normals need to be baked in tangent space but shaderref expects normals to 
+    /// be in world space.
+    void insertOutputNgUpstreamElems(OutputPtr translatedOutput, OutputPtr ngOutput);
```
jstone-lucasfilm, line 52: One additional suggestion that didn't come through in my last review:

Following the logic of your documentation above, would ShaderTranslator::insertUpstreamDependencies be a clearer name for this method?

Issue #468, opened 2020-08-24
Baker cannot find nodedefs


bsavery:
I have a mtlx file exported from Substance.  When trying to run through the texture baker script I get an error "LookupError: Could not find a nodedef for node 'standard_surface_constructor'".  I tried manually setting the --library or --path option to the included stdlib/stlib_defs.mtlx file.  Since I see that node defined as a "surfaceshader" and that is there.

Any ideas?
[MTL_Exports 2.zip](https://github.com/materialx/MaterialX/files/5120577/MTL_Exports.2.zip)

jstone-lucasfilm:
Hi Brian,

Good catch, and this is an area that we're actively working on.  For now, you need to provide the baking script with a "--path" argument, pointing to the root of your MaterialX installation (e.g. "--path /projects/MaterialX").

Going forward, we'd like to begin including the standard data libraries as part of MaterialX Python, which would remove the need for that step.

All the best,

-Jonathan

niklasharrysson:
Also, just want to mention that the nodedef you are looking for is in the
physically-based shader node library (pbrlib). The output *type* for
"standard_surface_constructor" is 'surfaceshader', which is a datatype in
stdlib, but the node definition itself, the node named <surface>, is found
in pbrlib. So make sure you load that library as well if not done already
(pbrlib/pbrlib_defs.mtlx).

Cheers,
/Niklas


Den tis 25 aug. 2020 kl 06:18 skrev Jonathan Stone <notifications@github.com

bsavery:
So actually I think the problem here is this line:  https://github.com/materialx/MaterialX/blob/fadc2e5938b556d12f54759d1cc7f275576cd9f1/python/Scripts/baketextures.py#L42

using cwd, I was running this from outside that directory. I believe this should be something more like:

os.path.join(os.path.abspath(__file__), '..', '..')

to have that script file executable from anywhere.

jstone-lucasfilm:
@bsavery That sounds like a good step forward, as it would allow baketextures.py to find its data libraries from the surrounding MaterialX repository.  Would you be interested in proposing that change in a pull request to MaterialX GitHub?

Going forward, we'd still like to include the stdlib/pbrlib/bxdf libraries in MaterialX Python itself, but your proposed change would remove this need in the short term.

bernardkwok:
Couple of things that you might want to look out for.
[...]

PR #470, opened 2020-08-25
Improve search path for baketextures script

bsavery:
Another possible enhancement here would be to add the directory of the "input_filename" as a library path as well (automatically).  Substance seems to include the custom nodedefs along side the exported mtlx file so this would make sense for that.


bsavery:
#468

bsavery:
What do you think about adding the local directory to the input file to the search path for custom libraries?  I think that makes sense.

jstone-lucasfilm:
That's an interesting question, and while we expect directly-included dependencies (e.g. XIncluded documents and referenced images) to be accessible from a content document via relative paths, we'd need to think through the question of whether implicitly-included dependencies (e.g. custom libraries for a studio pipeline) should be accessible in the same fashion.  Let's continue that discussion separately from this improvement, as it may not be quite as simple to resolve.

PR #483, opened 2020-09-13
Wedge updates for v1.38.

bernardkwok:
- Expose general parameters in advanced.
- Expose choice of parameter to sample (supports int, float, vec2, vec3, color3, vec4, color4) now vs only float before.
- Support more than tga in file dialog.
- Add in additional command line arguments to avoid having to interactively perform this:
  --cameraZoom <float> : Amount of user zoom to apply 1.3 seems to get a close fit.
  --envSampleCount <int> : Default is 16 but 64 is a good fidelity number.

Snapshot of new UI with sample material:
image

Sample result:
image

### DIFF
--- a/source/MaterialXView/Viewer.cpp
+++ b/source/MaterialXView/Viewer.cpp
@@ -1605,6 +1605,10 @@ bool Viewer::keyboardEvent(int key, int scancode, int action, int modifiers)
             _captureFilename = ng::file_dialog(filetypes, true);
             if (!_captureFilename.isEmpty())
             {
+                if (_captureFilename.getExtension().empty())
+                {
+                    _captureFilename.addExtension(mx::ImageLoader::TGA_EXTENSION);
+                }
                 _captureRequested = true;
             }
         }
@@ -1613,6 +1617,18 @@ bool Viewer::keyboardEvent(int key, int scancode, int action, int modifiers)
             _wedgeFilename = ng::file_dialog(filetypes, true);
             if (!_wedgeFilename.isEmpty())
             {
+                // There are issues with using the STB loader and png files
+                // so use another format if PNG specified or if no extension specified
+                const std::string extension = _wedgeFilename.getExtension();
+                if (extension.empty())
+                {                    
+                    _wedgeFilename.addExtension(mx::ImageLoader::TGA_EXTENSION);
+                }
+                else if (extension == mx::ImageLoader::PNG_EXTENSION)
+                {
+                    _wedgeFilename.removeExtension();
+                    _wedgeFilename.addExtension(mx::ImageLoader::TGA_EXTENSION);
+                }
                 _wedgeRequested = true;
             }
         }
--- a/source/MaterialXView/Viewer.cpp
+++ b/source/MaterialXView/Viewer.cpp
@@ -31,12 +31,13 @@
 
 const mx::Vector3 DEFAULT_CAMERA_POSITION(0.0f, 0.0f, 5.0f);
 const float DEFAULT_CAMERA_VIEW_ANGLE = 45.0f;
+const float DEFAULT_CAMERA_ZOOM = 1.0f;
+const int DEFAULT_ENV_SAMPLES = 16;
 
 namespace {
 
 const int MIN_ENV_SAMPLES = 4;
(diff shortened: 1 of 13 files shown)


source/MaterialXView/Viewer.cpp
```diff
@@ -905,6 +908,47 @@ void Viewer::createAdvancedSettings(Widget* parent)
...
+    ng::Label* wedgeLabel = new ng::Label(advancedPopup, "Wedge Rendering Options");
+    wedgeLabel->setFontSize(20);
+    wedgeLabel->setFont("sans-bold");
+
+    ng::Widget* wedgeMin = new ng::Widget(advancedPopup);
+    wedgeMin->setLayout(new ng::BoxLayout(ng::Orientation::Horizontal));
+    mx::UIProperties wedgeProp;
+    wedgeProp.uiSoftMin = mx::Value::createValue(0.0f);
+    wedgeProp.uiSoftMax = mx::Value::createValue(1.0f);
+    ng::FloatBox<float>* wedgeMinBox = createFloatWidget(wedgeMin, "Minimum Wedge Multiplier:",
```
jstone-lucasfilm, line 921: What is the motivation for the term "multiplier" in these option names?  I would expect them to be named "Wedge Minimum Value" and "Wedge Maximum Value", but I may be missing the intent here.

jstone-lucasfilm (commented):
This looks like a major improvement to wedge rendering, thanks @bernardkwok.  Just one question on naming conventions below.

source/MaterialXView/Viewer.cpp
bernardkwok: This is left over from the original intent that this be a multiplier on the value but didn't end up making sense. This should indeed be the absolute value and will make the change.

source/MaterialXView/Viewer.cpp
```diff
@@ -1547,31 +1591,30 @@ bool Viewer::keyboardEvent(int key, int scancode, int action, int modifiers)
...
-
-    // Render a wedge for the current material.
-    if (key == GLFW_KEY_W && action == GLFW_PRESS)
-    {
-        _wedgeFilename = ng::file_dialog({ { mx::ImageLoader::TGA_EXTENSION, mx::ImageLoader::TGA_EXTENSION } }, true);
-        if (!_wedgeFilename.isEmpty())
+        else
         {
-            if (_wedgeFilename.getExtension() != mx::ImageLoader::TGA_EXTENSION)
+            _wedgeFilename = ng::file_dialog(filetypes, true);
```
jstone-lucasfilm, line 1613: I like this generalization of filetypes in frame/wedge capture, but I think we still need a fallback extension for the common case where the user enters a string with no extension in the file browser (e.g. "output").  I actually ran into this myself when testing the new functionality just now.  :)

In the past I've used TGA as this default extension, as stb_image runs into edge cases with the PNG extension.

bernardkwok: Yes, I also ran into this. Ill restore logic to check for an extension and if none then use tga. The PNG problem I also ran into and have disallowed if using stb for unit tests. Maybe I'll disallow png here as well as it crashes.

jstone-lucasfilm:
This change is looking nearly ready, and my two remaining concerns are with the UI for wedge rendering:

- When I test the new UI on the Standard Surface Default material, the combo box for Wedge Property is so large that I can't select "specular_roughness".  One option would be to make this a string rather than a combo box, since the user can easily find the list of available properties in the Property Editor.
- It's challenging to discover the option for modifying the Wedge Property, since it's on a different page from the other wedge options.  Now that you've added a Wedge Rendering Options section, should Wedge Property be there as well?

bernardkwok:
Yes. I've run into the same issue for selection and it occurs for all pop-up menus. I'll look at alternatives.
For organization,  I think it's a good idea to keep all the options in one place but the input to wedge is dynamic based on selected material so will need to rejig this a bit.

jstone-lucasfilm:
Sounds good, thanks @bernardkwok!  The simpler approach, where the wedge property is a simple string (e.g. "specular_roughness"), would also be fine for now, and wouldn't introduce dependencies between two different combo boxes.

PR #493, opened 2020-10-02
Parameter to input 1.38 update Part 1.

bernardkwok:
- Convert Parameters to Inputs marked as "uniform".  Note that according to the spec all "filename" and "string" types are uniform. **All other inputs are marked with `uniform="false"`**. 
- Modify addInput() to make name, type required arguments and add in a `isUniform` argument (default=`false`) to allow a single method to create and set whether something is uniform. Note that according to the spec all `filename` and `string` types are uniform. 
- Add new `getIsUniform()` and `setIsUniform()` APIs.
- Replace all `get/set/isA<Parameter>()` calls with new API calls.
- Remove `getPrimaryShaderParameters()` API from Material Element.
- Update Core, ShaderGen + tests. RTS passes for GLSL, MDL, and Vanilla OSL.

**Note** : While 1.38 is WIP, we keep the version at 1.37 so that the upgrade will be appied.

* stdlib (defintion + node graph implementations) after upgrade. inputs with `uniform` attributes are converted `parameters`
[1_37_to_1_38_updated .mtlx.txt](https://github.com/materialx/MaterialX/files/5319974/1_37_to_1_38_updated.mtlx.txt)


jstone-lucasfilm:

PR #510, opened 2020-10-29
Backport to LucasFilm v1.38 branch.

bernardkwok:
- Wedge rendering support for unit tests.
- Small sync updates.

### DIFF
--- a/source/MaterialXRender/ShaderRenderer.h
+++ b/source/MaterialXRender/ShaderRenderer.h
@@ -118,7 +118,7 @@ class ShaderRenderer
 
     /// Save the current contents of the off-screen hardware buffer to disk.
     /// @param filePath Path to file to save rendered image to.
-    virtual void saveImage(const FilePath& filePath) = 0;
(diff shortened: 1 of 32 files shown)


source/MaterialXRender/ShaderRenderer.h
```diff
@@ -118,7 +118,7 @@ class ShaderRenderer
 
     /// Save the current contents of the off-screen hardware buffer to disk.
     /// @param filePath Path to file to save rendered image to.
-    virtual void saveImage(const FilePath& filePath) = 0;
+    virtual void saveImage(const FilePath& filePath, ConstImagePtr image, bool verticalFlip) = 0;
```
jstone-lucasfilm, line 121: One note for @bernardkwok:  It looks like this changelist transforms ShaderRenderer::saveImage into a pure image-saving method, one that no longer references the contents of the renderer's off-screen buffer.  If that's the intent, then perhaps we should make it a simple helper function of the Image class, rather than a ShaderRenderer method.  What are your thoughts?

jstone-lucasfilm:
This pull request has been superseded by #522, so I'm closing it out.

PR #523, opened 2020-11-13
Fix bug with codegen of custom BSDF/EDF graphs.

niklasharrysson:
This changelist addresses a problem in codegen of nodegraphs of BSDF and EDF type. For HW shaders the suffixes used for functions, depending on closure context, was being applied to _all_ functions, not just BSDF/EDF functions. This change makes the use of suffixes and extra arguments depend on node type so it only applies to BSDF/EDF functions.

### DIFF
--- a/source/MaterialXGenShader/HwShaderGenerator.h
+++ b/source/MaterialXGenShader/HwShaderGenerator.h
@@ -230,7 +230,7 @@ using HwShaderGeneratorPtr = shared_ptr<class HwShaderGenerator>;
 /// @class HwClosureContext
 /// Class representing a context for closure evaluation on hardware targets.
 /// On hardware BSDF closures are evaluated differently in reflection, transmission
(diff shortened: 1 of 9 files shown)


source/MaterialXGenShader/HwShaderGenerator.h
```diff
@@ -263,25 +263,37 @@ class HwClosureContext : public GenUserData
...
+        auto it = _suffix.find(nodeType);
+        return it != _suffix.end() ? it->second : EMPTY_STRING;
+    }
 
 protected:
     const int _type;
-    Arguments _arguments;
-    string _suffix;
+    std::unordered_map<const TypeDesc*, Arguments> _arguments;
+    std::unordered_map<const TypeDesc*, string> _suffix;
```
niklasharrysson, line 295: Make the arguments and suffixes settable per node type.

jstone-lucasfilm:

PR #524, opened 2020-11-16
Use newer OpenImageIO interface

marsupial:
OIIO dropped some implicit string conversions a while ago, so be more explicit.

@lgritz: CLA should be covered via Imageworks.

### DIFF
--- a/source/MaterialXRender/OiioImageLoader.cpp
+++ b/source/MaterialXRender/OiioImageLoader.cpp
@@ -46,7 +46,7 @@ bool OiioImageLoader::saveImage(const FilePath& filePath,
     }
 
     bool written = false;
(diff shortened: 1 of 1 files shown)


lgritz (commented):
LGTM

Long ago, `create()` took a `std::string&`, which would have implicitly converted from a FilePath.  At some point (a couple years ago, actually) create() switched to a string_view that can implicitly convert from std::string, but unfortunately, C++ is not smart enough to do the two-step of FilePath -> std::string -> string_view. So I think Roman's fix here is the correct and expedient approach.

jstone-lucasfilm (approved):
This looks good to me.  Thanks for the fix, @marsupial!

PR #525, opened 2020-11-16
Allow building against an external NanoGUI installation.

marsupial:
Adds a CMake option **MATERIALX_LOCAL_NANOGUI** to enable building against an existing NanoGUI installation.  The NanoGUI submodule actually brings in NanoGUI, NanoVG, glfw, eignen, and pybind11, which elongates the build process and hides the dependencies from package-management which may want to track dependencies for usage/audit/etc.

@lgritz: CLA should be covered via Imageworks.

### DIFF
--- a/source/MaterialXView/CMakeLists.txt
+++ b/source/MaterialXView/CMakeLists.txt
@@ -1,36 +1,64 @@
-if (NOT IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/NanoGUI/ext/glfw/src")
-    message(FATAL_ERROR "Building the MaterialX viewer requires the NanoGUI submodule "
-        "to be present. Update your repository by calling the following:\n"
-        "git submodule update --init --recursive")
+option(MATERIALX_NANOGUI_EXTERNAL "Build aginst an external install of NanoGUI (NANOGUI_ROOT may also need to be set)" OFF)
+
+if (MATERIALX_NANOGUI_EXTERNAL)
+    find_path(NANOGUI_INCLUDE_DIRS
+        NAMES
+            nanogui/nanogui.h
+        HINTS
+            "${NANOGUI_ROOT}/include"
+            "$ENV{NANOGUI_ROOT}/include"
+        )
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This is definitely a worthwhile improvement, but I'd like to make sure we're not missing an opportunity to align our interfaces for specifying the locations of external packages.

For PyBind11, we currently support external packages via the `MATERIALX_PYTHON_PYBIND11_DIR` option in CMake.  Would it be sufficient to implement your proposed feature with a similar `MATERIALX_NANOGUI_DIR` option, or would there be disadvantages in that approach for your use case?

marsupial:
Not gonna stand too firm on this, (especially as it's pretty easy/trivial to extend the **HINTS** to also use **MATERIALX_NANOGUI_DIR**), but I would like to bring up that it does seem MATERIALX_XXX_DIR vars are somewhat re-inventing the wheel: https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html#variable:%3CPackageName%3E_ROOT

From what I see, MATERIALX_PYTHON_PYBIND11_DIR is a fallback when `find_package(pybind11)` fails.
The use of **NANOGUI_ROOT** is more because NanoGUI _does not_ currently ship CMake helpers.
But if(when) it does, scripts to invoke CMake via `-DNANOGUI_ROOT=xxx` would continue to work with a `find_package(nanogui)` replacing the two `find_path` and `find_library` calls.

Let me know.

jstone-lucasfilm:
All good points, and I can see a logical case for moving away from MATERIALX_XXX_DIR controls in the future, especially in cases where find_package can be used effectively.

For this specific improvement, what would you think of renaming the new flag `MATERIALX_NANOGUI_EXTERNAL`, and making its default value OFF.  This would create a stronger parallel with existing MaterialX flags, without creating overlap with a future CMake helper for the NanoGUI package.

marsupial:
Dunno if the repo has merge/squash enabled, so just rolled that change in.

jstone-lucasfilm (approved):
Looks good, thank you!

PR #568, opened 2021-01-23
Mx camera

meshula:
This commit reimplements the NanoGui camera controls into the MaterialX Viewer using native MaterialX types.

The removes the dependency on Eigen, and opens the door to more sophisticated camera behaviors, and also enables migrating to the most up to date version of NanoGui, which no longer has large external dependencies such as Eigen. 

A minimal Quaternion class is introduced to MaterialX types, and the few necessary operations necessary for the implementation of arcball type control for cameras.


jstone-lucasfilm:
This change looks fantastic, thanks @meshula!  Since we'll soon be merging the v1.38 branch into master, I've been focusing active development on that branch, including the recent removal of the dependency on NanoGUI::GLShader.  Would it be possible to propose this change to the v1.38 branch as well, so that it can be included along with the other viewer improvements?

meshula:
Submitted a PR to v1.38. PS, I'll update the PR to address the Linux failure shortly.

dbsmythe:
Jonathan and Nick- what would you think of adding an actual Quaternion type to the official 1.38 spec?  As well as math operations on sensible combinations of quaternion / matrix / vector / float types?

-Doug

meshula:
Motivation for including Quaternion might come from a need for spatial operations, like surface displacements, or say hair grooms, or some kind of fancy direction dependent effect. Certainly if a camera were to be represented in a shader, a quaternion + position would be in some cases nicer to pass around than a matrix.

jstone-lucasfilm:
@dbsmythe That's definitely a worthwhile discussion, though I'd recommend delaying it until v1.39 so that we can push forward on the release of the current specification and codebase.

Historically, we've avoided making type-based distinctions between different interpretations of vectors (e.g. points, normals, Euler angles, quaternions), leaving these semantics up to the client.  If we continued that approach in v1.39, we might consider adding quaternion operations (e.g. generation from axis-angle, conversion to and from matrix), as supplemental MaterialX nodes, with the underlying data type remaining vector4.

jstone-lucasfilm:
Now that @meshula has created a v1.38 version of this pull request (#569), I'll close out the original.

dbsmythe:
Note that I am not necessarily advocating adding a Quaternion type to the main MaterialX spec.  Normally, I'd lean toward keeping the list of standard supported types to a relatively lean set of necessary types, since we expect all implementations to support all those standard types.  I'm not sure how useful a quaternion type would be in pattern and shader node inputs and outputs, especially in the PBR world where the renderer handles all the light transport.  So this is more a question of are there any use cases for a quaternion type in shading that I hadn't considered.

lgritz:
Possibly relevant FYI: The OpenEXR project is close to releasing version 3.0. A big change is that Imath has been split into a separate repo -- the half, vector/matrix, and yes, quaternion, classes no longer require pulling all the rest of the exr file format library in as a dependency. It's almost completely header-only now, and has been stylistically modernized (e.g., everything is now maximally constexpr, noexcept, and switches all kinds of custom doodads to using std:: equivalents that were added in C++11).

jstone-lucasfilm:
@lgritz At a minimum, we'll want to set up implicit conversions between MaterialX and Imath vectors (e.g. using the Wenzel Jakob approach in Mitsuba 2), but I'm also open to the idea of harmonizing on the Imath vector library within MaterialX once the OpenEXR/Imath split is finalized.

lgritz:
Yes, by far the most important thing is pain-free implicit conversion of the types. If that's in place, it matters a lot less if a project wants a different custom vector.

lgritz:
@jstone-lucasfilm Do you suppose you could point me to the exact spot that defines the idiom you think is best to apply for the implicit conversion? We'll definitely put that on the Imath side, too.

jstone-lucasfilm:
@lgritz Here's a link to our earlier conversation on this topic, where we discussed a VectorSize member variable that could be used for implicit conversions between vector classes.

https://github.com/materialx/MaterialX/pull/365#issuecomment-593127024

Let me know if this high-level approach seems reasonable, perhaps using a more unique member variable name for robustness.  Once Imath has selected an appropriate marker, I'm happy to add the matching marker on the MaterialX side, and we can make this an official recommendation for future projects.

lgritz:
I wonder if we could do even better, and drop the requirement that the other vector classes have this nonstandard VectorSize definition. Suppose "our" class is has element type T and length N. Informally, I would consider it a good  bet that an unknown external class V is compatible if it has a subscript operator, typeof(V[0]) is T, and sizeof(V) == N*sizeof(T). This could be expressed with the right enable_if clauses, and as well, could also handle the case of classes that lack subscript operator but have data elements .x, .y, .z that are all type T and they are in successive contiguous memory locations.

jstone-lucasfilm:
@lgritz That sounds very compelling, and would allow conversion from legacy vector classes that predate this strategy for compatibility.  If that approach works well for Imath, I'll definitely follow it in the MaterialX library, and we'll have an initial pair of libraries that are bidirectionally convertible.

lgritz:
I'll give it a try when I get a chance and let you know.

lgritz:
Just to properly cross-reference, here's some exploration I'm doing on this in Imath:  https://github.com/AcademySoftwareFoundation/Imath/pull/91

PR #569, opened 2021-01-24
Refactor viewer camera using MaterialX types

meshula:
Here's a request to the v1.3 branch to go along with the removal of NanoGui::GLShader from the MaterialXViewer


jstone-lucasfilm:
Original description from @meshula:

This commit reimplements the NanoGui camera controls into the MaterialX Viewer using native MaterialX types.

The removes the dependency on Eigen, and opens the door to more sophisticated camera behaviors, and also enables migrating to the most up to date version of NanoGui, which no longer has large external dependencies such as Eigen.

A minimal Quaternion class is introduced to MaterialX types, and the few necessary operations necessary for the implementation of arcball type control for cameras.

meshula:
Thanks for the revs, with that, the new code fits in "natively" as it were :)

jstone-lucasfilm (approved):
Thanks for this excellent change, @meshula!  This brings us within striking distance of upgrading the viewer to the latest NanoGUI, and makes it more straightforward to improve our camera interactions in the future.

PR #571, opened 2021-01-26
PBS library updates for MaterialX v1.38

niklasharrysson:
This change list contains the following changes to _pbrlib_, from our list of agreed changes for v1.38.

- Our dielectric BSDF nodes which previously was split into separate BRDF and BTDF nodes are now unified into a single spherical BSDF distribution, same as MDL, with a scatter_mode parameter to control the two hemispheres.

- The conductor_bsdf has been changed from an artistic parameterization (Gulbrandsen) into the physical / complex IOR parameterization. Instead, the node artistic_ior can be used to convert from artistic to physical parameters when needed. This makes it more aligned with MDL and avoids forcing the user into using Gulbrandsen's parameterization. The node complex_ior is thus no longer needed and has been removed.

- Removing the base input on the vertically layerable nodes, that connected a BSDF for the layer below. In favor of always using the layer operator to connect nodes into layer stacks, for consistency. 

- Removing the interior input on the dielectric transmission node, that connected a VDF for specifying volumetric properties of the medium. In favor of always using the layer operator to connect nodes into layer stacks, for consistency.

- A general naming convention change for BSDF nodes to use "_bsdf" as suffix for all nodes, where we previously used "_brdf" or "_btdf" depending on which hemisphere it represented.

- diffuse_brdf has been renamed oren_nayar_diffuse_bsdf

- diffuse_btdf has been renamed translucent_bsdf

- Document upgrade logic has been added to upgrade old documents according to above changes


jstone-lucasfilm:

PR #574, opened 2021-01-29
Merge update from ADSK to v.138

bernardkwok:
### Definitionss
- Update to current version of standard surface defaults
- Fix for USDPreviewSurface metallic input naming
- Fixes for definition version checking including removing version checking on nodegraphs which should not be done. Upgrade path removes these.

### Document Upgrade 
- Upgrade path for artistic_ior
- Fix upgrade path to make sure certain integer inputs are properly marked as uniforms. (e.g. frameoffset on an image node,and index number for geometric properties.)

### Code Generation
- Fixes to make MDL unit tests valid again.
- Fix for color management transform injection for targets which inherit from base targets.

### Nodegraph Connectins
- Fix for traversal from a material in a nodegraph to a shader outside a nodegraph. 

### Texturing
- Fix for constant texture baking.
- Fix for fallback code for single IBL loading in viewer

### Test Results
[jan_29_2021_render_tests.pdf](https://github.com/materialx/MaterialX/files/5895157/jan_29_2021_render_tests.pdf)


libraries/bxdf/standard_surface.mtlx
```diff
@@ -5,9 +5,9 @@
   -->
   <nodedef name="ND_standard_surface_surfaceshader" node="standard_surface" nodegroup="pbr"
             doc="Autodesk standard surface shader" version="1.0.1" isdefaultversion="true">
-      <input name="base" type="float" value="0.8" uimin="0.0" uimax="1.0" uiname="Base" uifolder="Base"
+      <input name="base" type="float" value="1.0" uimin="0.0" uimax="1.0" uiname="Base" uifolder="Base"
```
bernardkwok, line 8: Standard surface default values changes.

libraries/bxdf/usd_preview_surface.mtlx
```diff
@@ -12,7 +12,7 @@
     <input name="emissiveColor" type="color3" value="0, 0, 0" />
     <input name="useSpecularWorkflow" type="integer" value="0" />
     <input name="specularColor" type="color3" value="0, 0, 0" />
-    <input name="metalness" type="float" value="0" />
+    <input name="metallic" type="float" value="0" />
```
bernardkwok, line 15: Fix for match USD naming.

resources/Materials/TestSuite/pbrlib/bsdf/artistic_ior.mtlx
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
```
bernardkwok, line 1: artistic_ior upgrade test.

resources/Materials/TestSuite/stdlib/definition/surfacematerial_definition.mtlx
```diff
@@ -4,7 +4,7 @@
    <nodedef name="ND_mymaterial" node="mymaterial" nodegroup="material" target="mytarget" version="1.3" isdefaultversion="true">
     <output name="out" type="material"/>
   </nodedef>
-  <nodegraph name="NG_mymaterial" nodedef="ND_mymaterial" target="mytarget" version="1.3">
+  <nodegraph name="NG_mymaterial" nodedef="ND_mymaterial" target="mytarget" >
```
bernardkwok, line 7: One of a few fixes for improper nodegraph version tagging.

resources/Materials/TestSuite/stdlib/nodegraph_inputs/surfacematerial_nodegraph_to_surfaceshader.mtlx
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0"?>
```
bernardkwok, line 1: Test for nodegraph with surfacematerial connected to upstream surfaceshader. Part of nodegraph connection work.

resources/Materials/TestSuite/stdlib/texture/image.mtlx
```diff
@@ -41,10 +41,4 @@ Basic image unit test with one image node for each variation in input type.
   </image>
   <output name="image_float_output" type="float" nodename="image_float" />
 
-   <image name="image_default" type="color3">
```
bernardkwok, line 44: MDL will error on during compilation as image file is not found, so remove test.

resources/Materials/TestSuite/stdlib/version/multiple_version_test.mtlx
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0"?>
```
bernardkwok, line 1: Definition version test file.

source/MaterialXCore/Definition.cpp
```diff
@@ -30,8 +30,7 @@ namespace
         for (InterfaceElementPtr interface : candidates)
         {
             if (!interface->isA<NodeGraph>() ||
-                !targetStringsMatch(interface->getTarget(), target) ||
-                !nodedef.isVersionCompatible(interface))
+                !targetStringsMatch(interface->getTarget(), target))
```
bernardkwok, line 33: nodegraph should not have a version check.

source/MaterialXCore/Definition.cpp
```diff
@@ -126,11 +125,11 @@ bool NodeDef::validate(string* message) const
 
 bool NodeDef::isVersionCompatible(ConstElementPtr elem) const
 {
-    if (getVersionIntegers() == elem->getVersionIntegers())
+    if (getVersionString() == elem->getVersionString())
```
bernardkwok, line 128: Fix to compare actual string as version number utility can fail in the general string version case and only handles 
versions in the format "<x>.<y>".

source/MaterialXCore/Document.cpp
```diff
@@ -1216,13 +1222,80 @@ void Document::upgradeVersion()
             {
                 node->setCategory(SUBSURFACE_BRDF.second);
             }
+            else if(nodeCategory == ARTISTIC_IOR)
```
bernardkwok, line 1225: artistic_ior upgrade

source/MaterialXCore/Document.cpp
```diff
@@ -1258,28 +1331,44 @@ void Document::upgradeVersion()
...
+                    if (elementPtr->isA<NodeDef>())
                     {
-                        // Only strings and filename types should be set as uniforms.
+                        // Strings and filename types should always be uniforms.
                         const string& inputType = input->getType();
                         if (inputType == FILENAME_TYPE_STRING || inputType == STRING_TYPE_STRING)
                         {
                             input->setIsUniform(true);
                         }
+                        // Some integer inputs should be set as uniforms
```
bernardkwok, line 1352: Special casing to mark some integer inputs as uniforms (mainly affects image, and geometry inputs)

source/MaterialXCore/Material.cpp
```diff
@@ -62,28 +62,25 @@ std::unordered_set<NodePtr> getShaderNodes(const NodePtr& materialNode, const st
     for (const InputPtr& input : inputs) 
     {
         // Scan for a node directly connected to the input.
+        // Note that this will handle traversing through interfacename associations.
         //
-        const string& inputShader = input->getNodeName();
-        if (!inputShader.empty())
+        NodePtr shaderNode = input->getConnectedNode();
```
bernardkwok, line 67: Fix for traversing from materials inside nodegraphs to shaders using logic previously added.

source/MaterialXCore/Util.cpp
```diff
@@ -164,4 +165,15 @@ string parentNamePath(const string& namePath)
     return EMPTY_STRING;
 }
 
+ElementPtr changeChildCategory(ElementPtr parent, ElementPtr origChild, const string& category)
```
bernardkwok, line 168: This is code we use externally to change node categories for node translation.

source/MaterialXGenShader/ColorManagementSystem.cpp
```diff
@@ -39,29 +39,31 @@ void ColorManagementSystem::loadLibrary(DocumentPtr document)
...
+    ImplementationPtr impl = getImplementation(transform);
     if (!impl)
     {
         throw ExceptionShaderGenError("No implementation found for transform: ('" + transform.sourceSpace + "', '" + transform.targetSpace + "').");
     }
 
     // Check if it's created and cached already,
     // otherwise create and cache it.
-    ShaderNodeImplPtr nodeImpl = context.findNodeImplementation(implName);
+    ShaderNodeImplPtr nodeImpl = context.findNodeImplementation(impl->getName());
```
bernardkwok, line 61: Fix for handling targets which inherit from base ones like "arnold" which inherits from "genosl". Will check up inhertance list for parent targets to match. Fixes color management failing for these targets.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -162,6 +162,13 @@ void TextureBaker::optimizeBakedTextures(NodePtr shader)
         return;
     }
 
+    // Early exist if not optimizing
```
bernardkwok, line 165: Fix for when optimize constants is turned off. Used to not add an image input, but also ended up not writing any values as well during baking.

source/MaterialXView/Viewer.cpp
```diff
@@ -390,15 +390,15 @@ void Viewer::loadEnvironmentLight()
...
+    mx::ImagePtr envIrradianceMap = nullptr;
     if (!_normalizeEnvironment && !_splitDirectLight)
     {
         mx::FilePath envIrradiancePath = _envRadianceFilename.getParentPath() / IRRADIANCE_MAP_FOLDER / _envRadianceFilename.getBaseName();
         envIrradianceMap = _imageHandler->acquireImage(envIrradiancePath);
     }
 
     // If not found, then generate an irradiance map via spherical harmonics.
-    if (!envIrradianceMap)
+    if (!envIrradianceMap || envIrradianceMap == _imageHandler->getInvalidImage())
```
bernardkwok, line 401: This code was not being triggered since a fallback "invalid image" was being returned. Add the new api + check here to allow it to work.

source/MaterialXCore/Util.cpp
jstone-lucasfilm: Thanks for the clarification, and that makes sense.  I may go ahead and merge this new method with the existing updateChildSubclass in Document.cpp, so that we have just one implementation of this logic in the codebase.  The merged version could potentially be a method on the Element class.

source/MaterialXRenderGlsl/TextureBaker.cpp
jstone-lucasfilm: I think we still need to check for uniform images when optimizeConstants is set to false, so that we can still return 4x4 textures instead of full-sized textures when individual UDIMs are constant.  The idea behind "optimizeConstants" is that some tools don't support constant values, only images, but there are still optimizations that are valuable for tools that have this limitation.

jstone-lucasfilm (approved):
This looks excellent, thanks @bernardkwok!  Once we've had a chance to validate the full codebase with production assets, I believe we'll be ready to merge this branch back to main.

PR #588, opened 2021-02-10
Various bugfixes related to 1.38 changes. (#1109)

niklasharrysson:
* Removed some MDL uniform qualifiers.

* Removed unsafe conditional optimization.

* Allow thin-film layered over conductor (result is not correct and needs more work, but this setup is now valid).


jstone-lucasfilm:

PR #589, opened 2021-02-11
Upgrade material examples to v1.38

jstone-lucasfilm:
- Upgrade all documents in the resources/Materials/Examples folder to v1.38, making them user-facing examples of the latest syntax.
- Make sure the base and base_color inputs are bound in Standard Surface examples, making them independent of recent interface changes.
- Remove IOR settings from UsdPreviewSurface metals, bringing them in line with common usage of the metallic/roughness model.
- Remove two Standard Surface examples that are closer in spirit to unit tests, and are covered by the existing test suite.


niklasharrysson (approved):
Looks good! Great to get these upgraded.

PR #611, opened 2021-03-11
Fix external nanogui build

marsupial:
9197ee9b re-introduced linking via *-lnanogui* , which is redundant with the CMake variable *NANOGUI_LIBRARIES* for a local nano-gui build, and causes errors with an external NanoGUI install.  76a30db6 is the original merge where it was removed.


jstone-lucasfilm:
Thanks for the catch, @marsupial!  This looks like a good fix, and I'll plan to merge this into the v1.38.1 development build.

PR #613, opened 2021-03-17
Image handling improvements

jstone-lucasfilm:
- Simplify ImageHandler::acquireImage, making it non-virtual and removing an unused parameter.
- Add protected helper method ImageHandler::loadImage.
- Streamline the implementation of MaterialX::createImageStrip.

### DIFF
--- a/source/MaterialXRender/ImageHandler.cpp
+++ b/source/MaterialXRender/ImageHandler.cpp
@@ -105,55 +105,32 @@ bool ImageHandler::saveImage(const FilePath& filePath,
     return false;
 }
 
-ImagePtr ImageHandler::acquireImage(const FilePath& filePath, bool)
+ImagePtr ImageHandler::acquireImage(const FilePath& filePath)
 {
-    FilePath foundFilePath = _searchPath.find(filePath);
-    string extension = stringToLower(foundFilePath.getExtension());
-    bool extensionSupported = false;
-    for (ImageLoaderPtr loader : _imageLoaders[extension])
+    // Resolve the input filepath.
+    FilePath resolvedFilePath = filePath;
+    if (_resolver)
     {
-        extensionSupported = true;
-        ImagePtr image;
-        try
-        {
-            image = loader->loadImage(foundFilePath);
-        }
-        catch (std::exception& e)
-        {
-            std::cerr << "Exception in image I/O library: " << e.what() << std::endl;
-        }
-        if (image)
-        {
-            // Generated shaders interpret 1x1 textures as invalid images, so valid 1x1
-            // images must be resized.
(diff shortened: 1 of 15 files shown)


source/MaterialXRender/ImageHandler.cpp
```diff
@@ -184,6 +161,53 @@ void ImageHandler::releaseRenderResources(ImagePtr)
...
+    for (ImageLoaderPtr loader : _imageLoaders[extension])
+    {
+        ImagePtr image;
+        try
+        {
+            image = loader->loadImage(filePath);
+        }
+        catch (std::exception& e)
+        {
+            std::cerr << "Exception in image I/O library: " << e.what() << std::endl;
```
bernardkwok, line 176: For a side discussion: would like to resolve how to route error messages to places other than stderr. I think we did something for texture baker which allowed you to set the output stream which would allow routing to a stringstream.

source/MaterialXRender/ImageHandler.h
```diff
@@ -166,17 +166,13 @@ class ImageHandler
...
     /// @return if save succeeded
-    virtual bool saveImage(const FilePath& filePath,
-                           ConstImagePtr image,
-                           bool verticalFlip = false);
+    bool saveImage(const FilePath& filePath, ConstImagePtr image, bool verticalFlip = false);
 
     /// Acquire an image from the cache or file system.  If the image is not
     /// found in the cache, then each image loader will be applied in turn.
     /// @param filePath File path of the image.
-    /// @param generateMipMaps Generate mip maps if supported.
```
bernardkwok, line 176: Noticed that this option has been removed. It would be good to still allow this but perhaps keep this as a option which can be set on the ImageHandler ? We sometimes don't want mipmaps to save on memory. Or perhaps this only applies when binding for hardware usage ?

bernardkwok (commented):
Looks good to me. I just have a query about removing control over mipmap generation.

source/MaterialXRender/ImageHandler.h
jstone-lucasfilm: Initially, I thought the same thing, but this flag didn't actually have any effect on the results of ImageHandler::acquireImage.  It's the method ImageHandler::createRenderResources that decides whether mipmaps should be generated, and that method is called independently of ImageHandler::acquireImage.

source/MaterialXRender/ImageHandler.cpp
jstone-lucasfilm: That's a good point, and I like the idea of extending the TextureBaker pattern to ImageHandler in the future, where the output stream can be modified by the client.

PR #614, opened 2021-03-23
Added initial draft for OSL closures

niklasharrysson:
A first draft for adding OSL closures to represent the MaterialX PBS library in OSL.

This PR is created to be a forum for discussion. When the proposal is finalized a real PR should be made to https://github.com/AcademySoftwareFoundation/OpenShadingLanguage instead.

### DIFF
--- a/libraries/stdlib/osl/stdosl.h
+++ b/libraries/stdlib/osl/stdosl.h
@@ -573,6 +573,168 @@ closure color volume_matte() BUILTIN;
 closure color subsurface(float eta, float g, color mfp, color albedo) BUILTIN;
 #endif
 
+// ******************* MATERIALX PBS LIBRARY CLOSURES - DRAFT BEGIN ******************* //
+
+// -------------------------------------------------------------//
+// BSDF closures                                                //
+// -------------------------------------------------------------//
+​
+// Constructs a diffuse reflection BSDF based on the Oren-Nayar reflectance model.
+// A roughness of 0.0 gives Lambertian reflectance.
+//
+closure color oren_nayar_diffuse_bsdf(normal N, color albedo, float roughness) BUILTIN;
+​
+// Constructs a diffuse reflection BSDF based on the corresponding component of 
+// the Disney Principled shading model.
+//
+closure color burley_diffuse_bsdf(normal N, color albedo, float roughness) BUILTIN;
+​
+// Constructs a reflection and/or transmission BSDF based on a microfacet reflectance
+// model and a Fresnel curve for dielectrics. The two tint parameters control the 
+// contribution of each reflection/transmission lobe. The tints should remain 100% white
+// for a physically correct dielectric, but can be tweaked for artistic control or set
+// to 0.0 for disabling a lobe.
+// The closure may be vertically layered over a base BSDF for the surface beneath the
+// dielectric layer. By chaining multiple dielectric_bsdf closures you can describe a surface
+// with multiple specular lobes. If transmission is enabled (transmission_tint > 0.0) the
+// closure may be layered over a VDF closure describing the surface interior to handle
+// absorption and scattering inside the medium.
+//
+closure color dielectric_bsdf(normal N, vector U, color reflection_tint, color transmission_tint, float roughness_x, float roughness_y, float ior, string distribution) BUILTIN;
+​
+// Constructs a reflection BSDF based on a microfacet reflectance model.
+// Uses a Fresnel curve with complex refraction index for conductors/metals.
+// If an artistic parametrization is preferred the artistic_ior() utility function
+// can be used to convert from artistic to physical parameters.
+//
+closure color conductor_bsdf(normal N, vector U, float roughness_x, float roughness_y, color ior, color extinction, string distribution) BUILTIN;
+​
+// Constructs a reflection and/or transmission BSDF based on a microfacet reflectance model
+// and a generalized Schlick Fresnel curve. The two tint parameters control the contribution
+// of each reflection/transmission lobe.
+// The closure may be vertically layered over a base BSDF for the surface beneath the
+// dielectric layer. By chaining multiple dielectric_bsdf closures you can describe a surface
+// with multiple specular lobes. If transmission is enabled (transmission_tint > 0.0) the
+// closure may be layered over a VDF closure describing the surface interior to handle
+// absorption and scattering inside the medium.
+//
+// TODO:
+// - Transmission handling for this node has not been fully defined yet in MaterialX.
+//   In particular how is IOR for refractions derived from the f0, f90 parameterization?
+//   Do we just derive it from f0? For an artist it seems hard to control refractions that way.
+//   @Jonathan Stone you mentioned Lucasfilm (Naty Hoffman) had some feedback and ideas for how to handle this?
+//
+closure color generalized_schlick_bsdf(normal N, vector U, color reflection_tint, color transmission_tint, float roughness_x, float roughness_y, color f0, color f90, float exponent, string distribution) BUILTIN;
+​
+// Constructs a translucent (diffuse transmission) BSDF based on the Lambert reflectance model.
+//
+closure color translucent_bsdf(normal N, color albedo) BUILTIN;
+
+// Constructs a closure that represents straight transmission through a surface.
+//
+// NOTE:
+//  - This is not a node in the MaterialX library, but the surface shader constructor
+//    node has an 'opacity' parameter to control textured cutout opacity.
+//
+closure color transparent_bsdf() BUILTIN;
+​
+// Constructs a BSSRDF for subsurface scattering within a homogeneous medium.
+//
+// TODO:
+// - Do we want/need this dedicated BSSRDF closure? We have had discussions before about
+//   describing this more rigorously as a BSDF layered over a VDF instead.
+//   In the MaterialX library we want this as a node for the user friendly interface,
+//   but for back-end implementation maybe a dedicated closure is not needed?
+//
+closure color subsurface_bssrdf(normal N, color albedo, color radius, float anisotropy) BUILTIN;
+​
+// Constructs a microfacet BSDF for the back-scattering properties of cloth-like materials.
+// This closure may be vertically layered over a base BSDF, where energy that is not reflected
+// will be transmitted to the base closure.
+//
+closure color sheen_bsdf(normal N, color albedo, float roughness) BUILTIN;
+​
+// Adds an iridescent thin film layer over a microfacet base BSDF. This must be layered over
+// another base BSDF using the layer() closure, as this is a modifier and cannot be used as a
+// standalone closure.
+//
+// TODO:
+// - This might be better to represent as optional extra arguments on the closures that do support
+//   thin-film iridescence (dielectric_bsdf, conductor_bsdf and generalized_schlick_bsdf)?
+//
+closure color thin_film_bsdf(float thickness, float ior) BUILTIN;
+​
+​
+// -------------------------------------------------------------//
+// EDF closures                                                 //
+// -------------------------------------------------------------//
+​
+// Constructs an EDF emitting light uniformly in all directions.
+//
+closure color uniform_edf(color emittance) BUILTIN;
+
+// Constructs an EDF emitting light inside a code around the normal direction.
+//
+closure color conical_edf(color emittance, normal N, float inner_angle, float outer_angle) BUILTIN;
+
+// Constructs an EDF emitting light according to a measured IES light profile.
+//
+closure color measured_edf(color emittance, normal N, string file) BUILTIN;
+​
+​
+// -------------------------------------------------------------//
+// VDF closures                                                 //
+// -------------------------------------------------------------//
+​
+// Constructs a VDF for pure volumetric absorption.
+//
+closure color absorption_vdf(color absorption) BUILTIN;
+​
+// Constructs a VDF scattering light for a participating medium, based on the Henyey-Greenstein
+// phase function. Forward, backward and uniform scattering is supported and controlled by the
+// anisotropy input.
+//
+closure color anisotropic_vdf(color absorption, color scattering, float anisotropy) BUILTIN;
+​
+​
+// -------------------------------------------------------------//
+// Layering closures                                            //
+// -------------------------------------------------------------//
+​
+// Vertically layer a layerable BSDF such as dielectric_bsdf, generalized_schlick_bsdf or
+// sheen_bsdf over a BSDF or VDF. The implementation is target specific, but a standard way
+// of handling this is by albedo scaling, using "base*(1-reflectance(top)) + top", where
+// reflectance() calculates the directional albedo of a given top BSDF.
+//
+// TODO:
+// - This could also be achived by closure nesting where each layerable closure takes
+//   a closure color "base" input instead.
+// - One advantage having a dedicated layer() closure is that in the future we may want to
+//   introduce parameters to describe the sandwitched medium between the layer interfaces.
+//   Such parameterization could then be added on this layer() closure as extra arguments.
+// - Do we want/need parameters for the medium here now, or do we look at that later?
+//
+closure color layer(closure color top, closure color base) BUILTIN;
+​
+// Mix two closures according to a weight. Performs horizontal layering by linear interpolation
+// between the two inputs: "bg∗(1−weight) + fg∗weight".
+//
+closure color mix(closure color bg, closure color fg, float weight);
+​
+​
+// -------------------------------------------------------------//
+// Utility functions                                            //
+// -------------------------------------------------------------//
+​
+// Converts the artistic parameterization reflectivity and edge_color to complex IOR values.
+// To be used with the conductor_bsdf() closure.
+//
+void artistic_ior(color reflectivity, color edge_color, output color ior, output color extinction);
+​
+
+// ******************* MATERIALX PBS LIBRARY CLOSURES - DRAFT END ******************* //
+
+
 // Renderer state
 int backfacing () BUILTIN;
 int raytype (string typename) BUILTIN;


libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,169 @@ closure color volume_matte() BUILTIN;
...
+//   introduce parameters to describe the sandwitched medium between the layer interfaces.
+//   Such parameterization could then be added on this layer() closure as extra arguments.
+// - Do we want/need parameters for the medium here now, or do we look at that later?
+//
+closure color layer(closure color top, closure color base) BUILTIN;
+​
+// Mix two closures according to a weight. Performs horizontal layering by linear interpolation
+// between the two inputs: "bg∗(1−weight) + fg∗weight".
+//
+closure color mix(closure color bg, closure color fg, float weight);
```
lgritz, line 723: closure mix already exists.

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,169 @@ closure color volume_matte() BUILTIN;
...
+// inside the medium.
+//
+closure color dielectric_bsdf(normal N, vector U, color reflection_tint, color transmission_tint, float roughness_x, float roughness_y, float ior, string distribution) BUILTIN;
+​
+// Constructs a reflection BSDF based on a microfacet reflectance model.
+// Uses a Fresnel curve with complex refraction index for conductors/metals.
+// If an artistic parametrization is preferred the artistic_ior() utility function
+// can be used to convert from artistic to physical parameters.
+//
+closure color conductor_bsdf(normal N, vector U, float roughness_x, float roughness_y, color ior, color extinction, string distribution) BUILTIN;
```
fpsunflower, line 611: Question about the microfacet closures -- are the `roughness_*` arguments meant to be the "alpha" values of the distribution? or is this codifying the `alpha=roughness^2` convention?

It would be good to mention in the comment which values of `distribution` an implementation is expected to support.

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,169 @@ closure color volume_matte() BUILTIN;
...
+// NOTE:
+//  - This is not a node in the MaterialX library, but the surface shader constructor
+//    node has an 'opacity' parameter to control textured cutout opacity.
+//
+closure color transparent_bsdf() BUILTIN;
+​
+// Constructs a BSSRDF for subsurface scattering within a homogeneous medium.
+//
+// TODO:
+// - Do we want/need this dedicated BSSRDF closure? We have had discussions before about
```
fpsunflower, line 645: One thing to consider is that depending on the random walk implementation (especially the boundary handling) - the albedo mapping to volume parameters can be very different.

To me that suggests that the albedo mapping should happen inside the renderer which in turn suggests there should be a dedicated closure for subsurface that can capture the intended "artist friendly" albedo/radius parameter and let the renderer infer the volume parameters that will produce that effect.

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,169 @@ closure color volume_matte() BUILTIN;
...
+​
+// Constructs an EDF emitting light uniformly in all directions.
+//
+closure color uniform_edf(color emittance) BUILTIN;
+
+// Constructs an EDF emitting light inside a cone around the normal direction.
+//
+closure color conical_edf(color emittance, normal N, float inner_angle, float outer_angle) BUILTIN;
+
+// Constructs an EDF emitting light according to a measured IES light profile.
```
fpsunflower, line 681: IES profiles are usually defined for point light sources. By attaching the IES profile within the shader, it implicitly becomes a property of a specific point on the surface which has the effect of "blurring" the profile. Is this what is intended?

It also makes the light potentially quite a bit harder to sample. Some techniques for efficient IES profile sampling assume a constant profile across the surface.

I'm curious if any implementation are supporting this node already.

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,169 @@ closure color volume_matte() BUILTIN;
...
+​
+​
+// -------------------------------------------------------------//
+// Utility functions                                            //
+// -------------------------------------------------------------//
+​
+// Converts the artistic parameterization reflectivity and edge_color to complex IOR values.
+// To be used with the conductor_bsdf() closure.
+//
+void artistic_ior(color reflectivity, color edge_color, output color ior, output color extinction);
```
fpsunflower, line 733: Is the intent to provide the implementation of this function?

If left up to the renderer to provide, it would potentially make the `.oso` files non-portable.

aconty: Also, why the output arg? Can we make the function return the result?

libraries/stdlib/osl/stdosl.h
aconty: Not the roughness^2 convention per se, but I'd vote for a linearized one. And the linearization could change with the distribution. Is up to the implementation to do the roughness -> alpha mapping depending on the distribution.

I would decide a reference distribution like GGX with alpha=roughness^2 and try to keep the other distributions roughness response as close to it as possible?

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,169 @@ closure color volume_matte() BUILTIN;
...
+​
+// Constructs a VDF for pure volumetric absorption.
+//
+closure color absorption_vdf(color absorption) BUILTIN;
+​
+// Constructs a VDF scattering light for a participating medium, based on the Henyey-Greenstein
+// phase function. Forward, backward and uniform scattering is supported and controlled by the
+// anisotropy input.
+//
+closure color anisotropic_vdf(color absorption, color scattering, float anisotropy) BUILTIN;
```
aconty, line 698: absorption_vdf seems redundant when you have this. Also, absorption_vdf is not really a VDF strictly speaking. Should we just have one VDF like anisotropic_vdf? Maybe with (color attenuation, color albedo, ...) parametrization? You can always pass 0 albedo if you just want absorption.

Question: should we expect the params to be premultiplied by density? I would think yes.

At SPI we also chose to have a medium closure that sets all params for the medium enclosed by the current interface (when shading surfaces). It is very useful for glass/liquid objects. Do we want that?

libraries/stdlib/osl/stdosl.h
niklasharrysson: About the roughness, in our current implementation the roughness is squared outside, before given to the closure (following vanilla OSL and MDL). In MaterialX we normally connect a node for remapping a `float roughness` and `float anisotropy` into the pair `(roughness_x, roughness_y)` and the `roughness^2` is done as part of this node.

But I agree it would be better to let the implementation handle this, so the given roughness argument is expected to always give a perceptionally linear response, and the caller don't have to care about this.

niklasharrysson: About the microfacet distributions. On our side we have only required GGX so far. In Arnold I think we support GGX and Beckmann, but GGX is the only one used in for example Autodesk Standard Surface. Curious to hear what others have for microfacet distribution needs?

And yes, we should document the set we decide on.

libraries/stdlib/osl/stdosl.h
niklasharrysson: Good points, we can keep this in if it makes sense. On MaterialX side we will keep this in as a node in the library anyway, there was a request from ILM to keep this in as a artist friendly construct to add SSS. So as long as this can be translated to OSL one way or another we are happy on our side.

libraries/stdlib/osl/stdosl.h
niklasharrysson: Yes, this EDF was included in the PBS library mainly for constructing light shaders. In MaterialX there is a concept of light shaders, where for a point light it could be described as a light shader attached to a locator, and not necessarily a surface. I suppose you could use this EDF to a surface shader as well, but yes then it would be blurring the profile over the surface. The only implementation supporting this node as of now is Nvidia MDL which has a dedicated EDF closure for IES profiles.

I'm not sure OSL was intended to be used for light shaders of local lights in this manner? If not we can remove this for now. In any case the use of this EDF would be very rare and it can definitely be placed at the bottom of the backlog.

libraries/stdlib/osl/stdosl.h
niklasharrysson: Yes, this is the parameter remapping from Gulbrandsen (Artist Friendly Metallic Fresnel), and the implementation will be provided.

niklasharrysson: About the return values, do you mean returning a struct, since there are two floats to return?

libraries/stdlib/osl/stdosl.h
niklasharrysson: Yep, the absorption_vdf if just a subset of anisotropic_vdf and it can be removed unless there is a performance benefit of having this as a special case. But I think you are right we should just remove it.

Your proposal sounds interesting, maybe you could write it out to clarify the full parameterization. Same thing for the special medium closure. And I hope others will chime in as well :)

libraries/stdlib/osl/stdosl.h
aconty: We only support GGX at SPI, it was a great simplification of the codebase :) We also use the (roughness, anisotropy) parametrization to the closures.

libraries/stdlib/osl/stdosl.h
aconty: In the beginning we thought of using OSL for light shading too, and we tried. But we quickly saw it wasn't practical for us and we just define plain emission in the shader. The render is responsible for profiles.

libraries/stdlib/osl/stdosl.h
aconty: Sorry, I didn't notice there were two. Looks good.

libraries/stdlib/osl/stdosl.h
aconty: Sure, this is our medium closure prototype:

closure color medium(color sigma_a, color sigma_s,  float anisotropy, float ior, int priority) BUILTIN;

And it is designed so you return it together with say, dielectric:

Ci = dielectric(N, ...) + medium(sigma_a, ...)

and the semantics is that you are telling the render this is a closed object with that interior medium. You can then use medium tracking with priorities to override relative IOR at the interface if your object is say, underwater. Note that the IOR of the medium is absolute, while the one on the surface interface is relative.

In theory you could throw away the ior param to the BSDFs and drive it entirely from the medium tracking. We haven't done so as most of our shaders don't define a medium, we just override the one in the BSDF when the medium is present.

libraries/stdlib/osl/stdosl.h
portsmouth: The current SSS radius parameter is essentially the desired mean-free-path (MFP) in each color channel. This isn’t really directly related to the rendered color I think, and generally it's quite unintuitive to have a color parameter whose channels contain lengths. (While the provided "albedo" is a vector of dimensionless numbers giving the per-channel scattering albedo, so more easily interpretable as a genuine color).

It could be better to specify a single "depth" which gives the MFP (explicitly in units of scene length), and then an absorption color whose meaning is “the transmittance of a slab of medium of that thickness”. This is the color you would see if white light was transmitted through a 1 MFP thick slab of medium. Then the specified extinction is taken to be -log(absorption color) / MFP.

That is the scheme we use in Arnold's standard surface and standard volume, i.e. there is a transmission_depth and a transmission_color, which are converted into the RGB extinction in the way described.

libraries/stdlib/osl/stdosl.h
portsmouth: I think you might as well supply the reflectivity and edge_color directly to the conductor BSDF, and let the implementation decide how it wants to convert these (which have well-defined meaning as reflected color at normal and near-grazing incidence) internally to the conductor IOR and k.

libraries/stdlib/osl/stdosl.h
portsmouth: Arnold also has a (dielectric) medium closure, which is basically the same as Alejandro described. I think it could be useful to have this explicit in the model here. As discussed in Slack, one could actually replace the SSS with the concept of the interior medium, and map the artist-friendly SSS parameters to the medium parameters.

Though I'm not sure about the adding of dielectric and medium closures denoting the combination of the BSDF and the underlying medium. The BSDF and its surrounding media should ideally be tied together in some explicit way.

niklasharrysson: I agree that using addition for combining them seems a bit unintuitive. Using the layer() closure for this as well might be more logical conceptually:
`Ci = layer(dielectric(N, ...), medium(sigma_s, ...))`
Stating to the renderer we are placing a medium underneath the BSDF interface.

niklasharrysson: About the `medium` closure itself, you are saying the implementation for this would be different from the `anisotropic_vdf` closure, right? With the special handling of IOR and nested priority? So it would make sense to have _both_ the anisotropic_vdf closure and this medium closure as part of the standard closures? If so we should consider to make this medium closure a dedicated MaterialX nodes as well.

portsmouth: I would think of the `medium` mostly just as some container specifying the (homogeneous) properties of the bulk, i.e.:

- IOR
- priority (optional?)
- dispersion (optional?)
- volume/SSS properties

Where the volume properties are:

- sigma_a, sigma_s   (or equivalently, extinction and albedo)
- phase_function

The only distribution function ("VDF") involved here is the phase function, which is usually parametrized by a single anisotropy value. Alternatively, one could just specify the anisotropy value with the phase function some hard-coded form (probably Henyey-Greenstein) or implementation specific.

aconty: Or we could merge everything in a single vdf closure that would add IOR and priority to the volume properties. If you are shading an actual volume those would be unused (although somebody may want to go fancy and bend rays with IOR in the air). The homogeneous/heterogeneous character would come from the context where the closure is evaluated. Unless you guys feel it is important for that distinction to be explicit.

Btw, I'm leaning towards extinction/albedo parametrization for volume. I think we always end up computing those for rendering (sigma_t/albedo).

libraries/stdlib/osl/stdosl.h
MasterZap: I kinda agree with @portsmouth here.

libraries/stdlib/osl/stdosl.h
MasterZap: @portsmouth is right, and this is probably my fault, since I defined it thusly in the spiritual predecessor to the Autodesk Standard Surface. In my poor old head, it kinda made sense to just have the mean free paths scaled by the color, but I think my "mental intent" was exactly what Jamie says, i.e. "if you have this thick slab of something, this is the color you get". Which Jamies math does, and mine doesn't :P 
In my defense, one of my motto's is "I'm not a mathematician - I'm a magician" :)

libraries/stdlib/osl/stdosl.h
niklasharrysson: The `edge_color` here is not a real color, it's the 'g' parameter in Gulbrandsens parameterization, which has a strange indirect and nonlinear relationship to the final color. So I'm not sure we can make such a generalization and let the implementation choose if this is to be interpreted as f90 or 'g'.

We actually had this (Gulbrandsens parameterization) on the `conductor_bsdf` node in the first version of the MaterialX PBS library. But at Siggraph last year a lot of good points where raised about issues with this parameterization (mainly from Naty at the PBS course), so we decided to change it to use the physical eta/kappa parameters by default, and then anyone can remap that if desired as we do with the `artistic_ior` function.

Following Siggraph last year we started a discussion on this topic over at the Standard Surface github repo:
https://github.com/Autodesk/standard-surface/issues/12

On our side we haven't had the bandwidth to invest more time on that discussion yet (other than deciding to move away from using Gulbrandsens model by default). Maybe the Fresnel parameter discussion should continue in that other thread. Until we have resolved that I feel keeping the physical eta/kappa parameterization is maybe the safest bet for now.

Also, note that we already have the `generalized_schlick_bsdf` closure, with full artistic control over f0 and f90. This can be used for both dielectrics and metals if you prefer to have more control (with the risk of getting non-physical results)

libraries/stdlib/osl/stdosl.h
niklasharrysson: @portsmouth Is this new parameterization for subsurface_bssrdf matching what you had in mind?

libraries/stdlib/osl/stdosl.h
niklasharrysson: @portsmouth @aconty Is this new parameterization of anisotropic_vdf matching what you had in mind? I also added ior & priority as documented optional parameters for homogeneous media.

libraries/stdlib/osl/stdosl.h
```diff
@@ -601,13 +608,32 @@ closure color burley_diffuse_bsdf(normal N, color albedo, float roughness) BUILT
...
 closure color dielectric_bsdf(normal N, vector U, color reflection_tint, color transmission_tint, float roughness_x, float roughness_y, float ior, string distribution) BUILTIN;
 ​
 // Constructs a reflection BSDF based on a microfacet reflectance model.
 // Uses a Fresnel curve with complex refraction index for conductors/metals.
 // If an artistic parametrization is preferred the artistic_ior() utility function
 // can be used to convert from artistic to physical parameters.
 //
+//  \param  N                 Normal vector of the surface point beeing shaded.
+//  \param  U                 Tangent vector of the surface point beeing shaded.
+//  \param  roughness_x       Surface roughness in the U direction. Valid range [0,1] with a perceptually linear response over the range.
```
fpsunflower, line 631: Some anisotropic parameterizations give roughness values beyond 1 given a roughness/anisotropy factor in `[0,1]`. For example:

```
alpha_x = roughness^2 * (1 - anisotropy)
alpha_y = roughness^2 * (1 + anisotropy)
```

The alpha values range from `[0,2]` so the corresponding `roughness_x/y` range from `[0,sqrt(2)]` right ?

The anisotropic parameterization in the disney model also goes >1 for the individual roughnesses.

So maybe the bit about valid range could just be left out ?

libraries/stdlib/osl/stdosl.h
```diff
@@ -654,6 +689,10 @@ closure color subsurface_bssrdf(normal N, color albedo, float sss_depth, color s
 // This closure may be vertically layered over a base BSDF, where energy that is not reflected
 // will be transmitted to the base closure.
 //
+//  \param  N           Normal vector of the surface point beeing shaded.
+//  \param  albedo      Surface albedo.
+//  \param  roughness   Surface roughness [0,1].
+//
 closure color sheen_bsdf(normal N, color albedo, float roughness) BUILTIN;
 ​
 // Adds an iridescent thin film layer over a microfacet base BSDF. This must be layered over
```
fpsunflower, line 698: I think requiring the use of the `layer` closure will potentially make the implementation rather awkward. Having extra arguments to the 3 closures that accept it would be easier (they could be optional arguments).

But maybe its better to wait until we can attempt the implementation in `testrender` to decide.

libraries/stdlib/osl/stdosl.h
```diff
@@ -673,10 +715,18 @@ closure color thin_film_bsdf(float thickness, float ior) BUILTIN;
...
 // Constructs an EDF emitting light uniformly in all directions.
 //
+//  \param  emittance   Radiant emittance of light leaving the surface.
+//
 closure color uniform_edf(color emittance) BUILTIN;
 
 // Constructs an EDF emitting light inside a cone around the normal direction.
 //
+//  \param  emittance   Radiant emittance of light leaving the surface.
+//  \param  N           Cone direction vector.
```
fpsunflower, line 725: Just curious what the role of `N` is here. Bump mapped emitters? I haven't seen any mesh light implementation that tries to accurately model the role of shading normals - let alone bump-mapped shading normals. I'm curious if anyone else feels like this is worth requiring.

For example - this introduces all the inconsistencies of shading normals to the emissive context. Some portion of the hemisphere around `N` will be self-shadowed against the hemisphere around `Ng`.

We had closures like this in the original OSL spec a long time ago but wound up removing them because they were hard to implement exactly. Besides, controls like spot angles are usually desired relative to the center of the light -- not across its area. The angular falloff of a spotlight typically should not change when you change the area of the light. But when described as a closure there would be an implicit "blur" by the surface area of the mesh you've assigned to.

fpsunflower (commented):
Aside from my small comments below -- I think this is looking pretty good.

We eventually want this header to live in the OSL repo, right ?

libraries/stdlib/osl/stdosl.h
aconty: I'm ok with relying on layer(). Our latest implementation is moving away from nested closures because GPU doesn't like recursion. The layer() model would be more convenient. The way we did it might not be mature enough, but I think layer() gives you enough freedom. You could even keep nesting recursively if you wanted.

On the other hand, nesting closures makes it more cumbersome if you want to do it without recursion. What did you have in mind?

libraries/stdlib/osl/stdosl.h
aconty: +1

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,226 @@ closure color volume_matte() BUILTIN;
...
+// Constructs a VDF scattering light for a participating medium, based on the Henyey-Greenstein
+// phase function. Forward, backward and uniform scattering is supported and controlled by the
+// anisotropy input.
+//
+//  \param  albedo      Volume single scattering albedo.
+//  \param  extinction  Volume extinction coefficient.
+//  \param  anisotropy  Scattering anisotropy [-1,1]. Negative values give backwards scattering, positive values give forward scattering, 
+//                      and 0.0 gives uniform scattering.
+//  \param  ior         Optional float parameter for refraction index of a homogeneous medium.
+//  \param  priority    Optional int parameter for priority of a homogeneous medium (for nested dielectrics).
```
aconty, line 746: Is it too verbose if we make these formal parameters instead of optional? But I admit we've been using optional args for new features added after an API has already been published, so when I see a new API I tend to visualize everything as formal param. Maybe because OSL doesn't have syntax for declaring keyword args and you have to read the comments.

aconty:
This is looking very good!

niklasharrysson:
> We eventually want this header to live in the OSL repo, right ?
@fpsunflower Yes absolutely. I probably should have made this PR directly to OSL in the first place.. But as soon as this initial design has been stabilized I will make a proper PR to the OSL repo (and I think we're getting close). Then we can finalize it there and anyone who wants to contribute can start adding implementations in testrender.

I'll make sure to merge this PR first, saving all discussions here, and will link to this from the PR in OSL so all conversation history is preserved.

libraries/stdlib/osl/stdosl.h
niklasharrysson: I think Chris was only referring to the implementation of thin film here. Since thin film is not a real closure, and more of a closure modifier (modifying Fresnel), implementing this as optional parameters for the closures that supports it would probably be easier than implementing it through the layer operator (as noted in the TODO comment below).

This is what we do in Arnold on our side, if thickness is set and > 0.0 the thin film Fresnel effect is enabled.

libraries/stdlib/osl/stdosl.h
niklasharrysson: N here is mainly to set the direction of the cone when used in a light shader context when the shader is attached to a locator and not a surface. However, since the locator needs a transform anyway this is a bit redundant as we could instead define a single fixed direction and just add rotation to the transform.

In addition, for the same reasons we removed the measured_edf closure from this draft we could remove this one as well. It's main reason for being part of the Mtlx library is for defining light shaders for spot lights. As with the measured_edf closure this is supported in Nvidia MDL. But since OSL is currently not designed to describe light shaders with closures this way we can just remove this for now, and take a new look at this later if really needed.

libraries/stdlib/osl/stdosl.h
niklasharrysson: I agree it's not ideal one has to read the docs to know what the full parameterization is. On the other hand the ior and priority parameters would only be used in the case of a dielectric / homogenous medium, so it's a bit awkward having to set values for those in other contexts.

Does OSL support default values on parameters of built-in functions? If so we could make them formal parameters with default values, so you only set them when needed.

If not I suppose another option is to define a wrapper function for homogenous media with the full parameterization declared:

```
closure color media_vdf(color albedo, color extinction, float anisotropy, float ior, int priority)
{
    return anisotropic_vdf(albedo, extinction, anisotropy, "ior", ior, "priority", priority);
}
```

Otherwise, as you mentioned, the most elegant solution would be to add a keyword args syntax for OSL. I think I read something about this topic in the meeting notes from last weeks OSL meeting.

Getting off topic here but just wanted to mention that in Nvidia's MDL there is such a syntax (they call it _named arguments_ ) that looks like this:

```
float foo(float a = 7.0, float b = 42.0)
{
    return a + b;
}

float x1 = foo()             // equals 49.0
float x2 = foo(4.0, 5.0)     // equals  9.0
float x3 = foo(a:4.0, b:5.0) // equals  9.0
float x4 = foo(b:6.0, a:4.0) // equals 10.0
float x5 = foo(a:3.0)        // equals 45.0
float x6 = foo(b:8)          // result 15.0
```

lgritz: I like that syntax a lot. We talked about it at the last OSL TSC meeting, and I'm in favor of adding it.

libraries/stdlib/osl/stdosl.h
niklasharrysson: For the example above I think roughness_x/y should still be [0,1], right? If it's [0,sqrt(2)] you would get the alpha values range [0,4], unless I'm missing something.

But in general it feels like we should be able to have a defined range [0,1] where 0 means perfectly smooth and 1 means fully diffuse, and the implementation should try to follow this behavior, in the best way possible depending on the distribution used. And independent from an anisotropic parameterization, if used on the outside.

libraries/stdlib/osl/stdosl.h
aconty: Oh, right

libraries/stdlib/osl/stdosl.h
aconty: Meanwhile, your wrapper proposal sounds good to me.

That syntax is something we should have done a long time ago. The current ```"key", value);``` thing is ugly. But I would choose ```key = value);``` instead of ```key:value);``` if you give me the option :)

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,221 @@ closure color volume_matte() BUILTIN;
...
+//  \param  N           Normal vector of the surface point beeing shaded.
+//  \param  albedo      Surface albedo.
+//  \param  sss_depth   Mean-free path in units of scene length.
+//  \param  sss_color   Scattering color / transmittance. The desired color resulting from white light transmitted a distance of 'sss_depth'
+//                      through the surface.
+//  \param  anisotropy  Scattering anisotropy [-1,1]. Negative values give backwards scattering, positive values give forward scattering, 
+//                      and 0.0 gives uniform scattering.
+//  \param  label       Optional string parameter to name this component. For use in AOVs / LPEs.
+//
+closure color subsurface_bssrdf(normal N, color albedo, float sss_depth, color sss_color, float anisotropy) BUILTIN;
```
portsmouth, line 701: - Is the supplied albedo really intended to be the "surface" albedo, or the scattering albedo (i.e. scattering coefficient divided by absorption coefficient, of the SSS medium)?   In (ADSK) Arnold the scattering albedo is specified (the standard surface parameter is called "subsurface_color").  The surface exit/entry events are modelled as diffuse white transmission lobes (which in itself is an approximation ignoring the surrounding layer structure, but a reasonable one).

- Are we supplying the raw volumetric parameters of the medium, rather than artist friendly "effective" colors? There is some albedo remapping in (ADSK) Arnold which tries to ensure that the supplied scattering albedo roughly matches the resulting color.

- Nitpick: it should probably be "mean-free-path", not "mean-free path".

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,221 @@ closure color volume_matte() BUILTIN;
...
+​
+​
+// -------------------------------------------------------------//
+// VDF closures                                                 //
+// -------------------------------------------------------------//
+​
+// Constructs a VDF scattering light for a participating medium, based on the Henyey-Greenstein
+// phase function. Forward, backward and uniform scattering is supported and controlled by the
+// anisotropy input.
+//
```
portsmouth, line 734: I'm not super keen on the term "VDF",  as what we are describing here is not a single function but really all the properties of the "medium" through which light is propagating (in the sense of the term "participating medium" in graphics, or [optical medium](https://en.wikipedia.org/wiki/Optical_medium) in physics).

But maybe it is a reasonable loose terminology (as one could consider all of albedo, extinction and anisotropy as representing the cross sections for scattering/absorption, which are generally functions of incident/output direction).

libraries/stdlib/osl/stdosl.h
fpsunflower: `roughness` and `anisotropy` are both in `[0,1]`. Therefore `alpha_x` is in `[0,1]` but `alpha_y` is in `[0,2]`.

Assuming that `roughness_y=sqrt(alpha_y)`, then the range for that one is `[0,sqrt(2)]`.

I'm not sure this really impacts much since the "requirement" of the values being in `[0,1]` is more of a suggestion, but just mentioning it in case.

I am curious how many other parameterizations of anisotropy are out there. I believe there is mostly this one and the disney model:

```
    float aspect = sqrt(1-anisotropic*.9);
    float ax = sqr(roughness)/aspect;
    float ay = sqr(roughness)*aspect;
```

The disney approach (if I am doing my math right) has an upper bound for `roughness_x` of ~1.778.

libraries/stdlib/osl/stdosl.h
niklasharrysson: Great points, thanks @portsmouth. The first point is a copy-paste error in the docs and I'll fix that.

I agree it would be more clear to use the word "transmission", so maybe we should just change both names to transmission_depth and transmission_color. It feels artist friendly and intuitive to say this is the expected color of light transmitted this distance through the medium.

For the third point I personally like it if we can keep it as artist friendly as possible. I have a feeling this discussion about volumetric closures will result in an update of the corresponding nodes in the MaterialX library. And at least on that side we want to keep it as artist friendly as possible. So a remapping needed to ensure this is fine by me.

libraries/stdlib/osl/stdosl.h
niklasharrysson: Yep, I understand your point. My personal opinion is that it's better to have a simple and consistent naming of things (even if it's not 100% correct from a physical or mathematical point of view), then to have a complicated and inconsistent naming just to make sure the correct terminology from physics is used everywhere.

At least in the MaterialX library we are generalizing quite a bit, and consider a VDF to be "anything that has to do with light interacting with volumes / participating media", an EDF to be "anything that has to do with light emission" and a BSDF "anything that has to do with light interacting with surfaces". And there are some obvious incorrect uses of this, like for the node subsurface_bsdf which contains both surface and volume interactions, but we still call it a BSDF because it's used as a BSDF by artists. So an artist don't have to bother learning the difference between a BSSRDF and a BSDF.

libraries/stdlib/osl/stdosl.h
niklasharrysson: If we are adding the media_vdf wrapper (or even a separate built-in closure for this) would it then make sense to use the same parameterization as for subsurface_bssrdf, with the artist-friendly transmission_depth and transmission_color? For a closure that is dedicated for dielectric homogenous media, this seems much easier to control.

libraries/stdlib/osl/stdosl.h
niklasharrysson: Thanks for those additional details! Makes sense, and I have removed the conical_edf as well for now.

There is the whole subject of lights standardization that I suppose we will get to eventually, including what's done in USD / UsdLux etc. So let's defer this to a later time.

libraries/stdlib/osl/stdosl.h
niklasharrysson: Right, as long as the user-facing roughness is in [0,1] the ranges used under the hood is less important. I will update the doc strings.

For anisotropic parameterizations, the one you describe above is the only one we use in the MaterialX library. It's the same one being used in Autodesk Standard Surface.

libraries/stdlib/osl/stdosl.h
MasterZap: Nerdery: Technically, IES profiles describe what happens at a distance from the light that is 5x the lights own largest physical size axis. Meaning, they do not define near-field behavior, only far-field behavior.
I've been through this exact mess in about 3 renderers.... 

Some cheat one way, ny just applying the IES profile per sample, and hope the emitting area is small enough so that the distant result is "okay".
Some cheat the other way, and somehow apply distant magic like @fpsunflower mentions. 

Neither solution is ideal, THB. And out in reality, the emission profile of the light is a complicated result from it's reflector geometry and internal shadowing surfaces, and ain't nobody wanna model *that* for real :P

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,237 @@ closure color volume_matte() BUILTIN;
...
+//                              coefficient of the medium.
+//  \param  transmission_color  Desired color resulting from white light transmitted a distance of 'transmission_depth' through the medium.
+//                              Together with transmission_depth this determines the extinction coefficient of the medium.
+//  \param  anisotropy          Scattering anisotropy [-1,1]. Negative values give backwards scattering, positive values give forward scattering, 
+//                              and 0.0 gives uniform scattering.
+//  \param  ior                 Refraction index of the medium.
+//  \param  priority            Priority of this medium (for nested dielectrics).
+//  \param  label               Optional string parameter to name this component. For use in AOVs / LPEs.
+//
+closure color medium_vdf(color albedo, float transmission_depth, color transmission_color, float anisotropy, float ior, int priority) BUILTIN;
```
niklasharrysson, line 760: @portsmouth @aconty With the use of optional params for ior and priority, as well as the alternative of using transmission_depth/transmission_color to control the extinction coefficient, I felt it was justified to instead break this out to a separate VDF for homogenous media. Let me know what you think?

aconty: Yeah, that makes sense.

libraries/stdlib/osl/stdosl.h
```diff
@@ -573,6 +573,168 @@ closure color volume_matte() BUILTIN;
...
+// closure may be layered over a VDF closure describing the surface interior to handle
+// absorption and scattering inside the medium.
+//
+// TODO:
+// - Transmission handling for this node has not been fully defined yet in MaterialX.
+//   In particular how is IOR for refractions derived from the f0, f90 parameterization?
+//   Do we just derive it from f0? For an artist it seems hard to control refractions that way.
+//   @Jonathan Stone you mentioned Lucasfilm (Naty Hoffman) had some feedback and ideas for how to handle this?
+//
+closure color generalized_schlick_bsdf(normal N, vector U, color reflection_tint, color transmission_tint, float roughness_x, float roughness_y, color f0, color f90, float exponent, string distribution) BUILTIN;
```
nhoffman-lucasfilm, line 670: How is external IOR / IOR layering tracked? If it's done behind the scenes by passing the IOR ratio to the closure, that may not be enough information for the generalized Schlick BSDF closure.

nhoffman-lucasfilm (commented):
What was the reasoning behind changing thin-film interference from a separate BSDF to a set of optional parameters on all BSDFs? I couldn't find any discussion about that here or on the Slack.

libraries/stdlib/osl/stdosl.h
niklasharrysson: @nhoffman-lucasfilm That's a great question. I believe this will be tracked through the use of the medium_vdf closure that describes the medium which the light travels through before hitting a new interface. The medium_vdf has an `ior` parameter that is an absolute IOR value for the medium.

For the primary intersection I guess air/vacuum is assumed outside. I'm not sure how it's handled for situations where the camera starts inside, like a camera under water, but I suppose a media closure could be attached to the camera.

For cases where BSDFs are layered without the use of media in-between I think all the needed information should be available still, since IOR could be derived from the parameters on each BSDF.

But maybe people from Imageworks, Adsk/Arnold or other teams that have implemented OSL closures in practice can fill in more details here?

niklasharrysson:
> What was the reasoning behind changing thin-film interference from a separate BSDF to a set of optional parameters on all BSDFs? I couldn't find any discussion about that here or on the Slack.

This was discussed in one of the threads marked as resolved, here: 
https://github.com/materialx/MaterialX/pull/614#discussion_r611825067

libraries/stdlib/osl/stdosl.h
aconty: We attach medium data to every ray (if present). If a camera is inside some medium we find the object and evaluate the shader to grab what you're calling medium_vdf here. We also compute relative IOR outside the shader. Basically we intercept the BSDF creation from closure data and override the IOR parameters.

May be worth mentioning that we have an optional parameter to all bsdfs called "force_eta" (float from 0 to 1). This allows the shader, and the artist, to override the dynamic IOR. For instance, if the look changes too much for them when underwater, they can tone down the change.

portsmouth: We track the IOR of the bulk medium as the ray transmits through dielectric boundaries, via the usual priority system approach. The interior bulk medium IOR is taken from the specular lobe in the standard surface model (i.e. this lobe effectively defines the IOR of the bulk). We don't take into account IOR jumps between the various layers in the model though (e.g the coat BSDF "above" the specular layer, is not aware of the specular IOR), which is probably a reasonable approach considering the closure mixture model is a somewhat ad-hoc representation of the layered structure anyway.

portsmouth: We don't currently handle the "camera under water" situation, unfortunately. Though it seems that determining this is a matter for the renderer to handle, e.g. by casting a ray to infinity and tracking the media.

libraries/stdlib/osl/stdosl.h
portsmouth: Hm, it seems that `anisotropic_vdf` is basically specifying a (homogeneous) volumetric medium, and `medium_vdf` is doing that plus specifying IOR and priority (of the "embedding" dielectric medium), except with a different way of parameterizing the extinction coefficient..  Wouldn't it be easier to just have one `vdf` for the volume properties (maybe using the transmission parameterization, which is nicer than raw extinction),  which you pass into the full `medium_vdf` to define the general case where the volume is embedded in dielectric?

libraries/stdlib/osl/stdosl.h
nhoffman-lucasfilm: "We also compute relative IOR outside the shader. Basically we intercept the BSDF creation from closure data and override the IOR parameters" - in other words the ratio (external_ior / bulk_ior) is passed into the closure's "ior" parameter? In that case, for the generalized_schlick_bsdf closure a different parameter adjustment would be needed, adjusting f0 using something like the equations from slides 107-108 or slide 111 in https://blog.selfshadow.com/publications/s2020-shading-course/hoffman/s2020_pbs_hoffman_slides.pdf (a similar adjustment might need to be made for the f90 parameter - I've spent a little bit of time on that but don't have a formula for that one ready yet). Would this kind of closure-specific parameter adjustment be feasible, or is does that logic need to be agnostic to which closure is used?

nhoffman-lucasfilm: Related question - is there a way to determine the reflection's "sidedness" inside the closure? When using IOR as a parameter, internal vs. external reflection can be easily determined from the relative IOR - if it is greater or small than 1. But 1/eta and eta produce the same value for f0, so an extra bit (literally) of information is needed to disambiguate those two cases for the generalized_schlick_bsdf closure.

aconty: Not overriding the closure parameter itself, but the BSDF internal object that gets created on the renderer side. We do some translation for non-ior parametrizations, but we haven't polished that a lot, it just works. Also, this doesn't affect internal BSDF layering, we don't track stuff there.

About sidedness, our convention: The shader always puts the IOR of the front surface regardless of what side is being rendered. The implementation is responsible for doing 1/eta if on the backside. We did this to remove the burden from the shaders, but other people might have other opinions.

libraries/stdlib/osl/stdosl.h
niklasharrysson: To my mind `anisotropic_vdf` is a closure for general participating media (both homogenous and heterogeneous), whereas `media_vdf` is a specialization for homogenous dielectric media. So a specialization for a common case, where it more efficiently can handle strictly homogenous media, as well as extra logic for priority etc. with nested dielectrics. It's similar to how we have `subsurface_bssrdf` as a separate closure specialized for that use case.

I suppose we could change the parameterization of `anisotropic_vdf` to use the transmission parameterization instead of raw extinction here as well, if that makes sense for this more general case?

However I'm not sure about removing these parameters from `media_vdf` to replace them with a closure input (for plugging in an anisotropic_vdf, if I understood your proposal correctly). To my mind this would make `media_vdf` harder to implement and more confusing to use for OSL programmers as it allows plugging any closure into this input.

niklasharrysson:
A PR with these changes has now been made to OSL master, here: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1371

Please continue the discussion there if there are outstanding questions. When that PR has been merged you are all very welcome to contribute to the reference implementation of the closures in OSL testrender.

@jstone-lucasfilm I suggest we merge this PR in now, so all conversation history is saved here.

PR #615, opened 2021-03-29
MDL texture lookup code generation updates

bernardkwok:
Issues Addressed
- Fixes erroneous generation of a texture_2d() reference with "/" as an argument when no value is specified.
- Make texture_2d() lookups input arguments as opposed to always inlined.

### DIFF
--- a/source/MaterialXGenMdl/MdlSyntax.cpp
+++ b/source/MaterialXGenMdl/MdlSyntax.cpp
@@ -37,13 +37,19 @@ class MdlFilenameTypeSyntax : public ScalarTypeSyntax
 
     string getValue(const Value& value, bool /*uniform*/) const override
     {
+        const string outputValue = value.getValueString();
+        if (outputValue.empty() || outputValue == "/")
+        {
+            return getDefaultValue(true);
+        }
+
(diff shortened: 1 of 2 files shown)


source/MaterialXGenMdl/MdlSyntax.cpp
```diff
@@ -37,13 +37,19 @@ class MdlFilenameTypeSyntax : public ScalarTypeSyntax
 
     string getValue(const Value& value, bool /*uniform*/) const override
     {
+        const string outputValue = value.getValueString();
+        if (outputValue.empty() || outputValue == "/")
+        {
+            return getDefaultValue(true);
```
niklasharrysson, line 43: Instead of always passing in true, should the argument `uniform` be passed in?

niklasharrysson (approved):
LGTM, and I only had a minor remark above. Thanks Bernard

jstone-lucasfilm:
These changes look good to me, thanks @bernardkwok.  Let's move forward with this for now, allowing us to take advantage of the MDL improvements in environments such as Omniverse, and we can continue refining as needed.

PR #618, opened 2021-04-23
Allow building shared libraries on Windows

### LINKED ISSUE #610: Unable to build shared libraries under Windows
This seems to have been a conscious choice, but I'm wondering what the impediment is to building shared libraries on Windows? This seems like a problem for integrating MatX into DCCs that need to run on Windows... Specifically if Houdini wants to add MatX support to USD and to some other part(s) of Houdini, we'd need to link MatX into multiple libraries that would load simultaneously which at best is a waste or time and memory, and at worst will cause crashes like in #485...

Are there any other known Windows-specific limitations to MatX?

marktucker:
These changes allow MaterialX to be built under Windows with MATERIALX_BUILD_SHARED_LIBS set to true. New "MX_<LIB>_API" symbols exist for each library to allow classes and functions to be explicitly exported from the resulting DLLs.

Fixes #610


source/MaterialXFormat/Export.h
```diff
@@ -0,0 +1,22 @@
...
+#ifndef MATERIALX_FORMAT_EXPORT_H
+#define MATERIALX_FORMAT_EXPORT_H
+
+#include <MaterialXCore/Library.h>
+
+/// @file
+/// Macros for declaring imported and exported symbols.
+
+#if defined(MATERIALX_FORMAT_EXPORTS)
+    #define MX_FORMAT_API MATERIALX_SYMBOL_EXPORT
```
bernardkwok, line 15: A quick question here. Is there a need to have separate macros for each library. For example, can there just be an `MX_API` instead of `MATERIALX_SYMBOL_EXPORT` and not have `MX_CORE_API`, `MX_FORMAT_API` etc.

Just wondering if there a case where only specific libraries will have their symbols exported but not others ?

marktucker: The separate macros per-library are not too allow only specific libraries to export their symbols. It's because the different MaterialX libraries import symbols from each other. So Format imports Core symbols while exporting Format symbols. So when building the Format library, MX_FORMAT_API is set to "export", but MX_CORE_API is set to "import".

bernardkwok: Yes. Jonathan explained this context information as well -- which I didn't see. Please disregard this comment  :)

jstone-lucasfilm:
This looks excellent to me, and major thanks to @marktucker for taking the time to develop and refine this new functionality.  I'll give others a chance to weigh in, but I have no objections to merging this to MaterialX v1.38.1.

meshula:
Congrats on landing this!

PR #626, opened 2021-05-11
Added hybrid texture/procedural brick example

dlarsson-adobe:


jstone-lucasfilm:
This is a great contribution to the MaterialX project, a material example with both procedural elements and a thoughtful artistic interface.  Thanks @dlarsson-adobe!

PR #630, opened 2021-05-15
Default interface value fix

bernardkwok:
Fix to check for any associated `interfacename` input when creating a shader node.
Also fixes to always used resolved value for value initialization on ShaderNode and ShaderGraphs.

Unit test results:
[interfacename_tests.pdf](https://github.com/materialx/MaterialX/files/6488356/interfacename_tests.pdf)
of note is the specific graph which had issues:
image


bernardkwok:
Close this for now as more discussion is required.

PR #633, opened 2021-05-17
Improved support for compound nodegraphs

jstone-lucasfilm:
- Update shader code and UI generation to support compound nodegraphs with input bindings.
- Convert the standard_surface_brick_procedural and standard_surface_marble_solid examples to compound nodegraphs with input bindings, giving them an improved UI in the property editor.
- Add viewer option "Show All Inputs", which allows unbound shader inputs to be displayed in the property editor.

Standard Surface Brick in MaterialXView with UI:
StandardSurface_BrickProcedural_UI

Standard Surface Marble in MaterialXView with UI:
StandardSurface_MarbleSolid_UI

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -6,12 +6,8 @@
 #include <MaterialXGenShader/ShaderGraph.h>
 
 #include <MaterialXGenShader/GenContext.h>
-#include <MaterialXGenShader/ShaderNodeImpl.h>
 #include <MaterialXGenShader/ShaderGenerator.h>
-#include <MaterialXGenShader/TypeDesc.h>
 #include <MaterialXGenShader/Util.h>
-
-#include <MaterialXCore/Document.h>
 
 #include <iostream>
 #include <queue>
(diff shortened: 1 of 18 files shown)


source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -36,33 +32,39 @@ ShaderGraph::ShaderGraph(const ShaderGraph* parent, const string& name, ConstDoc
...
+        ValuePtr portValue = nodeInput->getResolvedValue();
+        if (!portValue)
         {
-            ShaderGraphInputSocket* inputSocket = nullptr;
-            const string& portValue = port->getResolvedValueString();
-            std::pair<const TypeDesc*, ValuePtr> enumResult;
-            const string& enumNames = port->getAttribute(ValueElement::ENUM_ATTRIBUTE);
-            const TypeDesc* portType = TypeDesc::get(port->getType());
-            if (context.getShaderGenerator().getSyntax().remapEnumeration(portValue, portType, enumNames, enumResult))
+            InputPtr interfaceInput = nodeInput->getInterfaceInput();
```
bernardkwok, line 41: Though we need this at the node level, I don't believe that it's possible for a nodegraph to have an Input with an "interfacename". I think I earlier has this logic as well. Probably best to not add this in unless there is a case for this?

source/MaterialXRender/Util.cpp
```diff
@@ -231,51 +238,90 @@ unsigned int getUIProperties(ConstValueElementPtr nodeDefElement, UIProperties&
...
+    // Generated an ordered map of shader inputs.
+    using ShaderInputPair = std::pair<InputPtr, ShaderPort*>;
+    std::map<int, ShaderInputPair> shaderInputMap;
+    for (ShaderPort* variable : block.getVariableOrder())
     {
-        return getUIProperties(valueElement, uiProperties);
-    }
-    return 0;
-}
+        if (!variable->getValue())
```
bernardkwok, line 249: I'm not quite sure we want to prune these out. e.g. I think this will prune out empty filenames on image nodes.

bernardkwok (commented):
Looks great! 
Just the comment that we should not look for `interfacename` associations on ShaderGraph inputs as it's currently not somehting possible.

source/MaterialXGenShader/ShaderGraph.cpp
jstone-lucasfilm: Good call, and I think you're correct that this case cannot occur.  I'll remove the test.

source/MaterialXRender/Util.cpp
jstone-lucasfilm: In this case, I believe getValue() will only return nullptr if no value at all is bound to the shader port, so we should be safe in the case of an empty filename.

bernardkwok: I think this goes towards another issue logged that input filenames which have no value are not showing up in the UI.
As I'm not sure yet if this is the cause of pruning vs pruning of uniforms curing code gen what you have is the same logic as before just a better early out. All looks good. Thanks.

bernardkwok (approved):

PR #637, opened 2021-05-31
Merge from ADSK to LucasFilm main

bernardkwok:
Proposed updates from Autodesk fork back to LucasFilm main. 

### Features
- Add support for local for numierc Value types.
- Addition of new unlit shader surface shader to standard pbr library.
- Add in new baking option to allow for baking to image using non-normalized texture coordinates. Required sometime for real-world unit mapping.
- Allow for "gamma" shader to be used instead of sRGB buffers. This allows for some color output transformation without relying on hardware SRGB buffers which are not support in MESA configurations. Gamma control is part of MaterialXView.
- Enable arm64 for macOS Xcode 12 CI build.

### Fixes / Additions
- Enhance hardware transparency check including handling when transparency is defined by an graph definition.
- Add token resolving to generic string resolvers. Fix code generation to use resolved values.
- Fixes so that derived targets can use units and color management properly. (Was not finding base class targets)
- Add `requiresLighting()` to shader generators to allow derived class to use this logic for custom direct lighting.
- Fixes to node definition "flattening" to take into account upstream nodegraphs, baking from surface shaders and handling baking when multiple children within a nodegraph are flattened.
- Minor MDL fix for texture pathing.
- Update unit test configuration to have more modular sets of options texture baing and wedging.

### Test Results
[unit_tests.pdf](https://github.com/materialx/MaterialX/files/6573016/unit_tests.pdf)


source/MaterialXCore/Element.cpp
```diff
@@ -427,16 +427,24 @@ StringResolverPtr Element::createStringResolver(const string& geom) const
...
         }
-    }
+
+    // Add element tokens
+    addTokens(resolver);
 
     return resolver;
+    }
+
+void Element::addTokens(StringResolverPtr& resolver) const
```
jstone-lucasfilm, line 440: With Element::addTokens as a virtual method, I'm finding it challenging to reason about the expected behavior for any specific Element subclass, and I wonder if this new feature might be more straightforward as a single, non-virtual method named `StringResolved::addTokenSubstitutions(ConstElementPtr elem)`.  This would make it clearer (to my eyes) what tokens will be added for any given input Element, including traversal to siblings and related interfaces.  What are your thoughts?

libraries/pbrlib/pbrlib_defs.mtlx
```diff
@@ -246,8 +246,46 @@
...
     <output name="out" type="lightshader" />
   </nodedef>
-
+   
   <!--
+    Node: unlit_surface
+    Uniform emission surface shader. Variants exist for the following input types: color3, color4, vec2, vec3, vec4, and float.
+  -->
+  <nodedef name="ND_unlit_surface_color4" node="unlit_surface" version="1.0" isdefaultversion="true" nodegroup="pbr" doc="Uniform emission surface shader which accepts a color4 input" >
+      <input name="color" type="color4" value="1, 1, 1, 1" />
```
jstone-lucasfilm, line 255: Since the new "unlit_surface" node supports scalar and vector types in addition to colors, would it make more sense for its first input to be named "value" instead of "color"?  This would mirror the naming convention for the first input of the "constant" node, which also supports scalar and vector types.

jstone-lucasfilm (commented):
Thanks for putting this proposal together, @bernardkwok!  I had two initial thoughts on the new features that are added in this changelist.

libraries/pbrlib/pbrlib_defs.mtlx
bernardkwok: This one is interesting as the second input is "opacity" so thought that "color" makes sense as even though the type may be vec3 it's interpreted as a color. 

"value" could work as well as I have no strong leaning either way but it would be "value" + "opacity"

source/MaterialXCore/Element.cpp
bernardkwok: I think these are some of the points we discussed previously:

1. One of the reasons for this set up was that each element type knows how a token should be added. The alternative would mean the single routine would need to know about derived types and branch on logic.  For example: geomInfo's "<>" and application data "[]" usage would need to be in one piece of logic.

2. Each string resolver usage would need to know how to find which elements to consider for adding tokens and then make this call. As this call would not be "required", it could be omitted and results could be inconsistent. Also having to make an explicit new call would defeat the goal of making token substitution "transparent" to the user -- i.e. current integrations don't need to do any new work.

bernardkwok: I do agree that it is a bit hard to see the overall logic. Perhaps this could be documented ?

PR #643, opened 2021-06-23
Add initial Lama BxDF node definitions and tests.

kngo-lucasfilm:
Note: some of the connections have been commented out as they're not yet supported e.g. using custom bsdfs as the top layer or layering over a vdf.


libraries/bxdf/lama/lama_layer.mtlx
```diff
@@ -0,0 +1,50 @@
...
+    </multiply>
+    <layer name="layer" type="BSDF">
+      <input name="top" type="BSDF" nodename="mul" />
+      <input name="base" type="BSDF" interfacename="materialBase" />
+    </layer>
+    <output name="out" type="BSDF" nodename="layer" />
+  </nodegraph>
+
+  <!-- LamaLayer for EDFs -->
+  <!-- TODO: uncomment once layering of EDFS is supported.
```
niklasharrysson, line 30: I'm curious what vertical layering of emission means in practice? Since there are no reflection or absorption of light involved at the top component (as in the BSDF case), how would you attenuate the base component?

So as far as I know there are no plans to add an EDF/EDF version of the `layer` node in MaterialX. Is there a use-case for this that we are missing with using a `mix` of EDF's instead?

libraries/bxdf/lama/lama_dielectric.mtlx
```diff
@@ -0,0 +1,167 @@
...
+      <input name="tangent" type="vector3" nodename="tangent_rotate_normalize" />
+      <input name="distribution" type="string" value="ggx" />
+      <input name="scatter_mode" type="string" value="R" />
+    </dielectric_bsdf>
+
+    <!-- BSDF -->
+    <add name="dielectric_bsdf" type="BSDF">
+      <input name="in1" type="BSDF" nodename="reflection_bsdf" />
+      <input name="in2" type="BSDF" nodename="transmission_layer" />
+    </add>
```
niklasharrysson, line 161: To my mind it would be more accurate to use a BSDF `layer` node than a BSDF `add` node here. To make sure the base is attenuated according to the top component's reflectance.

Using an `add` node implies that the base transmission BSDF must include an inverse Fresnel factor. This is not the case for the implementation of dielectric_bsdf in "T" scatter mode (afaik).

libraries/bxdf/lama/lama_layer.mtlx
kngo-lucasfilm: @niklasharrysson thanks for pointing this out. I've removed the EDF layering bits. These were just placeholders added for completeness, but it's now clear that it doesn't really make sense to have them.

jstone-lucasfilm (approved):
This looks ready to merge, and we can continue to refine the MaterialX Lama graphs and their documentation in future updates.

Thanks for this excellent work, @kngo-lucasfilm!  I also want to call out Andre Mazzone, Sam Cordingley, and Chong Deng on the ILM Shading team for their critical contributions to this project, and @niklasharrysson for his key insights on structure and code generation.

PR #656, opened 2021-07-22
Fix issue with generation of input metadata in OSL

niklasharrysson:
This change list fixes an issue with metadata generation for inputs of type Vector2, Vector4 and Color4. These types are custom structs in OSL, and metadata declarations with value assignments for structs are not supported. The OSL compiler raise a syntax error in these cases.

With this change we prevent metadata generation for these types, so it's just a workaround. A proper fix would be to support this on OSL side, and preferably let these data types become first class native data types in OSL.


source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -451,39 +466,43 @@ void OslShaderGenerator::emitShaderInputs(const VariableBlock& inputs, ShaderSta
...
             if (metadata)
             {
                 for (size_t j = 0; j < metadata->size(); ++j)
                 {
                     const ShaderMetadata& data = metadata->at(j);
-                    const string& delim = (widgetMetadata || j < metadata->size() - 1) ? Syntax::COMMA : EMPTY_STRING;
-                    const string& dataType = _syntax->getTypeName(data.type);
-                    const string dataValue = _syntax->getValue(data.type, *data.value, true);
-                    emitLine(dataType + " " + data.name + " = " + dataValue + delim, stage, false);
+                    if (METADATA_TYPE_BLACKLIST.count(data.type) == 0)
```
niklasharrysson, line 481: This guard prevents generation of code like:
```
    vector2 uvtiling = {3, 3}
    [[
        vector2 min = {0, 0},      // Struct assignments in this context gives syntax error
        vector2 max = {25, 25}  //
    ]],
```

Instead the input will be generated without the metadata:
```
    vector2 uvtiling = {3, 3},
```

lgritz:
Yeah, sorry, currently OSL metadata can only be simple "atomic" types (int, string, float, color, etc.), but vector2 and friends are implemented a structs.

There is generally an intent (mostly due to wanting to support MaterialX better) to migrate these to be first-class atomic types in OSL. It's just taking a back seat at the moment to some other big projects such as fleshing out the GPU/Cuda/OptiX support and SIMD batch shading.

niklasharrysson:
@lgritz No worries, we can use this workaround for now. Good to hear there are plans to make these first-class atomic types in the future.

jstone-lucasfilm (approved):

PR #658, opened 2021-07-23
GLSL code structure cleanup

bernardkwok:
- remove implicit casts
- support that vertex inputs may not be in structures


libraries/bxdf/usd_preview_surface.mtlx
```diff
@@ -12,7 +12,7 @@
     <input name="useSpecularWorkflow" type="integer" value="0" />
     <input name="specularColor" type="color3" value="0, 0, 0" />
     <input name="metallic" type="float" value="0" />
-    <input name="roughness" type="float" value="0.01" />
+    <input name="roughness" type="float" value="0.5" />
```
jstone-lucasfilm, line 15: This update to the signature of UsdPreviewSurface probably deserves its own changelist, making it more straightforward to track in the future, rather than including it as part of the GLSL refactoring.

bernardkwok: I'll open a separate PR for this. Removed from here.

jstone-lucasfilm (approved):
This looks good to me, thanks @bernardkwok!

PR #660, opened 2021-07-25
Fix default value for USD preview surface roughness implementation

bernardkwok:
Fix to use correct value for roughness of USD preview surface.
Was set to clearcoat-roughness value incorrecly 0.01 vs 0.5 (correct valut according to usd spec.)

### DIFF
--- a/libraries/bxdf/usd_preview_surface.mtlx
+++ b/libraries/bxdf/usd_preview_surface.mtlx
@@ -12,7 +12,7 @@
     <input name="useSpecularWorkflow" type="integer" value="0" />
     <input name="specularColor" type="color3" value="0, 0, 0" />
     <input name="metallic" type="float" value="0" />
-    <input name="roughness" type="float" value="0.01" />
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:
This looks like a valuable fix, thanks @bernardkwok!  Would we also need this change to be applied to the usd_preview_surface_default.mtlx example material?

https://github.com/materialx/MaterialX/blob/main/resources/Materials/Examples/UsdPreviewSurface/usd_preview_surface_default.mtlx

jstone-lucasfilm (approved):

bernardkwok:
> This looks like a valuable fix, thanks @bernardkwok! Would we also need this change to be applied to the usd_preview_surface_default.mtlx example material?
> 
> https://github.com/materialx/MaterialX/blob/main/resources/Materials/Examples/UsdPreviewSurface/usd_preview_surface_default.mtlx

Yes. I didn't see that we explicitly put the default value in here. I can put up another change. Thanks for catching this.

PR #663, opened 2021-07-25
Fix real-world unit target inheritance

bernardkwok:
Target inheritance for unit system patched. Similar to previous patch for color system.
Inherited targets will properly find ancestor implementations.


jstone-lucasfilm (approved):

Issue #674, opened 2021-08-12
CurveAdjust implementation missing


swillisart:
Hello! I'm writing a MaterialX [Maya File Translator ](https://help.autodesk.com/view/MAYAUL/2022/ENU/?guid=Maya_SDK_cpp_ref_class_m_px_file_translator_html)
And I just ran into a critical snag on the exporting of remap nodes as CurveAdjust.
![image](https://user-images.githubusercontent.com/33715288/129119728-0b5c00f9-0f82-4012-9751-f633356b9580.png)
The resulting example when loaded into MaterialXView yields an error. 
```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">
  <surfacematerial name="Plastic" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_plastic" />
  </surfacematerial>
  <standard_surface name="SR_plastic" type="surfaceshader">
    <input name="base" type="float" value="1" />
    <input name="base_color" type="color3" nodegraph="NG_remapColor1" output="out" />
    <input name="specular_roughness" type="float" value="0.32467532157897949" />
  </standard_surface>
  <nodegraph name="NG_remapColor1">
    <curveadjust name="remap1" type="color3">
      <input name="in" type="color3" nodename="remapColor1" />
      <parameter name="knots" type="vector2array" value="[0.0, 0.5], [1.0, 1.0]" />
    </curveadjust>
    <ramplr name="remapColor1" type="color3">
        <input name="valuel" type="color3" value="0.0, 0.0, 0.0" />
        <input name="valuer" type="color3" value="0.5, 0.0, 0.0" />
        <input name="texcoord" type="vector2" defaultgeomprop="UV0" />
    </ramplr>
    <output  name="out" type="color3" nodename="remap1" />
  </nodegraph>
</materialx>
```
![image](https://user-images.githubusercontent.com/33715288/129120282-7e44d5b8-c1cc-43dd-9b94-1fae5b469821.png)

And when reading though the code it seems MaterialX does not actually have an implementation available and is commented out. Is this something actively being looked into with Autodesk or are you accepting PR's for curve lookups?

Best regards, 
-Sean

bernardkwok:
Hi Sean,
Thanks for bringing this up and it is true there is no curveadjust support currently, and from the Autodesk side it is not actively being looked at.

[...]

PR #675, opened 2021-08-14
Reference implementation of remapvalue and curveadjust.

swillisart:
This is a rough draft for discussion and reference implementation of 2 new nodes for applying curve lookups on color3 data in OSL.
For simplicity I compared linear interpolation as there are 4 interpolation modes available per "knot".
(Arnold Maya)
image
I'm able to create the GLSL shaders if this is something we wish to continue with.

One deviation from existing direction in the test adjustment.mtlx is this uses values as an array of color3 instead of vector2 array's. Remapping channels separately is necessary and most applications as they apply unique curves per color channel.
(Nuke)
image

Attached is the working Maya file for comparison that I used for testing / generating the image above. (Requires Arnold)
[maya_arnold_osl_reference.zip](https://github.com/materialx/MaterialX/files/6987065/maya_arnold_osl_reference.zip)

I'm not a C++ developer by trade so consider this as an example for feedback.
not sure if this is this related to [Color Array Support #387](https://github.com/autodesk-forks/MaterialX/issues/387) ?


jstone-lucasfilm:
Hi @swillisart!

Thanks for getting this started, and I agree that it will be valuable to move forward with proper code generation for the remapvalue and curveadjust nodes.

My sense, though, is that the most challenging part of this improvement will be on the C++ side, where we need to update our code generators to handle the dynamic arrays that these two nodes leverage.  That's been the main hold-up in implementing these nodes, that our code generators don't yet support dynamic arrays, and we'll need some additional design and discussion to move forward with that.

To my mind, this seems like a good topic for an upcoming MaterialX TSC meeting, where we can discuss the different options for supporting these nodes (and others like them) in code generators across GLSL, OSL, and MDL.

swillisart:
Based the TSC discussion I will go ahead and update this reference using the node for flattened single-sequence arrays. Will also ensure it compiles before submitting the update.

jstone-lucasfilm:
It sounds like we now have a path forward for this work, based on recent work from SideFX on ramp nodes in MaterialX.  See the notes from the MaterialX TSC on November 9th for more details:

https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1636486452084200

I'll close out this pull request for now, and we can continue the discussion in #674 as this work is refined and integrated into MaterialX v1.39.

PR #681, opened 2021-08-23
Move ScopedTimer to MaterialXRender

jstone-lucasfilm:
This changelist refactors the ScopedTimer class and moves it from MaterialXTest to MaterialXRender, making timing facilities available across rendering implementations.


bernardkwok (commented):
Looks good to me. Thanks for migrating.

PR #692, opened 2021-09-11
Add rational curve fit for GGX albedo

jstone-lucasfilm:
This changelist adds a rational curve fit for GGX directional albedo, as a potential replacement for the classic Karis mobile approximation.  The new rational curve is generally more accurate at grazing angles than the Karis approximation, while maintaining comparable shading complexity.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -39,14 +39,16 @@ float mx_ggx_smith_G(float NdotL, float NdotV, float alpha)
     return 2.0 / (lambdaL / NdotL + lambdaV / NdotV);
 }
 
-// https://www.unrealengine.com/blog/physically-based-shading-on-mobile
+// Rational curve fit approximation for GGX directional albedo.
 vec3 mx_ggx_dir_albedo_curve_fit(float NdotV, float roughness, vec3 F0, vec3 F90)
 {
-    const vec4 c0 = vec4(-1, -0.0275, -0.572, 0.022);
-    const vec4 c1 = vec4( 1,  0.0425,  1.04, -0.04 );
-    vec4 r = roughness * c0 + c1;
-    float a004 = min(r.x * r.x, exp2(-9.28 * NdotV)) * r.x + r.y;
-    vec2 AB = vec2(-1.04, 1.04) * a004 + r.zw;
+    vec4 r = vec4(-0.03056, 0.91808, 1.0, 1.0) +
+             vec4(5.73633, -2.21942, 4.23817, -0.05271) * NdotV +
+             vec4(8.03663, -0.02122, 4.90296, 10.64834) * roughness +
+             vec4(-16.71761, 0.05680, -6.25996, -5.42338) * NdotV * roughness +
+             vec4(6.94236, 1.34623, 6.45697, 11.04905) * mx_square(NdotV) +
+             vec4(0.40356, -0.05725, 3.81770, 21.41884) * mx_square(roughness);
+    vec2 AB = r.xy / r.zw;
     return F0 * AB.x + F90 * AB.y;
 }
 


jstone-lucasfilm:
Here's a comparison of three different methods for computing GGX directional albedo, shown with the Standard Surface Plastic material:

Karis Mobile Curve (Original):
StandardSurfacePlastic_KarisCurve

Rational Curve (New):
StandardSurfacePlastic_RationalCurve

Monte Carlo (Reference):
StandardSurfacePlastic_MonteCarlo

jstone-lucasfilm:
And here's the corresponding comparison of the three methods for Standard Surface Default:

Karis Mobile Curve (Original):
StandardSurfaceDefault_KarisCurve

Rational Curve (New):
StandardSurfaceDefault_RationalCurve

Monte Carlo (Reference):
StandardSurfaceDefault_MonteCarlo

jstone-lucasfilm:
Here's a comparison of roughness wedges for the Standard Surface Default material, using three different methods to compute GGX directional albedo:

Monte Carlo (Reference):
StandardSurfaceDefault_MonteCarlo

Karis Curve (Original), with difference from Monte Carlo:
StandardSurfaceDefault_KarisCurve
StandardSurfaceDefault_KarisCurveDiff

Rational Curve (New), with difference from Monte Carlo:
StandardSurfaceDefault_RationalCurve
StandardSurfaceDefault_RationalCurveDiff

jstone-lucasfilm:
Here's a comparison of roughness wedges for the Standard Surface Default material, using three different methods to compute GGX directional albedo:

Monte Carlo:
StandardSurfaceDefault_MonteCarlo

Karis Curve (with difference from Monte Carlo):
StandardSurfaceDefault_KarisCurve
StandardSurfaceDefault_KarisCurveDiff

Rational Curve (with difference from Monte Carlo):
StandardSurfaceDefault_RationalCurve
StandardSurfaceDefault_RationalCurveDiff

niklasharrysson (approved):
Oh that's nice! Looks like a great improvement.

fpsunflower:
I am very interested to see this since I was recently working on the exact same problem 😄 


I did some tests with this approximation and am not getting great results. It is possible that I am interpreting something wrong though since my code uses a slightly different convention.

Looking at a pure GGX metal (F0=1, F90=1) I expect to get the following table for the energy missing `1-(AB.x+AB.y)` in your. code: 

ggx_ref

As expected at low roughness the curve is all black (no energy lost), and as roughness grows we have more and more energy lost across viewing angles.


Using the code above I get this:

ggx_mx_approx

Which seems to over-estimate the energy lost at low roughness values (and this produces over-bright spheres in a furnace test for me).

The approximation I came up with recently gives this result:

ggx_new_epic_approx

In my tests this is close enough to give results matching a table fit.

Can you post the results of a multiple-scattering pure GGX metal on your end? It is quite possible I've missed something.

jstone-lucasfilm:
Thanks @fpsunflower, and I think you're right that this new curve shows some energy loss at lower roughness values, which gives us some pointers on where improvements are needed.  I'd be interested in more details on your new approximation, and it could be useful to try visual comparisons of the two in a single rendering environment.

Here are three renders of roughness wedges for Standard Surface Gold, comparing the Monte Carlo, Rational Curve, and Karis Curve approaches.  Although the new Rational Curve shows significant benefits over our original solution for higher roughness values, I do notice some energy loss at lower roughness values:

Monte Carlo:
StandardSurfaceGold_AlbedoMonteCarlo

Rational Curve (with difference from MC):
StandardSurfaceGold_AlbedoRationalCurve
StandardSurfaceGold_AlbedoRationalCurveDiff

Karis Curve (with difference from MC):
StandardSurfaceGold_AlbedoKarisCurve
StandardSurfaceGold_AlbedoKarisCurveDiff

fpsunflower:
Your test seems to suggest your curve is working well, so maybe I am missing something.

Can you repeat the test with F0=F90=1 instead of yellow? Do you have a way to give your program a constant gray environment map so we can directly visualize the energy loss/gain? It helps if you can make the background the same shade of gray so that the spheres disappear when the model is energy conserving.

jstone-lucasfilm:
@fpsunflower I'll give that test a try, and just to make sure we're computing energy compensation in comparable ways, here's the GLSL function that handles this in MaterialX:

https://github.com/materialx/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L124

jstone-lucasfilm:
Here is the same sequence of rendered images for Standard Surface Chrome, which has a metalness of 1 and a specular_color of (1, 1, 1).  As before. the rendered results are in the general ballpark of Monte Carlo, but we do lose more energy than I'd like at low roughness values:

Monte Carlo:
StandardSurfaceChrome_AlbedoMonteCarlo

Rational Curve (with difference from MC):
StandardSurfaceChrome_AlbedoRationalCurve
StandardSurfaceChrome_AlbedoRationalCurveDiff

Karis Curve (with difference from MC):
StandardSurfaceChrome_AlbedoKarisCurve
StandardSurfaceChrome_AlbedoKarisCurveDiff

I'd need to do a bit of work to set up proper furnace tests with a constant environment and matching background, but that's a step that we've been wanting to take for a while.

fpsunflower:
That still looks quite good. 🤔 

In my local furnace tests there's a noticeable energy gain with your fit (because its over-estimating how much energy is lost). Your energy compensation logic matches what I'm using - so that isn't it.

jstone-lucasfilm:
I wonder if there's some undetected drift between our Monte Carlo computation of GGX directional albedo and the ground truth.  Here's the logic that we use to compute this, in case you spot an issue there that we've overlooked:

https://github.com/materialx/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L69

jstone-lucasfilm:
@fpsunflower Here's the data table that's generated by our Monte Carlo logic on the GPU, in case you'd like to compare it against your own generated data.  The red and green channels represent GGX directional albedo, while the blue channel represents sheen directional albedo:

[AlbedoTable.zip](https://github.com/materialx/MaterialX/files/7165584/AlbedoTable.zip)

fpsunflower:
See follow up in #700 -- the fits presented here work pretty well - main thing missing above was that `Roughness=Alpha` in this code which I wasn't taking into account above.

PR #693, opened 2021-09-13
ESSL code generator addition

bernardkwok:
### Web merge part 2: ESSL code generator

Previous check-in has modified other GLSL code snippets and shader node code generators to produce ESSL and GLSL compliant code such as avoiding explicit casts. This formally introduces a new generator to which shares existing GLSL code but is injected in into the final shader in a custom manner.

Note that there is no new C++ or Python module and that ESSL capabilities have been added to the existing
MaterialXGenGLSL library.

This change includes the following:
1. Refactors some of the underlying code in the GLSL generator to be modularized into virtual methods which can be overridden 
by the derived ESSL shader class.
2. Some pertinent points including modifying the language directives, and not initializing any uniforms.
3. From a usage perspective:
    - Include Python wrappers for this generator and add in a Python utility to allow arbitrary language output. (`generateshader`)
    -  Add the ability to output ESSL source code from MaterialXView
4. ESSL implementation checking uses `generateshader` as part of CI.
5. `glslang` is setup as part of CI with `glslangValidator` being invoked for ESSL and GLSL syntax checking of vertex and pixel shader stages. (Note: This covers all of the `pbr` library nodes + some stdlib nodes based on usage in the examples materials. 
basic infrastructure is now available.)
(Note: Actual WASM testing is not part of this check-in but will use this generator. ) As part of this CI has been modified a bit to reduce the number of builds for pushes vs PRs.
6. Some minor additional fixes for GLSL / ESSL source code implementation compliance in the pbr library.

[CI generated shaders, including ESSL (Zip file)](https://github.com/materialx/MaterialX/files/7212875/Shaders_Linux_GCC_11_Python37.zip)


.github/workflows/main.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           compiler: gcc
           compiler_version: "11"
           python: 3.7
+          test_generate: ON
           test_render: ON
```
jstone-lucasfilm, line 40: In our cloud-based render tests, it looks like "Viewer_BrassAverage.png" ends up as a solid grey square after this changelist!

python/Scripts/generateShader.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python
+'''
+Utility to generate the shader for materials found in a MaterialX document. One file will be generated
```
jstone-lucasfilm, line 3: I really like the idea of adding this new script to the codebase.  One minor adjustment I would recommend is to name it "generateshader.py", for letter-case consistency with our existing scripts.

source/MaterialXRenderGlsl/GLContext.cpp
```diff
@@ -20,7 +20,7 @@ namespace MaterialX
 
 #if defined(_WIN32)
 
-GLContext::GLContext(SimpleWindowPtr window, HardwareContextHandle sharedWithContext) :
+GLContext::GLContext(const SimpleWindowPtr window, HardwareContextHandle sharedWithContext) :
```
jstone-lucasfilm, line 23: Unintended change here?

jstone-lucasfilm (commented):
This looks very promising, thanks @bernardkwok!  I've added a few initial thoughts and comments.

.github/workflows/main.yml
bernardkwok: This is an accident of changing some variables to float in the glsl code which should not have be float.
I will correct this.

source/MaterialXRenderGlsl/GLContext.cpp
bernardkwok: Yes. Will revert

libraries/pbrlib/genglsl/mx_dielectric_bsdf.glsl
```diff
@@ -21,7 +21,15 @@ void mx_dielectric_bsdf_reflection(vec3 L, vec3 V, vec3 P, float occlusion, floa
     vec2 safeRoughness = clamp(roughness, M_FLOAT_EPS, 1.0);
     float avgRoughness = mx_average_roughness(safeRoughness);
 
-    FresnelData fd = tf.thickness > 0.0 ? mx_init_fresnel_dielectric_airy(ior, tf.thickness, tf.ior) : mx_init_fresnel_dielectric(ior);
+    FresnelData fd;
```
bernardkwok, line 24: Reverted other changes which we're not needed / incorrect.
This is the only implicit cast code fix required.

.github/workflows/main.yml
bernardkwok: Fix now (current render result)
Viewer_BrassAverage

bernardkwok:
Hi @meshula , Some of the ground work changes for ESSL will help with MSL support, but this change isn't necessary a new
mechanism per se but more of adding in additional entry points to allow variation in what code get's injected. e.g. uniforms are not initialized in the ESSL generator but are in the GLSL one.

.github/workflows/main.yml
```diff
@@ -109,7 +113,18 @@ jobs:
           echo "CC=clang" >> $GITHUB_ENV
           echo "CXX=clang++" >> $GITHUB_ENV
         fi
-
+    
+    - name: Build GLSlangValidator (Windows)
```
bernardkwok, line 117: Base level validation of GLSL / ESSL using glslang.

.github/workflows/main.yml
```diff
@@ -154,6 +169,21 @@ jobs:
...
+      if: matrix.test_generate == 'ON'
+      run: |
+        mkdir build/generate
+        python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path "." --outputPath "build/generate/" --target "glsl" 
+        python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path "." --outputPath "build/generate/" --target "osl"
+        python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path "." --outputPath "build/generate/" --target "mdl"
+        python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path "." --outputPath "build/generate/" --target "essl" 
+
+    - name: Generation Validation Tests
+      if: matrix.test_validate == 'ON' && runner.os == 'Windows'
```
bernardkwok, line 182: Test for GLSL and ESSL generation.

python/Scripts/generateshader.py
```diff
@@ -0,0 +1,161 @@
...
+    else:
+        print('--- No validation performed')
+
+def main():
+    parser = argparse.ArgumentParser(description='Generate shader code for each material / shader in a document.')
+    parser.add_argument('--path', dest='paths', action='append', nargs='+', help='An additional absolute search path location (e.g. "/projects/MaterialX")')
+    parser.add_argument('--library', dest='libraries', action='append', nargs='+', help='An additional relative path to a custom data library folder (e.g. "libraries/custom")')
+    parser.add_argument('--target', dest='target', default='glsl', help='Target shader generator to use (e.g. "genglsl"). Default is genglsl.')
+    parser.add_argument('--outputPath', dest='outputPath', help='File path to output shaders to.')
+    parser.add_argument('--validator', dest='validator', nargs='?', const=' ', type=str, help='Name of executable to perform source code validation.')
```
bernardkwok, line 38: @jstone-lucasfilm , I think this should be generic enough for this script to pass in
an argument for the validator + validator args. I separated the 2 as the script checks for
existence of the validator then builds the command to run.

source/MaterialXGenEssl/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+        get_filename_component(_source_path "${_source_rel}" PATH)
+        string(REPLACE "/" "\\" _source_path_msvc "${_source_path}")
+        source_group("${prefix}\\${_source_path_msvc}" FILES "${_source}")
+    endforeach()
+endfunction(assign_source_group)
+
+assign_source_group("Header Files" ${materialx_header})
+assign_source_group("Source Files" ${materialx_source})
+
+add_library(MaterialXGenEssl STATIC
```
jstone-lucasfilm, line 20: I've discussed this with @bernardkwok off-line, and my recommendation would be to add the new ESSL generator and syntax classes to the existing MaterialXGenGlsl library.  This would allow all GLSL generation functionality to be accessed from a single C++/Python library, and would help to focus developers on the more central distinction between GLSL, OSL, and MDL generators.

source/MaterialXView/CMakeLists.txt
```diff
@@ -97,3 +100,11 @@ install(TARGETS MaterialXView
     RUNTIME DESTINATION bin)
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/MaterialXView.pdb"
     DESTINATION "${CMAKE_INSTALL_PREFIX}/bin/" OPTIONAL)
+    
+# Copy over libraries and resources to local build area
```
jstone-lucasfilm, line 104: It would be better if we can avoid this duplication, and directly access the canonical data libraries and resources from the viewer.  One benefit of this approach is in shader development, where it's possible to modify the canonical data libraries in place and hot-reload them into the viewer, rather than maintaining two separate copies that can drift out of sync.

.github/workflows/main.yml
```diff
@@ -175,3 +205,10 @@ jobs:
       with:
         name: Renders_${{ matrix.name }}
         path: build/render/
+
+    - name: Upload Shaders
```
jstone-lucasfilm, line 209: Now that you've added cloud-based validation of GLSL shaders to our script, I would recommend removing the more expensive uploading of shader artifacts, which are less likely to be read by developers.  Ideally I'd like to extend the pattern you've set up for GLSL, and eventually add cloud-based validation of OSL and MDL shaders.

jstone-lucasfilm (commented):
This work is coming along really well, and I've written up a few high-level recommendations below.

.github/workflows/main.yml
bernardkwok: For now the validation (via glslang) is a syntax one only and does not actually compile the shader. Thus any errors there will not be caught. I think for shader artifacts we need a image compare against baseline but tbd how to do this.

source/MaterialXView/CMakeLists.txt
bernardkwok: I actually have another cmake change to use a common location for all binaries. Can omit this copy for now.

.github/workflows/main.yml
jstone-lucasfilm: That's good to know, though ideally I'd like to keep the number of uploaded artifacts as small as possible, since uploads are one of the most common causes of stalls during builds.

To keep things as simple as possible, what would you think about moving all of the new generateshader.py calls to our "Python Tests" step, and removing the "Upload Shaders" step?  I think it's fine for generateshader.py to run on all targets, since it's a relatively quick step, so we likely don't need a separate flag for shader generation in our builds.

source/MaterialXView/CMakeLists.txt
```diff
@@ -97,3 +97,4 @@ install(TARGETS MaterialXView
     RUNTIME DESTINATION bin)
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/MaterialXView.pdb"
     DESTINATION "${CMAKE_INSTALL_PREFIX}/bin/" OPTIONAL)
+
```
jstone-lucasfilm, line 100: Unintended change here?

source/MaterialXView/Viewer.cpp
```diff
@@ -868,6 +880,8 @@ void Viewer::createAdvancedSettings(Widget* parent)
     referenceQualityBox->setCallback([this](bool enable)
     {
         _genContext.getOptions().hwDirectionalAlbedoMethod = enable ? mx::DIRECTIONAL_ALBEDO_MONTE_CARLO : mx::DIRECTIONAL_ALBEDO_TABLE;
+        // No Albedo Table support for Essl yet.
```
jstone-lucasfilm, line 883: For ESSL code, I'd recommend always rendering with DIRECTIONAL_ALBEDO_CURVE_FIT, since it's the most performant of the three options.  Monte Carlo sampling for directional albedo would likely be too heavyweight in the contexts where ESSL code is run.

source/MaterialXView/CMakeLists.txt
bernardkwok: Fixed now.

source/MaterialXView/Viewer.cpp
bernardkwok: This was a copy and paste mistake. The initialization is to CURVE FIT so this will always use curve fit as table support does not exist currently.

source/MaterialXGenGlsl/GlslShaderGenerator.cpp
```diff
@@ -275,7 +275,7 @@ ShaderPtr GlslShaderGenerator::generate(const string& name, ElementPtr element,
     ScopedFloatFormatting fmt(Value::FloatFormatFixed);
 
     // Make sure we initialize/reset the binding context before generation.
-    HwResourceBindingContextPtr resourceBindingCtx = context.getUserData<HwResourceBindingContext>(HW::USER_DATA_BINDING_CONTEXT);
+    const HwResourceBindingContextPtr resourceBindingCtx = getResourceBindingContext(context);
```
jstone-lucasfilm, line 278: Why the added const modifier on this variable?

source/MaterialXGenGlsl/GlslResourceBindingContext.cpp
```diff
@@ -32,6 +32,19 @@ void GlslResourceBindingContext::emitDirectives(GenContext& context, ShaderStage
 {
     ShaderGenerator& generator = context.getShaderGenerator();
 
+    // write shader stage directives for Vulkan compliance
+    if( _separateBindingLocation )
```
jstone-lucasfilm, line 36: Minor: It looks like the coding style here doesn't match the codebase, where we usually have a space after the `if`, and no spaces inside the parentheses.  Once Clang 13 is released, we're planning to add a clang-format file for MaterialX, which should allow for automated style alignment when needed.

.github/workflows/main.yml
```diff
@@ -152,8 +164,20 @@ jobs:
         python Scripts/mxupdate.py ../resources/Materials/TestSuite/stdlib/upgrade --yes
         python Scripts/mxvalidate.py ../resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx --verbose
         python Scripts/mxdoc.py ../libraries/pbrlib/pbrlib_defs.mtlx
+        cd ..
+        mkdir build/generate
```
jstone-lucasfilm, line 168: Since we're not consuming these generated shaders, it might make more sense to create this dedicated folder in `Generate Validation Tests`, and to simply output shaders to the source folder in these cross-platform Python tests.

.github/workflows/main.yml
```diff
@@ -109,7 +110,18 @@ jobs:
           echo "CC=clang" >> $GITHUB_ENV
           echo "CXX=clang++" >> $GITHUB_ENV
         fi
-
+
```
jstone-lucasfilm, line 113: Minor: Unintended whitespace changes here?

jstone-lucasfilm (commented):
This is looking nearly ready to merge, and I have just a few minor recommendations below.

.github/workflows/main.yml
bernardkwok: This should be okay. Adds a space between steps.

.github/workflows/main.yml
bernardkwok: Mostly I did this as it otherwise "pollutes" the Python directory and when run locally it shows up as git changes.

source/MaterialXGenGlsl/GlslResourceBindingContext.cpp
```diff
@@ -86,15 +105,15 @@ void GlslResourceBindingContext::emitResourceBindings(GenContext& context, const
...
+                                                                ShaderStage& stage, const std::string& structInstanceName,
+                                                                const std::string& arraySuffix)
 {
     ShaderGenerator& generator = context.getShaderGenerator();
     const Syntax& syntax = generator.getSyntax();
 
-    // Glsl struct need to be aligned. We make a best effort to base align struct members and add
-    // padding if required
+    // Glsl structures need to be aligned. We make a best effort to base align struct members and add
+    // padding if require.
```
jstone-lucasfilm, line 116: It looks like the original grammar here is correct, but this change makes it incorrect?  :)

.github/workflows/main.yml
jstone-lucasfilm: The original newline is good, but we don't need the additional 4 spaces as well...  :)

.github/workflows/main.yml
jstone-lucasfilm: I see other examples above, though, where we're already modifying files from Materials/TestSuite in place.  For example, the call to mxupdate.py at the top of this block.  Perhaps we could write the output shaders to a similarly safe location in the Materials folder?

bernardkwok: I've changed the behavour so the default output path is the path of the input document. This should
match behaviour for other utilities. The CI script then does not need this artificial creation of another directory.

jstone-lucasfilm: Thank you, Bernard!

jstone-lucasfilm (approved):

Issue #695, opened 2021-09-16
PBR mix node default value inconsistency


dleex:
The default value for mix parameter on mix nodes for BSDF, EDF, and VDF are different than those for standard data types. According to 1.38 specs the value is supposed to be 0 for both std and pbr nodes, but in pbrlib_defs.mtlx, the default value is set to 1.
We'd prefer if PBR mix node's default mix is set to 0 for consistency's sake (in Houdini we have just one mix node that handles both std and pbr data types, but if they're supposed to have different default values, we may have to split them into two).

niklasharrysson:
@dleex Good catch! I have fixed this in PR #697.

PR #697, opened 2021-09-17
Fix inconsistency in mix nodes default value

niklasharrysson:
This change list addresses issue #695


jstone-lucasfilm:
Good catch, @dleex, and thanks for the quick fix, @niklasharrysson!

PR #698, opened 2021-09-19
Improved rational curve fit for sheen albedo

jstone-lucasfilm:
This changelist improves the rational curve fit for sheen directional albedo, generating results that are significantly closer to Monte Carlo reference for low roughness values.

Additionally, it increases the size of the directional albedo table to 128, allowing a correct match for sheen direction albedo when table-based rendering is used.


jstone-lucasfilm:
Here's a comparison of roughness wedges for the Standard Surface Velvet material, using three different methods to compute sheen directional albedo:

Monte Carlo:
StandardSurfaceVelvet_SheenAlbedoMonteCarlo

New Rational Curve, with difference from MC:
StandardSurfaceVelvet_SheenAlbedoRationalNew
StandardSurfaceVelvet_SheenAlbedoRationalNewDiff

Old Rational Curve, with difference from MC:
StandardSurfaceVelvet_SheenAlbedoRationalOld
StandardSurfaceVelvet_SheenAlbedoRationalOldDiff

PR #699, opened 2021-09-20
Update MaterialXView to handle multiple meshes.

bernardkwok:
This is the code written to support GLTF loading in the Khronos branch,
As discussed this should be generally useful for any loader that loads multiple meshes each with their own partitions and have it work in the viewer. (which currently only handles the 1st mesh).


jstone-lucasfilm (approved):
This change looks good to me, thanks @bernardkwok!

PR #700, opened 2021-09-21
Improved rational curve fit for GGX albedo

jstone-lucasfilm:
This changelist improves the visual quality of the rational curve fit for GGX albedo, generating results that are very close to Monte Carlo ground truth.  Since this improvement comes at the cost of additional shader instructions, it's worthwhile to discuss the tradeoffs before moving forward.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -42,12 +42,19 @@ float mx_ggx_smith_G(float NdotL, float NdotV, float alpha)
 // Rational curve fit approximation for GGX directional albedo.
 vec3 mx_ggx_dir_albedo_curve_fit(float NdotV, float roughness, vec3 F0, vec3 F90)
 {
-    vec4 r = vec4(-0.03056, 0.91808, 1.0, 1.0) +
-             vec4(5.73633, -2.21942, 4.23817, -0.05271) * NdotV +
-             vec4(8.03663, -0.02122, 4.90296, 10.64834) * roughness +
-             vec4(-16.71761, 0.05680, -6.25996, -5.42338) * NdotV * roughness +
-             vec4(6.94236, 1.34623, 6.45697, 11.04905) * mx_square(NdotV) +
-             vec4(0.40356, -0.05725, 3.81770, 21.41884) * mx_square(roughness);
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Here are three renders of roughness wedges for Standard Surface Gold, comparing the Monte Carlo, New Rational Curve, and Karis Curve approaches:

Monte Carlo:
StandardSurfaceGold_AlbedoMonteCarlo

New Rational Curve, with difference from MC:
StandardSurfaceGold_AlbedoRationalNew
StandardSurfaceGold_AlbedoRationalNewDiff

Karis Curve, with difference from MC:
StandardSurfaceGold_AlbedoKarisCurve
StandardSurfaceGold_AlbedoKarisCurveDiff

fpsunflower:
I just took this fit for a spin but am getting similar results as before. The reference directional albedo image (NdotV, roughness) is:

ggx_ref

And the fit in this PR gives me:

ggx_approx_pr700

So again, at low roughness values the value seems fairly far from the ground truth. And using it in renders produces a noticeable energy gain in a furnace test.

I am curious if you tried increasing the sample count of the monte carlo reference? I haven't studied the code in detail to see where the difference might be -- but I did notice its using regular distribution sampling instead of visible normal sampling which means it might be a fairly noisy estimate (except the sampling pattern is locked so the "noise" may not be visible).

Otherwise maybe there is a subtle difference in how we are implementing the GGX lobe?

fpsunflower:
I took a closer look at this (also in follow up to the comments I made on #692). Both fits actually perform quite well, the main thing I was missing was that the code expects `Roughness = Alpha` while my code was using the convention `Alpha = Roughness^2`.

With that factored in, the proposed fits are looking good. At high roughness I still see a bit of energy loss, but it's definitely in the acceptable range.

I am going to see if I can get my improved fit contributed to the project. They might be a bit more expensive because they call `pow` and do divisions, so will need to be benchmarked in context.

jstone-lucasfilm:
Thanks for the feedback, @fpsunflower!  Agreed on the subtle energy loss at high roughness values, and as a next step we're interested in improving the accuracy of our Monte Carlo computation, potentially using VNDF sampling for better convergence.  Once your own proposed fit is available, we'd be very interested in testing it in the MaterialX codebase, where it would be straightforward to render roughness wedges with Standard Surface materials for validation.

PR #702, opened 2021-09-27
File template naming update for TextureBaker

mjyip-lucasfilm:
Added file template naming and updated naming for baked texture filenames.
* Template variables include: `$ASSET`, `$INPUT`, `$EXTENSION`, `$NAMEPATH`, `$MATERIAL`, `$SHADINGMODEL`, `$UDIM`, `$UDIMPREFIX`
* Variable values are inferred accordingly
  * For example, `$ASSET` is inferred from the output socket's `activeSourceUri`
* Note `$UDIMPREFIX` and `$UDIM` is only added when udims are detected
* Default filename template is `$MATERIAL_$SHADINGMODEL_$INPUT$UDIMPREFIX$UDIM.$EXTENSION`
* Adding `.$EXTENSION` in template is optional (`.$EXTENSION` will be added if not specified)


source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -101,20 +103,55 @@ TextureBaker::TextureBaker(unsigned int width, unsigned int height, Image::BaseT
...
+        {
+            bakedImageName.replace(i, pair.first.length(), replacement);
+        }
+    }
+    return _outputImagePath / bakedImageName;
+}
+
+StringMap TextureBaker::initializeFileTemplateMap(OutputPtr output, NodePtr shader, const string& udim)
+{
+    string inputName = output->getName();
```
jstone-lucasfilm, line 141: In this logic, would it be possible to get the unique name of the associated _input_ of the shader node, rather than the modified name of the output?  Since we don't require any particular naming convention for output elements, I can imagine situations where it might have an ambiguous or unclear name, while shader node inputs are required to match the canonical input names in the shader nodedef.

jstone-lucasfilm (commented):
This looks excellent, thanks @mjyip-lucasfilm!  I had just one comment below, relating to the logic for gathering the input name.

source/MaterialXRenderGlsl/TextureBaker.cpp
mjyip-lucasfilm: I have incorporated the feedback into the code.

mjyip-lucasfilm: Thanks for the feedback!

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -101,20 +103,53 @@ TextureBaker::TextureBaker(unsigned int width, unsigned int height, Image::BaseT
...
+}
+
+StringMap TextureBaker::initializeFileTemplateMap(InputPtr input, NodePtr shader, const string& udim)
+{
+    string inputName = input->getName();
+    FilePath assetPath = FilePath(shader->getActiveSourceUri());
+    assetPath.removeExtension();
+    StringMap filenameTemplateMap;
+    filenameTemplateMap["$ASSET"] = assetPath.getBaseName();
+    filenameTemplateMap["$INPUT"] = _bakedInputOverlapMap.count(inputName)? _bakedInputOverlapMap[inputName] : inputName;
```
jstone-lucasfilm, line 146: To simplify this new logic, you might consider naming the new map `_bakedInputMap`, and setting it consistently for all baked inputs, not just those that share their baked texture with another input.  That would allow this logic to be written more simply as:

```
filenameTemplateMap["$INPUT"] = _bakedInputMap.count(inputName);
```

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -145,13 +180,13 @@ void TextureBaker::bakeShaderInputs(NodePtr material, NodePtr shader, GenContext
         return;
     }
 
-    std::set<OutputPtr> bakedOutputs;
+    std::unordered_map<OutputPtr, InputPtr> bakedOutputs;
```
jstone-lucasfilm, line 183: Now that this container is a map from output to input elements, it might be more clear to name it `bakedOutputMap`.

mjyip-lucasfilm: Fixed!

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -101,20 +103,54 @@ TextureBaker::TextureBaker(unsigned int width, unsigned int height, Image::BaseT
...
+{
+    FilePath assetPath = FilePath(shader->getActiveSourceUri());
+    assetPath.removeExtension();
+    StringMap filenameTemplateMap;
+    filenameTemplateMap["$ASSET"] = assetPath.getBaseName();
+    filenameTemplateMap["$INPUT"] = _bakedInputMap[input->getName()];
+    filenameTemplateMap["$EXTENSION"] = _extension;
+    filenameTemplateMap["$MATERIAL"] = _material->getName();
+    filenameTemplateMap["$NAMEPATH"] = createValidName(input->getNamePath());
+    filenameTemplateMap["$SHADER"] = shader->getName();
```
jstone-lucasfilm, line 149: Now that you've added `$MATERIAL` and `$SHADINGMODEL`, I'd recommend removing the original `$SHADER`, which returns the name of the shading node.  In practice, this name tends to be relatively arbitrary, and is often just derived from the name of the material node.

mjyip-lucasfilm: Fixed!

jstone-lucasfilm (approved):
Looks good to me, thanks @mjyip-lucasfilm!

PR #706, opened 2021-10-02
Use VNDF sampling for GGX directional albedo

jstone-lucasfilm:
This changelist updates the Monte Carlo computation for GGX directional albedo to use VNDF sampling (http://jcgt.org/published/0007/04/01/paper.pdf), creating a notable improvement in its accuracy.  Additionally, the rational quadratic fit for for GGX directional albedo has been rebuilt from the latest Monte Carlo data, creating a smoother, more accurate fit.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
@@ -45,7 +45,7 @@ vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 roughness, int distrib
         vec3 F = mx_compute_fresnel(VdotH, fd);
 
         // Compute the geometric term.
-        float G = mx_ggx_smith_G(NdotL, NdotV, mx_average_roughness(roughness));
+        float G = mx_ggx_smith_G2(NdotL, NdotV, mx_average_roughness(roughness));
         
         // Add the radiance contribution of this sample.
         // From https://cdn2.unrealengine.com/Resources/files/2013SiggraphPresentationsNotes-26915738.pdf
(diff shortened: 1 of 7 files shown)


jstone-lucasfilm:
Here are three renders of roughness wedges for Standard Surface Gold, comparing a Monte Carlo reference render (NDF, 4096 samples), a new Monte Carlo render (VNDF, 64 samples), and the updated Rational Quadratic fit:

Monte Carlo Reference (NDF, 4096 Samples):
StandardSurfaceGold_AlbedoMonteCarlo_NDF_4096_Samples

Monte Carlo New (VNDF, 64 Samples):
StandardSurfaceGold_AlbedoMonteCarlo_VNDF_64_Samples

Rational Quadratic Fit:
StandardSurfaceGold_AlbedoRationalQuadratic

Difference of Monte Carlo New from Reference:
StandardSurfaceGold_AlbedoMonteCarlo_VNDF_64_Samples_Diff

Difference of Rational Quadratic Fit from Monte Carlo New:
StandardSurfaceGold_AlbedoRationalQuadratic_Diff

fpsunflower:
Very nice! The updated fit looks much better in my tests.

fpsunflower:
You may have already played with this. But does the improved input data let you use a lower order rational approximation for the fit? Or are the extra terms still needed?

jstone-lucasfilm:
Thanks, @fpsunflower!  I'll need to run additional tests to find the ideal trade-off of complexity versus quality, but I have found that a rational quadratic with only 8 constants has nearly the same RMSE as this version with 9 constants.  The error is distributed differently for each fit, and I'll need to confirm that there aren't any visible artifacts with those proposed simplifications, but it's a worthwhile area for additional exploration.

PR #707, opened 2021-10-04
Fix search path for XML stream loading

bernardkwok:
The XML stream loading path does not take into consideration search paths.
This affects cases such as when an xinclude is in being loaded from memory.
This makes the functionality orthogonal to the file based XML API.

Note: this in memory xinclude workflow can be used for loading for web browsing. 

Additional small API addition to set current file path.


jstone-lucasfilm (approved):

Issue #737, opened 2021-10-26
Request to enhance the discovery of "material" nodes


bernardkwok:
**Workflows to address**:
1. Finding all material nodes in a document.
2. Finding all the material node connected to a `materialassign`

**Request** : to be able to not only find an atomic node which outputs a material, but also the case where:
a. A nodegraph has a `material` output type.
b. If not already handled a top level output connected to a node or nodegraph which outputs a `material`.

**Existing support:**
c. GraphElement::getMaterialNodes() just returns immediate child notes of type material. Does not consider nodegraphs.
d. MaterialAssign::getReferencedMaterials() which finds sibling nodes which match the "material" name attribute on the assignment.

Note: findRenderableElements() performs the traversal for 1, but is not connected to 2.

**Desired support**:  References to one or more Nodes or NodeGraphs which have one or more outputs which
are of type `material`.  Return values in proposed single API utility are:
- A list of NodePtr  or NodeGraphPtr if querying from a `Document` 
- A single NodePtr or NodeGraphPtr if querying via a `materialassign`
- A list of NodePtr which are connected to an `output` if querying from a `nodegraph`

Attached test file with possible configurations:
[material_node_discovery.mtlx.txt](https://github.com/materialx/MaterialX/files/7428456/material_node_discovery.mtlx.txt)

1. A built in materialnode at the top level : assigned (connected to `materialassign`) or not.
![image](https://user-images.githubusercontent.com/14275104/139126193-1f23c9b1-f320-4af6-a00c-337451f9bd28.png)
![image](https://user-images.githubusercontent.com/14275104/139126140-886a2be5-40e1-410d-9eee-4e1853161c0a.png)

2. A graph with a `material` output : assigned (connected to `materialassign`) or not.
![image](https://user-images.githubusercontent.com/14275104/139126348-96f97d57-82c1-4d55-91a6-d19bb59c6b49.png)
![image](https://user-images.githubusercontent.com/14275104/139126452-7f21e52c-89b8-4bad-bbf6-cde43ec5acf1.png)
where the nodegraph looks like this:
![image](https://user-images.githubusercontent.com/14275104/139126546-cee9ff84-87df-4c8f-aeeb-0b5bf54eba80.png)

3. A custom defined materialnode, with the implementation as a nodegraph (the node shown above) : assigned (connected to `materialassign`) or not.
[...]

PR #743, opened 2021-11-03
Add material discovery utility

kwokcb:
### New methods
* `NodeGraph::getMaterialOutputs` : return any outputs connected to an graph material node
* `Node::getMaterialOutputs` : return any outputs of type material.
* `MaterialAssign::getMaterialOutputs` : return any outputs on assigned nodegraphs which have outputs of type material. Uses the Nodegraph method.
* `Document::getMaterialOutputs` : return any outputs on nodegraphs which have outputs of type material. Uses the Nodegraph method. Note that implementation nodegraphs are ignored. 

### Configurations Supported.
Fixes: #737
Refer to that issue for graphical view of configurations handled.

### Test Results
(For test file used)

*Material nodes discovered which are referenced by `materialassigns`*
- Material node: top_level_material_assigned
- Material node: top_level_material_in_graph_assigned/surfacematerial1
- Material node: top_level_material_def_assigned

*Material nodes discovered which are within a nodegraph*
- Material node: top_level_material_in_graph_no_assign/surfacematerial1
- Material node on definition: NG_material_def_material_1_0/surfacematerial1
- Material node: top_level_material_in_graph_assigned/surfacematerial1

*Material nodes or nodegraphs discovered within a Document*
This scans at the document level first, and then for each nodegraph scans inside. The difference
when scanning at document level is that nodegraph implementations are skipped so `NG_material_def_material_1_0/surfacematerial1` is not return.
- NodeGraph Material Node: top_level_material_in_graph_no_assign/surfacematerial1
- NodeGraph Material Node: top_level_material_in_graph_assigned/surfacematerial1
- Node Material : top_level_material_no_asssign
- Node Material : top_level_material_assigned
- Node Material : top_level_material_def
- Node Material : top_level_material_def_assigned


source/MaterialXCore/Document.cpp
```diff
@@ -7,8 +7,10 @@
 
 #include <MaterialXCore/Util.h>
 #include <MaterialXCore/Version.h>
+#include <MaterialXCore/Node.h>
```
jstone-lucasfilm, line 10: It looks like these two added headers (Node.h and unordered_set) are no longer needed?

source/MaterialXCore/Document.cpp
```diff
@@ -305,6 +307,30 @@ ValuePtr Document::getGeomPropValue(const string& geomPropName, const string& ge
     return value;
 }
 
+std::vector<OutputPtr> Document::getMaterialOutputs(bool skipIncludes) const
```
jstone-lucasfilm, line 310: Is there any valid case where we _wouldn't_ want to skip XIncluded content?  As a user, I would only expect this method to find material outputs in the content file, not in the underlying data libraries.

source/MaterialXCore/Document.h
```diff
@@ -170,6 +170,13 @@ class MX_CORE_API Document : public GraphElement
         removeChildOfType<GeomPropDef>(name);
     }
 
+    /// @}
+    /// @name Material Outputs
+    /// @{
+    
+    /// Return outputs on nodegraps which are of type "material"
```
jstone-lucasfilm, line 177: I would recommend phrasing this as "Return material-type outputs for all nodegraphs in the document."

source/MaterialXCore/Look.cpp
```diff
@@ -4,9 +4,10 @@
 //
 
 #include <MaterialXCore/Look.h>
-
 #include <MaterialXCore/Document.h>
 
+#include <unordered_set>
```
jstone-lucasfilm, line 9: As above, it looks like these edits to the included headers are no longer needed.

source/MaterialXCore/Material.cpp
```diff
@@ -3,8 +3,12 @@
 // All rights reserved.  See LICENSE.txt for license.
 //
 
+#include <MaterialXCore/Document.h>
```
jstone-lucasfilm, line 6: Same as above.

source/MaterialXCore/Node.h
```diff
@@ -326,6 +326,13 @@ class MX_CORE_API NodeGraph : public GraphElement
     }
     virtual ~NodeGraph() { }
 
+    /// @name Material References
+    /// @{
+    
+    /// Return any outputs on the nodegraph which are of type "material"
```
jstone-lucasfilm, line 332: Following the pattern above, I would phrase this as "Return all material-type outputs of the nodegraph."

source/MaterialXTest/MaterialXCore/Material.cpp
```diff
@@ -7,9 +7,87 @@
...
+
+#include <unordered_set>
 
 namespace mx = MaterialX;
 
+namespace
+{
+
+// Utility function for testing material discovery
+std::vector<mx::InterfaceElementPtr> getMaterials(mx::ElementPtr root, bool skipIncludes)
```
jstone-lucasfilm, line 21: I can see where this complex unit test was warranted for the original material discovery logic, but maybe we can now omit this, since the added functionality is a clear-cut collection of getMaterialNodes methods?  I would recommend simplifying this, and simply validating that Document::getMaterialNodes returns the expected set of nodes for your example above.

jstone-lucasfilm (commented):
Thanks for this pull request, @bernardkwok!  I've added some notes and recommendations below.

source/MaterialXCore/Document.cpp
kwokcb: I wasn't sure if it's always considered that an `xinclude` references a "library" not or if the author just wanted to split up the content file into multiple files. For simplicity I will remove this argument for now. (Adding an argument more backward compatible than removing one.)

source/MaterialXTest/MaterialXCore/Material.cpp
kwokcb: Actually the request was to be able to query from a nodegraph (including nodegraphs for noddefs), a document and materialassign as roots for different workflows. 
Also I think it's better to test all new apis. I'll see about simplifying this however as some cross checks are probably not needed anymore.

source/MaterialXCore/Document.cpp
kwokcb: Could be. I'll do a clean up here and other places you've pointed out.

source/MaterialXCore/Document.h
kwokcb: Will do.

source/MaterialXCore/Node.h
kwokcb: Will do.

resources/Materials/TestSuite/stdlib/materials/material_node_discovery.mtlx
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0"?>
+<materialx version="1.38">
+  <nodedef name="ND_material_def_material_1_0" node="material_def" version="1.0" isdefaultversion="true">
+    <output name="out" type="material" value="" />
+  </nodedef>
+  <standard_surface name="top_level_shader" type="surfaceshader" nodedef="ND_standard_surface_surfaceshader" version="1.0.1" xpos="620" ypos="950" />
```
jstone-lucasfilm, line 6: What is the motivation for linking this node to a specific NodeDef in the standard libraries, rather than controlling the NodeDef matching through node name and version number?  This seems like an unusual pattern that we would want to discourage in production usage, since only the node name and version number is robust to future changes in the library, but maybe there's an advantage here that I'm missing.

resources/Materials/TestSuite/stdlib/materials/material_node_discovery.mtlx
```diff
@@ -0,0 +1,54 @@
...
+  <nodegraph name="NG_material_def_material_1_0" xpos="925" ypos="1596" nodedef="ND_material_def_material_1_0">
+    <standard_surface name="standard_surface2" type="surfaceshader" nodedef="ND_standard_surface_surfaceshader" version="1.0.1" xpos="-414.603" ypos="326.959" />
+    <surfacematerial name="surfacematerial1" type="material" nodedef="ND_surfacematerial" xpos="-67.8807" ypos="21.3596">
+      <input name="surfaceshader" type="surfaceshader" nodename="standard_surface2" />
+      <input name="displacementshader" type="displacementshader" value="" />
+    </surfacematerial>
+    <output name="out" type="material" nodename="surfacematerial1" />
+  </nodegraph>
+  <material_def name="top_level_material_def" type="material" nodedef="ND_material_def_material_1_0" version="1.0" xpos="925" ypos="1787" />
+  <nodegraph name="top_level_material_in_graph_assigned" xpos="315" ypos="468" ldx_io_nodes="&amp;lt;?xml version=&quot;1.0&quot;?&amp;gt;&#10;&amp;lt;materialx version=&quot;1.38&quot;&amp;gt;&#10;  &amp;lt;layout name=&quot;io&quot; type=&quot;&quot; /&amp;gt;&#10;&amp;lt;/materialx&amp;gt;&#10;">
```
jstone-lucasfilm, line 33: What is the meaning of the `ldx_io_nodes` attribute?  Perhaps it's an application specific artifact that we should remove from the test suite?

source/MaterialXCore/Document.cpp
```diff
@@ -305,6 +305,30 @@ ValuePtr Document::getGeomPropValue(const string& geomPropName, const string& ge
     return value;
 }
 
+std::vector<OutputPtr> Document::getMaterialOutputs() const
+{
+    vector<OutputPtr> materialOutputs;
+
+    const std::string documentUri = getSourceUri();
```
jstone-lucasfilm, line 312: Within the MaterialX libraries, I would recommend using `string` instead of `std::string`, since it's more compact and gives us more flexibility in the future.

source/MaterialXCore/Look.cpp
```diff
@@ -4,7 +4,6 @@
 //
 
 #include <MaterialXCore/Look.h>
-
```
jstone-lucasfilm, line 7: Unintended whitespace change here?

source/MaterialXCore/Look.cpp
```diff
@@ -125,6 +124,17 @@ NodePtr MaterialAssign::getReferencedMaterial() const
     return resolveRootNameReference<Node>(getMaterial());
 }
 
+std::vector<OutputPtr> MaterialAssign::getMaterialOutputs() const
```
jstone-lucasfilm, line 127: As above, I would recommend `vector` instead of `std::vector` here.

source/MaterialXTest/MaterialXCore/Material.cpp
```diff
@@ -54,3 +58,143 @@ TEST_CASE("Material", "[material]")
...
+{
+    mx::DocumentPtr doc = mx::createDocument();
+
+    mx::FileSearchPath searchPath;
+    const mx::FilePath currentPath = mx::FilePath::getCurrentPath();
+    searchPath.append(currentPath / mx::FilePath("libraries"));
+    searchPath.append(currentPath / mx::FilePath("resources/Materials/TestSuite"));
+
+    searchPath.append(mx::FilePath::getCurrentPath() / mx::FilePath("libraries"));
+    loadLibraries({ "targets", "stdlib", "pbrlib" }, searchPath, doc);
```
jstone-lucasfilm, line 72: What is the motivation for loading the standard libraries during material discovery?  Usually I think of that step only being required in code generation, or in validation of nodes against their nodedefs.

source/MaterialXTest/MaterialXCore/Material.cpp
```diff
@@ -54,3 +58,143 @@ TEST_CASE("Material", "[material]")
...
+        {
+            const std::string& nodeName = docNode->getNamePath();
+            CHECK(documentMaterialNodePaths.count(nodeName));
+            foundNames.insert(nodeName);
+        }
+        // Note: We are only comparing counts excluding any nodegraph
+        // implementations which contain material nodes.
+        CHECK(documentMaterialNodePaths.size() == foundNames.size());
+    }
+}
```
jstone-lucasfilm, line 200: Missing a newline here

source/MaterialXTest/MaterialXCore/Material.cpp
```diff
@@ -54,3 +58,143 @@ TEST_CASE("Material", "[material]")
...
+    mx::readFromXmlFile(doc, filename, searchPath);
+
+    mx::StringSet foundNames;
+
+    // 1. Find all materials referenced by material assignments.
+    //    Tests both finding direct node connections as well as traversing through
+    //    connected nodegraphs
+    mx::StringSet assignmentMaterialPaths =
+    {
+        "top_level_material_assigned", // reference to material node from standard library
```
jstone-lucasfilm, line 84: Could we potentially remove this string set, and just check the count of discovered paths?  It seems ideal to minimize the coupling between unit tests and specific test documents, allowing element names to change over time without our tests falling out of sync.

Any additional simplification that can be made to this unit test would also be appreciated, since its complexity seems out of balance with the simplicity of the newly-added methods.  :)

source/MaterialXCore/Look.cpp
kwokcb: Removing extra blank line.

resources/Materials/TestSuite/stdlib/materials/material_node_discovery.mtlx
kwokcb: I generated this where the nodedef is always written out to improve search performance for nodedef association. Not needed here so removing.

resources/Materials/TestSuite/stdlib/materials/material_node_discovery.mtlx
kwokcb: Application artifact. Removed.

source/MaterialXTest/MaterialXCore/Material.cpp
kwokcb: Just copy and paste of test setup. Removed.

source/MaterialXTest/MaterialXCore/Material.cpp
kwokcb: Removed the test on node interfaces + specific names and only kept the ones for outputs. Don't think there is anything simpler now :).

jstone-lucasfilm (approved):
This looks great, thanks @bernardkwok!

PR #745, opened 2021-11-04
Refactor TextureBaker and add support for baking material subsets

mjyip-lucasfilm:
### TextureBaker Refactor
* Removed `createBakeDocuments`. 
* Created function `bakeMaterialToDoc` (contains functionality nested in the loop of `createBakeDocuments` which iterated through each `renderablePath`).
* Moved public facing `bakeMaterial` to protected `generateBakedDocumentFromShader`.


mjyip-lucasfilm:
@jstone-lucasfilm I am not sure why I can't add you as a reviewer but I will tag you in this pull request thread for now. Let me know your thoughts. Thanks!

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -103,73 +93,6 @@ TextureBaker::TextureBaker(unsigned int width, unsigned int height, Image::BaseT
     _frameCaptureImage->createResourceBuffer();
 }
 
-size_t TextureBaker::findVarInTemplate(const string& filename, const string& var, size_t start)
```
jstone-lucasfilm, line 106: For the clarity of reviewing this change and tracking it over time, I would recommend defining these functions in their original order within the source file, so that GitHub doesn't think they've all been deleted and then re-added in a different location.  It's fine if they're declared in a different, more canonical order in the header, since that establishes the public API for the class and its methods.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -21,18 +21,8 @@ namespace {
...
-    std::vector<TypedElementPtr> elems;
-    findRenderableElements(doc, elems);
-    for (TypedElementPtr elem : elems)
-    {
-        renderablePaths.push_back(elem->getNamePath());
-    }
-    return renderablePaths;
-}
+const string SHADER_PREFIX = "SR_";
+const string UDIM_PREFIX = "_";
```
jstone-lucasfilm, line 25: Here I'd recommend the name DEFAULT_UDIM_PREFIX for this constant, in order to clarify that it can be overridden by the user.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -310,7 +233,220 @@ void TextureBaker::optimizeBakedTextures(NodePtr shader)
...
+            resolvedRenderablePath = resolver->resolve(materialPath, FILENAME_TYPE_STRING);
+        }
+        ElementPtr elem = doc->getDescendant(resolvedRenderablePath);
+        if (!elem || !elem->isA<Node>())
+        {
+            continue;
+        }
+        NodePtr materialNode = elem->asA<Node>();
+
+        vector<NodePtr> shaderNodes = getShaderNodes(materialNode);
```
jstone-lucasfilm, line 294: Could this computation of `shaderNode` be moved outside of the material tag loop?  It seems slightly odd for the variable to be declared outside the loop but defined internally, leaving the last definition of the variable to be used by subsequent logic.

source/MaterialXRenderGlsl/TextureBaker.h
```diff
@@ -243,6 +248,9 @@ class MX_RENDERGLSL_API TextureBaker : public GlslRenderer
     // Generate a texture filename for the given graph output.
     FilePath generateTextureFilename(const StringMap& fileTemplateMap);
 
+    // Create document that links shader outputs to a material.
+    DocumentPtr generateBakedDocumentFromShader(NodePtr shader, const StringVec& udimSet);
```
jstone-lucasfilm, line 252: Perhaps it would be clearer to name this method `generateNewDocumentFromShader`, since it doesn't actually perform any texture baking work.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -310,7 +233,220 @@ void TextureBaker::optimizeBakedTextures(NodePtr shader)
     }
 }
 
-DocumentPtr TextureBaker::bakeMaterial(NodePtr shader, const StringVec& udimSet)
+DocumentPtr TextureBaker::bakeMaterialToDoc(DocumentPtr doc, const FileSearchPath& searchPath, const string& materialPath, 
+                                              const StringVec udimSet, string& documentName)
```
jstone-lucasfilm, line 237: It doesn't look like the `documentName` argument is used by the method?

jstone-lucasfilm (commented):
This looks like a great start, and I've added some notes and recommendations below.

source/MaterialXRenderGlsl/TextureBaker.cpp
mjyip-lucasfilm: I set the value for `documentName` on line 317. The original code would return a `std::pair` of the (`documentName`, `bakedDoc`). I thought it would be more intuitive to return `bakedDoc` and to pass `documentName` by reference and modify the value within the code.

jstone-lucasfilm: Ah, that makes sense.  Ok, let me take a second look at the interface of this method today, taking all of the latest changes into account.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -482,67 +480,73 @@ BakedDocumentVec TextureBaker::createBakeDocuments(DocumentPtr doc, const FileSe
...
+    // Compute the material tag set.
+    StringVec materialTags = udimSet;
+    if (materialTags.empty())
+    {
+        materialTags.push_back(EMPTY_STRING);
+    }
+
+    if (_outputStream)
+    {
+        *_outputStream << std::endl << "Working on material: " << materialPath << std::endl;
```
jstone-lucasfilm, line 493: I like the idea of displaying this "processing" message at the start of each material bake, so that users can follow the progress of a multi-material bake command.  For the extra newline that separates the messages for each material, I would recommend placing it after the final "wrote baked document" message, rather than before the initial processing message.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -482,67 +480,73 @@ BakedDocumentVec TextureBaker::createBakeDocuments(DocumentPtr doc, const FileSe
...
+    }
+
+    NodePtr shaderNode;
```
jstone-lucasfilm, line 496: What are your thoughts on this `shaderNode` variable, which is declared outside of the material tag loop, but defined multiple times within it?  If it depends only upon the `materialPath` that is passed into the method, then perhaps it could be computed entirely outside of the loop, thus simplifying this logic?

source/MaterialXRenderGlsl/TextureBaker.h
```diff
@@ -203,13 +209,12 @@ class MX_RENDERGLSL_API TextureBaker : public GlslRenderer
     /// Optimize baked textures before writing.
     void optimizeBakedTextures(NodePtr shader);
 
-    /// Write the baked material with textures to a document.
-    DocumentPtr bakeMaterial(NodePtr shader, const StringVec& udimSet);
+    // TODO: Write images to memory rather than to disk
```
jstone-lucasfilm, line 212: Thanks for adding this TODO, which is an important reminder of the future work needed here.  To keep the public API of this method simple and straightforward to read, I would recommend moving the TODO into the method itself, just above the block of code that handles the writing of baked images to disk.

jstone-lucasfilm (approved):
Thanks @mjyip-lucasfilm, and these look like significant improvements to the API for texture baking!

PR #747, opened 2021-11-06
Refactoring of BSDF handling in shader generation

niklasharrysson:
This change list is a refactor of some concepts for BSDF handling in MaterialX shader generation. The two main changes are:

**1. Propagation of BSDF throughput**

In BSDF evaluation the result has been separated into _response_ and _throughput_. Response is the normal fraction of light reflected, while throughput is the remaining fraction of light transmitted after reflection and absorption. For opaque substances this throughput is always zero, since all light that is not reflected is then absorbed. A typical example is conductors/metals. But for dielectric media light can transmit, possibly through multiple vertical layers, and here the throughput is used to track the fraction of light remaining as it arrives at each interface in a layer stack.

This makes it possible to calculate vertical layering on any two BSDF's or mixtures of BSDF's. This was not possible before when vertical layering was done by transforming the BSDF layer operator graph into a nesting of BSDFs, and only a limited number of BSDFs support such nested layering.

The new changes applies mainly to GLSL codegen. For now it also applies to OSL codegen, since we currently calculate vertical layering in shader code, outside from the closures. For longer term, when the OSL reference closures are implemented, this should be handled inside the closures instead. For MDL only BSDF nesting can be used for layering, so this change does not apply there. Other improvements can be done on MDL side to support more layering configuration, by transforming the layering graphs, but his is future work and not part of this change list. If an unsupported configuration is found MDL will for now just render the top BSDF as if it was opaque.

**2. Propagation of BSDF parameters**

Setting of BSDF parameters for thin-film was previously done by direct assignment from the top thin-film layer to the base BSDF layer it applies to. Hence layering of thin-film could only be done directly on top of an atomic BSDF node.

In this change list we instead propagate these closure parameters through the BSDF network. For example, if the thin-film is applied to a BSDF mix node the parameters will propagate to both BSDF components upstream. If it's applied to a BSDF layer node it will be applied to only the top component upstream. Hence it's possible to apply thin-film to both atomic BSDFs, mixtures and layer stacks. This can be extended to include other layering parameters in the future.

This change also applies mainly to GLSL and OSL. For MDL there is an explicit node handling thin-film so it does not apply there.

**Other changes**

- Some improvements towards supporting VDF's with the BSDF-over-VDF version of the layer operator.
- Cleanup of legacy elements in test suite documents.
- Adding missing class export directives.


jstone-lucasfilm:
This is fantastic work, thanks @niklasharrysson!  I'll continue looking at the new code for any issues, but from a high level I'd like to merge this to main as soon as is practical, so that it can be a foundation for future development on MaterialX PBS.

source/MaterialXGenMdl/Nodes/ClosureCompoundNodeMdl.h
```diff
@@ -0,0 +1,30 @@
...
+class MX_GENMDL_API ClosureCompoundNodeMdl : public CompoundNodeMdl
+{
+  public:
+    static ShaderNodeImplPtr create();
+
+    void emitFunctionDefinition(const ShaderNode& node, GenContext& context, ShaderStage& stage) const override;
+    void emitFunctionCall(const ShaderNode& node, GenContext& context, ShaderStage& stage) const override;
+
+protected:
+    string _returnStruct;
```
jstone-lucasfilm, line 25: It looks like this member variable shadows one with an identical name in `CompundNodeMdl`, so we should either merge them into a single variable, or give this one a unique name.

source/MaterialXGenMdl/Nodes/ClosureLayerNodeMdl.h
```diff
@@ -0,0 +1,43 @@
...
+
+    void emitFunctionCall(const ShaderNode& node, GenContext& context, ShaderStage& stage) const override;
+
+    /// String constants
+    static const string TOP;
+    static const string BASE;
+};
+
+
+/// Layarable BSDF node.
```
jstone-lucasfilm, line 32: Minor spelling note, should be "layerable".

source/MaterialXGenMdl/Nodes/ClosureCompoundNodeMdl.cpp
```diff
@@ -0,0 +1,184 @@
+//
+// TM & (c) 2020 Lucasfilm Entertainment Company Ltd. and Lucasfilm Ltd.
```
jstone-lucasfilm, line 2: For any new files in the changelist, we should mark the copyright date as "2021", following our convention that copyrights state the year in which the file was contributed to the codebase.

source/MaterialXGenMdl/Nodes/ClosureCompoundNodeMdl.h
niklasharrysson: Good catch!

jstone-lucasfilm (approved):

PR #749, opened 2021-11-08
Fix the incorrect package version in CMake file

kwokcb:
The package version number set in CMake is incorrect. 
It is using the cmake version vs the MaterialX Version


kwokcb:
@jstone-lucasfilm , this is a fix from @JGamache-autodesk , than I'm putting up the PR for.

jstone-lucasfilm:
Good catch, thanks @bernardkwok and @JGamache-autodesk!

PR #753, opened 2021-11-11
linux: ensure libMaterialXRenderHw has a dependency on libXt

pmolodo:
After linking a program against a MaterialX build I made recently on linux, I got this error on execution:

```
Error: /blah/libMaterialXRenderHw.so: undefined symbol: topLevelShellWidgetClass
```

On inspection, the problem was that `libMaterialXRenderHw.so` wasn't declaring a dependency on `libXt.so`. When I inspected the Cmake source, I found this was because RenderHw wasn't doing a `find_package(X11)` - so that even though it calls `target_link_libraries(${X11_Xt_LIB})`, it does nothing because `${X11_Xt_LIB}` hasn't been populated yet.


jstone-lucasfilm:
Thanks for writing up this fix, @pmolodo, and it looks like a very worthwhile improvement.  If you haven't already, make sure to fill out a Contribute License Agreement (either individual or corporate), so that we can move forward and merge the change to MaterialX main:

https://github.com/materialx/MaterialX/blob/main/CONTRIBUTING.md

pmolodo:
Ah, I had not - just submitted one now, though!

jstone-lucasfilm (approved):
Looks good to me, thanks @pmolodo!

PR #754, opened 2021-11-11
Update to support nodedef/nodegraph associations via implementation elements

bernardkwok:
Add ability to find a nodedef / nodegraph definition via an implementation.
Update "standard surface" to allow for 2 versions to be specified in the library.


jstone-lucasfilm (approved):
Thanks for this change, @bernard, and it looks great to me.  This seems like a key step forward with respect to maintaining the history of shading models and other MaterialX nodes over time.

PR #758, opened 2021-11-15
Merge of some additional fixes after BSDF refactor

niklasharrysson:
* Fix issue with getting implementations for nodedefs with inheritance.
* Fix regression with vertex data prefix.


jstone-lucasfilm (approved):
Looks good, thanks for these fixes!

Issue #728, opened 2021-10-16
cmake files are installed into a wrong location


yurivict:
They are installed into ```$PREFIX/cmake``` when a standard location is ```$PREFIX/lib/cmake```.

[Here](https://github.com/materialx/MaterialX/blob/main/CMakeLists.txt#L231) and [here](https://github.com/materialx/MaterialX/blob/main/CMakeLists.txt#L234).

jstone-lucasfilm:
Good observation, @yurivict, and I can confirm that the OpenEXR project follows the pattern you recommend here, where CMake files are installed into `$CMAKE_INSTALL_PREFIX/lib/cmake`.  I'm adding @bernardkwok to the discussion, since this logic was originally written on the Autodesk side, but the proposed change seems logical to me.

Feel free to put together a pull request for this change if you have the bandwidth; otherwise, we'll take a look on our side once all parties have had a chance to weigh in.

jasonliu--:
> IMO this is a very involved and important issue to be left to outside people to submit a PR. Many files of different kind are affected. The project should define an internal variable for the path of all files and update all relevant places accordingly.
> 
> Currently the project just disregards UNIX file layout conventions.

I agree, it's not just the CMake files which disregard the FHS conventions, but rather the entire project. Another example would be the fact that MaterialX places many files into the `resources/` and `libraries/` directories, which MaterialX's build files assume are located in `$CMAKE_INSTALL_PREFIX`. For my MacPorts package, I have to relocate those folders into a directory that conforms to the FHS... in the case of my package, I decided to place them into `/opt/local/share/MaterialX` (MacPorts uses `/opt/local` as its standard installation location; other build systems might use `/usr/local`, etc.). The reason why @yurivict is correct in saying that the overall issue is "very involved" and that "the project should define an internal variable for the path of all files" is because if I change the value of `$CMAKE_INSTALL_PREFIX` at the start of the built, MaterialX will _still_ place everything into `/opt/local/resources` and `/opt/local/libraries`, seemingly ignoring whatever I set `$CMAKE_INSTALL_PREFIX` to (I actually get around the issue by manually moving the `resources/` and `libraries/` directories in to the `share` folder _after_ the output products get generated by MaterialX's build process). In addition, because the MaterialX viewer uses `resources/` and `libraries/` to house its default materials, I actually have to go in and patch the viewer's source code to add `/opt/local/share/MaterialX` to the viewer's default search paths.

[...]

PR #762, opened 2021-11-17
Install cmake files into lib/cmake, not just cmake.

yurivict:
cmake files are commonly installed into ```lib/cmake```.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,7 +228,7 @@ export(EXPORT MaterialX
 
 # Package for using the installed build
 install(EXPORT MaterialX
-        DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake
+        DESTINATION ${MATERIALX_INSTALL_LIB_PATH}/cmake
         FILE MaterialXConfig.cmake)
 install(FILES "${CMAKE_BINARY_DIR}/cmake/MaterialXConfigVersion.cmake"
-        DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake)
+        DESTINATION ${MATERIALX_INSTALL_LIB_PATH}/cmake)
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks for this proposal, @yurivict, which is closely related to the discussion in #728.

In that same discussion, @meshula suggested using `${CMAKE_INSTALL_LIB_DIR}` instead of the hardcoded string `lib`, and that seems like a worthwhile refinement to make here.

https://github.com/materialx/MaterialX/issues/728#issuecomment-955572417

yurivict:
I updated this PR.

yurivict:
Ad it turned out ```CMAKE_INSTALL_LIB_DIR``` isn't a standard cmake variable so I changed it back to ```lib```.

jstone-lucasfilm:
Sorry about that, @yurivict, I think we may have misspelled the variable name in our recommendation.  It looks like the standard variable is `CMAKE_INSTALL_LIBDIR` without the underscore between `LIB` and `DIR`.

Here's an example of that variable in the OpenEXR project for reference:

https://github.com/AcademySoftwareFoundation/openexr/blob/ee35a19888d20c5e794b8e7edb82b93accd978c7/cmake/CMakeLists.txt#L240

jstone-lucasfilm:
Stepping back from the naming convention in CMake, though, it looks like we already have a variable in MaterialX that defines this installation path:

https://github.com/materialx/MaterialX/blob/72ae8c546e964afa65a0ddd8661d66a236c7ccd1/CMakeLists.txt#L43

Should we perhaps be using `MATERIALX_INSTALL_LIB_PATH` here instead of the raw CMake version?

yurivict:
No problem. I've updated the patch again.

jstone-lucasfilm:
Thanks @yurivict, and I've gone ahead and made the additional update from `CMAKE_INSTALL_LIBDIR` to `MATERIALX_INSTALL_LIB_PATH`, so that the installed library paths are consistent across MaterialX.  That should make it straightforward to change the installed library path globally, if needed in the future, just by changing the default value of `MATERIALX_INSTALL_LIB_PATH`.

PR #765, opened 2021-11-19
Fix viewer performance for multi-UDIM assets

jstone-lucasfilm:
This changelist fixes a regression in the performance of the MaterialX viewer for multi-UDIM assets, which was caused by unbinding geometries between the rendering of each mesh partition.

Additionally, it adds the helper method GeometryHandler::findParentMesh, which is used to clarify mesh and shader binding logic in Viewer::renderFrame.


jstone-lucasfilm:
As a note for the future, we might consider refactoring multi-mesh support (#699) by handling geometry existing bindings in the `GlslProgram::bindMesh` method, so that the viewer and other clients of MaterialXRenderGlsl don't need to manage this on their own.

PR #767, opened 2021-11-22
Deterministic codegen

niklasharrysson:
This change list adds some changes to make codegen deterministic, always producing the same code from the same input data.

* Change connection storage on ShaderOutput from set to vector to make connection ordering deterministic.
* Make ShaderGraph topological sort order independent of std::unorder_map implementation.
* Add test for deterministic codegen.

### DIFF
--- a/source/MaterialXGenShader/ShaderNode.cpp
+++ b/source/MaterialXGenShader/ShaderNode.cpp
@@ -45,16 +45,32 @@ ShaderInput::ShaderInput(ShaderNode* node, const TypeDesc* type, const string& n
 
 void ShaderInput::makeConnection(ShaderOutput* src)
 {
-    breakConnection();
-    _connection = src;
-    src->_connections.insert(this);
+    // Make sure this is a new connection.
+    if (src != _connection)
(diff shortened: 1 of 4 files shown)


source/MaterialXGenShader/ShaderNode.cpp
```diff
@@ -45,16 +45,32 @@ ShaderInput::ShaderInput(ShaderNode* node, const TypeDesc* type, const string& n
...
 }
 
 void ShaderInput::breakConnection()
 {
     if (_connection)
     {
-        _connection->_connections.erase(this);
+        // Find and erase this input from the connected output's connection vector.
+        ShaderInputVec& connected = _connection->_connections;
+        ShaderInputVec::iterator it = std::find(connected.begin(), connected.end(), this);
```
niklasharrysson, line 68: Breaking a connection is now worst case O(n) with regards to the number of inputs connected to. But this number is very small. In the large majority of cases there is just a single input connected downstream.

jstone-lucasfilm (approved):

PR #776, opened 2021-12-06
Add reference GLSL implementation for Worley Noise nodes.

code-monkey:
Add reference GLSL implementation for Worley Noise nodes.  This adds the reference implementation to `lib/mx_noise.gls`l and new `.glsl` functions that call into the noise library.  `stdlib_genglsl_impl.mtlx` was modified to call the new functions.
This is the implementation as discussed on the ASWF slack.

### DIFF
--- a/libraries/stdlib/genglsl/lib/mx_noise.glsl
+++ b/libraries/stdlib/genglsl/lib/mx_noise.glsl
@@ -319,3 +319,206 @@ vec3 mx_fractal_noise_vec3(vec3 p, int octaves, float lacunarity, float diminish
     }
     return result;
 }
+
+float mx_8bits_to_01(uint bits) { return float(bits) / float(0xff); }
+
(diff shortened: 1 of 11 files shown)


jstone-lucasfilm:
This changelist looks great to me, @code-monkey, and I don't see any issues with merging it to main.  As a next step, we should evaluate what would be required to port it to OSL and MDL, and I'm adding @niklasharrysson and @bernardkwok for their thoughts on this.

niklasharrysson:
@code-monkey @jstone-lucasfilm This looks good! There should not be any problems porting this to OSL and MDL, and I would be happy to do those ports when I'm back working again in the new year.

jstone-lucasfilm:
Thanks for the confirmation, @niklasharrysson, and it sounds like we're ready to merge this initial change to MaterialX main.

Issue #772, opened 2021-11-30
Client project fails to link against shared MaterialX build on Windows


JGamache-autodesk:
To reproduce:
- On windows, build MaterialX with `-DMATERIALX_BUILD_SHARED_LIBS=ON`
- Try to build either USD or MayaUSD against the produced shared library
Result: Build fails to find global constants like mx::EMPTY_STRING at link time

Cause:
The header [MaterialXCore/Library.h](https://github.com/materialx/MaterialX/blob/main/source/MaterialXCore/Library.h#L23) requires MATERIALX_BUILD_SHARED_LIBS to be defined to select the correct shared import/export definitions.

Fix:
Make MaterialXCore/Library.h a configurable file (just like Version.h.in) that toggles between the static or shared import/export definitions based on the value of MATERIALX_BUILD_SHARED_LIBS at configure time. With this in place, the call to `add_definitions(-DMATERIALX_BUILD_SHARED_LIBS)` in the main [CMakeLists.txt](https://github.com/materialx/MaterialX/blob/main/CMakeLists.txt#L152) should no longer be required

JGamache-autodesk:
Another better fix, similar to what OpenColorIO is doing, is to have the generated CMake config reflect that variable:

This requires changing all library CMakeLists.txt to expose the compile definition as "PUBLIC", which means downstream projects need to define it as well. For example, adding this to MaterialXCore/CMakeLists.txt:
```
if (MATERIALX_BUILD_SHARED_LIBS)
target_compile_definitions(MaterialXCore PUBLIC -DMATERIALX_BUILD_SHARED_LIBS)
endif()
```
Will end up in the generated ${INSTALL_DIR}/lib/cmake/MaterialXConfig.cmake as:
```
set_target_properties(MaterialXCore PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "MATERIALX_BUILD_SHARED_LIBS"
)
```
And libraries using the generic `find_package(MaterialX) will automatically work correctly.

jstone-lucasfilm:
Thanks for posting this issue, @JGamache-autodesk, and your second proposal sounds like a great solution to me!

jstone-lucasfilm:
This has been addressed in pull request #780, so I'm closing out this issue.  Thanks for both the original report and the fix!

PR #780, opened 2021-12-09
Update CMake config file

kwokcb:
Update #772 
- Stores shared library state for downstream Windows clients
- Adds utility paths found in legacy FindMaterialX.cmake
- Fixes config file location to follow CMake guidelines
Tested by building USD on Windows using generated config files of a
MaterialX built with shared libraries.


jstone-lucasfilm:
This looks great to me, thanks @bernardkwok and @JGamache-autodesk.

ix-dcourtois:
Hello, I updated our MaterialX version to 1.38.3 and I hit a problem with this commit. Not sure if I should open a new issue, but the problem with this commit is that the pathes in the targets.cmake file are now absolute.
Which means the whole package is not redistributable

JGamache-autodesk:
Note: this commit is incomplete and leads to hardcoded paths in MaterialXTargets.cmake, which is fixed (sadly post 1.38.3) in PR 805.

PR #784, opened 2021-12-10
Re-expose virtual callback to emit light functions

kwokcb:
This callback was removed in PR #747. We restore it so USD can keep
using it in their codegen.


jstone-lucasfilm (approved):
Looks good, thanks @kwokcb and @JGamache-autodesk.

Issue #773, opened 2021-11-30
Allow build to set custom namespace


kwokcb:
To avoid possible clashes, it has been recommended to allow for custom namespace setting.
Currently MaterialX has a hard-coded name space but this could be settable via a Cmake option
as is done for OCIO. Then each integration can set it's own namespace to help avoid clashes if MaterialX
is pulled in via multiple integrations.

Note that this could be a possible avenue to avoid things like global symbol clashes as mentioned in #772.

This is a check-in in OCIO repo for namespace setting:
https://github.com/AcademySoftwareFoundation/OpenColorIO/commit/b3f6e9d1dd22b7b86bdcd2820c4e6a69b353ad9e
Note: that namespace can also be used to encode version so the API entry points can be versioned:
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/CMakeLists.txt#L203-L214

jstone-lucasfilm:
Thanks for posting this issue, @kwokcb, and it's been resolved in #785.

PR #785, opened 2021-12-10
Custom namespace and library naming support

kwokcb:
Update #773


python/Scripts/mxcustomnamespacefixer.py
```diff
@@ -0,0 +1,139 @@
+# Fixup code namespaces to be customizable:
+# As run for fixing namespaces in codebase for Github issue #773
```
jstone-lucasfilm, line 2: Although this is a useful script to keep around, I wouldn't recommend adding it to MaterialX main, since it's not a script we'd expect other teams to leverage in their own pipelines.

source/MaterialXCore/Definition.cpp
```diff
@@ -7,8 +7,8 @@
 
 #include <MaterialXCore/Document.h>
 
-namespace MaterialX
-{
+MATERIALX_NAMESPACE_BEGIN
+
```
jstone-lucasfilm, line 11: I'd recommend removing these double newlines after calls to MATERIALX_NAMESPACE_BEGIN, and setting the standard that only a single newline is needed after these calls.

CMakeLists.txt
```diff
@@ -57,6 +57,27 @@ if (MATERIALX_BUILD_GEN_MDL)
     set(MATERIALX_INSTALL_MDL_MODULE_PATH ${MATERIALX_INSTALL_STDLIB_PATH} CACHE FILEPATH "Install path for mdl module.")
 endif()
 
+###############################################################################
```
jstone-lucasfilm, line 60: For consistency with other CMake scripts in MaterialX, I would recommend removing these hash separators, and letting the text comments stand on their own.

jstone-lucasfilm (commented):
This looks like a great step forward, thanks @kwokcb and @JGamache-autodesk, and I have just minor recommendations below for consistency.

source/MaterialXCore/Definition.cpp
JGamache-autodesk: Done.

jstone-lucasfilm (approved):
This looks ready to merge, thanks for putting together this change, @JGamache-autodesk!

PR #798, opened 2021-12-16
Improved document validation

jstone-lucasfilm:
- Add validation for node interfaces when standard libraries are present in a document.
- Fix errors caught by the new validation logic.
- Remove the older example Looks.mtlx, which no longer passes validation or reflects current usage.


resources/Materials/TestSuite/stdlib/upgrade/1_36_to_1_37.mtlx
```diff
@@ -117,10 +117,6 @@ Upgrade path test from 1.36 to 1.37
     <input name="in" type="vector3" value="1.0, 0.5, 0.0"/>
     <input name="mat" type="matrix33" value="0.5, 0.0, 0.0, 0.0, 2.0, 0.0, 0.0, 0.0, 0.5"/>
   </transformnormal>
-  <transformnormal name="normal_transform34" type="vector4">
```
bernardkwok, line 120: Curious as to why this case is no longer tested ?

source/MaterialXTest/MaterialXFormat/XmlIo.cpp
```diff
@@ -225,32 +225,3 @@ TEST_CASE("Load content", "[xmlio]")
     mx::DocumentPtr nonExistentDoc = mx::createDocument();
     REQUIRE_THROWS_AS(mx::readFromXmlFile(nonExistentDoc, "NonExistent.mtlx", mx::FileSearchPath(), &readOptions), mx::ExceptionFileMissing&);
 }
-
-TEST_CASE("In memory xincludes", "[xmlio]")
```
bernardkwok, line 229: The test itself here is for xinclude logic, and happens to use the Looks.mtlx file.
In this case the test should remain but just reference any other file and test for a child other than `testLooks1.

python/MaterialXTest/main.py
```diff
@@ -27,20 +27,19 @@
 
 _fileDir = os.path.dirname(os.path.abspath(__file__))
 _libraryDir = os.path.join(_fileDir, '../../libraries/stdlib/')
-_exampleDir = os.path.join(_fileDir, '../../resources/Materials/Examples/Syntax/')
```
bernardkwok, line 30: Nice to see the removal of testing with Syntax files.

resources/Materials/TestSuite/stdlib/upgrade/1_36_to_1_37.mtlx
jstone-lucasfilm: With the new validation logic, what I discovered is that we never truly supported the upgrade of this particular legacy node in 1.36, since this upgrade path would require a change of node signature (not just a change of name).  If needed, we can add support for that upgrade pathway, but I've removed it from our test as it's not currently a supported feature.

source/MaterialXTest/MaterialXFormat/XmlIo.cpp
jstone-lucasfilm: If you have the bandwidth to help refactor this test, without requiring references to Looks.mtlx, that would be much appreciated!

PR #807, opened 2022-01-05
Fix shader compilation in albedo table generation

ashwinbhat:
The albedo table shader fails to compile because DIRECTIONAL_ALBEDO_METHOD is not defined.
DIRECTIONAL_ALBEDO_METHOD get’s defined if and only if a lighting nodes are found (see GlslShaderGenerator::requiresLighting)

Minor update to this to also check for hwWriteAlbedoTable in the Context.


jstone-lucasfilm (approved):
Thanks for putting together this fix, @ashwinbhat!

PR #810, opened 2022-01-16
Initial implementation of Worley noise for OSL

niklasharrysson:
Adds an initial implementation of Worley noise for OSL.

The hash is not yet identical to the one used by the GLSL and MDL implementations (an unsigned int right shift operator is missing in OSL). As discussed we will ignore this for now and we will refine this later to get identical results.

Example of the differences in results between GLSL and OSL.

worleynoise3d_float_out_glsl
worleynoise3d_float_out_osl

worleynoise3d_vector2_out_glsl
worleynoise3d_vector2_out_osl

worleynoise3d_vector3_out_glsl
worleynoise3d_vector3_out_osl


jstone-lucasfilm (approved):

lgritz:
Near as I can tell, mx_hash_int is implementing the exact same thing as the hash function inside OSL. OSL currently only exposes it to shaders for one int (`hash(int)`), but the 2 and 3 argument versions are implemented internally (and used for various noise functions), but simply not exposed directly to shaders. That seems fairly trivial to add, and that would eliminate the need for all the code in this patch that is "below" mx_hash_int.

libraries/stdlib/osl/mx_funcs.h
```diff
@@ -469,6 +469,303 @@ vector4 fBm( point position, int octaves, float lacunarity, float diminish, stri
...
+    // Either Euclidian or Distance^2
+    return dot(diff, diff);
+}
+
+float mx_worley_distance3(vector p, int x, int y, int z, int xoff, int yoff, int zoff, float jitter, int metric)
+{
+    int3 hash = mx_hash_int3(x+xoff, y+yoff, z+zoff);
+    vector off = vector(mx_8bits_to_01(hash.x),
+                        mx_8bits_to_01(hash.y),
+                        mx_8bits_to_01(hash.z));
```
lgritz, line 593: Note that the very definition of cellnoise (which is in MaterialX, and I presume is identical to that of OSL) serves the same purpose as this -- returning a deterministic uncorrelated [0,1] range value for each "cell" [i,i+1)^N in the volume. And that appears to be exactly what is at the heart of this construct and how it's used.

I feel like maybe an opportunity was missed here, where the specification of the worley functions in MX could have simply stipulated that the random point positions within each integer lattice cell were *by definition* identical to the cellnoise results for the contents of the cell. That would allow the top level of the mx_worley_noise_TYPE functions to call cellnoise to get the offsets and thus eliminate the majority of the lines of code in this module, as well as guaranteeing use of any carefully optimized cellnoise implementations in the {CPU, SIMD, or GPU} inside OSL or any other languages. Expanding the entire implementation out in shader source code down to every last bit twiddle seems very unlikely to be as efficient as the versions we've already optimized behind the cellnoise implementations.

The whole idea of cellnoise, back from when I first implemented it in PRMan (circa 1996?) was to be a "primitive function" from which you could build up higher level things like worley patterns, without having to get lost in the bit-by-bit details. This is literally what it was born to do!

Is it too late to allow for changes in the canonical pattern in order to pave the way for all the benefits of having cellnoise and worley overlap in their implementation and results?

jstone-lucasfilm: That's a great observation, and it's still possible to align the definitions of Worley noise and cellnoise in the way that you propose.  I'd be interested in thoughts from @code-monkey and @niklasharrysson on this idea, and I'll plan to bring this up at today's TSC meeting.

niklasharrysson: Great points @lgritz! And I also agree it would be nice to align this. I think we may need some input from @code-monkey about the implementation details, and if/how the use of cellnoise would differ from the GLSL implementation.

If it can't be done that way we have the other option of exposing the 2D and 3D hash functions as you mentioned above.

code-monkey: In MaterialX, there are only versions of cellnoise which return a single float value.  They boil down to a simple hash function on the cell location.  It would be good to extend the GLSL noise library to include versions that returned a vec2/vec3/vec4 for the "offset" inside the cell.  If OSL versions already exist for these functions, it would be ideal to use them for multiple reasons.  These would be perfect for the Worley noise implementation (as Larry points out).
Maybe a future spec of MaterialX could also expose the cellnoise functions that return vector types.

The MaterialX  GLSL implementation for Worley Noise would need to be patched with the OSL hashing functions.  It's not too late to change the reference implementation (IMO).

PR #811, opened 2022-01-19
Improved OSL render testing

jstone-lucasfilm:
- Clarify the names of OSL test options, giving them a consistent MATERIALX_OSL prefix.
- Move the list of render test paths from code to data, adding a "renderTestPaths" option to options.mtlx.
- Manually convert relative paths in OSL render tests, working around an issue in testrender on Windows.
- Add support for launching MaterialXTest directly from Visual Studio.
- Fix typos in the transparency regression check for MaterialXGenShader.

### DIFF
--- a/resources/Materials/TestSuite/_options.mtlx
+++ b/resources/Materials/TestSuite/_options.mtlx
@@ -87,11 +87,11 @@
     <!-- Transforms UVs of loaded geometry -->
     <input name="transformUVs" type="matrix44" value="1.0f, 0.0f, 0.0f, 0.0f, 0.0f, -1.0f, 0.0f, 0.0f, 0.0f, 0.0f, 1.0f, 0.0f, -0.235f, 1.0f, 0.0f, 1.0f" />
 
-    <!-- External testing: extra comma separated list of library paths -->
-    <input name="externalLibraryPaths" type="string" value="" />
+    <!-- List of extra library paths for generator and render tests -->
+    <input name="extraLibraryPaths" type="string" value="" />
 
-    <!-- External testing: extra comma separated list of test root paths -->
-    <input name="externalTestPaths" type="string" value="" />
+    <!-- List of document paths for render tests -->
+    <input name="renderTestPaths" type="string" value="resources/Materials/Examples/StandardSurface,resources/Materials/Examples/UsdPreviewSurface,resources/Materials/TestSuite" />
 
     <!-- Wedge rendering options -->
     <input name="wedgeFiles" type="string" value="wedge_conductor.mtlx,wedge_conductor.mtlx" />
--- a/source/MaterialXRenderOsl/OslRenderer.cpp
+++ b/source/MaterialXRenderOsl/OslRenderer.cpp
@@ -83,7 +83,7 @@ void OslRenderer::renderOSL(const FilePath& dirPath, const string& shaderName, c
     string shaderPath = shaderFilePath.asString();
 
     // Set output image name.
-    string outputFileName = shaderPath + "_osl.exr";
+    string outputFileName = shaderPath + "_osl.png";
     _oslOutputFileName = outputFileName;
 
     // Use a known error file name to check
@@ -217,7 +217,7 @@ void OslRenderer::shadeOSL(const FilePath& dirPath, const string& shaderName, co
     string shaderPath = shaderFilePath.asString();
 
     // Set output image name.
-    string outputFileName = shaderPath + ".testshade.exr";
+    string outputFileName = shaderPath + ".testshade.png";
     _oslOutputFileName = outputFileName;
 
     // Use a known error file name to check
--- a/source/MaterialXRenderOsl/OslRenderer.cpp
+++ b/source/MaterialXRenderOsl/OslRenderer.cpp
@@ -5,17 +5,15 @@
 
 #include <MaterialXRenderOsl/OslRenderer.h>
 
-#include <MaterialXFormat/File.h>
-
-#include <MaterialXFormat/Util.h>
-
 #include <MaterialXGenOsl/OslShaderGenerator.h>
+
+#include <MaterialXFormat/File.h>
+#include <MaterialXFormat/Util.h>
 
 #include <fstream>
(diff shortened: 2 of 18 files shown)


resources/Materials/TestSuite/_options.mtlx
```diff
@@ -87,11 +87,11 @@
...
 
-    <!-- External testing: extra comma separated list of library paths -->
-    <input name="externalLibraryPaths" type="string" value="" />
+    <!-- List of extra library paths for generator and render tests -->
+    <input name="extraLibraryPaths" type="string" value="" />
 
-    <!-- External testing: extra comma separated list of test root paths -->
-    <input name="externalTestPaths" type="string" value="" />
+    <!-- List of document paths for render tests -->
+    <input name="renderTestPaths" type="string" value="resources/Materials/Examples/StandardSurface,resources/Materials/Examples/UsdPreviewSurface,resources/Materials/TestSuite" />
```
kwokcb, line 94: This is great to externalize this!
Just want to check here that this is now all the test paths to check instead of extra ones. Thus if this is empty, nothing will be considered to run.

source/MaterialXRenderOsl/OslRenderer.cpp
```diff
@@ -170,35 +165,41 @@ void OslRenderer::renderOSL(const FilePath& dirPath, const string& shaderName, c
     }
     command += " > " + errorFile + redirectString;
 
-    int returnValue = std::system(command.c_str());
-
-    std::ifstream errorStream(errorFile);
-    StringVec result;
-    string line;
-    const string pngWarning("libpng warning: iCCP: known incorrect sRGB profile");
```
kwokcb, line 178: Does that mean that oiio error does not come up when running the test suite ?
Might want to keep ignoring this in case it shows up when folks are using a different version of oiio ?

source/MaterialXTest/Main.cpp
```diff
@@ -25,6 +29,16 @@ int main(int argc, char* const argv[])
     }
 #endif
 
+    mx::FilePath resourcesPath = mx::FilePath::getCurrentPath() / "resources";
```
kwokcb, line 32: I'm curious why this is required ?

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -178,10 +178,10 @@ TEST_CASE("GenShader: Transparency Regression Check", "[genshader]")
         "Materials/Examples/StandardSurface/standard_surface_default.mtlx", 
         "Materials/Examples/StandardSurface/standard_surface_glass.mtlx",
         "Materials/TestSuite/libraries/metal/brass_wire_mesh.mtlx",
-        "Materials/TestStuie/pbrlib/surfaceshader/transparency_nodedef_test.mtlx",
-        "Materials/TestStuie/pbrlib/surfaceshader/transparency_test.mtlx",
+        "Materials/TestSuite/pbrlib/surfaceshader/transparency_nodedef_test.mtlx",
```
kwokcb, line 181: Hmm. This was my bad. Thanks for fixing this.

source/MaterialXTest/MaterialXRender/RenderUtil.cpp
```diff
@@ -134,24 +134,11 @@ bool ShaderRenderTester::validate(const mx::FilePathVec& testRootPaths, const mx
 
     mx::ScopedTimer ioTimer(&profileTimes.ioTime);
     mx::FilePathVec dirs;
-    if (options.externalTestPaths.size() == 0)
+    for (const auto& root : options.renderTestPaths)
```
kwokcb, line 137: Never mind about my previous comment about what get's run.
It is as assumed that only these paths are considered for testing now.
Much more flexible and does not force running the testsuite files.

source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
```diff
@@ -175,6 +177,14 @@ bool OslShaderRenderTester::runRenderer(const std::string& shaderName,
             }
             CHECK(shader->getSourceCode().length() > 0);
 
+            // Convert relative paths to absolute, using hardcoded logic for now.
```
kwokcb, line 180: This feels a bit too hard-coded. I added a flattenFilenames() utility a while back, but
this is an in-place Document modification, which is too much here. Still feels better to make this generic to resolve these paths vs string substitution more generally ?

resources/Materials/TestSuite/_options.mtlx
jstone-lucasfilm: That's correct, @bernardkwok, and this now represents the full set of paths to be used for render testing (across all languages).

source/MaterialXRenderOsl/OslRenderer.cpp
jstone-lucasfilm: I haven't encountered this error in the test suite so far, though you could be right that it may persist with some versions of OpenImageIO.  My recommendation would be to start with this approach, and then add back error suppression based on what we encounter in the latest codebase.

source/MaterialXTest/Main.cpp
jstone-lucasfilm: Good question, and this handles the test suite when it's run directly from Visual Studio, where CMake has no influence over the current working folder.  I should document this more effectively, so that the role of this code is clear.

source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
jstone-lucasfilm: Agreed that this solution is not great, but it's the only approach I found that generates correct textured renders in OSL on WIndows.  There's definitely room for improvement here, but this at least gets us working renders across platforms.

source/MaterialXTest/Main.cpp
jstone-lucasfilm: Thanks for the note, and I've now added a comment that clarifies this logic.

PR #813, opened 2022-01-20
add PyElement.getDescendant

pmolodo:
Addresses https://github.com/materialx/MaterialX/issues/812


jstone-lucasfilm (approved):
Looks great, @pmolodo, and thanks for the fix!

PR #814, opened 2022-01-24
Improvements to render test parity

jstone-lucasfilm:
This changelist implements a number of improvements to the parity of render tests between GLSL and OSL, including the following:

- Render with a unit sphere in both GLSL and OSL, adjusting the zoom in OSL to compensate for framing differences.
- Increase the environment map sample count in GLSL.
- Increase the rays per pixel for lit surfaces in OSL.
- Increase the tolerance for error messages in testrender.

### DIFF
--- a/source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
+++ b/source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
@@ -101,6 +101,7 @@ void GlslShaderRenderTester::registerLights(mx::DocumentPtr document,
     REQUIRE(envIrradiance);
     _lightHandler->setEnvRadianceMap(envRadiance);
     _lightHandler->setEnvIrradianceMap(envIrradiance);
+    _lightHandler->setEnvSamples(256);
 }
 
 //
--- a/resources/Materials/TestSuite/Utilities/closure_color_scene.xml
+++ b/resources/Materials/TestSuite/Utilities/closure_color_scene.xml
@@ -1,5 +1,6 @@
+<!-- Template for a lit scene in testrender -->
 <World>
-  <Camera eye="0, 0, 4"  dir="0,0,-1"  fov="70" />
+   <Camera eye="0, 0, 3" dir="0, 0, -1" fov="79.3" />
 
    <!-- Background environment map. -->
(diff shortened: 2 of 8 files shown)


jstone-lucasfilm:
[RenderTests_StandardSurface.pdf](https://github.com/materialx/MaterialX/files/7927054/RenderTests_StandardSurface.pdf)

source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
```diff
@@ -101,6 +101,7 @@ void GlslShaderRenderTester::registerLights(mx::DocumentPtr document,
     REQUIRE(envIrradiance);
     _lightHandler->setEnvRadianceMap(envRadiance);
     _lightHandler->setEnvIrradianceMap(envIrradiance);
+    _lightHandler->setEnvSamples(256);
```
kwokcb, line 104: I'm a bit worried about how long this will take to run with a lot of files.

Could be add a fidelity option into the `_options.mtlx` file which controls these kind of parameters. e.g. a "high" vs "low" quality setting ? "high" can include turning on shadow maps for GLSL for instance.

jstone-lucasfilm: I wouldn't expect a real-world performance impact from these extra samples, since they're performed on the GPU, and our render tests currently spend most of their time in file I/O and shader compilation.  For the future, I think you're right that a high/low option for quality could be valuable, especially if we end up running render tests on a software rasterizer one day.  But it's likely worthwhile to wait until we have a real-world use case for the "low" quality mode, and we can then tailor it specifically for the needs of that case.  For now, adding shadows wouldn't have a visual impact, since we're restricted to rendering spheres in order to match testrender, though this limitation is likely to be lifted in the future.

jstone-lucasfilm:
Here's a comparison of standard_surface_brass_tiled between GLSL and OSL, showing the visual impact of aligning cameras and render settings.

SR_brass1_glsl
SR_brass1_osl

Notice that there's still a slight UV coordinate mismatch between the two renders, which appears to be a geometric difference between the polygonal sphere in OpenGL and the parametric sphere in testrender.

PR #815, opened 2022-01-24
Fix up html generator for rendering tests

kwokcb:
Fix:
- Adding to source destination lists incorrectly, producing source on one row and destination on another.
- Cleaned up to use source to determine destination file name by replacing language to avoid name mismatches per row. If exists as source file, then append destination file. e.g. if `foo.glsl.png` found then  `foo.osl.png` is added. The second file is not checked until generation time.
- Fix to allow only one language to be shown (again) if both destination and source languages are the same.
-


jstone-lucasfilm:
Thanks for the fix, @bernardkwok!

PR #819, opened 2022-01-25
Fix to OBJ geometry: sphere and plane

kwokcb:
### Sphere Geometry
- Align geometry and UV origin border line with z-axis

### Plane Geometry
- Orient so UVs are not flipped in V

### Code fixes: 
Remove additional extra uv transform which was use to try and align UVs properly. 
Not required anymore.

### Results
[jan25_2022_GLSL_vs_OSL_tests.pdf](https://github.com/materialx/MaterialX/files/7935597/jan25_2022_GLSL_vs_OSL_tests.pdf)


jstone-lucasfilm:
This looks like a good step forward, thanks @bernardkwok!

PR #825, opened 2022-01-28
Add V-flip option for geometry loaders

kwokcb:
### Add V-flip option for geometry loaders
- Allows for customization on load allowing for tangent space to be computed on load
- Used for GLSL render unit testing when the specified geometry is the "stock" sphere or plane.
- Turn off texture flip as a consequence of flipping texture coordinates.
- Also simplified testing logic to either use the geometry file if specified or default to the stock sphere otherwise.


jstone-lucasfilm:

PR #829, opened 2022-02-02
Add new signatures for cellnoise to support higher dimensions of seed…

code-monkey:
…, but also to generate Vec3 results.  These methods are consistent with the OSL implementation of cellnoise.

Change Worley noise to use cellnoise to pick the offsets in the cell (rather than mx_hash_vec3()).  This is a more generic form that only uses public API calls in OSL so it can be implemented without any changes to the core OSL library.


niklasharrysson (approved):
@code-monkey Excellent! Thanks Mark

lgritz:
nice

jstone-lucasfilm (approved):
Looks good to me, thanks @code-monkey!

PR #830, opened 2022-02-03
Clarify usage of alpha and roughness

jstone-lucasfilm:
This changelist clarifies the usage of the terms alpha and roughness in the PBR libraries for GLSL and OSL, using the term 'alpha' to mean the square of artist-facing roughness in the context of specular BSDFs.

In the future, we should consider modifying the interfaces of specular BSDF nodes such as <dielectric_bsdf> and <conductor_bsdf> to follow this convention as well, since their "roughness" inputs actually represent alpha.

Additional changes:
- Align handling of input alpha between GLSL and OSL.
- Remove unused helper functions in the PBR library for OSL.


niklasharrysson (approved):
Nice clarification and cleanup. Thanks @jstone-lucasfilm !

I agree we can take it a step further and rename this on microfacet BSDF:s to make it fully consistent. We could then also improve the documentation of the nodes `roughness_anisotropy`, `roughness_dual` and `glossiness_anisotropy`, to clarify that these are doing conversions from roughness to alpha, and they should be used with the microfacet BSDF's to get a perceptually linear response.

PR #831, opened 2022-02-06
Refinements to GGX directional albedo

jstone-lucasfilm:
- Improve the precision of the computation in mx_ggx_dir_albedo_monte_carlo, moving the global component of the geometric term outside of the sample loop.
- Rebuild the constants in mx_ggx_dir_albedo_analytic using MATLAB.


jstone-lucasfilm:
Here's an updated comparison of GGX directional albedo textures, with the first image generated by a Monte Carlo simulation with 16K samples per pixel, and the second image generated by the latest analytic approximation:

Monte Carlo with 16K samples per pixel:
AlbedoTable_MonteCarlo_16384_Samples_Precise

Analytic approximation:
AlbedoTable_Analytic_16384_Samples_Precise_MatLab

PR #836, opened 2022-02-10
Port of new Worley noise implementation to OSL

niklasharrysson:
Adopting the new Worley noise implementation written by @code-monkey. The results are now identical between GLSL and OSL, and there are no longer any code duplication since the implementation uses OSL's built-in cellnoise functions.

worley2d

worley3d


lgritz:
Nice!

PR #838, opened 2022-02-11
Make fractal3d node consistent between GLSL and OSL

niklasharrysson:
Improve GLSL implementation to make fractal3d results consistent with OSL.

Before:
before

After:
after


jstone-lucasfilm (approved):
Looks good, thanks @niklasharrysson.  It's worth noting that this changelist only impacts the vector2 and vector4 forms of fractal3d, which are less commonly used, but it's still an important fix for alignment of procedurals between GLSL and OSL.

PR #842, opened 2022-02-15
Port noise improvements to MDL

niklasharrysson:
This change list ports the recent improvements to noise in GLSL/OSL to MDL. 

Results should now be consistent between all three languages.

Examples:
worley3d_float

worley3d_float3


jstone-lucasfilm (approved):
Brilliant, thanks @niklasharrysson!

PR #843, opened 2022-02-15
glsl mx_fresnel_conductor now uses physcially-correct function

BrianSharpe:
The previous implementation was an optimization, and gave incorrect results when "k" approached zero.
This change improves alignment with offline renderers

More details in the slack discussion here:
https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1644834192552769


niklasharrysson (approved):
Looks good! Thanks @BrianSharpe

jstone-lucasfilm (approved):
This looks good to me, @BrianSharpe, and thanks for the fix!

PR #845, opened 2022-02-17
Shader generation token resolution

kwokcb:
Update for token support make code generation (or any resolver call) correctly perform token substitution.
Can handle `<token>` elements within a node, a nodegraph and at the Document level. Those `<token>` which are closest in level to the input are considered higher priority than those at higher levels. 

1. StringResolver has new method called `addTokenSubstitutions()` to adding to the resolver mapping. This can be used on any resolver. It is called when geometry resolution is done (on resolver creation).  Note that no resolution occurs until the "resolve" is performed.
1. Added in validation for tokens on interface elements
2. Added unit test to test different levels of `<token>` resolution.
3. Added render test. Results are as follows: (all resolve properly -- the same result)

image


jstone-lucasfilm (approved):

PR #846, opened 2022-02-17
GLTF geometry loader using cgltf library.

kwokcb:
GLTF geometry only loader

* Used cgltf for loading
* Converts transform instancing into instances of geometry as this is not supported by the Mesh geometry classes and the viewer.
* Auto generates uvs, normals and tangents if not found.
* Does not handle materials or texture dependencies
* The viewer logic has been updated to handle multiple meshes.

Example using shader found under the Khronos MaterialX branch (https://github.com/KhronosGroup/MaterialX).
image

Multiple mesh example (Using sample MaterialX shaders)
image

Mesh example where uvs and tangents are generated on load.
image

Mesh example with multiple partitions converted to meshes
image

Mesh example with transform instancing + multiple partitions
image

Cesium Truck (has transform hierarchy under  nodes with meshes). Materials assignments from example materials.
cesium_ground_vehicle

Similar mesh + sub-hierarchy used for Sponza. Materials assignments from example materials.
Sponza3

All models tested are from: https://github.com/KhronosGroup/glTF-Sample-Models/ and https://github.com/CesiumGS/cesium


jstone-lucasfilm (approved):
This looks like a great addition to me, thanks @bernardkwok!

PR #852, opened 2022-03-01
Clarify Fresnel logic in GLSL

jstone-lucasfilm:
- Add constants for Fresnel model indices.
- Separate FresnelData parameters for physical and generalized Schlick models.
- Move shared initializations to an mx_init_fresnel_data function.


niklasharrysson (approved):
Looks great! Thanks for the cleanup @jstone-lucasfilm

Issue #835, opened 2022-02-09
Outputs from Multi-Output nodes not always passed correctly


klucknav:
It seems that when a multi-output node is directly providing the nodegraph output, only one of the outputs is actually used. There doesn't seem to be any issue if the outputs are passed through another node.

To see this in the MaterialXViewer, below is a test mtlx file with an ```<artistic_ior>``` node in the nodegraph. When the ```<artistic_ior>``` outputs are passed through a no-opt ```<dot>``` node, the sphere is the expected red color. However, when the nodegraph outputs are coming directly from the ```<artistic_ior>``` node, the sphere renders as a cyan color. 

The underlying issue can be seen in the main function of the generated Pixel Shader. The two different outputs are not being appropriately fed into ```IMPL_standard_surface_surfaceshader()```. Only the first output (ior in this case) is being used, and is passed in instead of the second output.

```XML
<?xml version="1.0"?>
<materialx version="1.38">

  <nodegraph name="test_ng">
    <artistic_ior name="artistic_ior" type="multioutput">
      <input name="reflectivity" type="color3" value="1,0,0" />
      <input name="edge_color" type="color3" value="1,1,1" />
    </artistic_ior>

    <!-- Option 1 - the expected red color -->
    <dot name="specular" type="color3">
      <input name="in" type="color3" nodename="artistic_ior" output="extinction" />
    </dot>
    <dot name="ior" type="color3">
      <input name="in" type="color3" nodename="artistic_ior" output="ior" />
    </dot>
    <output name="base_color_output" type="color3" nodename="specular"/>
    <output name="specular_color_output" type="color3" nodename="ior"/>

    <!-- Option 2 - a cyan color -->
    <!-- <output name="base_color_output" type="color3" nodename="artistic_ior" output="extinction" />
    <output name="specular_color_output" type="color3" nodename="artistic_ior" output="ior" /> -->
  </nodegraph>

  <standard_surface name="test_sr" type="surfaceshader">
    <input name="base_color" type="color3" output="base_color_output" nodegraph="test_ng" />
    <input name="specular_color" type="color3" output="specular_color_output" nodegraph="test_ng" />
  </standard_surface>
  <surfacematerial name="test_m" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="test_sr" />
  </surfacematerial>
</materialx>
```

kwokcb:
This is another example case found by Rasmus Bonnedal:
```xml
[...]

PR #853, opened 2022-03-01
Fix "output" qualifier traversal on nodegraph's output 

kwokcb:
Update #835

## Connection Setup
The situation is as follows:

```
[ngnode.output1]
[ngnode.output2]--->nodegraph.ngoutput-->node.input
```

During traversal `ngnode` is found when traversing from `node.input` as it goes through the nodegraph's output to find `ngnode`.There is however an `output` name which is used to specify the actual output to use. This it the output name specified on `node.input (nodegraph.ngoutput)` and was **not** be updated to (`ngnode.output2`).

This fix simply does the proper check to consider `ngoutput` and traverse upstream to get `output2` on `ngnode`.
 
 ## Test Results
 - Routing from red channel output of nodegraph. The original input color is `(0,1,0)` which is separated into individual channels. The green channel is routed to base which results in 1 * base color (white) giving a "white" result
image

 - Routing from blue channel output of nodegraph. The original input color is `(0,1,0)` which is separated into individual channels. The blue channel is routed to base which results in 0 * base color (white) giving a "black" result
image

 - Routing again from blue channel output of nodegraph. The original input color is `(0,0,1)` which is separated into individual channels. The blue channel is routed to base which results in 1 * base color (white) giving a "white" result. 
image


resources/Materials/TestSuite/stdlib/nodegraph_inputs/nodegraph_multioutput.mtlx
```diff
@@ -0,0 +1,31 @@
...
+      <input name="in" type="color3" interfacename="in" value="0, 1, 0"  />
+      <output name="outr" type="float" />
+      <output name="outg" type="float" />
+      <output name="outb" type="float" />
+    </separate3>
+    <output name="green_output" type="float" nodename="green_node" output="outg" />
+    <output name="blue_output" type="float" nodename="green_node" output="outb" />
+  </nodegraph>
+
+  <!-- Test to pick the appropraite output from the nodegraph. In this case this is the green output which is 1.0 or white -->
```
kwokcb, line 15: Add 2 shaders to route from G, and B which would not work before since it would ignore the "output" attribute names on nodegraph outputs and default to using the node and hence by default use the first output which is "outr".

jstone-lucasfilm (approved):

PR #855, opened 2022-03-05
Fix typo in MDL thin_surface

proog128:
This PR fixes a typo in the MDL shader generation.


jstone-lucasfilm (approved):
Thanks for the fix, @proog128!

Issue #839, opened 2022-02-11
Proposal: Add in native "unlit" shader


kwokcb:
Proposed new surface shader variant which does no compute any illumination.
This allows for any upstream pattern graph to be connected and used as a material.
Some applications include support for "modeling" and "NPR" materials.

Description:
The new node's interface would look like this. The possible nodegroup is "npbr" (non-pbr).
```
nodedef name="ND_surface_std" node="surface" nodegroup="npbr" doc="Construct a surface shader from emission and transmission values." version="1.0">
  <input name="emission" type="float" value="1.0" doc="Surface emission amount." />
  <input name="emission_color" type="color3" value="0,0,0" doc="Surface emission color." />
  <input name="transmission" type="float" value="0.0" doc="Surface transmission amount." />
  <input name="transmission_color" type="color3" value="0,0,0" doc="Surface transmission color." />
  <input name="opacity" type="float" value="1.0" doc="Surface cutout opacity." />
  <output name="out" type="surfaceshader" />
</nodedef>
```
where transmission can be used to perform blending with background elements as opposed to opacity which is just for cut-outs. The inverse of `transmission` is used for blending (this is similar to what is done for Standard Surface currently). A surface that is 100% transparent will not emit any color (not emit or reflect any light if it was a lit surface).

This is a variant on the existing surface shader,.
```
  <nodedef name="ND_surface" node="surface" nodegroup="pbr" doc="A constructor node for the surfaceshader type.">
    <input name="bsdf" type="BSDF" value="" doc="Distribution function for surface scattering." />
    <input name="edf" type="EDF" value="" doc="Distribution function for surface emission." />
    <input name="opacity" type="float" value="1.0" doc="Surface cutout opacity" />
    <output name="out" type="surfaceshader" />
  </nodedef>
  ```

Properties:
* This shader can be used anywhere a `surfaceshader` can current be used. Thus it can be connected to a `surfacematerial` and assigned.
* Is faster than using a "uniform edf" node 
* Does not require the `pbr` library to be present. Will be part of stdlib definitions.

Support:
* This can be the based node to support a new "non-pbr" library equivalant to the existing "pbr" library.
* Here common base utility nodes such as "view vector" or "up vector" would reside.  * Implementations which support pbr rendering do not need to support these new nodes.
* Below is a possible organization
![image](https://user-images.githubusercontent.com/49369885/153610373-9815033d-08d5-4996-a404-d35b183b40b7.png)

Examples:
[...]

PR #858, opened 2022-03-07
Unlit surface shader node implemenation

niklasharrysson:
Implementation of the "unlit" surface shader, as proposed in issue #839.

This change list adds the definitions and implementations for GLSL, OSL and MDL.

Some tests added to the test suite, testing visualization of a simple procedural with/without cutout opacity and with/without colored transmission:
unlit_surface_tests

Note that MDL rendering is not part of the testsuite yet and are here renderer separately with the MDL DXR sdk sample. Hence the difference in environment and geometry. Also note that for GLSL our renderer does not support colored transmission yet.


source/MaterialXView/Viewer.cpp
```diff
@@ -1415,7 +1415,11 @@ void Viewer::saveShaderSource(mx::GenContext& context)
         mx::TypedElementPtr elem = material ? material->getElement() : nullptr;
         if (elem)
         {
-            mx::ShaderPtr shader = createShader(elem->getNamePath(), context, elem);
+            const bool hasTransparency = mx::isTransparentSurface(elem, context.getShaderGenerator().getTarget());
+            mx::GenContext materialContext = context;
+            materialContext.getOptions().hwTransparency = hasTransparency;
+
+            mx::ShaderPtr shader = createShader(elem->getNamePath(), materialContext, elem);
```
niklasharrysson, line 1422: This code path was missing the check for transparency, so saved GLSL shaders would never include transparency handling.

niklasharrysson:
@jstone-lucasfilm I'm not sure with my EasyCLA is not picked up here, as it's signed and approved in the materialx-test repo.

niklasharrysson:
> @jstone-lucasfilm I'm not sure why my EasyCLA is not picked up here, as it's signed and approved in the materialx-test repo.

It seems I had my old Autodesk email address set locally on my laptop that I made commits from. Will try to rebase or create a new PR to make sure all commits are made from my current user / email.

source/MaterialXGenGlsl/Nodes/UnlitSurfaceNodeGlsl.cpp
```diff
@@ -0,0 +1,56 @@
...
+        shadergen.emitLineEnd(stage);
+
+        const string outColor = output->getVariable() + ".color";
+        const string outTransparency = output->getVariable() + ".transparency";
+
+        const ShaderInput* emission = node.getInput("emission");
+        const ShaderInput* emissionColor = node.getInput("emission_color");
+        shadergen.emitLine(outColor + " = " + shadergen.getUpstreamResult(emission, context) + " * " + shadergen.getUpstreamResult(emissionColor, context), stage);
+        
+        if (context.getOptions().hwTransparency)
```
kwokcb, line 37: Hi @niklasharrysson, I don't think you want to emit this based on the option as this makes it inconsistent between language implementations. Also, it would make it so you can't change the values which affect "outTransparency" without recompiling the shader.

niklasharrysson: Hey @kwokcb! Those are good points, but this convention has been used for a long time for our surface shading in GLSL., where transparency handling is not generated if the option is not set.

For example, see the PBR version of the surface constructor node here: 
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.cpp#L201

And also when emitting the final alpha blending of the fragment result here:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/GlslShaderGenerator.cpp#L643

So shader regeneration has always been required to introduce transparency to a surface shader in GLSL. While I agree it would be nice to improve on this, I think that should be done as part of a more holistic look at our transparency rendering for GLSL. Maybe it's worth adding a separate issue to track this.

niklasharrysson:
Closing this in favor of a new PR with proper EasyCLA setup: https://github.com/AcademySoftwareFoundation/MaterialX/pull/860)

Issue #859, opened 2022-03-08
Examine transparency handling for hardware code generation


kwokcb:
This issue has been logged for discussion purposes

## Background
- There is a code generation context setting on whether to consider "hardware transparency". By default this is set to false.
- The user must explicitly determine when this should be set to be true.
  - As a convenience a "is transparent" heuristic utility traverses the shade tree at a given shader root and examines specific inputs and input values to determine if the shader is transparent.
  - There have be requests to enhance this to recognize more cases. (the inputs considered are currently fixed)
- Depending on whether this option is set to true on not _different code_ is generated.

## Proposed Items To Examine:
- Whether the code to support transparency should always be emitted to avoid re-generating code which can be costly. 
  - If unique shaders are cached then 2 versions are required to support "transparent" and "opaque" variants, or the same shader may need to be regenerated + replaced on value changes.
  - A value change implies a shader regeneration requirement but there is no support for this mechanism of detection.
-  Whether the heuristic is too strict in checking for specific values to determine if a shade tree is (as opposed to could be transparent).
- `transmission` is not included in the final color computation and instead relies on h/w alpha blending which is a "cheap" way to draw transparent objects.  This should be re-examined as this may not be a given.
- Enhance heuristics to allow for additional inputs to be checked with custom logic. e.g. should `transmission_color` be checked for "unlit" shader.
- do we still want to add support for a transparency "hint" attribute as proposed in the current 1.39 specification ? A suggestion from @niklasharrysson is that we move such a hint to `<implementation>`s as this has as it's main usage hardware rasterizers.
- more?

## Appendix:
Some examples of such as switch:
- the PBR version of the surface constructor node here:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.cpp#L201
- when emitting the final alpha blending of the fragment result here:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/GlslShaderGenerator.cpp#L643
- for "unlit" shader support
https://github.com/AcademySoftwareFoundation/MaterialX/pull/860

kwokcb:
Based on specification discussion, the h/w logic should be updated to follow this heuristic:
```
hint (string): A hint to help code generators understand how input may be used.  The following hints are currently defined:
"transparency": the input is indicating a level of shading transparency.
[...]

PR #860, opened 2022-03-08
An unlit surface shader constructor node.

niklasharrysson:
Implementation of the "unlit" surface shader, as proposed in issue #839.

This change list adds the definitions and implementations for GLSL, OSL and MDL.

Some tests added to the test suite, testing visualization of a simple procedural with/without cutout opacity and with/without colored transmission:
unlit_surface_tests

Note that MDL rendering is not part of the testsuite yet and are here renderer separately with the MDL DXR sdk sample. Hence the difference in environment and geometry. Also note that for GLSL our renderer does not support colored transmission yet.

Signed-off-by: Niklas Harrysson <niklas.harrysson@gmail.com>


source/MaterialXGenGlsl/Nodes/UnlitSurfaceNodeGlsl.cpp
```diff
@@ -0,0 +1,56 @@
...
+            shadergen.emitLine(outTransparency + " = " + shadergen.getUpstreamResult(transmission, context) + " * " + shadergen.getUpstreamResult(transmissionColor, context), stage);
+
+            const ShaderInput* opacity = node.getInput("opacity");
+            const string surfaceOpacity = shadergen.getUpstreamResult(opacity, context);
+            shadergen.emitLine(outColor + " *= " + surfaceOpacity, stage);
+            shadergen.emitLine(outTransparency + " = mix(vec3(1.0), " + outTransparency + ", " + surfaceOpacity + ")", stage);
+        }
+        else
+        {
+            shadergen.emitLine(outTransparency + " = vec3(0.0)", stage);
```
kwokcb, line 50: Shouldn't this be 1.0 (meaning fully opaque) otherwise if blending is enabled this will not show up ?

source/MaterialXGenGlsl/Nodes/UnlitSurfaceNodeGlsl.cpp
```diff
@@ -0,0 +1,56 @@
...
+        
+        if (context.getOptions().hwTransparency)
+        {
+            const ShaderInput* transmission = node.getInput("transmission");
+            const ShaderInput* transmissionColor = node.getInput("transmission_color");
+            shadergen.emitLine(outTransparency + " = " + shadergen.getUpstreamResult(transmission, context) + " * " + shadergen.getUpstreamResult(transmissionColor, context), stage);
+
+            const ShaderInput* opacity = node.getInput("opacity");
+            const string surfaceOpacity = shadergen.getUpstreamResult(opacity, context);
+            shadergen.emitLine(outColor + " *= " + surfaceOpacity, stage);
```
kwokcb, line 45: Checking here that this doesn't quite match the osl implementation in that it does not mix transmission here?

source/MaterialXGenMdl/mdl/materialx/stdlib.mdl
```diff
@@ -32,6 +33,30 @@ import ::anno::*;
...
+
+export material mx_surface_unlit(
+    float mxp_emission = 0.0,
+    color mxp_emission_color = color(1.0, 1.0, 1.0),
+    float mxp_transmission = 0.0,
+    color mxp_transmission_color = color(1.0, 1.0, 1.0),
+    float mxp_opacity = 1.0
+)
+= let {
+    float trans = math::clamp(mxp_transmission, 0.0, 1.0);
```
kwokcb, line 46: Small comment: You seem to perform clamps here and for osl but not for glsl.

source/MaterialXView/Viewer.cpp
```diff
@@ -1412,7 +1412,11 @@ void Viewer::saveShaderSource(mx::GenContext& context)
         mx::TypedElementPtr elem = material ? material->getElement() : nullptr;
         if (elem)
         {
-            mx::ShaderPtr shader = createShader(elem->getNamePath(), context, elem);
+            const bool hasTransparency = mx::isTransparentSurface(elem, context.getShaderGenerator().getTarget());
+            mx::GenContext materialContext = context;
+            materialContext.getOptions().hwTransparency = hasTransparency;
```
kwokcb, line 1417: Leaving note that this is "fragile" and is value dependent on what isTransparentSurface() returns. 
- Currently opacity / transmissions are detected but transmission_color is not -- should this be fixed as part of this  PR ?
- Also recompiles only occur on load / save shader but not for value changes which can
change the result of this. Defer this to a new issue on how to deal with transparency
update and codegen. (#859)

source/MaterialXGenGlsl/Nodes/UnlitSurfaceNodeGlsl.cpp
niklasharrysson: Transparency is the complement of opacity, so 0.0 transparency means fully opaque :)

source/MaterialXGenGlsl/Nodes/UnlitSurfaceNodeGlsl.cpp
niklasharrysson: Yes, my thinking here is that the HW alpha blending would already do that for us. But just let me know if that's not the case. Again this logic is taken from the PBR version of the "surface" node, so if a change is needed it might be best to do as part of a deeper look at all transparency handling, so it's consistent for all surface shaders.

source/MaterialXGenMdl/mdl/materialx/stdlib.mdl
niklasharrysson: Since there is no explicit `1.0 - transmission` in this case I didn't add the clamp. But you are right, to be 100% consistent for values outside [0,1] I could add it here as well. Or I could remove the clamp everywhere, assuming the value is always in [0,1], which is should be by design.

source/MaterialXView/Viewer.cpp
niklasharrysson: Yes, as discussed this should be looked at in another change list, so thanks for adding a separate issue for that.

I don't think it's worth updating the general `isTransparentSurface()` function to detect more specific inputs at this point. It may be better to wait for the larger transparency overhaul. Ignoring the input `transmission_color` makes it less conservative since it's enough to have `transmission > 0` to detect that transmission handling is needed. So there may be false positives but at least we avoid false negatives.

source/MaterialXGenGlsl/Nodes/UnlitSurfaceNodeGlsl.cpp
kwokcb: I see so alpha blending gives you some "cheap" way to fake transmission. Let's add a note for this, but from past implementations of h/w transmission for rasterizers I kind of remember including transmission w/o reliance on alpha blending.

source/MaterialXGenMdl/mdl/materialx/stdlib.mdl
kwokcb: Whatever is the convention is fine. Just happen to notice this small thing.

source/MaterialXView/Viewer.cpp
kwokcb: Yes, `transmission_color` is secondary to `transmission` so agree let's defer this. Thanks.

jstone-lucasfilm (approved):

PR #869, opened 2022-03-14
Fix "show all inputs" option in MaterialXView

kwokcb:
Minor change so that:

- At load time node definition inputs are added to each node.
- This allows shader gen to take these into consideration.
- At UI building time the inputs will exist for usage.

As part of this fix the utility to copy inputs from node definitions to node instances to only include the minimal which is name, type, and value attributes.

The previous logic would do nothing as the inputs don't get passed through code generation and into
the uniform block which is used for iterate over for the UI. (Also the toggle never called the UI to update).
 
 Example: Brass-tiled
- Without "Show All Inputs" 

- With "Show All Inputs"

Example with default std surface input:
```
<?xml version="1.0"?>
  <standard_surface name="SR_default" type="surfaceshader">
  </standard_surface>
  <surfacematerial name="Default" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_default" />
  </surfacematerial>
</materialx>
```
The expansion includes all inputs in `standard_surface_default.mtlx` example, as well as additional inputs which are not specified but part of the definition.
```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709" xmlns:xi="http://www.w3.org/2001/XInclude">
  <standard_surface name="SR_default" type="surfaceshader">
    <input name="base" type="float" value="1.0" />
    <input name="base_color" type="color3" value="0.8, 0.8, 0.8" />
    <input name="diffuse_roughness" type="float" value="0" />
    <input name="metalness" type="float" value="0" />
    <input name="specular" type="float" value="1" />
    <input name="specular_color" type="color3" value="1, 1, 1" />
    <input name="specular_roughness" type="float" value="0.2" />
    <input name="specular_IOR" type="float" value="1.5" />
    <input name="specular_anisotropy" type="float" value="0" />
    <input name="specular_rotation" type="float" value="0" />
    <input name="transmission" type="float" value="0" />
    <input name="transmission_color" type="color3" value="1, 1, 1" />
    <input name="transmission_depth" type="float" value="0" />
    <input name="transmission_scatter" type="color3" value="0, 0, 0" />
    <input name="transmission_scatter_anisotropy" type="float" value="0" />
    <input name="transmission_dispersion" type="float" value="0" />
    <input name="transmission_extra_roughness" type="float" value="0" />
    <input name="subsurface" type="float" value="0" />
    <input name="subsurface_color" type="color3" value="1, 1, 1" />
    <input name="subsurface_radius" type="color3" value="1, 1, 1" />
    <input name="subsurface_scale" type="float" value="1" />
    <input name="subsurface_anisotropy" type="float" value="0" />
    <input name="sheen" type="float" value="0" />
    <input name="sheen_color" type="color3" value="1, 1, 1" />
    <input name="sheen_roughness" type="float" value="0.3" />
    <input name="coat" type="float" value="0" />
    <input name="coat_color" type="color3" value="1, 1, 1" />
    <input name="coat_roughness" type="float" value="0.1" />
    <input name="coat_anisotropy" type="float" value="0.0" />
    <input name="coat_rotation" type="float" value="0.0" />
    <input name="coat_IOR" type="float" value="1.5" />
    <input name="coat_normal" type="vector3" />
    <input name="coat_affect_color" type="float" value="0" />
    <input name="coat_affect_roughness" type="float" value="0" />
    <input name="thin_film_thickness" type="float" value="0" />
    <input name="thin_film_IOR" type="float" value="1.5" />
    <input name="emission" type="float" value="0" />
    <input name="emission_color" type="color3" value="1, 1, 1" />
    <input name="opacity" type="color3" value="1, 1, 1" />
    <input name="thin_walled" type="boolean" value="false" />
    <input name="normal" type="vector3" />
    <input name="tangent" type="vector3" />
  </standard_surface>
  <surfacematerial name="Default" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_default" />
    <input name="displacementshader" type="displacementshader" />
  </surfacematerial>
 </materialx>
```

### DIFF
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -669,15 +669,18 @@ void Node::addInputsFromNodeDef()
     NodeDefPtr nodeNodeDef = getNodeDef();
     if (nodeNodeDef)
     {
-        for (auto nodeDefInput : nodeNodeDef->getActiveValueElements())
+        for (auto nodeDefInput : nodeNodeDef->getActiveInputs())
         {
             const string& name = nodeDefInput->getName();
             InputPtr nodeInput = getInput(name);
             if (!nodeInput)
             {
                 nodeInput = addInput(name, nodeDefInput->getType());
-                nodeInput->setAttribute(ValueElement::VALUE_ATTRIBUTE,
-                                        nodeDefInput->getAttribute(ValueElement::VALUE_ATTRIBUTE));
+                const string& valueString = nodeDefInput->getValueString();
+                if (!valueString.empty())
+                {
+                    nodeInput->setValueString(valueString);
+                }
             }
         }
     }
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -657,10 +657,9 @@ InputPtr Node::addInputFromNodeDef(const string& name)
         if (nodeDefInput)
         {
             nodeInput = addInput(nodeDefInput->getName(), nodeDefInput->getType());
-            const string& valueString = nodeDefInput->getValueString();
-            if (!valueString.empty())
+            if (nodeDefInput->hasValueString())
             {
-                nodeInput->setValueString(valueString);
+                nodeInput->setValueString(nodeDefInput->getValueString());
             }
         }
     }
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -649,6 +649,12 @@ void NodeGraph::setNodeDef(ConstNodeDefPtr nodeDef)
 
 InputPtr Node::addInputFromNodeDef(const string& name)
 {
+    StringVec nonInstanceAttributes = { DOC_ATTRIBUTE, ValueElement::INTERFACE_NAME_ATTRIBUTE,
+                                        ValueElement::ENUM_ATTRIBUTE, ValueElement::ENUM_VALUES_ATTRIBUTE, 
+                                        ValueElement::UI_NAME_ATTRIBUTE, ValueElement::UI_FOLDER_ATTRIBUTE, ValueElement::UI_MIN_ATTRIBUTE,
+                                        ValueElement::UI_MAX_ATTRIBUTE, ValueElement::UI_SOFT_MIN_ATTRIBUTE, ValueElement::UI_SOFT_MAX_ATTRIBUTE,
+                                        ValueElement::UI_STEP_ATTRIBUTE, ValueElement::UI_ADVANCED_ATTRIBUTE, ValueElement::UNIFORM_ATTRIBUTE };
+
     InputPtr nodeInput = getInput(name);
     if (!nodeInput)
     {
@@ -658,9 +664,26 @@ InputPtr Node::addInputFromNodeDef(const string& name)
         {
             nodeInput = addInput(nodeDefInput->getName());
             nodeInput->copyContentFrom(nodeDefInput);
+            for (auto nonInstanceAttribute : nonInstanceAttributes)
+            {
+                nodeInput->removeAttribute(nonInstanceAttribute);
+            }
         }
     }
     return nodeInput;
+}
+
+void Node::addInputsFromNodeDef()
+{
+    NodeDefPtr nodeNodeDef = getNodeDef();
+    if (nodeNodeDef)
+    {
+        for (auto nodeDefValueElem : nodeNodeDef->getActiveValueElements())
+        {
+            const std::string& valueElemName = nodeDefValueElem->getName();
+            addInputFromNodeDef(valueElemName);
+        }
+    }
 }
 
 void NodeGraph::addInterfaceName(const string& inputPath, const string& interfaceName)
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -657,8 +657,11 @@ InputPtr Node::addInputFromNodeDef(const string& name)
         if (nodeDefInput)
         {
             nodeInput = addInput(nodeDefInput->getName(), nodeDefInput->getType());
-            nodeInput->setAttribute(ValueElement::VALUE_ATTRIBUTE, 
-                                    nodeDefInput->getAttribute(ValueElement::VALUE_ATTRIBUTE));
+            const string& valueString = nodeDefInput->getValueString();
+            if (!valueString.empty())
+            {
+                nodeInput->setValueString(valueString);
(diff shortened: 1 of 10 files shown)


source/MaterialXCore/Node.cpp
```diff
@@ -658,11 +664,28 @@ InputPtr Node::addInputFromNodeDef(const string& name)
         {
             nodeInput = addInput(nodeDefInput->getName());
             nodeInput->copyContentFrom(nodeDefInput);
+            for (auto nonInstanceAttribute : nonInstanceAttributes)
```
jstone-lucasfilm, line 667: It may be more straightforward here to provide a list of attributes to be _copied_ to instances, rather than copying all attributes and then removing the invalid ones afterwards.  This would be more robust as we add additional UI attributes in the future, not to mention situations where applications store custom attributes that we cannot know about.  What are your thoughts?

source/MaterialXCore/Node.h
```diff
@@ -160,6 +160,10 @@ class MX_CORE_API Node : public InterfaceElement
     /// If the input already exists on the node it will just be returned.
     InputPtr addInputFromNodeDef(const string& name);
 
+    /// Add inputs based on the corresponding associated node definition.
+    /// If the inputs already exists on the node it will just be returned.
```
jstone-lucasfilm, line 164: It looks like this second sentence isn't needed in the comment, as it only really applies to the original addInputFromNodeDef method.

source/MaterialXCore/Node.cpp
```diff
@@ -658,11 +664,28 @@ InputPtr Node::addInputFromNodeDef(const string& name)
...
 }
 
+void Node::addInputsFromNodeDef()
+{
+    NodeDefPtr nodeNodeDef = getNodeDef();
+    if (nodeNodeDef)
+    {
+        for (auto nodeDefValueElem : nodeNodeDef->getActiveValueElements())
+        {
+            const std::string& valueElemName = nodeDefValueElem->getName();
```
jstone-lucasfilm, line 683: Minor: This type can be written as "string" in this context.

jstone-lucasfilm (commented):
This looks like a good change, thanks @kwokcb, and I've written up some minor review notes.

source/MaterialXCore/Node.cpp
kwokcb: I don't particularly like adding and then removing after the fact so the list of attributes to copy definitely is better -- and it could possibly make it so the cache is not dirtied as much.

For the "minimal" attributes that should be copied over from a nodedef to a node, I suggest:
- name
- type
- value

source/MaterialXCore/Node.cpp
```diff
@@ -647,22 +647,38 @@ void NodeGraph::setNodeDef(ConstNodeDefPtr nodeDef)
...
     if (!nodeInput)
     {
-        NodeDefPtr nodeDef = getNodeDef();
+        NodeDefPtr nodeDef = fromNodeDef ? fromNodeDef : getNodeDef();
         InputPtr nodeDefInput = nodeDef ? nodeDef->getActiveInput(name) : nullptr;
         if (nodeDefInput)
         {
             nodeInput = addInput(nodeDefInput->getName());
-            nodeInput->copyContentFrom(nodeDefInput);
+            nodeInput->copyContentFrom(nodeDefInput, valueAttributes);
```
jstone-lucasfilm, line 663: In this situation, perhaps we don't even need these changes to the API of Element::copyContentFrom, and we could instead copy these two attributes in the most direct and efficient way:

```
const string& nodeDefType = nodeDefInput->getType();
const string& nodeDefValueString = nodeDefInput->getValueString();
nodeInput->setType(nodeDefType);
nodeInput->setValueString(nodeDefValueString);
```

This would avoid the additional overhead of checking for children of the input element, and would avoid adding new checks for existing users of Element::copyContentFrom.

source/MaterialXCore/Node.cpp
```diff
@@ -647,22 +647,38 @@ void NodeGraph::setNodeDef(ConstNodeDefPtr nodeDef)
     }
 }
 
-InputPtr Node::addInputFromNodeDef(const string& name)
+InputPtr Node::addInputFromNodeDef(const string& name, NodeDefPtr fromNodeDef)
```
jstone-lucasfilm, line 650: I see where this additional input could be used as an optimization, allowing many calls to addInputFromNodeDef to share a single getNodeDef query, but it does add complexity to common use cases, and the alternative implementation proposed below should be even faster.

jstone-lucasfilm (commented):
This changelist is starting to come together, and I have two additional recommendations I've written up below.

source/MaterialXCore/Node.cpp
```diff
@@ -656,13 +656,33 @@ InputPtr Node::addInputFromNodeDef(const string& name)
...
+    NodeDefPtr nodeNodeDef = getNodeDef();
+    if (nodeNodeDef)
+    {
+        for (auto nodeDefInput : nodeNodeDef->getActiveValueElements())
+        {
+            const string& name = nodeDefInput->getName();
+            InputPtr nodeInput = getInput(name);
+            if (!nodeInput)
+            {
+                nodeInput = addInput(name, nodeDefInput->getType());
```
kwokcb, line 678: Believe this is as compact as I can get this to be now :).
All the extra parameters are avoided now.

jstone-lucasfilm: This looks great, thanks Bernard!

jstone-lucasfilm:
This change looks very promising!  But when I test the viewer locally, I notice that materials aren't visually identical in the viewport when `Show All Inputs` is enabled.

As an example, try loading `standard_surface_plastic.mtlx` with this option both disabled and enabled.  In addition to the visual differences when `Show All Inputs` is enabled, it doesn't look like our Property Editor UI is correctly grouping shader inputs into visual categories when this option is set.

kwokcb:
> This change looks very promising! But when I test the viewer locally, I notice that materials aren't visually identical in the viewport when `Show All Inputs` is enabled.
> 
> As an example, try loading `standard_surface_plastic.mtlx` with this option both disabled and enabled. In addition to the visual differences when `Show All Inputs` is enabled, it doesn't look like our Property Editor UI is correctly grouping shader inputs into visual categories when this option is set.

This was a mistype on my part. Only inputs should be copied over. It was copying over outputs as well which causes a signature mismatch.

jstone-lucasfilm (approved):

PR #870, opened 2022-03-14
Add glTF BoomBox example

kwokcb:
- Add in BoomBox example from Ed Mackey. Minor name change to shader and material so it's easier to read results. Added occlusion connection based on [preliminary review](https://github.com/kwokcb/MaterialX/pull/14) 
- Add in `GltfPbr` directory for generation tests and render tests.

- As part of this fixed OSL tests to work if any new test file uses relative resource pathing by pre-processing these paths before hand. 
   - The new boombox tests uses resources which are both relative to the main gltf file and are thus co-located in the same folder, while at the same time accessible from the materialX file via relative pathing.
  - Remove hard-coded patch code to handle existing cases.

## Default test result. GLSL and OSL
image

## Boombox test result. GLSL and OSL
image

## Boombox viewer result. GLSL
image

### Test Run 
[glTF_test_run_Mar_15_2022.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/8256290/glTF_test_run_Mar_15_2022.pdf)


resources/Geometry/BoomBox.gltf
```diff
@@ -0,0 +1,172 @@
...
+  ],
+  "buffers": [
+    {
+      "uri": "BoomBox.bin",
+      "byteLength": 207816
+    }
+  ],
+  "images": [
+    {
+      "uri": "BoomBox_baseColor.png"
```
emackey, line 82: One minor concern I have is that this `*.gltf` file is not valid as presented here, because the images associated with the URIs here are all in the `Images` folder.  I realize the whole structure of this project keeps images away from geometry, but that's not going to allow most glTF loaders to parse this as-is.  Relative paths are allowed, although paths reaching into parent folders are strongly discouraged, as some implementations reject them for security.

kwokcb: I've moved the gltf file and images to be co-located so the references should be okay now.
This should hopefully also make it so that there is no additional packing step for resources for usage by the MaterialXView web viewer.
-----
As part of this @jstone-lucasfilm , I put in new logic for resolving relative pathing for OSL render tests. I removed the current patch logic as such. As part of this I wrote a simple utility to remove ".." qualifiers embedded in `FilePath`s which testrender and the MDL dxr test renderer don't handle.

jstone-lucasfilm (commented):
For this initial boombox example, I would recommend that we include just the GLTF geometry and MaterialX material, rather than a GLTF file containing both geometry and material data.  This should help to clarify how the example is meant to be tested, and will allow us to preserve our existing folder structure for examples, with the individual components separated into Geometry, Materials, and Images.

Going forward, I'm definitely interested in developing code that supports transformations between glTF and MaterialX materials, though this might be best handled in a higher-level codebase that includes MaterialX as a dependency, rather than in the MaterialX codebase itself.

kwokcb:
> For this initial boombox example, I would recommend that we include just the GLTF geometry and MaterialX material, rather than a GLTF file containing both geometry and material data. This should help to clarify how the example is meant to be tested, and will allow us to preserve our existing folder structure for examples, with the individual components separated into Geometry, Materials, and Images.
> 
> Going forward, I'm definitely interested in developing code that supports transformations between glTF and MaterialX materials, though this might be best handled in a higher-level codebase that includes MaterialX as a dependency, rather than in the MaterialX codebase itself.

The material, image, and geometry information are now separated with the material and image references removed from the glTF file. 

The transformation is not part of this and as discussed a wider audience including @emackey  et al should be included as to where to place / host this.

resources/Geometry/boombox.gltf
```diff
@@ -0,0 +1,116 @@
+{
```
kwokcb, line 1: Material and image free version of the boombox.

jstone-lucasfilm (commented):
This change looks great, thanks @bernardkwok!

What is the difference between the GLB format used by the original shader ball and the GLTF/BIN format pair used for the boombox?  Does the GLTF/BIN pair have advantages for this model, or would it make more sense to harmonize on GLB for glTF geometry files in MaterialX?

resources/Geometry/boombox.gltf
```diff
@@ -0,0 +1,116 @@
...
+    {
+      "primitives": [
+        {
+          "attributes": {
+            "TEXCOORD_0": 0,
+            "NORMAL": 1,
+            "TANGENT": 2,
+            "POSITION": 3
+          },
+          "indices": 4,
```
emackey, line 90: Invalid JSON, trailing comma here.

emackey:
> What is the difference between the GLB format used by the original shader ball and the GLTF/BIN format pair

@jstone-lucasfilm In general, the Binary form (GLB) is expected to be self-contained, for easier sharing and/or relocating the entire asset as a single file.  In contrast, the text/JSON form (*.gltf + *.bin + textures) requires multiple files to form a complete asset, and the root *.gltf file is expected to contain valid relative paths to all referenced files.

I can't say which is best for this project, but some things to consider are:

* The *.gltf (text/JSON) version is easier for developers to inspect and even hand-edit (only ever by using [VSCode](https://code.visualstudio.com/) with my glTF-tools extension installed of course).
* The text form also allows the material's images/textures to be stand-alone files, enabling easy inspection and editing, although they typically should be in the same folder or a child folder of the JSON file.
* The text form can be confusing for folks who copy/relocate/share it without understanding its dependencies on other files.

emackey:
> the original shader ball

On a different topic, I noticed that `shaderball.glb` has an image nearby called `shaderball_ao.png` that is neither referenced nor embedded in the GLB file as its occlusion map.  It might be nice to have that GLB as a *.gltf file that calls out its occlusion map explicitly.

The boombox has one of these occlusion images too, in the red channel (by glTF spec) of `BoomBox_occlusionRoughnessMetallic.png`, which is no longer referenced now that all material information has been removed from the corresponding `boombox.gltf` file.

I'm not sure what the correct answer is here, because one cannot reference these without including a material definition in the glTF file (which shaderball does, but boombox no longer does).

kwokcb:
I am curious here what is a "minimum" valid gltf file. If it always must have geometry and a material. 
If so then it guess a dummy material can be inserted but this means it is still a geometry + material, albeit the 
"incorrect" material. 

I guess what I'm wondering here that the split to have material + geometry separate is too 
an unnatural one. The other aspect that is hard to consolidate is material assignment which I think is not possible or
very awkward if artificially kept separate.

The workflow premise is thus: a gltf/glb may contain geometry+materials+animation etc an extractors can be written
to pull out pieces of it for editing but in the end you always want to pack to back in together again?

emackey:
> I am curious here what is a "minimum" valid gltf file.

This answer may be far more literal than you had in mind, but the minimum valid glTF file is:

```
{ "asset": {"version": "2.0"} }
```

For anything more complex, we have the [Khronos glTF Validator](https://github.khronos.org/glTF-Validator/).

> a dummy material can be inserted but this means it is still a geometry + material, albeit the
> "incorrect" material.

When there's no material, the implementation should use the defined default values for each parameter.  Those values include all `1.0` settings for base color, metallic, and roughness.  These defaults were chosen for good technical reasons, but certainly not aesthetic ones, and the resulting material is an unnaturally rough metal surface that reflects its environment rather strangely.  So the shader ball sets `metallicFactor` to zero, and supplies a gray base color to make things look a bit more neutral.  It could also reference the AO map in isolation without bringing a metallic/roughness texture into the material.

> I guess what I'm wondering here that if the split to have material + geometry separated is too
> an unnatural one.

Well now we're touching on some fun topics here.  Take a second look at the BoomBox: It has a metal antenna, a shiny plastic body, an emissive glowing play button, different plastics on the base... but how many materials, in a glTF or MaterialX sense?  Only one material.  Rasterizers don't want separate draw calls for all those parts just due to the antenna being made of metal or something like that.  The whole BoomBox can be drawn with a single draw call, with simply a texture identifying what's made of metal and what's plastic.  So a "material" is not really a physical material; it's a shader program configuration.  It needs to be different when the alpha blending mode is different or some other thing like that, not when the physical object is made of different stuff.

So the "BoomBox material" is not re-usable on any other geometry besides the BoomBox itself.  You can place it on the shader ball, and the whole shader ball looks like a distorted BoomBox.  This is not a "material" that a user would want in a library of materials.  It's single-use only.  And I think you'll find this holds true for many models that have been prepared for real-time rendering: They bake out all the different Substance-Painter-Preset-type materials into a smaller set (or single instance) of texture stacks and shader program configurations, that get labeled as glTF "materials" but are not reusable with any other UV atlas or geometry.

> The other aspect that is hard to consolidate is material assignment which I think is not possible or
> very awkward if artificially kept separate.

Exactly.  In well-optimized real-time models, you won't be able to swap the metal BoomBox antenna for a plastic one by changing the material assignments around.  There may be a half-dozen separate "materials", but each one might only work for its own particular set of geometry, with its own single-use UV atlas.

> The workflow premise is thus: a gltf/glb may contain geometry+materials+animation etc and extractors can be written
> to pull out pieces of it for editing but in the end you always want to pack to back in together again?

I don't think the intention will always be to reassemble it.  There may be a lot of uses for one-way conversion, either bringing an asset found online into the MaterialX system, or using MaterialX to craft an asset that gets baked out to glTF for distribution.

The latter might present an interesting bake challenge, for example if the user had assigned different physical materials from a library of different MaterialX materials to the BoomBox, but the geometry all fits onto a single UV atlas with a single blend mode and such settings, then could it "bake" multiple MaterialX materials into a single glTF material for that atlas?  Such techniques can yield a performance boost in the land of rasterizers, by reducing the number of separate draw calls to get the mesh on screen.

jstone-lucasfilm:
@emackey Great, thanks for those additional details!

@kwokcb For now, I would recommend aligning our geometry storage with the existing shaderball.glb, and use a single, geometry-only GLB file for the new boombox asset.  In the future, as we consider including both geometric and material content in other formats, we can revisit these choices and refine the organization of the resources folder.

kwokcb:
> @emackey Great, thanks for those additional details!
> 
> @kwokcb For now, I would recommend aligning our geometry storage with the existing shaderball.glb, and use a single, geometry-only GLB file for the new boombox asset. In the future, as we consider including both geometric and material content in other formats, we can revisit these choices and refine the organization of the resources folder.

Hi @jstone-lucasfilm, @emackey 
I've fixed the gltf syntax error and converted the the geometry to a single glb.
I've also re-localized images back with the material as they where before to keep them localized with the mtlx material file.

jstone-lucasfilm (approved):
Looks good, thanks @bernardkwok and @emackey!

PR #883, opened 2022-03-22
Improvements to testing and TextureBaker

ashwinbhat:
Testing updates:
- Create data model to define wedge and baking testing parameters in options.mtlx
- Add test case for material nodegraph translation 

TextureBaker updates:
- Add a uv min/max settings and update GlslRenderer::renderTextureSpace to set the uv range for the baking quad.
- Enable hashing for baked image names to avoid long path issues.


source/MaterialXTest/MaterialXRender/RenderUtil.h
```diff
@@ -128,6 +128,9 @@ class ShaderRenderTester
     // Code generation methods
     //
 
+    // Register any additional source code paths used by the generator
+    virtual void registerSourceCodeSearchPaths(mx::GenContext& /*context*/) {};
```
jstone-lucasfilm, line 132: I believe this method is now unused in MaterialXRender, and we can omit this from your changelist.  Let me know if there are subtle cases I'm missing, though.

source/MaterialXRenderGlsl/TextureBaker.cpp
```diff
@@ -475,7 +486,14 @@ DocumentPtr TextureBaker::bakeMaterialToDoc(DocumentPtr doc, const FileSearchPat
...
     DefaultColorManagementSystemPtr cms = DefaultColorManagementSystem::create(genContext.getShaderGenerator().getTarget());
     cms->loadLibrary(doc);
-    genContext.registerSourceCodeSearchPath(searchPath);
+    if (!_codeSearchPath.isEmpty())
+    {
+        genContext.registerSourceCodeSearchPath(_codeSearchPath);
+    }
+    for (const FilePath& path : searchPath)
+    {
+        genContext.registerSourceCodeSearchPath(path / "libraries");
```
jstone-lucasfilm, line 495: This looks like a reference to the older pattern where the string "libraries" was manually added to each location in the code search path.  This should no longer be needed in the latest build, and I would recommend double-checking that this logic is working as intended.

source/MaterialXRenderGlsl/TextureBaker.h
```diff
@@ -118,6 +118,12 @@ class MX_RENDERGLSL_API TextureBaker : public GlslRenderer
         return _outputImagePath;
     }
 
+    /// Set the "libraries" search path location.
```
jstone-lucasfilm, line 121: As above, the string "libraries" is no longer required in code search paths, so I would recommend double-checking that this is working as expected.

jstone-lucasfilm (commented):
There are some good improvements here, but I would recommend double-checking the changes involving code search paths, since these seem to be following an older pattern that is no longer supported.

ashwinbhat:
> There are some good improvements here, but I would recommend double-checking the changes involving code search paths, since these seem to be following an older pattern that is no longer supported.

The libraries logic seems ok. I need to confirm a few deployments on MacOS where libraries are nested into application bundle or integrated with USD. i.e. they are not relative to current path. I'll get back to you with my findings.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -165,6 +165,47 @@ TEST_CASE("GenShader: Graph + Nodedf Translation Check", "[genshader]")
     mx::writeToXmlFile(doc, "transparency_test_nodedefs.mtlx");
 }
 
+TEST_CASE("GenShader: Shader Translation", "[translate]")
```
kwokcb, line 168: Isn't there already a translation test so this would be redundant ?

source/MaterialXTest/MaterialXRender/RenderUtil.cpp
```diff
@@ -176,8 +176,9 @@ bool ShaderRenderTester::validate(const mx::FilePath optionsFilePath)
     _shaderGenerator->getUnitSystem()->setUnitConverterRegistry(registry);
 
     mx::GenContext context(_shaderGenerator);
-    context.registerSourceCodeSearchPath(currentPath);
+    context.registerSourceCodeSearchPath(searchPath);
```
kwokcb, line 179: What is the correct logic here? current path or search path.
Should current path for osl also be updated ?

source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
```diff
@@ -630,21 +630,23 @@ bool GlslShaderRenderTester::runRenderer(const std::string& shaderName,
...
-                                      unsigned int bakeWidth, unsigned int bakeHeight, bool bakeHdr, std::ostream& log)
+void GlslShaderRenderTester::runBake(mx::DocumentPtr doc, const mx::FileSearchPath& imageSearchPath, const mx::FileSearchPath& codeSearchPath, 
+                                     const mx::FilePath& outputFileName, const GenShaderUtil::TestSuiteOptions::BakeSetting& bakeOptions,
+                                     std::ostream& log)
 {
-    mx::ImageVec imageVec = _renderer->getImageHandler()->getReferencedImages(doc);
-    auto maxImageSize = mx::getMaxDimensions(imageVec);
-    bakeWidth = std::max(bakeWidth, maxImageSize.first);
-    bakeHeight = std::max(bakeHeight, maxImageSize.second);
+    const unsigned int bakeWidth = std::max(bakeOptions.resolution, (unsigned int) 2);
```
kwokcb, line 637: Is this correct. Shouldn't this compare max bakeoptions vs max dimensions on the set of images instead of 2 ?

source/MaterialXRenderGlsl/GLTextureHandler.cpp
```diff
@@ -56,21 +56,23 @@ bool GLTextureHandler::bindImage(ImagePtr image, const ImageSamplingProperties&
     }      
     _boundTextureLocations[textureUnit] = image->getResourceId();
 
+    GLenum targetType = image->getHeight() > 1 ? GL_TEXTURE_2D : GL_TEXTURE_1D;
```
kwokcb, line 59: I assume these target type changes are pulled from the CM code changes
I made ? I don't think they are necessary for any of the test changes here.
It may not be good to add this in as it's only for CM 1D LUTs I added this in.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
ashwinbhat: I couldn't find anything that tested translation API so added this. Can you point me to an existing test, perhaps I can update it to add usdpreviewsurface translation.

source/MaterialXTest/MaterialXRender/RenderUtil.cpp
ashwinbhat: Since this is for testing, currentPath is good.

source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
ashwinbhat: Yes, corrected it.

source/MaterialXRenderGlsl/GLTextureHandler.cpp
ashwinbhat: Yes these were part of CM changes, but I thought it would be a good change to add to the texture handler to support 1D textures. I've reverted it. I'll be removing CM changes from contrib so wanted to align the main with contrib.

ashwinbhat:
> > There are some good improvements here, but I would recommend double-checking the changes involving code search paths, since these seem to be following an older pattern that is no longer supported.
> 
> The libraries logic seems ok. I need to confirm a few deployments on MacOS where libraries are nested into application bundle or integrated with USD. i.e. they are not relative to current path. I'll get back to you with my findings.

I have done some cursory tests on MacOS, I think current code search  path is good. We need to ensure that deployments place the libraries in correct location, relative to the host code.

jstone-lucasfilm (approved):

PR #892, opened 2022-03-29
Improvements to support locale, units and additional tests.

ashwinbhat:
- Add utilities
   - UnitConverterRegistry::convertToUnit
   - hasNamespace
   - joinStrings
- Use C Locale for string stream operations. (for https://github.com/autodesk-forks/MaterialX/issues/1183)
- Add tests for unit and locale content.

### DIFF
--- a/resources/Materials/TestSuite/locale/numericformat.mtlx
+++ b/resources/Materials/TestSuite/locale/numericformat.mtlx
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<materialx version="1.38">
+  <standard_surface name="SR_numberformats" type="surfaceshader">
+    <input name="base_color" type="color3" value="0.18,0.18, 0.18"/>
+    <input name="specular_color" type="color3" value="0.05, 0.05,0.05" />
+    <input name="normal" type="vector3" value="1,0.5,1" />
+    <input name="specular_roughness" type="float" value="0.25" />
+    <input name="coat" type="float" value="1" />
+  </standard_surface>
+  <surfacematerial name="Number_formats" type="material">
+    <input name="surfaceshader" type="surfaceshader" nodename="SR_numberformats" />
+  </surfacematerial>
+</materialx>
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -169,20 +169,11 @@ void Document::initialize()
 }
 
 NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const string& nodeDefName, const string& node,
-                                         const string& version, bool isDefaultVersion, const string& group, string& newGraphName, const string& namespaceString)
+                                         const string& version, bool isDefaultVersion, const string& group, string& newGraphName)
 {
-    if (!nodeGraph || nodeDefName.empty() || node.empty())
+    if (getNodeDef(nodeDefName))
     {
-        throw Exception("Invalid input arguments specified for nodedef creation");
-    }
-
-    // Always store nodedef name as fully qualified name
-    bool isNameSpaced = !namespaceString.empty();
-    string qualifiedNodeDefName = isNameSpaced ? (namespaceString + NAME_PREFIX_SEPARATOR + nodeDefName) : nodeDefName;
-    
-    if (getNodeDef(qualifiedNodeDefName))
-    {
-        throw Exception("Cannot create duplicate nodedef: " + qualifiedNodeDefName);
+        throw Exception("Cannot create duplicate nodedef: " + nodeDefName);
     }
 
     NodeGraphPtr graph = nodeGraph;
@@ -195,9 +186,9 @@ NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const str
         graph = addNodeGraph(newGraphName);
         graph->copyContentFrom(nodeGraph);
     }
-    graph->setNodeDefString(qualifiedNodeDefName);
+    graph->setNodeDefString(nodeDefName);
 
-    NodeDefPtr nodeDef = addChild<NodeDef>(qualifiedNodeDefName);
+    NodeDefPtr nodeDef = addChild<NodeDef>(nodeDefName);
     nodeDef->setNodeString(node);
     if (!group.empty())
     {
@@ -213,13 +204,6 @@ NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const str
         {
             nodeDef->setDefaultVersion(true);
         }
-    }
-
-    // Set the namespace for the functional graph and nodedef
-    if (isNameSpaced)
-    {
-        graph->setNamespace(namespaceString);
-        nodeDef->setNamespace(namespaceString);
     }
 
     for (auto output : graph->getOutputs())
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -169,11 +169,20 @@ void Document::initialize()
 }
 
 NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const string& nodeDefName, const string& node,
-                                         const string& version, bool isDefaultVersion, const string& group, string& newGraphName)
+                                         const string& version, bool isDefaultVersion, const string& group, string& newGraphName, const string& namespaceString)
 {
-    if (getNodeDef(nodeDefName))
+    if (!nodeGraph || nodeDefName.empty() || node.empty())
     {
-        throw Exception("Cannot create duplicate nodedef: " + nodeDefName);
+        throw Exception("Invalid input arguments specified for nodedef creation");
+    }
+
+    // Always store nodedef name as fully qualified name
+    bool isNameSpaced = !namespaceString.empty();
+    string qualifiedNodeDefName = isNameSpaced ? (namespaceString + NAME_PREFIX_SEPARATOR + nodeDefName) : nodeDefName;
+    
+    if (getNodeDef(qualifiedNodeDefName))
+    {
+        throw Exception("Cannot create duplicate nodedef: " + qualifiedNodeDefName);
     }
 
     NodeGraphPtr graph = nodeGraph;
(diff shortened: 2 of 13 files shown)


resources/Materials/TestSuite/locale/numericformat.mtlx
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<materialx version="1.38" cms="ocio" colorspace="lin_rec709">
+  <nodedef name="ND_simple_srf_surface" node="simple_srf">
```
jstone-lucasfilm, line 3: We've recently removed our legacy material examples referencing "simple_srf", in favor of renderable examples that reference real shading models such as Standard Surface and glTF PBR.  Perhaps we should update these new examples to use the latest approach as well?

source/MaterialXCore/Node.cpp
```diff
@@ -140,15 +140,15 @@ Edge Node::getUpstreamEdge(size_t index) const
 
 OutputPtr Node::getNodeDefOutput(ElementPtr connectingElement)
 {
-    string outputName;
+    const string* outputName = nullptr;
```
jstone-lucasfilm, line 143: This change looks like an unintended revert of recent improvements to Node::getNodeDefOutput, so I'd recommend taking a closer look to see which of these changes are truly intended.

source/MaterialXFormat/XmlExport.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#include <MaterialXFormat/Util.h>
+
+MATERIALX_NAMESPACE_BEGIN
+
+//
+// XmlExportOptions methods
+//
+
+XmlExportOptions::XmlExportOptions() :
```
jstone-lucasfilm, line 16: What is the semantic difference between XmlExport and XmlWrite?  Could this instead be implemented as an extension of the XmlWrite system, rather than a separate parallel system?

jstone-lucasfilm (commented):
This looks promising, and I've written up a few initial thoughts below.

resources/Materials/TestSuite/locale/numericformat.mtlx
ashwinbhat: I'll update the tests to use a material node.

source/MaterialXFormat/XmlExport.cpp
ashwinbhat: There are a few additional options, I need to consolidate with RTMaterialX. I'll remove this for now and reintroduce in a separate PR

source/MaterialXCore/Document.cpp
```diff
@@ -170,12 +170,21 @@ void Document::initialize()
     setVersionIntegers(MATERIALX_MAJOR_VERSION, MATERIALX_MINOR_VERSION);
 }
 
-NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const string& nodeDefName, const string& node,
-                                         const string& version, bool isDefaultVersion, const string& group, string& newGraphName)
+NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const string& nodeDefName, const string& category,
```
kwokcb, line 173: @ashwinbhat, I'm don't this code is correct in general as it was added in for a specific Autodesk workflow.

source/MaterialXCore/Unit.cpp
```diff
@@ -243,4 +243,70 @@ void UnitConverterRegistry::write(DocumentPtr doc) const
     }
 }
 
+bool UnitConverterRegistry::convertToUnit(DocumentPtr doc, const string& unitType, const string& targetUnit)
```
kwokcb, line 246: This was a hack I put in for Arnold rendering to pre-convert anything with a unit in C++. I don't think this should go back as it only works in some cases.

source/MaterialXCore/Util.cpp
```diff
@@ -55,6 +57,11 @@ bool isValidName(const string& name)
     return it == name.end();
 }
 
+bool isValidNamespace(const string& name)
```
kwokcb, line 60: Not sure why this is being added back. Seems like a strange name for a method. This is to find if it's valid and **not** namespaced ?

source/MaterialXCore/Util.h
```diff
@@ -28,13 +28,20 @@ MX_CORE_API string createValidName(string name, char replaceChar = '_');
...
+
 /// Increment the numeric suffix of a name
 MX_CORE_API string incrementName(const string& name);
 
 /// Split a string into a vector of substrings using the given set of
 /// separator characters.
 MX_CORE_API StringVec splitString(const string& str, const string& sep);
 
+/// Merge a string vector into a single string with each element
+/// seprated by the given separator
```
kwokcb, line 42: Typo.

source/MaterialXRender/Util.cpp
```diff
@@ -239,7 +239,7 @@ unsigned int getUIProperties(InputPtr input, const string& target, UIProperties&
 }
 
 void createUIPropertyGroups(DocumentPtr doc, const VariableBlock& block, UIPropertyGroup& groups,
-                            UIPropertyGroup& unnamedGroups, const string& pathSeparator)
+                            UIPropertyGroup& unnamedGroups, const string& pathSeparator, bool showAllInputs)
```
kwokcb, line 242: This should not be added in as I removed this when fixing up this code :).

source/MaterialXTest/MaterialXCore/Unit.cpp
```diff
@@ -159,33 +161,53 @@ TEST_CASE("UnitDocument", "[unit]")
...
+                                results[input->getNamePath()] = mx::Value::createValue<mx::Vector4>(convertedValue)->getValueString();
                                 REQUIRE(originalval == reconvert);
                             }
                         }
                     }
                 }
             }
         }
+
+        if (!results.empty())
```
kwokcb, line 197: As I don't think this method should be added, this test does not need to change naturally...

kwokcb (changes requested):

source/MaterialXFormat/Util.cpp
```diff
@@ -200,4 +209,12 @@ void flattenFilenames(DocumentPtr doc, const FileSearchPath& searchPath, StringR
     }
 }
 
+
+bool isValidPath(const string& path)
```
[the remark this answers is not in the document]
ashwinbhat: Seems unused, will remove it from PR and fork.

source/MaterialXCore/Util.cpp
ashwinbhat: Change seems to have originated from this PR: https://github.com/autodesk-forks/MaterialX/pull/1191

source/MaterialXRender/Util.cpp
ashwinbhat: Will revert.

source/MaterialXCore/Document.cpp
ashwinbhat: @kwokcb I don't follow, the change here is to add a namespace? How is this Autodesk specific workflow.

source/MaterialXCore/Util.cpp
```diff
@@ -91,6 +98,16 @@ StringVec splitString(const string& str, const string& sep)
     return split;
 }
 
+string mergeStringVec(const StringVec& stringVec, const string& sep)
```
jstone-lucasfilm, line 101: This is a good addition to the string library, though I would recommend the name `joinStrings`, in the spirit of the `String.join` method in Python.

source/MaterialXCore/Value.cpp
```diff
@@ -94,7 +95,7 @@ template <class T> void stringToData(const string& str, enable_if_std_vector_t<T
 template <class T> void dataToString(const T& data, string& str)
 {
     std::stringstream ss;
-
+    ss.imbue(std::locale::classic());
```
jstone-lucasfilm, line 98: Make sure to preserve the newline here, after `ss.imbue` and before the code block that handles float formatting and precision.

source/MaterialXFormat/Util.cpp
```diff
@@ -139,7 +139,7 @@ StringSet loadLibraries(const FilePathVec& libraryFolders,
     return loadedLibraries;
 }
 
-void flattenFilenames(DocumentPtr doc, const FileSearchPath& searchPath, StringResolverPtr customResolver)
+void flattenFilenames(DocumentPtr doc, const FileSearchPath& searchPath, StringResolverPtr customResolver, const FilePathPredicate& skipFlattening)
```
jstone-lucasfilm, line 142: Two thoughts on this new argument to `flattenFilenames`:
- I don't yet see any usage of this argument within the MaterialX codebase, so I would double-check that it's truly needed.
- If the change is still needed, I would recommend `pathPredicate` as a clearer argument name.

jstone-lucasfilm: If you end up naming the argument `pathPredicate`, then it would also make sense to skip paths for which the predicate returns `false`, rather than skipping paths for which it returns `true`.  That would follow the usage of other predicate functions in the MaterialX codebase.

source/MaterialXCore/Document.cpp
```diff
@@ -169,11 +169,20 @@ void Document::initialize()
 }
 
 NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const string& nodeDefName, const string& node,
```
jstone-lucasfilm, line 171: This method doesn't seem to be used within the MaterialX codebase (other than in unit tests), and its purpose seems a bit ambiguous to me.  Perhaps this should instead be broken down into multiple simpler methods, rather than adding another option to its argument list?

source/MaterialXCore/Util.cpp
```diff
@@ -6,6 +6,8 @@
 #include <MaterialXCore/Types.h>
 
 #include <cctype>
+#include <sstream>
```
jstone-lucasfilm, line 9: Are these new includes truly needed here?  I don't see any new usage of std::stringstream, for example, which would require the sstream header to be present.

jstone-lucasfilm (commented):
This changelist is definitely starting to come together, though I had some questions below about Document::addNodeDefFromGraph and its role in MaterialXCore.

source/MaterialXCore/Document.cpp
ashwinbhat: I'll revert this for now and revisit. This change allowed defining a namespace while add nodes. I don't think this commonly used but the Autodesk nodes have namespaces. https://github.com/autodesk-forks/MaterialX/blob/adsk_contrib/dev/libraries/adsk/adsklib/adsklib_defs.mtlx#L106

source/MaterialXCore/Util.cpp
ashwinbhat: Not required will update

jstone-lucasfilm (approved):

ashwinbhat:
> This looks good to me, thanks @ashwinbhat!

Thanks @jstone-lucasfilm for fixing it up. Appreciate it!

PR #894, opened 2022-03-30
Additional improvements to surface_unlit

jstone-lucasfilm:
- Assign a unique name to the new node, in order to avoid name collisions in applications with dynamic node typing (i.e. where node types are inferred from connections).
- Set the default value of the emission scalar input to 1.


niklasharrysson (approved):
Looks good! Thanks @jstone-lucasfilm

PR #895, opened 2022-03-30
JavaScript viewer update

kwokcb:
## Updates:

- Add USD and glTF as loadable materials.
- Add in geometry switcher to select glTF files to load.
- Fix texture flipping.
- Fix when default url options not set.
- Update camera to more closely match desktop viewer.
- For boombox example add in images and image path mapping. Search pathing has not been added to viewer so adding to manual mapping.
- Fixed up geometric stream loading so there is fallback geometry if the load fails, and if there are no uvs or normals they are generated. This prevents tangent generation from failing if either of those does not exist and hence failing on a series of GL errors.
- Add some very basic parameter control.

### Example glTF boombox in JS and desktop viewer (left, and right respectively)
image

### Testing on Mac Safari
- Tested local git-hub page on Safari now that WebGL is supported :).
WebGL_Safari

### Latest UI 
- Overlayed loader UI + parameter UI
NewUI
- Same with parameter UI collapsed.
NewUI_collapsedUI


javascript/MaterialXView/source/index.js
```diff
@@ -21,7 +21,15 @@ let normalMat = new THREE.Matrix3();
 let viewProjMat = new THREE.Matrix4();
 let worldViewPos = new THREE.Vector3();
 
-const materialFilename = new URLSearchParams(document.location.search).get("file");
+// Get URL options. Fallback to defaults if not specified.
+let materialFilename = new URLSearchParams(document.location.search).get("file");
```
kwokcb, line 25: This is kind not efficient the way it's implemented. It reloads the entire page with option args which specify the material, and now the geometry file. Refactoring this is out-of-scope for this PR.

javascript/MaterialXView/source/index.js
```diff
@@ -184,11 +213,11 @@ function init()
         bbox.getBoundingSphere(bsphere);
 
         controls.target = bsphere.center;
-        camera.position.y = bsphere.radius * 0.65;
-        camera.position.z = bsphere.radius * 1.95;
+        camera.position.y = bsphere.center.y;
```
kwokcb, line 216: Small initial camera changes.

javascript/MaterialXView/webpack.config.js
```diff
@@ -5,10 +5,33 @@ const HtmlWebpackPlugin = require('html-webpack-plugin')
 
 const stdSurfaceMaterials = "../../resources/Materials/Examples/StandardSurface";
 const stdSurfaceMaterialsBaseURL = "Materials/Examples/StandardSurface";
-
-const materials = fs.readdirSync(stdSurfaceMaterials)
+let materials = fs.readdirSync(stdSurfaceMaterials)
   .map((fileName) => ({name: fileName, value: `${stdSurfaceMaterialsBaseURL}/${fileName}`}));
 
+const usdSurfaceMaterials = "../../resources/Materials/Examples/UsdPreviewSurface";
```
kwokcb, line 11: Add in USD and glTF materials to be available.

javascript/MaterialXView/webpack.config.js
```diff
@@ -19,7 +42,8 @@ module.exports = {
   plugins: [
     new HtmlWebpackPlugin({
       templateParameters: {
-        materials
+        materials,
+        geometry
```
kwokcb, line 46: Allows for access from javascript to build the html options.

jstone-lucasfilm:
@kwokcb This sounds like a good improvement!  Could you send out a link to your locally-built web viewer, so that we can see the changes in context?

kwokcb:
> @kwokcb This sounds like a good improvement! Could you send out a link to your locally-built web viewer, so that we can see the changes in context?

Hi @jstone-lucasfilm, I created a [dummy branch ](https://github.com/kwokcb/MaterialX/tree/jsviewer_update_deploy) of this to force [deployment](https://github.com/kwokcb/MaterialX/tree/gh-pages-testing) on my main github page here:
https://kwokcb.github.io/MaterialX/
It's a bit non-standard but it should work.

jstone-lucasfilm:
@bernardkwok I like the new functionality in this change, but it seems as if we've lost some of the design finesse in the original web viewer, especially in the handling of options at the top of the screen.  Here are two screenshots to give a sense of what I mean:

Original Web Viewer:
OriginalWebViewer

New Web Viewer:
NewWebViewer

Would it be possible to maintain the look of the original "material" option more closely in the addition of this new "geometry" option?

kwokcb:
> @bernardkwok I like the new functionality in this change, but it seems as if we've lost some of the design finesse in the original web viewer, especially in the handling of options at the top of the screen. Here are two screenshots to give a sense of what I mean:
> 
> Would it be possible to maintain the look of the original "material" option more closely in the addition of this new "geometry" option?

Yes, I have this fixed locally already. It's just a formatting of where the menus are so they overaly vs are above the viewing canvas.

jstone-lucasfilm:
The new material properties panel is a great addition, though it's slightly confusing that it only works for manually-loaded materials, not for the initial `standard_surface_default` material that launches in the viewer.  Also, perhaps `Properties` would be a better name than `Material` for this panel?

source/JsMaterialX/JsMaterialXGenShader/JsShaderNode.cpp
```diff
@@ -19,5 +19,9 @@ EMSCRIPTEN_BINDINGS(ShaderPort)
         .function("getVariable", &mx::ShaderPort::getVariable)
         .function("getType", &mx::ShaderPort::getType, ems::allow_raw_pointers())
         .function("getValue", &mx::ShaderPort::getValue)
+        .function("getPath", &mx::ShaderPort::getPath)
```
kwokcb, line 22: Reflection API is still incomplete but this helps with some basics including getting a path back to get things like UI attributes etc.

kwokcb:
> The new material properties panel is a great addition, though it's slightly confusing that it only works for manually-loaded materials, not for the initial `standard_surface_default` material that launches in the viewer. Also, perhaps `Properties` would be a better name than `Material` for this panel?

Hi @jstone-lucasfilm,
I should remove old images. It works the same for all materials including the initial one. I can rename the panel and also make it not collapsed by default.

jstone-lucasfilm:
@kwokcb These are some great improvements, and this looks nearly ready to merge.  Maybe we should start with the properties window _closed_ rather than open, so that the initial configuration works on mobile devices as well as desktop PCs?

kwokcb:
Probably last cleanup as this is about the limits of `dat-gui` usage (`im-gui` is a better toolkit). Also there is missing Javascript API which should be addressed afterwards (too large a change).
- Fixed up float parameters not showing up if 0,
- Added in some ui-folder grouping and naming.
- Property editor is closed on load. Can also use 'h' key to hide. 

image
image

jstone-lucasfilm (approved):
Great work, thanks @kwokcb!  I'll plan to merge this today, as it's definitely worthwhile to include in the upcoming 1.38.4 release.

PR #900, opened 2022-04-04
fixing UI min/max range for standard-surface SSS aniso parameter

BrianSharpe:
subsurface_anisotropy parameter on the ND_standard_surface_surfaceshader_100 node has incorrect uimin value.  Currently its at 0, it should instead be -1 (representing the Henyey-Greenstein G value).

### DIFF
--- a/libraries/bxdf/standard_surface.mtlx
+++ b/libraries/bxdf/standard_surface.mtlx
@@ -54,7 +54,7 @@
            doc="The mean free path. The distance which light can travel before being scattered inside the surface." />
     <input name="subsurface_scale" type="float" value="1" uimin="0.0" uisoftmax="10.0" uiname="Subsurface Scale" uifolder="Subsurface" uiadvanced="true"
            doc="Scalar weight for the subsurface radius value." />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
@BrianSharpe This change looks fine to me, but make sure you're part of a Contributor License Agreement, either through SideFX or your own account!

BrianSharpe:
Ok I've now signed the Contributor License Agreement as an individual.  Hopefully the EasyCLA checks will resolve themselves.
thanks!

jstone-lucasfilm:
@BrianSharpe That last "/easycla" comment should have done the trick, but there may be a subtle issue I'm missing.  I would recommend creating a support ticket using the link [above](https://github.com/AcademySoftwareFoundation/MaterialX/pull/900#issuecomment-1086984254), and they should be able to get this resolved quickly.

BrianSharpe:
ok, I've logged a ticket
https://jira.linuxfoundation.org/plugins/servlet/theme/portal/4/SUPPORT-9555

Cheers!

PR #909, opened 2022-04-11
Refactor Web MaterialXView

kwokcb:
## Changes
*  Create Viewer, Editor, Material and Scene classes to organize functionality
* Change material and geometry changes to not reload the entire page causing everything to be re-initialized. Instead perform incremental updates.
* Small fix to get any UI names if looking up an interface input

## Example
* Sample session capture to show how material and geometry can be independently re-loaded.
viewerCapture

* Sample capture of geometry change while material changes are preserverd
materialWithGeomChange


kwokcb (commented):
A lot of changes related to creating some logical classes and getting functionality and code restructured properly. This allowed for independent geometry and material loading refactoring
so the entire page does not a reload on changes.

javascript/MaterialXView/source/index.js
```diff
@@ -4,484 +4,116 @@
...
-        // Register lights with generation context
-        const lights = findLights(doc);
-        const lightData = registerLights(mx, lights, genContext);
-
-        const radianceTexture = prepareEnvTexture(loadedRadianceTexture, renderer.capabilities);
-        const irradianceTexture = prepareEnvTexture(loadedIrradianceTexture, renderer.capabilities);
-
-        const threeMaterial = generateMaterial(elem, gen, genContext, lights, lightData,
-            textureLoader, radianceTexture, irradianceTexture, gui);
+        viewer.getMaterial().loadMaterial(viewer, materialFilename);
```
kwokcb, line 92: As much as possible has been hidden now in the viewer class. The main calls are
loading a material or loading geometry, which are used in the callbacks when menu selection changes as well as here on page load.

jstone-lucasfilm:
This looks really promising, thanks @bernardkwok!  Could you post a link to a deployment of the updated site, so that we can get a sense of how it works in practice?

kwokcb:
> This looks really promising, thanks @bernardkwok! Could you post a link to a deployment of the updated site, so that we can get a sense of how it works in practice?

The deployment can be tried out here.](https://kwokcb.github.io/MaterialX/).

jstone-lucasfilm:
This looks excellent, thanks @bernardkwok.  I'll give it a day or two for others to comment, but my instinct is to merge this change ASAP.

PR #910, opened 2022-04-11
Add emissive strength to glTF PBR

proog128:
This PR adds support for the upcoming KHR_materials_emissive_strength (KhronosGroup/glTF#1994) extension to the glTF PBR node definition and implementation. The extension adds a parameter called "Emissive Strength". It is a scalar (uniform) multiplier to the emission color. The additional parameter is required in glTF because the glTF core spec limits emission color to 1 (KhronosGroup/glTF#1083).

The extension is currently in "Release Candidate" status, which means we do not expect huge changes anymore and implementations can adapt it.

### DIFF
--- a/libraries/bxdf/gltf_pbr.mtlx
+++ b/libraries/bxdf/gltf_pbr.mtlx
@@ -20,7 +20,7 @@
     <input name="clearcoat_normal" type="vector3" defaultgeomprop="Nworld" uiname="Clearcoat Normal" uifolder="Clearcoat" />
     <input name="emissive" type="color3" value="0, 0, 0" uimin="0, 0, 0" uimax="1, 1, 1" uiname="Emissive" uifolder="Emission" />
     <input name="emissive_strength" uniform="true" type="float" value="1" uimin="0" uiname="Emissive Strength" uifolder="Emission" />
-    <input name="thickness" uniform="false" type="float" uimin="0" uiname="Thickness" uifolder="Volume" />
+    <input name="thickness" uniform="false" type="float" value="0" uimin="0" uiname="Thickness" uifolder="Volume" />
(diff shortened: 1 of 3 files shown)


jstone-lucasfilm (approved):
This looks great to me, thanks @proog128.  I'll give others a day or two to comment, but this seems ready to merge.

libraries/bxdf/gltf_pbr.mtlx
```diff
@@ -19,6 +19,7 @@
     <input name="clearcoat_roughness" type="float" value="0" uimin="0" uimax="1" uiname="Clearcoat Roughness" uifolder="Clearcoat" />
     <input name="clearcoat_normal" type="vector3" defaultgeomprop="Nworld" uiname="Clearcoat Normal" uifolder="Clearcoat" />
     <input name="emissive" type="color3" value="0, 0, 0" uimin="0, 0, 0" uimax="1, 1, 1" uiname="Emissive" uifolder="Emission" />
+    <input name="emissive_strength" uniform="true" type="float" value="1" uimin="0" uiname="Emissive Strength" uifolder="Emission" />
     <input name="thickness" uniform="false" type="float" uimin="0" uiname="Thickness" uifolder="Volume" />
```
kwokcb, line 23: Hi @proog128, a minor thing not directly related. Both `thickness` and `attenutation_distance` have no default values so perhaps worth fixing up with this check-in ? Works for the example since it's explicitly specified in the instance of the node, otherwise not sure what value is used.

jstone-lucasfilm: That's a good question, and I don't believe we yet handle `inf` as a valid scalar value in MaterialX, though this seems like a natural feature to add in the future.  For now, I think you're right to omit this from the shading model declaration, and we can discuss how to handle this feature going forward.

kwokcb (approved):

PR #911, opened 2022-04-12
fixing UI min/max range for standard-surface SSS aniso parameter

BrianSharpe:


jstone-lucasfilm (approved):
Looks good to me, thanks @BrianSharpe!

PR #913, opened 2022-04-17
Fix environment preview for viewer

kwokcb:
The environment material used for 3d preview was not being recompiled when changing the environment.
Hence the scene was rendering using the new environment but the preview was not.


jstone-lucasfilm (approved):
Thanks for this fix, @kwokcb!

PR #915, opened 2022-04-19
IMP_UsdPrimvarReader reader fixes

BrianSharpe:
As per discussions on MaterialX slack channel, this implements...
- the "fallback" parameter in IMP_UsdPrimvarReader
- also makes the default parameter in ND_geompropvalue_integer non-uniform


jstone-lucasfilm (approved):
Looks good to me, thanks @BrianSharpe!

PR #917, opened 2022-04-20
Fix sample sphere geometry 

kwokcb:
An interior test plane was left inside the sphere geometry.
Removing it as it affects transparent objects as well as environment preview in MaterialXView.

## Viewer Result 
* Before with one of the test suite test shaders
<image src="https://user-images.githubusercontent.com/49369885/164279099-8891e012-206b-453a-808e-ea190a0964ad.png" width=480>
* After
<image src="https://user-images.githubusercontent.com/49369885/164279259-43502146-4ffd-47ed-bb0a-0f6eccd9d355.png" width=480>

## Test Suite Results (GLSL, OSL)
* Hard to see differences since plane is axis aligned with the view vector
[April_20_2022_Tests.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/8523473/April_20_2022_Tests.pdf)


jstone-lucasfilm (approved):
Thanks for the fix, @kwokcb!

PR #918, opened 2022-04-20
Initial refraction approximation in GLSL

jstone-lucasfilm:
This changelist introduces a rough approximation of refraction in GLSL, allowing transmissive materials such as glass to respond visually to changes in specular_IOR, specular_roughness, and transmission_color.

Because this initial approximation is limited to a single render pass, only the environment radiance map is currently visible in refractions, and opaque objects behind transmissive surfaces are not yet visible.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
@@ -34,7 +34,7 @@ vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 alpha, int distributio
 
         // Compute the half vector and incoming light direction.
         vec3 H = mx_ggx_importance_sample_NDF(Xi, alpha);
-        vec3 L = -reflect(V, H);
+        vec3 L = fd.refraction ? mx_refraction_solid_sphere(-V, H, fd.ior.x) : -reflect(V, H);
         
         // Compute dot products for this sample.
         float NdotH = clamp(H.z, M_FLOAT_EPS, 1.0);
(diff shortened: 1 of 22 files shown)


jstone-lucasfilm:
The following renders are from MaterialXView, showing the proposed refraction approximation in the context of transmissive Standard Surface materials:

Standard Surface Glass:
StandardSurfaceGlass_GlslRefraction

Standard Surface Glass Tinted:
StandardSurfaceTintedGlass_GlslRefraction

jstone-lucasfilm:
Thanks to @mjyip-lucasfilm for her significant contributions to this work!

jstone-lucasfilm:
Reference renders in GLSL and OSL:

[ReferenceRender_04_20_2022.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/8526914/ReferenceRender_04_20_2022.pdf)

libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
```diff
@@ -34,7 +34,7 @@ vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 alpha, int distributio
 
         // Compute the half vector and incoming light direction.
         vec3 H = mx_ggx_importance_sample_NDF(Xi, alpha);
-        vec3 L = -reflect(V, H);
+        vec3 L = fd.refraction ? mx_refraction_solid_sphere(-V, H, fd.ior.x) : -reflect(V, H);
```
niklasharrysson, line 37: Just a note for the future that we need to split this switch in two later to support thin-walled materials. We then need to have separate control of transmission and refraction on/off, since we can have transmission without refraction in that case, but still transmission with a roughness.

libraries/pbrlib/genglsl/mx_dielectric_bsdf.glsl
```diff
@@ -61,20 +54,29 @@ void mx_dielectric_bsdf_transmission(vec3 V, float weight, vec3 tint, float ior,
...
     vec3 comp = mx_ggx_energy_compensation(NdotV, avgAlpha, F);
     vec3 dirAlbedo = mx_ggx_dir_albedo(NdotV, avgAlpha, F0, 1.0) * comp;
     bsdf.throughput = 1.0 - dirAlbedo * weight;
 
-    bsdf.response = (scatter_mode == 2) ? tint * weight * bsdf.throughput : vec3(0.0);
+    // For now, we approximate the appearance of dielectric transmission as
+    // glossy environment map refraction, ignoring any scene geometry that
+    // might be visible through the surface.
+    fd.refraction = true;
+    vec3 Li = $refractionEnv ? mx_environment_radiance(N, V, X, safeAlpha, distribution, fd) : $refractionColor;
```
niklasharrysson, line 78: Out of curiosity why is there an option to replace the refracted environment with a uniform color? We don't have a corresponding control for environment reflections. Or is this just for debugging for now?

niklasharrysson (approved):
Looks great @jstone-lucasfilm. I only had some minor comments above.

libraries/pbrlib/genglsl/mx_dielectric_bsdf.glsl
jstone-lucasfilm: That option was primarily added for our GLSL/OSL reference renders, since OSL `testrender` refracts the constant-color _background_ instead of the distant environment.

niklasharrysson: Aha! Maybe we should fix this so the environment is visible in both reflection and transmission, for both GLSL and OSL test suite rendering. Lets do that in a separate PR though.

PR #919, opened 2022-04-22
Update UsdUVTexture to version 2.3

kwokcb:
# Update Notes
Update #923, #921

* 2.3 removes the `rgba` output
* 2.2 inherits from 2.3 since 2.3 removes an output. 
  * It's not possible to remove inputs or outputs via inheritance currently.
  * Issue: A full output signature is required even though inheritance is used. e.g. this is not allowed (just add 1 output):
  ```
  <nodedef name="ND_UsdUVTexture" node="UsdUVTexture" version="2.2" inherit="ND_UsdUVTexture_23">
	<output name="rgba" type="color4" />
  </nodedef>
  ```
* Clairification: There is no current support for `<nodegraph>` inheritance so a near duplicate implementation graph is required.  e.g. this is not allowed (seems like a validation defect):
   ```
   <nodegraph name="IMP_UsdUVTexture_22" nodedef="ND_UsdUVTexture" inherit="IMP_UsdUVTexture_23">   
       <output name="rgba" type="color4" nodename="image_bias" />
    </nodegraph>	
    ```
* Issue: It is not possible to add in a `<implementation>`  which associates the `nodedef` with the `nodegraph` as there
appears to be (validation) issues when the outputs access internal nodes.  This also appears to be a validation defect.
* Issue: Seems color management does not kick in for USDUvTexture. Historical and could be due to fact that node
is `multioutput` ?
* Issue: Version 2.3 introduces a new `sourceColorSpace` input. TBD if may this can map to `colorSpace` attribute.
* Issue: `wrapS` and `wrapT` inputs include `black`, and `useMetadata`. To see how to make these work.

# Testing
* Added new test to use UsdUVTexture as it was never proper of the test suite.
* Test usage of both versions in variants on the existing brass material test.

## RTS results:
<image src="https://user-images.githubusercontent.com/49369885/164601408-f03c6503-161a-4e62-a7a8-2f3a2ff7fa82.png" width=640>

## Viewer results:
   * Version 2.2
     <image src="https://user-images.githubusercontent.com/49369885/164601802-5353075a-1c0b-4643-ad47-f0ec3decc305.png" width=640>
   * Version 2.3
     <image src="https://user-images.githubusercontent.com/49369885/164601903-850ae2c2-296b-46bb-aa57-5082799a9bfb.png" width=640>


pablode:
Hi! Thanks for this PR, I was going to bring this issue, up, too.
There are some things I noticed which could be included in this PR (or a separate one):
- Version 2.3 introduces a new sourceColorSpace input. Maybe it's possible to map this to MaterialX's colorSpace attribute?
- The wrapS and wrapT inputs have the following enum values: `black,clamp,periodic`. The UsdUVTexture spec defines the enum values `black,clamp,repeat,mirror,useMetadata` instead. Would it be possible to remap these enum values to the MaterialX `constant,clamp,periodic,mirror` ones using nodes? Even if it's not possible, it seems like `mirror` could be added. Not sure about the `black` enum type, currently it seems to be ignored.

kwokcb:
> Hi! Thanks for this PR, I was going to bring this issue, up, too. There are some things I noticed which could be included in this PR (or a separate one):
> 
> * Version 2.3 introduces a new sourceColorSpace input. Maybe it's possible to map this to MaterialX's colorSpace attribute?
> * The wrapS and wrapT inputs have the following enum values: `black,clamp,periodic`. The UsdUVTexture spec defines the enum values `black,clamp,repeat,mirror,useMetadata` instead. Would it be possible to remap these enum values to the MaterialX `constant,clamp,periodic,mirror` ones using nodes? Even if it's not possible, it seems like `mirror` could be added. Not sure about the `black` enum type, currently it seems to be ignored.

Hi @pablode,
I think since there are numerous issues to address that these additional issues can also be noted here but best to log some formal issues to address them to not make this one overtly too large or complicated a change ? @jstone-lucasfilm what do you think.

jstone-lucasfilm:
Thanks for putting this change together, @kwokcb!  @pablode brings up some additional good points, and I agree the remaining ones are subtle enough that they deserve their own GitHub Issues and discussion.

Since there haven't previously been versions for this node, my instinct would be to consider the new "2.3" to be the first version that we maintain and track, rather than aiming to maintain compatibility with the historical "2.2".  If we followed that approach here, we could simplify this PR and remove the need for its complex inheritance structure.

What are your thoughts?

kwokcb:
> Thanks for putting this change together, @kwokcb! @pablode brings up some additional good points, and I agree the remaining ones are subtle enough that they deserve their own GitHub Issues and discussion.
> 
> Since there haven't previously been versions for this node, my instinct would be to consider the new "2.3" to be the first version that we maintain and track, rather than aiming to maintain compatibility with the historical "2.2". If we followed that approach here, we could simplify this PR and remove the need for its complex inheritance structure.
> 
> What are your thoughts?

Hi @jstone-lucasfilm,
I think it's good to establish how another variant of versioning (with `<output>` changes) can be specified. Also
this preserves previous versions of `USDUvTexture` since there is no upgrade path. Otherwise they will cause
an error on load if the `rgba` output was ever used. Basically similar approach to Standard Surface.

jstone-lucasfilm (approved):

PR #926, opened 2022-04-29
Relative shader includes

jstone-lucasfilm:
This changelist adds support for relative includes of shader source files in code generation, and makes all file includes relative within the MaterialX data libraries.

This allows clients to use arbitrary naming conventions for the root folder of their MaterialX data libraries, with "libraries" being the default name in the MaterialX distribution.

### DIFF
--- a/libraries/lights/genglsl/lights_genglsl_impl.mtlx
+++ b/libraries/lights/genglsl/lights_genglsl_impl.mtlx
@@ -2,12 +2,12 @@
 <materialx version="1.38">
 
   <!-- <point_light> -->
-  <implementation name="IM_point_light_genglsl" nodedef="ND_point_light" file="libraries/lights/genglsl/mx_point_light.glsl" function="mx_point_light" target="genglsl" />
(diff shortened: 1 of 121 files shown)


ix-dcourtois:
Hi, first off, thanks a lot for taking the time to make this change, it's highly appreciated!
I've reviewer the changes, and from what I can understand, I think that it would indeed allow us to upgrade without changing how the data library is deployed, compared to 1.38.3. I think we would have to reset the library prefix to an empty string, but otherwise the rest should work just fine.
I just hope that the changes to `emitBlock` and `emitInclude` in the `ShaderGenerator` class will not be too much of a bother for others (I've seen your announcement in the Slack channel, I hope other teams will take time to check this to avoid any surprise) (and maybe I should also start following that channel :p)
Regards,
Damien.

rafalSFX:
Using relative paths seems like a good solution, that should address the problem of an arbitrary installation structure. 

If the included files are not moved relative to the files that include them, all should work, as long as the preprocessor looks for the include files relative to the same directory (of the including file) too. Which I assume is the case for .mtlx, .osl, .glsl (etc) preprocessors.

Issue #922, opened 2022-04-26
Thin-film with metallic base in standard surface implementation


dleex:
It seems "thin_film_bsdf" is layerd on top of "specular_layer" then mixed with "metal_bsdf" (with "metalness") afterwards, so 100% metallic material won't have any thin-film. But according to the [spec](https://autodesk.github.io/standard-surface/), "thin-film logically sits between the coat and underlying layer" so it should be possible to have thin-film with metallic base.

niklasharrysson:
Good catch @dleex! We previously didn't support having "thin_film_bsdf" applied to a mix of BSDF nodes, but this was resolved in #747 . So I just missed to update the nodegraph for Standard Surface to use this. Should be easy to resolve. Thanks for reporting it.

jstone-lucasfilm:
Thanks to @dleex for this report, and to @niklasharrysson for the fix!

PR #927, opened 2022-05-01
Fix issue with thin-film in standard_surface implementation

niklasharrysson:
This change list improves the implementation of standard_surface. The thin-film BSDF operator is now applied to both the dielectric specular BSDF and the metal BSDF, following the Standard Surface specification.

This should resolve issue #922.


jstone-lucasfilm (approved):
This looks like a good fix, thanks @niklasharrysson!

Issue #921, opened 2022-04-26
Some library nodes are not part of a nodegroup


JGamache-autodesk:
Very minor. The following nodes do not have a nodegroup attribute:

ND_UsdPrimvarReader_*
ND_UsdTransform2d
ND_UsdUVTexture
ND_disney_brdf_2012_surface
ND_disney_bsdf_2015_surface
ND_lama_*
ND_standard_surface_to_UsdPreviewSurface

kwokcb:
Hi @JGamache-autodesk, I happen to be in the USD file so added USD nodgroups as part of this [PR](https://github.com/AcademySoftwareFoundation/MaterialX/pull/919)

kwokcb:
Note:, with USD check-in, the list that remains is this:
ND_disney_brdf_2012_surface
ND_disney_bsdf_2015_surface
ND_lama_*

jstone-lucasfilm:
@kwokcb For the MaterialX Lama nodes, we should definitely add the node groups for clarity.

For the two Disney Principled declarations, I've been considering the idea of removing these from the MaterialX data libraries, since they don't yet have BSDF graph definitions.  We can always restore these nodes in the future when graph definitions are available.

What are your thoughts?

kwokcb:
Hi @jstone-lucasfilm,
Agree with the Lama nodes. Of immediate advantage is to tag texture nodes as this is used during shader generation.
I can take some time to update them.

I guess the Disney declarations is more up to you. Definitions in themselves can still be used outside of implementation / shader generation. e.g. can still use them in translation graphs, USDShade etc -- or just to create graphs using these nodes -- so they seem worth keeping for these other purposes?

jstone-lucasfilm:
Good points, @kwokcb, and I think you're right that the Disney Principled interfaces are potentially valuable as they stand, even without full graph definitions.

Would you mind extending your latest Lama MaterialX improvement to include the Disney Principled nodes as well, so that we can close out this issue?

kwokcb:
Sure no problem. I'm going to update the `mxdoc` script as well.

jstone-lucasfilm:
Thanks for this report, @JGamache-autodesk, and thanks to @kwokcb for the fixes!

PR #929, opened 2022-05-03
Add missing `nodegroup` to a few Lama definitions.

kwokcb:
Update #921 
Few small updates for Lama nodes so all nodes have `nodegroup`  tags.


jstone-lucasfilm (approved):

PR #934, opened 2022-05-05
Transmission shading improvements

jstone-lucasfilm:
- Add support for generalized Schlick transmission in OSL.
- Minor improvements to dielectric transmission in OSL.
- Clamp input values to a stable range in mx_f0_to_ior.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -223,7 +223,7 @@ float mx_ior_to_f0(float ior)
 // Convert normal-incidence reflectivity to real-valued index of refraction.
 float mx_f0_to_ior(float F0)
 {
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm:
Reference renders in GLSL and OSL:
[MaterialX_ReferenceRenders_05_05_2022.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/8635599/MaterialX_ReferenceRenders_05_05_2022.pdf)

niklasharrysson (approved):
Looks good @jstone-lucasfilm. 

Since we only have the "scientific Fresnel" microfacet closure available in vanilla OSL, I think this is the best approximation we can find for now.

As discussed many times we need the proper MaterialX closures implemented in OSL/testrender to get further on this.

jstone-lucasfilm:
Thanks @niklasharrysson, and agreed on the importance of moving forward with the MaterialX closures for OSL.  That would be a good subject for an upcoming MaterialX TSC.

PR #937, opened 2022-05-08
Normal and tangent fixes

pablode:
Hi @jstone-lucasfilm!

I recently tried to test a .mtlx file generated by my converter and ran into some rendering artifacts in MaterialXView:

before

Further investigation revealed that the normals were not normalized after a matrix multiplication in the glTF loader.
I have also identified more situations where I believe normalization is necessary.

Plus, it seems like tangents and bitangents were incorrectly transformed with the normal matrix:
https://pbr-book.org/3ed-2018/Geometry_and_Transformations/Applying_Transformations#Normals

After the changes, the model renders as follows:

after

You can find the model here: [ToyCar.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/8646544/ToyCar.zip)

### DIFF
--- a/source/MaterialXGenGlsl/Nodes/BitangentNodeGlsl.cpp
+++ b/source/MaterialXGenGlsl/Nodes/BitangentNodeGlsl.cpp
@@ -26,6 +26,7 @@ void BitangentNodeGlsl::createVariables(const ShaderNode& node, GenContext&, Sha
     const int space = spaceInput ? spaceInput->getValue()->asA<int>() : OBJECT_SPACE;
     if (space == WORLD_SPACE)
     {
+        addStageUniform(HW::PRIVATE_UNIFORMS, Type::MATRIX44, HW::T_WORLD_MATRIX, vs);
         addStageUniform(HW::PRIVATE_UNIFORMS, Type::MATRIX44, HW::T_WORLD_INVERSE_TRANSPOSE_MATRIX, vs);
         addStageConnector(HW::VERTEX_DATA, Type::VECTOR3, HW::T_BITANGENT_WORLD, vs, ps);
     }
@@ -47,12 +48,23 @@ void BitangentNodeGlsl::emitFunctionCall(const ShaderNode& node, GenContext& con
         const string prefix = shadergen.getVertexDataPrefix(vertexData);
         if (space == WORLD_SPACE)
         {
+            ShaderPort* normal = vertexData[HW::T_NORMAL_WORLD];
+            if (!normal->isEmitted())
+            {
+                normal->setEmitted();
+                shadergen.emitLine(prefix + normal->getVariable() + " = normalize((" + HW::T_WORLD_INVERSE_TRANSPOSE_MATRIX + " * vec4(" + HW::T_IN_NORMAL + ", 0.0)).xyz)", stage);
+            }
+            ShaderPort* tangent = vertexData[HW::T_TANGENT_WORLD];
+            if (!tangent->isEmitted())
+            {
+                tangent->setEmitted();
+                shadergen.emitLine(prefix + tangent->getVariable() + " = normalize((" + HW::T_WORLD_MATRIX + " * vec4(" + HW::T_IN_TANGENT + ", 0.0)).xyz)", stage);
+            }
             ShaderPort* bitangent = vertexData[HW::T_BITANGENT_WORLD];
             if (!bitangent->isEmitted())
             {
                 bitangent->setEmitted();
-                shadergen.emitLine(prefix + bitangent->getVariable() +
-                    " = (" + HW::T_WORLD_INVERSE_TRANSPOSE_MATRIX + " * vec4(cross(" + HW::T_IN_NORMAL + ", " + HW::T_IN_TANGENT + "), 0.0)).xyz", stage);
+                shadergen.emitLine(prefix + bitangent->getVariable() + " = cross(" + normal->getVariable() + ", " + tangent->getVariable() + ")", stage);
             }
         }
         else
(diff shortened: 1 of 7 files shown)


jstone-lucasfilm:
@pablode Thanks for putting together this post, and these improvements look very promising.  I'm CC'ing @bernardkwok for his thoughts on the revisions to the cgltf loader and MeshStream::transform.

source/MaterialXGenGlsl/Nodes/BitangentNodeGlsl.cpp
```diff
@@ -47,12 +48,23 @@ void BitangentNodeGlsl::emitFunctionCall(const ShaderNode& node, GenContext& con
         const string prefix = shadergen.getVertexDataPrefix(vertexData);
         if (space == WORLD_SPACE)
         {
+            ShaderPort* normal = vertexData[HW::T_NORMAL_WORLD];
```
kwokcb, line 51: This can cause the emitted code to be unpredicable depending on whether the normals and or the tangents have already been emitted. 

I guess this computation change is also spawns from the fact that bitangents are not bound precomputed streams and thus are computed here.

source/MaterialXGenGlsl/Nodes/TangentNodeGlsl.cpp
```diff
@@ -25,7 +25,7 @@ void TangentNodeGlsl::createVariables(const ShaderNode& node, GenContext&, Shade
     const int space = spaceInput ? spaceInput->getValue()->asA<int>() : OBJECT_SPACE;
     if (space == WORLD_SPACE)
     {
-        addStageUniform(HW::PRIVATE_UNIFORMS, Type::MATRIX44, HW::T_WORLD_INVERSE_TRANSPOSE_MATRIX, vs);
+        addStageUniform(HW::PRIVATE_UNIFORMS, Type::MATRIX44, HW::T_WORLD_MATRIX, vs);
```
kwokcb, line 28: Why is this now using the world matrix ?

source/MaterialXGenGlsl/Nodes/TransformNormalNodeGlsl.cpp
```diff
@@ -12,6 +12,20 @@ ShaderNodeImplPtr TransformNormalNodeGlsl::create()
     return std::make_shared<TransformNormalNodeGlsl>();
 }
 
+void TransformNormalNodeGlsl::emitFunctionCall(const ShaderNode& node, GenContext& context, ShaderStage& stage) const
```
kwokcb, line 15: This appears to be an incomplete implementation. This derives from a general transform from space A to space B which does not appear to be handled anywhere here.

I'd suggest injecting normalization code after the regular vector transform code.

source/MaterialXRender/CgltfLoader.cpp
```diff
@@ -309,7 +310,7 @@ bool CgltfLoader::load(const FilePath& filePath, MeshList& meshList, bool texcoo
                                     float floatValue = (v < vectorSize) ? input[v] : 0.0f;
                                     normal[v] = floatValue;
                                 }
-                                normal = positionMatrix.transformNormal(normal);
+                                normal = normalMatrix.transformVector(normal).getNormalized();
```
kwokcb, line 313: Thanks for the normalization fix.

kwokcb (changes requested):
This code adds normalization and tries to keep the tangent basis correct but needs some tweaking. 

BTW, MaterialXView assumes everything is on world space already so should bind as such. Thus there really shouldn't be any object->world transformation going on AFAIK since object == world.

I didn't double check but explicit bitangent binding should occur so should also not be computed.

kwokcb:
Leaving a ping with @niklasharrysson  as he wrote the geometric binding code for any comments.

source/MaterialXGenGlsl/Nodes/TangentNodeGlsl.cpp
pablode: Because, as the commit message explains, tangents are not transformed using the normal matrix:
https://pbr-book.org/3ed-2018/Geometry_and_Transformations/Applying_Transformations#Normals

source/MaterialXGenGlsl/Nodes/TransformNormalNodeGlsl.cpp
pablode: The first line of this function calls the base implementation of `TransformVectorNodeGlsl`, transforming the vector. It is then normalized, like you are suggesting.

source/MaterialXGenGlsl/Nodes/BitangentNodeGlsl.cpp
pablode: Can you elaborate on this?

I understand that the normal already might have been emitted in SurfaceNodeGlsl:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/46de5f9153aec01243fce24590b15fd71da89f8e/source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.cpp#L81-L86

However, that's why the `if (!normal->isEmitted())` check exists here. And the calculation is the same. Also, only `T_IN_NORMAL/TANGENT` is used in the calculation which I expect to exist?

source/MaterialXGenGlsl/Nodes/TangentNodeGlsl.cpp
jstone-lucasfilm: That's a great catch, @pablode, and we'd overlooked that detail in earlier implementations/review of the tangent node.

jstone-lucasfilm (approved):
These fixes look good to me, thanks @pablode!

PR #940, opened 2022-05-08
Improve generated header setup in MaterialXCore

theblackunknown:
# Description

This PR improves how `MaterialXCore` generated header is handled in the CMake setup.
It write generated sources in the buildtree rather than the source tree, and uses `CMAKE_CURRENT_LIST_DIR` instead of `CMAKE_CURRENT_SOURCE_DIR`, then `MaterialXCore` `.gitignore` file is not needed anymore.
I have encountered issues with it while developing a materialx port for vcpkg C++ package manager (cf. microsft/vcpkg#24614).

Those changes will leads to more reliable and reproducible results.

### DIFF
--- a/source/MaterialXCore/CMakeLists.txt
+++ b/source/MaterialXCore/CMakeLists.txt
@@ -1,7 +1,7 @@
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Generated.h.in ${CMAKE_CURRENT_SOURCE_DIR}/Generated.h)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Generated.h.in ${CMAKE_CURRENT_BINARY_DIR}/Generated.h)
 
 file(GLOB materialx_source "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
-file(GLOB materialx_headers "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
+file(GLOB materialx_headers "${CMAKE_CURRENT_SOURCE_DIR}/*.h" "${CMAKE_CURRENT_BINARY_DIR}/*.h")
 
 add_library(MaterialXCore ${materialx_source} ${materialx_headers})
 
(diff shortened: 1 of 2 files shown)


theblackunknown:
Regarding the CLA check, I have both sent a signed PDF and use the Github check link to sign it.
Let me know how to update the check for it

jstone-lucasfilm:
@theblackunknown Thanks for putting together this propoal!  You can often resolve the EasyCLA detection by adding a comment including the following text:

/easycla

If that doesn't resolve the issue, then I would recommend writing up an LFX support ticket as described here:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/940#issuecomment-1120483941

theblackunknown:
CI Static Analysis returned an error about `MATERIALX_NAMESPACE_BEGIN` being a unknown macro, but this does not seems related to my changes, does not it ?

> MaterialXGenShader/Factory.h:14:1: error: There is an unknown macro here somewhere. Configuration is required. If MATERIALX_NAMESPACE_BEGIN is a macro then please configure it. [unknownMacro]
MATERIALX_NAMESPACE_BEGIN
^

theblackunknown:
@jstone-lucasfilm I figured out the EasyCLA issue ! 
Note that I have also updated the way `cppcheck` is done by relying on `CMake` generated compile commands rather than an explicit command-line configuration as it was done before, this should improve the robustness of cppcheck too.

theblackunknown:
It seems that I need your approval to run the CI and make sure this is working as expected

source/MaterialXCore/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Generated.h.in ${CMAKE_CURRENT_SOURCE_DIR}/Generated.h)
+configure_file(${CMAKE_CURRENT_LIST_DIR}/Generated.h.in ${CMAKE_CURRENT_BINARY_DIR}/Generated.h)
```
jstone-lucasfilm, line 1: I can see the need for the generated header to be in CMAKE_CURRENT_BINARY_DIR, and that part of the change seems like a clear improvement.

Can you provide some clarity on the motivation for referencing source files in CMAKE_CURRENT_LIST_DIR rather than CMAKE_CURRENT_SOURCE_DIR?  I'm not against this idea, but I want to make sure that I understand the logic behind the change.

theblackunknown: In this context it does not matter that much and both are equivalent, so I will probably revert this change to keep the diff minimal. 

My habit though it to use `CMAKE_CURRENT_LIST_DIR` when I can as it always refers to the CMake file currently being processed as the other one may not. 

Here is a page with a bit more details and examples if you want to dig this : https://cgold.readthedocs.io/en/latest/tutorials/cmake-sources/common.html#cmake-current-list-dir-vs-cmake-current-source-dir

jstone-lucasfilm (approved):
This looks good to me, thanks @theblackunknown!

PR #942, opened 2022-05-10
Enable JMC flag for Visual Studio

ashwinbhat:
Just my code (JMC) is a helpful VS compiler flag that enables a debugging feature to automatically step over calls to system, framework, and other non-user code.
see https://docs.microsoft.com/en-us/visualstudio/debugger/just-my-code?view=vs-2022#BKMK_C___Just_My_Code

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,11 +9,11 @@ cmake_minimum_required(VERSION 3.1)
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 set(CMAKE_MACOSX_RPATH ON)
+set(CMAKE_VS_JUST_MY_CODE_DEBUGGING ON)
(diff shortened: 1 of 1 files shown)


kwokcb:
Don't mean to be nosey :), but seems like this should be a build option. I've read this can have adverse affects on exception handling from dependent libraries.

jstone-lucasfilm:
@ashwinbhat @kwokcb Since I don't set this flag in my own Visual Studio environment, I'm open to ideas on how it would best be handled in MaterialX.  A build flag would be the most flexible, especially if there are tradeoffs involved in enabling it.

ashwinbhat:
Based on the documentation and checking with the internal team, I don't see any impact on behavior. 
This can be enabled via commandline using `-DCMAKE_VS_JUST_MY_CODE_DEBUGGING=ON` so no need to add it to CMake setting. I'll close this PR

PR #944, opened 2022-05-10
Upload MaterialX JavaScript

kwokcb:
* Set up so that CI will upload an artifact for Javascript / WASM libraries
* Add a package.json wrapper to the install directory. This file can be used to publish to npm. (via `npm publish`).
These steps are not part of CI otherwise arbitrary PRs would publish.  A test publish ends up with a result like this:
image

### DIFF
--- a/source/JsMaterialX/package.json
+++ b/source/JsMaterialX/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "jsmaterialx",
+  "version": "1.38.5",
+  "description": "JavaScript MaterialX Library",
+  "main": "JsMaterialXGenShader.js",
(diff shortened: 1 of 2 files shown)


source/JsMaterialX/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "jsmaterialx",
```
kwokcb, line 2: BTW, If we want to publish this, it might be good to do it "sooner". It tried to publish with the name `materialx` but someone already has a package with that name (totally unrelated).

jstone-lucasfilm (approved):

PR #947, opened 2022-05-12
MaterialXView : Add Basic Turntable

kwokcb:
# UI Changes
* Create new `Viewport Options` area
  * Add in ability to toggle turntable (rotation of mesh along axis)
  * Add option for number of steps within a single turntable revolution.
 
image

# Command Line Changes:
  ```
--enableTurntable[BOOLEAN]     Specify whether to enable turntable rendering of the scene
 --turntableSteps [INTEGER]     Specify the number of steps for a complete turntable rotation. Defaults to 360
```
### Example usage:
  * Command: `MaterialXView.exe --enableTurntable true --turntableSteps 9 --captureFilename blah.png`
  * Command output: 
```
Wrote turntable frame at angle: 0.000000 to file: blah_0000.png
Wrote turntable frame at angle: 40.000000 to file: blah_0001.png
Wrote turntable frame at angle: 80.000000 to file: blah_0002.png
Wrote turntable frame at angle: 120.000000 to file: blah_0003.png
Wrote turntable frame at angle: 160.000000 to file: blah_0004.png
Wrote turntable frame at angle: 200.000000 to file: blah_0005.png
Wrote turntable frame at angle: 240.000000 to file: blah_0006.png
Wrote turntable frame at angle: 280.000000 to file: blah_0007.png
Wrote turntable frame at angle: 320.000000 to file: blah_0008.png
```
# Interaction changes.
  * If turntable is enabled, the `F` key will write out a series of images from 0 to 360 degrees by the angle increment specified.
  * For example if the total steps is 36 then 10 (360/ 36) frames are written in the format `<filename>_####.<extension>`
    Where `####` is 0-padded frame number from 0 to the number of frames written.
  * The `P` (play) key is used to key turntable animation.
  * The web view now can also show the background environment (`V` key for both apps). 

# Animations
  * For now there are no movie or animated gif files written out. The files for a couple of examples were written to an animated gif and mpg (respectively) as a post process. (Note: Please save out the files to avoid choppy animation which is not inherently part of the saved files themselves).

TT_Capture

  https://user-images.githubusercontent.com/49369885/168165395-fdca7a41-5b7b-4bb3-bbe3-a790752e6159.mp4
  
  * Capture from Web Viewer
  
 https://user-images.githubusercontent.com/49369885/169410523-e2a20947-e52f-4ecd-b3da-5e500ea48350.mp4

### CODE REVIEW

File: source/MaterialXView/Viewer.cpp [hunk 1]
```diff
@@ -895,6 +889,57 @@ void Viewer::createAdvancedSettings(Widget* parent)
         _lightHandler->setEnvSampleCount(MIN_ENV_SAMPLE_COUNT * (int) std::pow(4, index));
     });
 
+    ng::Label* viewLabel = new ng::Label(advancedPopup, "Viewing Options");
```
kwokcb, line 892: I thought it makes sense to pull these out as they aren't really rendering options.

source/MaterialXView/Viewer.cpp
```diff
@@ -2452,7 +2562,11 @@ mx::ImagePtr Viewer::getShadowMap()
             blurSamplingProperties.uaddressMode = mx::ImageSamplingProperties::AddressMode::CLAMP;
             blurSamplingProperties.vaddressMode = mx::ImageSamplingProperties::AddressMode::CLAMP;
             blurSamplingProperties.filterType = mx::ImageSamplingProperties::FilterType::CLOSEST;
-            for (unsigned int i = 0; i < _shadowSoftness; i++)
+
+            // This is expensive to compute every frame for a turntable render.
+            // Compute this however if a capture is requisted.
+            unsigned int softness = (_turntableEnabled && !_captureRequested) ? 0 : _shadowSoftness;
```
kwokcb, line 2568: I didn't examine this but this code is quite slow and makes the updates very laggie.

jstone-lucasfilm: @kwokcb In my experience, it's more common for turntable systems to rotate the _camera_ around a fixed mesh, rather than rotating the mesh itself.  This would eliminate the problem you ran into here, as shadows would no longer need to be recomputed on each frame of the turntable render.  Would that potentially be a better overall approach to implementing this new feature?

jstone-lucasfilm: Both approaches to rendering turntables are certainly valid, and there are cases where a user would prefer to keep the camera static and rotate the mesh.  But to robustly implement this form of turntable rendering in MaterialXView, we'd need to address the underlying performance limitation by moving shadow rendering entirely to the GPU, rather than implementing it as a hybrid GPU/CPU pass.

That shadow optimization would be valuable in any case, but I wouldn't recommend trying to work around it by disabling individual shadow passes during turntable rendering, since this effectively reduces the visual quality of turntable renders below the quality of our standard viewport renders.

kwokcb: I can remove this to not mess with quality vs frame rate. BTW other options slow frame rate (like upping the number of environment samples). "Interactive degradation" options could be added if that's a route worth pursuing.

Perhaps we could try this option first, as a camera based turntable has different characteristics.

Camera rotation also means the view of the background environment changes so to me is more of a camera "fly-by".
Currently the camera is free to move about so a rotating mesh can be evaluated from different angles.
Finally I think you may end up with some unexpected camera viewpoints if someone does transform the camera before running a turntable. I guess you could fix the "look at" point to the origin, or scene center but rotations on the camera could end up with strange results. 

Does this sound reasonable?

jstone-lucasfilm: Good points about mesh vs camera rotation, and I think you're right to focus on mesh rotation for now.  In that case, I would recommend keeping your change as minimal as possible, and leaving performance optimizations as a future improvement.  My sense is that optimizing shadow rendering would make the most difference, removing the CPU step that copies final pixels into a texture and making this a pure GPU operation.

kwokcb: If I have some time, I can take a look at how to create a all hardware "pipeline" but probably not until later.
I know we don't want to formalize MaterialX as a "pipeline" specifier but I may be tempted to try it out in this context.
To solve this would also to allow MaterialX to generate shaders for "multi-pass" web rendering (nothing is enabled now).

source/MaterialXView/Main.cpp
```diff
@@ -14,6 +14,8 @@ const std::string options =
 "    --mesh [FILENAME]              Specify the filename of the OBJ mesh to be displayed in the viewer\n"
 "    --meshRotation [VECTOR3]       Specify the rotation of the displayed mesh as three comma-separated floats, representing rotations in degrees about the X, Y, and Z axes (defaults to 0,0,0)\n"
 "    --meshScale [FLOAT]            Specify the uniform scale of the displayed mesh\n"
+"    --enableTurntable[BOOLEAN]     Specify whether to enable turntable rendering of the scene\n"
+"    --turntableAmount [FLOAT]      Specify the amount to incrementatlly rotate the scene when turntable rendering is enabled (defaults to 1 degree)\n"
```
jstone-lucasfilm, line 18: This might be better stated as "--turntableSteps", where the current rotation is computed each frame from the current and total step counts.  This would avoid the potential for floating-point precision issues when incrementing rotation angles on a per-frame basis.

kwokcb: Good suggestion. Updated now. Thanks.

source/MaterialXView/Viewer.cpp
kwokcb: @jstone-lucasfilm. I ran this with on a Mac M1 machine and there is only a very small perceivable lag for the shadow filtering option turned on. Can probably log something once I figure out what the characteristics are.

source/MaterialXView/Main.cpp
```diff
@@ -265,6 +281,8 @@ int main(int argc, char* const argv[])
                                             screenColor);
         viewer->setMeshRotation(meshRotation);
         viewer->setMeshScale(meshScale);
+        viewer->setTurntableIncrement(360.0f / static_cast<float>(turntableSteps));
```
jstone-lucasfilm, line 284: In addition to describing the user input in terms of turntable steps, I was also imagining that the internal logic would operate in terms of integral current/total steps, eliminating the need to accumulate the rotation angle as a series of floating-point additions.  This would remove the potential for floating-point precision issues when the increment angle doesn't turn out to be exactly representable as a float.

source/MaterialXView/Viewer.cpp
```diff
@@ -1691,6 +1736,7 @@ bool Viewer::keyboard_event(int key, int scancode, int action, int modifiers)
             {
                 _captureFilename.addExtension(mx::ImageLoader::PNG_EXTENSION);
             }
+
```
jstone-lucasfilm, line 1739: Unintended line added here?

source/MaterialXView/Viewer.cpp
```diff
@@ -2084,8 +2130,48 @@ void Viewer::bakeTextures()
...
+        _turntableRotation = fmod(_turntableRotation + _turntableIncrement, 360.0f);
+    }
+
+    _meshRotation[1] = currentRotation;
+}
+
+void Viewer::draw_contents()
+{    
+    mx::ScopedTimer frameTime;
+    frameTime.startTimer();
```
jstone-lucasfilm, line 2173: I don't see where the new frameTime and _frameTime are yet used.  Maybe they were part of an unrelated change, and are not needed in this one?

jstone-lucasfilm (commented):
This looks like a good start, thanks @kwokcb, and I've written up some thoughts and recommendations.

source/MaterialXView/Main.cpp
kwokcb: Yes, I was thinking on this as well. 
One feedback item is for the UI the angle is + or - for forward or reverse, so the UI would present total steps -360 to 360 or step size -180 to 180. Does this sounds reasonable or should angle still be presented in the UI and steps deduced from this ?

source/MaterialXView/Viewer.cpp
kwokcb: Yes. This pulled over from some frame-rate display code. I'll remove.

source/MaterialXView/Viewer.cpp
```diff
@@ -209,6 +210,10 @@ Viewer::Viewer(const std::string& materialFilename,
     _searchPath(searchPath),
     _libraryFolders(libraryFolders),
     _meshScale(1.0f),
+    _turntableEnabled(false),
+    _turntableSteps(360),
+    _turntableStep(0),
+    _turntableRotation(0.0f),
```
jstone-lucasfilm, line 216: @kwokcb This might be clearer if we removed the shadow variable `_turntableRotation`, computing it as needed from `_turntableStep`, rather than maintaining both variables and keeping them in sync.

kwokcb: Makes sense. Removed now.

jstone-lucasfilm (approved):
This looks good to me, and thanks for contributing this new feature, @kwokcb!

PR #948, opened 2022-05-13
Fix build error when OSL or MDL is disabled in cmake options

niklasharrysson:
If MDL or OSL is disabled by build options Viewer.cpp would not compile.


jstone-lucasfilm (approved):

Issue #961, opened 2022-05-24
Regression: failures for baking in render test suite 


kwokcb:
If you run render tests for GLSL, these errors occur. They appear to be new so marking as regressions:
```
genglsl_render_log.txt:Failed to write baked image: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\4689911269797761771
genglsl_render_log.txt:Failed to write baked image: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\17205681858176880487
genglsl_render_log.txt:Failed to write baked image: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\1315322837353390033
genglsl_render_log.txt:Failed to write baked image: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\5369917747197254097
genglsl_render_log.txt:Failed to write baked image: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\6290293205101015369
genglsl_render_log.txt:Failed to write baked image: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\12308328058536465686
genglsl_render_log.txt:Wrote baked document: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\mapped_surfaceshader_baked_MappedShader.mtlx
genglsl_render_log.txt:Wrote baked document: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\mapped_surfaceshader_baked_UnitMappedShader.mtlx
genglsl_render_log.txt:Wrote baked document: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\mapped_surfaceshader_baked_ColorSpaceShader.mtlx
genglsl_render_log.txt:Wrote baked document: resources\Materials\TestSuite\pbrlib\surfaceshader\mapped_surfaceshader\mapped_surfaceshader_baked_NormalMapShader.mtlx
genglsl_render_log.txt:Failed to write baked image: resources\Materials\TestSuite\pbrlib\surfaceshader\normalmapped_surfaceshader\1557509004446421018
genglsl_render_log.txt:Wrote baked document: resources\Materials\TestSuite\pbrlib\surfaceshader\normalmapped_surfaceshader\normalmapped_surfaceshader_baked.mtlx
```
[...]

PR #963, opened 2022-05-25
Fix TextureBaker image filename bug

ashwinbhat:
Fix texturebaker regression from image filename hashing.
https://github.com/AcademySoftwareFoundation/MaterialX/issues/961


jstone-lucasfilm (approved):

PR #968, opened 2022-05-28
Update Worley noise to be backwards compatible with OSL v1.9.10

niklasharrysson:
This change list updates the `vector` accessor syntax from .x .y .z to subscript operator [], to be compatible with older OSL versions that don't support .x .y .z accessors. Since we state 1.9.10 as the minimum supported OSL version.


jstone-lucasfilm (approved):
Thanks for this update, @niklasharrysson!

PR #970, opened 2022-05-29
Add alpha to transparency detection

jstone-lucasfilm:
This changelist adds "alpha" to the list of shader inputs that are checked in transparency detection functions (e.g. mx::isTransparentSurface), making them inclusive of the conventions of the glTF PBR shading model.  Materials with non-zero or spatially-varying values for "alpha" will now be considered transparent by renderers that leverage these functions.

### DIFF
--- a/source/MaterialXGenShader/Util.cpp
+++ b/source/MaterialXGenShader/Util.cpp
@@ -91,6 +91,7 @@ namespace
         // Inputs on a surface shader which are checked for transparency
         const OpaqueTestPairList inputPairList = { {"opacity", 1.0f},
                                                    {"existence", 1.0f},
+                                                   {"alpha", 0.0f},
                                                    {"transmission", 0.0f} };
 
 


kwokcb:
Hi @jstone-lucasfilm I'm was thinking to add in the semantic checking that @dbsmythe and @niklasharrysson and I were discussing in the spec comments. I think this might be a cleaner chang than using input names. What do you think?

jstone-lucasfilm:
@kwokcb Agreed that transparency metadata is the right path forward in MaterialX 1.39, but we'll need a solution in the current codebase as well, in order to address issues such as https://github.com/PixarAnimationStudios/USD/issues/1882.

JGamache-autodesk:
I might be mistaken, but is it possible that the value should have been 1.0 instead of 0.0 since we want the fully opaque value, which is 1.0 for an alpha channel?

jstone-lucasfilm:
Agreed, @JGamache-autodesk, and I've updated this as part of @pablode's proposed fix to the glTF PBR default values (#971).

PR #972, opened 2022-05-29
Fix incorrect transform in genosl

jstone-lucasfilm:
This changelist updates the transform(matrix, vector4) method in genosl to match a corresponding fix in the OSL codebase.  See https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1513 for more details.


lgritz (approved):

rasmusbonnedal:
@jstone-lucasfilm 
I think this code depends on the old behaviour and will break with this change:
https://github.com/jstone-lucasfilm/MaterialX/blob/e47a3b3f14860d1b7e19fa49fa899fed973a73ea/libraries/stdlib/genosl/mx_rotate_vector3.osl#L18
https://github.com/jstone-lucasfilm/MaterialX/blob/e47a3b3f14860d1b7e19fa49fa899fed973a73ea/libraries/stdlib/genosl/mx_ap1_to_rec709_color3.osl#L5
https://github.com/jstone-lucasfilm/MaterialX/blob/e47a3b3f14860d1b7e19fa49fa899fed973a73ea/libraries/stdlib/genosl/mx_ap1_to_rec709_color4.osl#L5
https://github.com/jstone-lucasfilm/MaterialX/blob/e47a3b3f14860d1b7e19fa49fa899fed973a73ea/libraries/stdlib/genosl/mx_g22_ap1_to_lin_rec709_color3.osl#L8
https://github.com/jstone-lucasfilm/MaterialX/blob/e47a3b3f14860d1b7e19fa49fa899fed973a73ea/libraries/stdlib/genosl/mx_g22_ap1_to_lin_rec709_color4.osl#L8

kwokcb:
> 

Hi @rasmusbonnedal,
I replied in Slack, but yes, I believe these will break. Running the test suite for OSL rendering before and after should see what's changed (in case there is more).  Also I'd suggest we change these to be vec3 computations as AFAIK no input colorspace transform modifies the 4th channel (alpha) (at least not in the OCIO implementations).

rasmusbonnedal:
@kwokcb I agree! I searched for `transform` in all `*.{osl,h}` in the repo and these were all which uses vector4.

jstone-lucasfilm:
Attaching reference renders with the latest proposed changes:

[MaterialX_ReferenceRenders_05_30_2022.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/8800605/MaterialX_ReferenceRenders_05_30_2022.pdf)

libraries/stdlib/genosl/mx_ap1_to_rec709_color3.osl
```diff
@@ -1,7 +1,7 @@
 void mx_ap1_to_rec709_color3(color _in, output color result)
 {
     vector4 outColor = vector4(_in[0], _in[1], _in[2], 0.);
-    matrix m = transpose(matrix(1.705079555511475, -0.1297005265951157, -0.02416634373366833, 0., -0.6242334842681885, 1.138468623161316, -0.1246141716837883, 0., -0.0808461606502533, -0.008768022060394287, 1.148780584335327, 0., 0., 0., 0., 1.));
+    matrix m = matrix(1.705079555511475, -0.1297005265951157, -0.02416634373366833, 0., -0.6242334842681885, 1.138468623161316, -0.1246141716837883, 0., -0.0808461606502533, -0.008768022060394287, 1.148780584335327, 0., 0., 0., 0., 1.);
```
kwokcb, line 4: Hi @jstone-lucasfilm ,

Believe the recommendation here from @rasmusbonnedal is to make this 3x3 matrix and call the vec3 version of transform() which is built in.

As far as I know vec4 (alpha) is only used for a few output (display) transforms and no input transforms, which is the same as alpha=0 in a vec4 computation.

rasmusbonnedal:
I think rotate3d needs to add a transpose: https://github.com/jstone-lucasfilm/MaterialX/blob/e47a3b3f14860d1b7e19fa49fa899fed973a73ea/libraries/stdlib/genosl/mx_rotate_vector3.osl#L18

This is before the change in `transpose(matrix, vector4)`:
rotate3d_before_change

This is after:
rotate3d_after_change

jstone-lucasfilm:
@rasmusbonnedal I was curious about this as well, and I couldn't find an existing test material that would confirm this one way or another.  If you're confident that the original version was correct, and a transpose is now needed for parity, would you have the bandwidth to write up a pull request for this change to the MaterialX GitHub?

jstone-lucasfilm:
That's great news, @rasmusbonnedal, and thanks for confirming!

PR #973, opened 2022-05-31
RTS compare utility: Multiple folder comparison support

kwokcb:
- Add the ability to have difference source folders for different source languages. This allows for the same or different languages to be compared against different runs. Note that the folder structure must still be the same. e.g. can compare OSL in one build with one renderer vs OSL in another build with another renderer.
- Cleanup so that destination files will just try and match source ones to avoid file name mismatch comparisons.


jstone-lucasfilm (approved):
Looks good to me, @kwokcb, and thanks for this update!

Issue #864, opened 2022-03-10
Implementation of thin-walled / double-sided materials


klucknav:
It looks like the only implementation for `thin_surface` nodes is in MDL. Missing both a GLSL and OSL implementation. Unfortunately, without the GLSL implementation, we cannot use this node in either MaterialX View or Usdview with Storm. And the OSL implementation would be needed to use with Usdview with Prman.

niklasharrysson:
Hi @klucknav, I just want to add some context on this.

We have refrained from working on an implementation of `thin_surface` so far, because its definition is debatable. When we defined the PBR library it seemed like a natural way to separate closed surfaces with volume (the `surface` node) from thin-walled surfaces (the `thin_surface` node). But in hindsight it's not a good approach.

The first problem is that there is no natural way to have a user controlled switch between the two, for example as needed for an uber shader. You could use a surfaceshader `mix` node to switch between the one or the other, but then you introduce the concept of having a surface that could be 50% closed and 50% thin-walled, which doesn't makes sense.

In addition, we now have the "unlit" version of `surface`, which then by definition becomes a closed volume unlit surface. So we would then need an "unlit" version of `thin_surface` as well, which is quite messy.

So we have come to the conclusion that we need a better way to define closed vs. thin-walled surfaces. A proposal we are considering is to move the definition of closed vs thin-walled to the `material` level. Adding inputs on the `surfacematerial` constructor node for both frontfacing and backfacing surfaceshaders, as well as a boolean switch to say if it's thin-walled or not.

The `thin_surface` node would then be removed and replaced with this more robust definition.

niklasharrysson:
Here is a proposal to discuss for the `surfacematerial` node to make it support thin-walled / double-sided materials:

```
  <!-- Surface material -->
  <nodedef name="ND_surfacematerial" node="surfacematerial" nodegroup="material">
    <input name="surface" type="surfaceshader" value="" />
    <input name="backface" type="surfaceshader" value="" />
    <input name="thin_walled" type="boolean" value="false" />
    <input name="displacement" type="displacementshader" value="" />
    <output name="out" type="material" />
  </nodedef>
```

If `thin_walled` is enabled the material is thin-walled and double-sided, where `surface` and `backface` defines the surfaceshaders on each side respectively. If `thin_walled` is disabled (the default) the material is a closed surface with an interior volume, where `surface` defines the single surfaceshader and `backface` is ignored.

[...]

Issue #620, opened 2021-05-04
Material isn't getting applied


numega:
I believe the following file should lead to a green object in MaterialXView, but right now I get a black object. Additionally, I see the following output: "OpenGL error after program unbind geometry: 1282". If I replace the `interfacename="colorA"` and `interfacename="colorB"` lines with a value then everything shows up properly.

```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">
    <nodedef name="NG_mymaterial" node="mymaterial">
        <input name="colorA" type="color3" value="0, 0.5, 0"/>
        <input name="colorB" type="color3" value="0, 0.5, 0"/>
        <output name="out" type="material"/>
    </nodedef>
    <nodegraph name="blah" nodedef="NG_mymaterial">
        <add name="add1" type="color3">
            <input name="in1" type="color3" interfacename="colorA"/>
            <input name="in2" type="color3" interfacename="colorB"/>
        </add>
        <UsdPreviewSurface name="SR_default" type="surfaceshader">
            <input name="diffuseColor" type="color3" nodename="add1"/>
        </UsdPreviewSurface>
        <surfacematerial name="m1" type="material">
            <input name="surfaceshader" type="surfaceshader" nodename="SR_default"/>
        </surfacematerial>
        <output name="out" type="material" nodename="m1"/>
    </nodegraph>
    <mymaterial name="Material1" type="material"/>
</materialx>
```

numega:
The OpenGL error is due to GlslProgram::unbindGeometry() being called for each assignment of the material. You can reproduce this issue with the shaderball.obj file in resources/Geometry. As a workaround I added the following at the beginning of the function:

```
void GlslProgram::unbindGeometry()
{
    if (_vertexArray == GlslProgram::UNDEFINED_OPENGL_RESOURCE_ID)
        return;

    glBindVertexArray(_vertexArray);

```

bernardkwok:
What you have should show up green without modification. 

From an initial look, it appears to be an issue with having the `<surfacematerial>` in the nodegraph definition not picking up the default values properly for code generation. 

For example this does pick up default values properly:
```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">
  <nodedef name="NG_mymaterial" node="mymaterial">
    <input name="colorA" type="color3" value="0, 0.5, 0" />
    <input name="colorB" type="color3" value="0, 0.5, 0" />
[...]

PR #981, opened 2022-06-06
Support for code generation from material nodes.

niklasharrysson:
This change list adds support for generating shaders directly from `material` type nodes. Previously material nodes were terminals from which we first extracted shader nodes to generate code from. Now we can generate code from these nodes directly and include custom logic for them.

This change is foundational work towards having support for double-sided materials, which will be added in a future change list (issue tracked here https://github.com/AcademySoftwareFoundation/MaterialX/issues/864).


niklasharrysson:
Test suite renderings for GLSL and OSL after the change.
[tests_new_compressed.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/8846257/tests_new_compressed.pdf)

source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
```diff
@@ -40,7 +40,7 @@ class TangentOsl : public mx::ShaderNodeImpl
         BEGIN_SHADER_STAGE(stage, mx::Stage::PIXEL)
             shadergen.emitLineBegin(stage);
             shadergen.emitOutput(node.getOutput(), true, false, context, stage);
-            shadergen.emitString(" = normalize(vector(N.z, 0, -N.x))", stage);
+            shadergen.emitString(" = normalize(vector(N[2], 0, -N[0]))", stage);
```
niklasharrysson, line 43: Fix to support OSL back to 1.9.10.

source/MaterialXView/Viewer.cpp
```diff
@@ -1223,19 +1223,7 @@ void Viewer::loadDocument(const mx::FilePath& filename, mx::DocumentPtr librarie
         {
             mx::TypedElementPtr renderableElem = elem;
             mx::NodePtr node = elem->asA<mx::Node>();
-            if (node && node->getType() == mx::MATERIAL_TYPE_STRING)
```
niklasharrysson, line 1226: No need for special treatment of material nodes now.

jstone-lucasfilm:
This looks like a great step forward, thanks @niklasharrysson!  I should note that it appears to successfully address report #620, which was written up last year by @numega.  I'll plan to take a closer look and merge the change if no issues are found.

jstone-lucasfilm:
Updated reference renders, with transparency on BrassWireMesh material:
[MaterialXReferenceRenders_06_07_2022.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/8857070/MaterialXReferenceRenders_06_07_2022.pdf)

jstone-lucasfilm (approved):

numega:
Awesome! Thanks for taking care of this @niklasharrysson

Issue #964, opened 2022-05-25
LightCompound shaders produce invalid code


kwokcb:
This is spawned from issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/960.

The sample light compound node is being added as a library definition for shader generation tests.
However it currently produces invalid code when the LightNodeGlsl function emitter is invoked. It produces duplicate code as well as uses undefined variables.

 It appears to be from this code refactor:
 https://github.com/AcademySoftwareFoundation/MaterialX/pull/747
 
 The initial triage analysis is as follows:
 
1. The light compound node has been generalized to be callable from different contexts if required.
2. The logic has been generalized to push and pop contexts as needed. The compound node will emit texture calls and then light shader calls within each required context.
3. In the current case it's the default context. (0)
4. When a light node is visited to emit it's function, it is first added to the dictionary of emitted calls for the given context (0).
5. Then the code is emitted for light direction, falloff etc.
6. From within the light node emit call there is call to emit the node again. What used to occur is that there was a specific [iteration of the graph](https://github.com/AcademySoftwareFoundation/MaterialX/pull/747/files#diff-f97fcd65b701d9df5db9d33be293486728647a235b2abb1fe929c898995a3463L57-L59) to emit calls in order with the last result being stored into a special "emission" variable.
7. This no longer occurs instead the node is called to emit itself again and thus causes duplicate code to be emitted -- this is due to pushing an "emission" context (3). It's only done one more time as right after the function emission is added to the dictionary to avoid repeating again, otherwise it could recurse infinitely.
8. As no output variable is created to contain the final result, the call to use this variable will fail.

So the question is what is the fix ? I think part of the reasoning for 7 is to not special case anymore but the logic is not correct as it's recursing on itself and due to filtering to only visit light shader "roots" in 2) not all nodes in the graph would be visited (I think).

(I have left a branch [here](https://github.com/AcademySoftwareFoundation/MaterialX/compare/main...kwokcb:lightcompound_debug?expand=1) with my investigation / "hack" to make it output something valid.)

If light nodes are not of importance than it is still advised to not include this as part of the default library definition as it can be emitted if direct lighing is enabled.

PR #984, opened 2022-06-08
Fix regressions with light shader node in GLSL

niklasharrysson:
This change list fixes a regression with light shader generation for GLSL. This should fix issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/964.

Thanks @ashwinbhat for finding the issue and @bernardkwok for logging it.


jstone-lucasfilm (approved):

PR #990, opened 2022-06-14
Add look support to Web MaterialXView

kwokcb:
## Features
* Parse looks found in MaterialX files and use this to perform assignment. In the case where no assignment matches any geometry in the scene then the first material is assigned to the entire scene. Supports rudimentary regexpr string matching of names.
* Keep track of material assignments and compile unique set of shaders once and assign after 1) material load, 2) geometry change. The latter allows for material to be loaded first and geometry second to have the same result as the reverse.
* Will load UI for all materials found via looks. Add back in "hide" controls as the material list can take up more space.

## Results
https://user-images.githubusercontent.com/49369885/173661975-1f73147d-64ea-4517-8887-5c7fc9e72302.mp4

## FIxes / Updates
* Remove hard-coded hacks to re-path images only to the Images/ folder. Add rudimentary local search pathing for textures.  Required to support MTLX file local texture folders such as found for the chessboard and boombox.
* Chessboard, wood and brass tiled files with looks now have proper assignments.
* Fix image installation to install all files properly based on original location.
* Add a glTF sample file which has materials converted to MTLX and uses collections which was never tested in the viewer.
* Fix material dropdown filtering to only include files with a ".mtlx" suffix to filter out image sub-folders.
* Add additional testing for looks which was missing to ensure the api is robust. 
* Add in findRenderableMaterials()  exposure to be able to get back all material nodes in a document.

## Found (leave to be resolved as separate issues)
* Spurious divide by zero errors for some shaders. (Issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/992)
* Some placement issue with image calibration shader. (Issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/991)


javascript/MaterialXView/source/viewer.js
```diff
@@ -79,6 +83,7 @@ export class Scene
         // Always reset controls based on camera for each load. 
         orbitControls.reset();
         this.updateScene(viewer, orbitControls);
+        viewer.getMaterial().updateMaterialAssignments(viewer);
```
kwokcb, line 86: @jstone-lucasfilm, This makes it possible to load geometry first or corresponding materials first and get the same result. IMHO this is better than the desktop version
which uses the current "active" material to assign to loaded geometry. Also there is no concept of "active" material since all materials are accessible at the same time.

javascript/MaterialXView/source/viewer.js
```diff
@@ -158,17 +161,65 @@ export class Scene
...
+
         const scene = this.getScene();
         const camera = this.getCamera();
         scene.traverse((child) => {
-            if (child.isMesh) {
-                child.material = material;
-                child.material.needsUpdate = true;
+            if (child.isMesh) 
+            {
+                const dagPath = this.getDagPath(child);
```
kwokcb: I will keep dag path support a separate update for now as first the glTF files needs to be fixed to have pathing which match the materialassign notation and hierarchy, and the glTF loader for desktop needs to also set proper dag path names.

javascript/MaterialXView/source/viewer.js
```diff
@@ -360,33 +461,135 @@ export class Material
...
+        for (let matassign of this._materials)
+        {
+            if (matassign.getShader())
+            {
+                assigned += viewer.getScene().updateMaterial(matassign);
+            }
+        }
+        if (assigned == 0 && this._materials.length)
+        {
+            this._defaultMaterial = new MaterialAssign(this._materials[0].getMaterial(), ALL_GEOMETRY_SPECIFIER);
```
kwokcb, line 585: Assign to first material if no material assignments match geometry. Similar to desktop except desktop has concept of "active" material which does not exist here.

source/MaterialXTest/MaterialXFormat/XmlIo.cpp
```diff
@@ -223,6 +223,22 @@ TEST_CASE("Load content", "[xmlio]")
     mx::readFromXmlFile(parentDoc, "resources/Materials/TestSuite/libraries/metal/brass_wire_mesh.mtlx", searchPath);
     REQUIRE(parentDoc->getNodeDef("ND_TestMetal") != nullptr);
 
+    // Read a string with a relative Xinclude
```
kwokcb, line 226: Some testing which was remove I think when the Syntax directory was removed a while back. Add back in searchpath usage for xincludes when parsing a document which is a string / stream (which is used for Web viewer).

jstone-lucasfilm:
@kwokcb This change looks really promising, and I think it's definitely worthwhile to include in the upcoming 1.38.5.

Can you explain in more detail the motivation for modifying `shaderball.glb`?  Since this is our default geometry across MaterialX viewers, I'd like to make sure I fully understand any proposed changes to it, and we should be careful not to break the assumptions of external tools leveraging the same model.

jstone-lucasfilm:
@kwokcb It sounds wise to revert this change for now, so that we're minimizing new changes in preparation for the 1.38.5 release.  In the future, let's take a closer look at the scene pathing in our GLB files, and come up with a consistent convention that aligns with usage in other projects.

jstone-lucasfilm (approved):

PR #995, opened 2022-06-17
Moved setConnectedOutput from Input to PortElement

igor-elovikov:
From the slack discussion:

> Currently `setConnectedOutput`  is implemented only in `Input`  but not in `Output` meaning that if you want to connect some named output (from node that have several output connections) to the graph output you just can't do that. Only with some workaround and pass through node (like `add` with zero) because you can indicate specific output on the node input.

After browsing the source (originally I just used prebuilt python bindings) I've also discovered that I can use `Output::setOutputString` but still it's probably better if those functions are available on `PortElement`

### DIFF
--- a/source/MaterialXCore/Interface.cpp
+++ b/source/MaterialXCore/Interface.cpp
@@ -65,6 +65,95 @@ NodePtr PortElement::getConnectedNode() const
     return graph ? graph->getNode(getNodeName()) : nullptr;
 }
 
+void PortElement::setConnectedOutput(ConstOutputPtr output)
+{
+    if (output)
+    {
+        setOutputString(output->getName());
+        ConstElementPtr parent = output->getParent();
+        if (parent->isA<NodeGraph>())
+        {
+            setNodeGraphString(parent->getName());
+        }
+        else
+        {
+            removeAttribute(NODE_GRAPH_ATTRIBUTE);
+        }
+    }
+    else
+    {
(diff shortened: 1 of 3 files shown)


jstone-lucasfilm (approved):
Looks good to me, thanks @igor-elovikov!

source/MaterialXCore/Interface.cpp
```diff
@@ -65,6 +65,95 @@ NodePtr PortElement::getConnectedNode() const
     return graph ? graph->getNode(getNodeName()) : nullptr;
 }
 
+void PortElement::setConnectedOutput(ConstOutputPtr output)
```
kwokcb, line 68: I forgot to leave a comment. This isn't exactly "safe" as if there is a connection to a node only `nodename` is note cleared leave a possible invalid input->node or output->node connection to the first output on the upstream node.

kwokcb: Also moving this to PortElement, I'm not sure if there is specific configurations of output->output connections which should not be allowed (type is not checked here).

jstone-lucasfilm: @kwokcb Feel free to propose improvements to this, and my own recommendation would be to check for invalid configurations at document validation time (e.g. in Element::validate and its specializations), rather than aiming to enforce correctness at document construction time.

PR #1001, opened 2022-06-21
Add in flipV on geometry support

### LINKED ISSUE #991: Calibration shader texture placement incorrect in Web MaterialXView
The placement of the texture is incorrect when you use: 
`standard_surface_greysphere_calibration.mtlx`
in the web viewer.

Most likely it is missing explicit binding of some parameters as one key difference is that there is no
default value binding for ESSL.

Current incorrect result:
<image src="https://user-images.githubusercontent.com/49369885/173891907-12535bf8-18d5-448b-b580-3603bea1fba9.png" width=320>

## Initial Triage:
* Seems the bindings are occuring properly.
* The issue appears with V direction is negated for image `offset` as shown in snapshot.
Screen Shot 2022-06-15 at 11 05 20 PM

kwokcb:
Fixes #991.

* Add in a flip "V" for geometry load. This is a generally useful option.
* The option is currently always enabled for geometry which is loaded. This is to be consistent in orientation with desktop and orientation used for 2d placement.
* Images when loaded to GPU have an option to be flipped. This is now enabled based on the geometry orientation option (which is true).

### Results (Desktop on left)
web_desktop_orentation_fix2
web_desktop_orentation_fix1

### DIFF
--- a/javascript/MaterialXView/source/helper.js
+++ b/javascript/MaterialXView/source/helper.js
@@ -126,7 +126,7 @@ function fromMatrix(matrix, dimension)
  * @param {mx.Uniforms} uniforms
  * @param {THREE.textureLoader} textureLoader
  */
-function toThreeUniform(type, value, name, uniforms, textureLoader, searchPath)
+function toThreeUniform(type, value, name, uniforms, textureLoader, searchPath, flipY)
 {
     let outValue;  
     switch (type)
@@ -159,7 +159,7 @@ function toThreeUniform(type, value, name, uniforms, textureLoader, searchPath)
                 let fullPath = searchPath + IMAGE_PATH_SEPARATOR + value;
                 const texture = textureLoader.load(fullPath);
                 // Set address & filtering mode
-                setTextureParameters(texture, name, uniforms);
+                setTextureParameters(texture, name, uniforms, flipY);
                 outValue = texture;
             } 
             break;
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm:
@kwokcb It's exciting to see textured rendering working in the Web Viewer, but why would this flip only be needed in JavaScript, and only specifically for the shaderball mesh?  If the problem were specific to the shaderball, then I'd expect the same flip to be required in C++.  And if the problem were specific to JavaScript, then I'd expect it to be required for all meshes, not just this one example.

kwokcb:
> @kwokcb It's exciting to see textured rendering working in the Web Viewer, but why would this flip only be needed in JavaScript, and only specifically for the shaderball mesh? If the problem were specific to the shaderball, then I'd expect the same flip to be required in C++. And if the problem were specific to JavaScript, then I'd expect it to be required for all meshes, not just this one example.

I did some more investigation and all geometry needs to be flipped to be consistent with desktop, but also all textures need to be flipped as well.

Desktop always flips V for geometry and assumes bottom-to-top for shader and texture orientation.
Web (maybe just Three only) flips textures on load (https://threejs.org/docs/#api/en/textures/Texture.flipY) so is inconsistent.

i.e. Seems to boil down to having a consistent orientation convention for:
1. Shader code
2. Geometry
3. Textures

jstone-lucasfilm (approved):
This looks good to me, @kwokcb, and thanks for tracking this issue to its source!

Issue #827, opened 2022-02-01
Proposal: Add Support for Vulkan


kwokcb:
# Proposal For Vulkan Support In MaterialX

# Shader Generation Ecosystem
The following shows how Vulkan code generation would fit into the current eco-system (at time of writing). Note that the intention is to reuse current GLSL code with customizations provided by a new Vulkan code generator classs which is derived from the base GLSL generator.
![image](https://user-images.githubusercontent.com/49369885/152036774-344f677f-956e-41ec-b1fb-c167d96027f3.png)

# Code Generation / Rendering Support Timeline
The following is a possible timeline. 
* The next patch release (at time of writing is 1.38.4) which would be the target for creating a new generator along with appropriate source code validation.
* Naturally rendering would come later with a possible target of a patch release or as part of the 1.39 release.
* The initial goal is only to support code generation for a rasterizer and a simple sample renderer to use this code. One main reason is that the current GLSL code emits code which only works for rasterization. _It is to be determined what the appropriate generation will be for a path tracer_.
* There will be no new source code implementations required. Instead a new code generator will emit the appropriate variation of GLSL required for Vulkan compliance. This is similar to how the existing ESSL generator is structured.
* This means that there are _no additional tooling dependencies_ such as SPIRV for code generation.  
* The tooling for code syntax validation will remain `glslValidator`. This is currently part of the the CI workflow and as desired will be an option for local unit test validation (unit test suite validation is currently too expensive to run on CI).
* _OCIO_ support in MaterialX is a related dependency, for which Vulkan support adds an additional requirement. This support will not be part of this proposal, except to note that the base colorspace transforms provided as part of the MaterialX core distribution should be supported.
![image](https://user-images.githubusercontent.com/49369885/152037234-2a71b681-31a0-4a9e-87c0-cb9a06db963e.png)

# Details:
## Handling Vulkan GLSL

- KHR_vulkan_glsl extension on OpenGL GLSL
- Minor differences in functions, main changes in input and resource binding.
- Default uniforms not allowed
  - e.g., uniform float base=1.0 is not allowed
- Use layout identifier 
  - e.g., layout (set=M, binding=N) uniform sampler2D variableName;
- Use of sampler2D replaced by sampler and texture2D with separate binding points
- Uniform Buffers to group uniforms by access and frequency of update
- Advanced usage (for future exploration)
-  Storage Images (Image2d)
  -  Push constants
  -  Descriptor Sets
  -  Storage buffers

[...]

PR #1002, opened 2022-06-22
Vulkan GLSL shader generation

ashwinbhat:
Vulkan Shaders use SPIRV but they are written using a variant of GLSL. 
This variant of GLSL that conforms to Vulkan specification  has a few semantic differences. see [KHR_vulkan_glsl](https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_vulkan_glsl.txt)

This new VkShaderGenerator builds upon existing GLSLShaderGenerator and add Vulkan semantics in the generated GLSL code. 

The MaterialX Test code exercises the VkShaderGenerator. In a future change, we will introduce validation and test rendering. 

Partially addresses https://github.com/AcademySoftwareFoundation/MaterialX/issues/827


jstone-lucasfilm (approved):

PR #1004, opened 2022-06-23
Fix concatenation of FilePath for concurrent "."

### LINKED ISSUE #1003: Regression: GLSL validation test is failing in CI
The shader generation validation test for GLSL appears to be failing when using the Python script used to generate code.
This appears to be a regression. Though in CI, the failure does not mark the run as failing (to be fixed).
https://github.com/kwokcb/MaterialX/runs/7011955715?check_suite_focus=true

The basic command to test is:
`python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx --path . --target glsl `

If you have glslValidator installed it will catch the error (as occurs in CI)
`python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx --path . --target glsl --validator glslangValidator.exe"`

It appears as though `lib/mx_microfacet.glsl` file is included twice resulting in duplicate definitions.

This does work in the viewer (desktop or web) and C++ RTS so it could be the setup done in the Python script differs in some way
 from the C++ version. 
 
 As a patch, the setup for the Python script could be changed but this is a regression as this used to work as is.

kwokcb:
Fixes #1003

## Change

* Add in new `FilePath::getNormalized()` which will normalize and return a new `FilePath`. For now only sequential "." paths are removed. Possible additions include collapsing intermediate ".." paths.
* This means that we no longer have a  mismatch between "./<foo>" and "././<foo>" as they are both the same paths
but just happen to hold different path strings. New test added for this.
* Added to `GenContext::resolveSourceFile()` to avoid duplicate source code includes during shader generation as the `FilePath` is used as the unique dictionary key.
* Added to `FileFormat flattenFilenames()` utility to make to closer to a true "flatten".

## Updated results
```
python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path . --target glsl --validator glslangValidator.exe --vulkanGlsl True --validatorArgs="-V --aml"
- Set up CMS ...
- Set up Units ...
- Shader output path: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurface
-- Generate code for node: SR_brass1
--- Wrote pixel shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.glsl.frag
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.glsl.frag -V --aml
--- Wrote vertex shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.glsl.vert
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.glsl.vert -V --aml
Validation passed for shader:  SR_brass1
-- Generate code for node: SR_Greysphere_Calibration
--- Wrote pixel shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.glsl.frag
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.glsl.frag -V --aml
--- Wrote vertex shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.glsl.vert
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.glsl.vert -V --aml
Validation passed for shader:  SR_Greysphere_Calibration
```
and
```
python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path . --target essl --validator glslangValidator.exe
- Set up CMS ...
- Set up Units ...
- Shader output path: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurface
-- Generate code for node: SR_brass1
--- Wrote pixel shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.essl.frag
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.essl.frag
--- Wrote vertex shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.essl.vert
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_brass1.essl.vert
Validation passed for shader:  SR_brass1
-- Generate code for node: SR_Greysphere_Calibration
--- Wrote pixel shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.essl.frag
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.essl.frag
--- Wrote vertex shader to: D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.essl.vert
----- Run: glslangValidator.exe D:\Work\materialx\bernard_MaterialX\resources\Materials\Examples\StandardSurfaceSR_Greysphere_Calibration.essl.vert
Validation passed for shader:  SR_Greysphere_Calibration
```


jstone-lucasfilm:
@kwokcb Thanks for this changelist, and it's definitely a step in the right direction.

Instead of modifying the rules for `FilePath` concatenation, though, I would recommend that we create an initial `FilePath::normalize` method that merges all sequential `.` paths, and we can then call this method in the context of deduplicating shader includes.  This would make for a better separation between FilePath concatenation (which a user wouldn't expect to apply contextual rules to the original path), and FilePath normalization, which is expected to understand and apply those rules on demand.

This `FilePath::normalize` method could then be extended over time to handle common cases such as `..` parent paths, with the same understanding that the user needs to invoke this method when they want contextual rules to be applied.

kwokcb:
> @kwokcb Thanks for this changelist, and it's definitely a step in the right direction.
> 
> Instead of modifying the rules for `FilePath` concatenation, though, I would recommend that we create an initial `FilePath::normalize` method that merges all sequential `.` paths, and we can then call this method in the context of deduplicating shader includes. This would make for a better separation between FilePath concatenation (which a user wouldn't expect to apply contextual rules to the original path), and FilePath normalization, which is expected to understand and apply those rules on demand.
> 
> This `FilePath::normalize` method could then be extended over time to handle common cases such as `..` parent paths, with the same understanding that the user needs to invoke this method when they want contextual rules to be applied.

@jstone-lucasfilm,
I was wondering which approach you'd prefer. Let's go with this. I'm wondering if "resolve" would be a be better 
interface name, or "flatten". I think these match other places where path resolution is handled ?  
Also for the method, I assume it can follow the same signature of returning a new FilePath instead of working in-place.

kwokcb:
Reading through `canonical`, and `normpath` (from [Python](https://docs.python.org/3/library/os.path.html#os.path.normpath))
I think `normalize()` is a good fit for what it does. When absolute path conversion is supported then `canonical` could be `normalize` + `absolute` path conversion.

source/MaterialXFormat/File.cpp
```diff
@@ -122,10 +122,17 @@ FilePath FilePath::operator/(const FilePath& rhs) const
         throw Exception("Appended path must be relative.");
     }
 
+    const string DOT_PATH = ".";
```
jstone-lucasfilm, line 125: Can we now remove this special handling of dot paths in FilePath concatenation, since the client can normalize their paths on demand?

source/MaterialXFormat/File.cpp
```diff
@@ -141,6 +148,32 @@ bool FilePath::exists() const
 #endif
 }
 
+FilePath FilePath::normalize() const
```
jstone-lucasfilm, line 151: This naming convention looks good to me, but perhaps we should name the method "getNormalized" as we do for vectors, making it clear that the original FilePath is left unmodified.

source/MaterialXFormat/Util.cpp
```diff
@@ -172,6 +172,7 @@ void flattenFilenames(DocumentPtr doc, const FileSearchPath& searchPath, StringR
                 for (size_t i = 0; i < searchPath.size(); i++)
                 {
                     FilePath testPath = searchPath[i] / resolvedValue;
+                    testPath.normalize();
```
jstone-lucasfilm, line 175: Wouldn't we need to call `testPath = testPath.getNormalized()` to actually affect the result?

jstone-lucasfilm (commented):
This looks really promising, and I've added a few thoughts and recommendations.

source/MaterialXFormat/Util.cpp
kwokcb: Typo fixed.

source/MaterialXFormat/File.cpp
kwokcb: Yes. Removed now.

python/Scripts/generateshader.py
```diff
@@ -43,13 +40,11 @@ def main():
...
-        print(err)
-        sys.exit(0)
+        print('Generation failed: "', err, '"')
+        sys.exit(-1)
 
     stdlib = mx.createDocument()
-    filePath = os.path.dirname(os.path.abspath(__file__))
-    searchPath = mx.FileSearchPath(os.path.join(filePath, '..', '..', 'libraries'))
-    searchPath.append(os.path.dirname(opts.inputFilename))
+    searchPath = mx.FileSearchPath()
```
kwokcb, line 47: Leave it up to the user to set the search path.

source/MaterialXFormat/File.cpp
```diff
@@ -141,6 +141,32 @@ bool FilePath::exists() const
 #endif
 }
 
+FilePath FilePath::getNormalized() const
+{
+    FilePath path(*this);
```
jstone-lucasfilm, line 146: What's the motivation for copy constructing the result path from the original, since it will be cleared below?  Perhaps we can just use a default constructor below?

source/MaterialXTest/MaterialXFormat/File.cpp
```diff
@@ -141,3 +141,22 @@ TEST_CASE("Flatten filenames", "[file]")
     resolvedPathString = image2->getInputValue("file")->getValueString();
     CHECK(resolvedPathString == (rootPath / TEST_FILE_PREFIX_STRING / TEST_IMAGE_STRING2).asString(mx::FilePath::FormatPosix));
 }
+
+TEST_CASE("Path normalizeation test", "[file]")
```
jstone-lucasfilm, line 145: Minor spelling note: "normalization"

jstone-lucasfilm (commented):
This is looking good, and I had just a few additional recommendations.

source/MaterialXFormat/File.cpp
kwokcb: This is to copy over any other data members including type. I was thinking of explicitly copying over type, but this seems safer in case there are any changes to FilePath members in the future.

.github/workflows/main.yml
```diff
@@ -217,8 +217,8 @@ jobs:
       run: |
         vcpkg/vcpkg install glslang --triplet=x64-windows
         glslangValidator.exe -v
-        python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx --path . --target glsl --validator glslangValidator.exe --vulkanGlsl True --validatorArgs="-V --aml"
-        python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx --path . --target essl --validator glslangValidator.exe
+        python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path . --target glsl --validator glslangValidator.exe --vulkanGlsl True --validatorArgs="-V --aml"
```
kwokcb, line 220: Change to make sure that relative search path to file is tested.

source/MaterialXFormat/File.cpp
jstone-lucasfilm: Ah, good point, and this choice makes sense to me.

jstone-lucasfilm (approved):
This looks good to me, @kwokcb, and thanks for the fix!

PR #1009, opened 2022-06-26
Improvements to FilePath::getNormalized

jstone-lucasfilm:
- Add support for parent path references.
- Extend unit tests to cover additional cases.
- Clarify documentation.

### DIFF
--- a/source/MaterialXFormat/File.cpp
+++ b/source/MaterialXFormat/File.cpp
@@ -39,7 +39,8 @@ const string VALID_SEPARATORS = "/\\";
 const char PREFERRED_SEPARATOR_WINDOWS = '\\';
 const char PREFERRED_SEPARATOR_POSIX = '/';
 
-const string DOT_STRING = ".";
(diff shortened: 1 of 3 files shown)


source/MaterialXFormat/File.cpp
```diff
@@ -137,10 +138,15 @@ FilePath FilePath::getNormalized() const
     FilePath res;
     for (const string& str : _vec)
     {
-        if (!res.isEmpty() && res[res.size() - 1] == DOT_STRING && str == DOT_STRING)
+        if (str == CURRENT_PATH_STRING)
         {
             continue;
         }
+        if (str == PARENT_PATH_STRING && !res.isEmpty() && res[res.size() - 1] != PARENT_PATH_STRING)
```
kwokcb, line 145: Looks great. Thanks for adding this in. Should help with deployments which don't like ".." specifies. Believe at least I think it will help MDL.

kwokcb (approved):

PR #1013, opened 2022-06-28
Logic fixes for output connections

### LINKED ISSUE #998: Removal of connections will not clear nodename correctly
The set method to clear the output on a PortElement should have it's logic fixed to remove
any `nodename` on the port the same way `nodegraph` is cleared.

This is a minor logic fix which was not covered when the "get" code was enhanced.

kwokcb:
Fixes #998

- Change `resolveRootNameReference()` to be resolveNameReference() which allows for an optional root to be used. Used to find sibling references for updated connection logic.
- Patch up logic for `setConnectedOutput()` to clear the appropriate nodegraph / nodename on an input/output.
- Clean up logic for `getConnectedOutput()` which could miss connections on sibling nodes, not in or on a nodegraph.
- Minor shader translator test and exception cleanup


source/MaterialXCore/Interface.cpp
```diff
@@ -92,10 +95,19 @@ OutputPtr PortElement::getConnectedOutput() const
...
+    // Want to find a node or nodegraph at the same level as the 
+    // parent element to this port. 
+    ConstElementPtr parent = getParent();
+    ConstElementPtr root = parent ? parent->getParent() : nullptr;
+
     // Look for an output in a nodegraph
     if (hasNodeGraphString())
     {
-        NodeGraphPtr nodeGraph = resolveRootNameReference<NodeGraph>(getNodeGraphString());
+        NodeGraphPtr nodeGraph = resolveRootNameReference<NodeGraph>(getNodeGraphString(), root);
```
kwokcb, line 106: Check at sibling node level for node/nodegraph instead of just checking at the root which isn't actually correct as you could inadvertently find a node/nodegraph 2 levels removed. Though nested nodegraphs are not supported yet, this code should still behave correctly when it is.

source/MaterialXCore/Interface.cpp
```diff
@@ -92,10 +95,19 @@ OutputPtr PortElement::getConnectedOutput() const
...
+        if (!nodeGraph)
+        {
+            nodeGraph = resolveRootNameReference<NodeGraph>(getNodeGraphString());
```
kwokcb, line 109: If no sibling found check at root.

source/MaterialXCore/Interface.cpp
```diff
@@ -116,37 +128,29 @@ OutputPtr PortElement::getConnectedOutput() const
     else if (hasNodeName())
     {
         const string& nodeName = getNodeName();
-        ConstElementPtr startingElement = getParent();
-        if (startingElement)
+        NodePtr node = resolveRootNameReference<Node>(nodeName, root);
```
kwokcb, line 131: This now just mirrors the nodegraph code as the only difference is that it's looking for a node vs nodegraph. Previous code would miss siblings outside of nodegraphs.

source/MaterialXGenShader/ShaderTranslator.cpp
```diff
@@ -144,15 +144,18 @@ void ShaderTranslator::connectTranslationOutputs(NodePtr shader)
...
+        if (!shader)
     {
         return;
     }
 
     const string& sourceCategory = shader->getCategory();
     if (sourceCategory == destCategory)
     {
-        throw Exception("The source shader \"" + shader->getNamePath() + "\" category is already \"" + destCategory + "\"");
+        return;
```
kwokcb, line 158: Remove so as to not throw as it's not invalid. There is just nothing to do.
Removes false unit test failures if you run the tests twice as the translated file will throw here and fail the test.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -180,14 +180,17 @@ TEST_CASE("GenShader: Shader Translation", "[translate]")
...
         try
         {
             shaderTranslator->translateAllMaterials(doc, "UsdPreviewSurface");
+            translated = true;
         }
         catch (mx::Exception &e)
         {
             std::cout << "Failed translating: " << (testPath / mtlxFile).asString() << ": " << e.what() << std::endl;
         }
+        REQUIRE(translated);
```
kwokcb, line 193: File could be valid but not translated. Catch this now.

source/MaterialXCore/Interface.cpp
```diff
@@ -116,37 +128,29 @@ OutputPtr PortElement::getConnectedOutput() const
...
-            if (startingElement->isA<NodeGraph>())
-            {
-                startingElement = startingElement->getParent();
-            }
-            if (startingElement)
+            node = resolveRootNameReference<Node>(nodeName);
+        }
+        if (node)
+        {
+            std::vector<OutputPtr> outputs = node->getOutputs();
```
kwokcb, line 138: This logic has not changed. As an explicit output is requested and if the output is implicit it will not be found.

source/MaterialXCore/Element.h
```diff
@@ -761,14 +761,16 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
     /// name, and attributes.
     string asString() const;
 
-    /// Resolve a reference to a named element at the root scope of this document,
+    /// Resolve a reference to a named element under a given root. If
+    /// If no root is provided then the root scope of this document is used,
     /// taking the namespace at the scope of this element into account.
-    template <class T> shared_ptr<T> resolveRootNameReference(const string& name) const
+    template <class T> shared_ptr<T> resolveRootNameReference(const string& name, ConstElementPtr root = nullptr) const
```
jstone-lucasfilm, line 767: Now that this method supports name resolution at arbitrary scopes, we should likely choose a new method name that isn't restricted to root scope.  Perhaps something along the lines of the following:

```
/// Resolve a reference to a named element at the scope of the given parent,
/// taking the namespace at the scope of this element into account.  If no parent
/// is provided, then the root scope of the document is used.
template <class T> shared_ptr<T> resolveNameReference(const string& name, ConstElementPtr parent= nullptr) const
```

jstone-lucasfilm (commented):
This looks promising, thanks @bernardkwok, and I have one recommendation related to method names.

source/MaterialXCore/Element.h
kwokcb: This sounds good. I've made the change. This makes me think to keep a note to see if other places can be made more robust by not always assuming root.

source/MaterialXCore/Element.h
```diff
@@ -761,14 +761,16 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
     /// name, and attributes.
     string asString() const;
 
-    /// Resolve a reference to a named element at the root scope of this document,
-    /// taking the namespace at the scope of this element into account.
-    template <class T> shared_ptr<T> resolveRootNameReference(const string& name) const
+    /// Resolve a reference to a named element at the scope of the given parent,
+    /// taking the namespace at the scope of this element into account.  If no parent
+    /// is provided, then the root scope of the document is used.
+    template <class T> shared_ptr<T> resolveNameReference(const string& name, ConstElementPtr parent= nullptr) const
```
jstone-lucasfilm, line 767: Normally I would recommend that we add this new default argument to our Python binding for `resolveNameReference`, so that callers can use both the default and explicit forms of the method.

But I wonder if a better approach here would actually be to make `resolveNameReference` a protected method of Element, since it's not really a user-facing feature that would be expected as part of the Element API.  If we instead took this second approach, then I'd recommend that we simplify the changelist by removing the Python binding for this method as well.

What are your thoughts?

jstone-lucasfilm (commented):
This looks close to the mark, and I had one additional thought on the `resolveNameReference` method.

source/MaterialXCore/Element.h
kwokcb: Hi @jstone-lucasfilm 
The main implementation issue is that not all uses of this API are via derived classes so would need to work around this somehow. Perhaps for now, the Javascript and Python bindings can be removed first ?

kwokcb: I've made it protected. ShaderRef upgrade used this code so I just copied a few lines over.

source/MaterialXCore/Element.h
kwokcb: Hi @ix-dcourtois (Damien)
Yes, this is something that has come up before and worth discussing as I also come from a history of it being "taboo" to break API compatibility :). Let's open an issue. 

@jstone-lucasfilm, I guess we can start with what you see as the precedence for API compatibility.

BTW, for this case, it will not "break" as this method was never meant to be public and is now renamed and hidden.

jstone-lucasfilm: @ix-dcourtois Completely agreed that changing the signatures of methods in the public API should be done as rarely as possible, since it has the potential to impact so many users.  In this case, we're considering the change because `resolveRootNameReference` never quite made sense as a public API method, and it's my sense that it was swept into the Python/JS bindings for completeness, rather than being a method that developers would want to use.  If you're aware of use cases for this method in developer code, let us know; otherwise we're thinking of making this a "protected" method that would only be visible within the MaterialX libraries themselves.

ix-dcourtois: No complaints for me, I was just pointing out the issue, we're not actually using this one in our code (we had the problem with other methods in which parameters were added or removed, I don't remember, but this one going to protected is ok, at least for us)

source/MaterialXTest/MaterialXCore/Node.cpp
```diff
@@ -164,6 +164,7 @@ TEST_CASE("Flatten", "[nodegraph]")
 
     // Flatten all root-level nodes.
     doc->flattenSubgraphs();
+    doc->validate();
```
jstone-lucasfilm, line 167: Wouldn't we need to check the return value of this validate call with a `REQUIRE`?

source/MaterialXCore/Document.cpp
```diff
@@ -20,7 +20,9 @@ NodeDefPtr getShaderNodeDef(ElementPtr shaderRef)
     if (shaderRef->hasAttribute(NodeDef::NODE_DEF_ATTRIBUTE))
     {
         string nodeDefString = shaderRef->getAttribute(NodeDef::NODE_DEF_ATTRIBUTE);
-        return shaderRef->resolveRootNameReference<NodeDef>(nodeDefString);
+        ConstElementPtr root = shaderRef->getRoot();
```
jstone-lucasfilm, line 23: Since we're now writing out this code explicitly, I would recommend calling `DocumentPtr doc = shaderRef->getDocument()` here, which would allow us to call `doc->getNodeDef()` in the two calls below.

kwokcb: Unfortunately, `Document::getNodeDef()` is not equivalent to resolve logic as it does not try and use the qualified name. I don't think I want to change that method and am hesitant to add a new method just for this small change.

kwokcb:
Thanks for the fix to Document. I didn't think of call getNodDef() twice with the qualified and then non-qualified name :).

jstone-lucasfilm (approved):
This change looks good to me, thanks @kwokcb!

PR #1014, opened 2022-06-29
Python binding for VkShaderGenerator

ashwinbhat:
Example usage:
python\Scripts\generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_gold.mtlx --path .. **--target vulkan** --validator c:\VulkanSDK\1.3.204.1\Bin\glslangValidator.exe

### DIFF
--- a/source/PyMaterialX/PyMaterialXGenGlsl/PyGlslShaderGenerator.cpp
+++ b/source/PyMaterialX/PyMaterialXGenGlsl/PyGlslShaderGenerator.cpp
@@ -42,7 +42,7 @@ void bindPyGlslResourceBindingContext(py::module &mod)
 
 void bindPyEsslShaderGenerator(py::module& mod)
 {
-    py::class_<mx::EsslShaderGenerator, mx::HwShaderGenerator, mx::EsslShaderGeneratorPtr>(mod, "EsslShaderGenerator")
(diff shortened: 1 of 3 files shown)


source/PyMaterialX/PyMaterialXGenGlsl/PyGlslShaderGenerator.cpp
```diff
@@ -48,3 +49,15 @@ void bindPyEsslShaderGenerator(py::module& mod)
         .def("getTarget", &mx::EsslShaderGenerator::getTarget)
         .def("getVersion", &mx::EsslShaderGenerator::getVersion);
 }
+
+// Glsl Vulkan shader generator bindings
+
+void bindPyVkShaderGenerator(py::module& mod)
+{
+    py::class_<mx::VkShaderGenerator, mx::HwShaderGenerator, mx::VkShaderGeneratorPtr>(mod, "VkShaderGenerator")
```
jstone-lucasfilm, line 57: Assuming the second argument in this list represents the base class, shouldn't it be `mx::GlslShaderGenerator` rather than `mx::HwShaderGenerator`?  (It looks like we should apply the same change to the bindings for `mx::EsslShaderGenerator` above as well!)

jstone-lucasfilm (commented):
This looks like a great change, @ashwinbhat, and I had just one recommendation below.

jstone-lucasfilm (approved):

PR #1019, opened 2022-07-05
Fix numerical warnings in GLSL

jstone-lucasfilm:
This changelist adds guards to the GLSL implementation of thin-film rendering, addressing numerical warnings and visual artifacts in situations where eta values decrease between interfaces.  Without these guards, the computation requires the computation of an imaginary square root, which is undefined behavior in GLSL.


kwokcb:
I can't comment on the logic for the IOR, but it does make sense to clamp eta2 locally within the function. It could further down [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L287) where the issue is. (I didn't trace, as this only shows up at program link time).

jstone-lucasfilm:
Good catch, @niklasharrysson, and indeed that second guard was completely unneeded.  Removed!

niklasharrysson (approved):
Excellent! Thanks @jstone-lucasfilm

Issue #1021, opened 2022-07-06
NG_place2d_vector2 generates skew result when used for transform UV texture coordinates


vernalchen:
The place2d node transforms the uv coordinates in the order of -pivot, scale, rotate, translate, pivot. This will result in skew results when the rotation is not zero, while scale x and scale y has different values, which is unacceptable.
Instead, the sequence of transformation should be (in the order of application): -pivot, translate, rotate, scale, pivot.

sample mtlx:
```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">
  <standard_surface name="XID_Default_A0590D2507C54822B1AC8C0DD42DF3EA" type="surfaceshader">
    <input name="base" type="float" value="1" />
    <input name="metalness" type="float" value="0" />
    <input name="base_color" type="color3" output="generic_diffuse_output" nodegraph="XID_Default_A0590D2507C54822B1AC8C0DD42DF3EA_nodegraph" />
    <input name="specular" type="float" value="0" />
    <input name="specular_roughness" type="float" value="0.7" />
    <input name="specular_IOR" type="float" value="1.5" />
    <input name="thin_walled" type="boolean" value="false" />
  </standard_surface>
  <surfacematerial name="XID_Default_A0590D2507C54822B1AC8C0DD42DF3EA_shader" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="XID_Default_A0590D2507C54822B1AC8C0DD42DF3EA" />
  </surfacematerial>
  <nodegraph name="XID_Default_A0590D2507C54822B1AC8C0DD42DF3EA_nodegraph">
    <texcoord name="texcoord1" type="vector2" />
    <place2d name="place2d" type="vector2">
      <input name="texcoord" type="vector2" nodename="texcoord1" />
      <input name="offset" type="vector2" value="0.8, 0" />
      <input name="rotate" type="float" value="-110" />
      <input name="scale" type="vector2" value="0.4, 0.2" />
      <input name="pivot" type="vector2" value="0, 0" />
    </place2d>
    <image name="generic_diffuse" type="color3">
      <input name="texcoord" type="vector2" nodename="place2d" />
      <input name="file" type="filename" value="2x1inch.png" colorspace="srgb_texture" />
      <input name="uaddressmode" type="string" value="periodic" />
      <input name="vaddressmode" type="string" value="periodic" />
    </image>
[...]

PR #1027, opened 2022-07-11
Provide a frame transform for UV texture coordinates

vernalchen:
A proposed fix for #1021
Adding a new `place2dframe` node that changes the order of the operations from
`place2d`: -pivot, scale, rotate, translate, pivot
into
`place2dframe`: -pivot, translate, rotate, scale, pivot
test_place2d_out_glsl:
test_place2d_out_glsl
test_place2dframe_out_glsl:
test_place2dframe_out_glsl

### DIFF
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -2333,8 +2333,23 @@
   <!--
     Node: <place2d> Supplemental Node
     Transform incoming UV texture coordinates for 2D texture placement.
+    Sequence of transformation (in the order of application): -pivot, scale, rotate, translate, pivot.
   -->
   <nodedef name="ND_place2d_vector2" node="place2d" nodegroup="math">
+    <input name="texcoord" type="vector2" value="0.0, 0.0" />
+    <input name="pivot" type="vector2" value="0.0,0.0" />
+    <input name="scale" type="vector2" value="1.0,1.0" />
+    <input name="rotate" type="float" value="0.0" unittype="angle" unit="degree" />
+    <input name="offset" type="vector2" value="0.0,0.0" />
+    <output name="out" type="vector2" defaultinput="texcoord" />
+  </nodedef>
+
+  <!--
+    Node: <place2dframe> Supplemental Node
+    Transform incoming UV texture coordinates for 2D texture placement.
+    Sequence of transformation (in the order of application): -pivot, translate, rotate, scale, pivot.
+  -->
+  <nodedef name="ND_place2dframe_vector2" node="place2dframe" nodegroup="math">
     <input name="texcoord" type="vector2" value="0.0, 0.0" />
     <input name="pivot" type="vector2" value="0.0,0.0" />
     <input name="scale" type="vector2" value="1.0,1.0" />
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -2332,8 +2332,8 @@
 
   <!--
     Node: <place2d> Supplemental Node
-    Transform incoming UV texture coordinates for 2D texture placement.
-    Sequence of transformation (in the order of application): -pivot, scale, rotate, translate, pivot.
+    Transform incoming UV texture coordinates from one 2D frame of reference to another.
+    The order of transformation is: -pivot, scale, rotate, translate, pivot.
   -->
   <nodedef name="ND_place2d_vector2" node="place2d" nodegroup="math">
     <input name="texcoord" type="vector2" value="0.0, 0.0" />
@@ -2346,17 +2346,10 @@
 
   <!--
     Node: <place2dframe> Supplemental Node
-    Transform incoming UV texture coordinates for 2D texture placement.
-    Sequence of transformation (in the order of application): -pivot, translate, rotate, scale, pivot.
+    Transform incoming UV texture coordinates from one 2D frame of reference to another.
+    The order of transformation is: -pivot, translate, rotate, scale, pivot.
   -->
-  <nodedef name="ND_place2dframe_vector2" node="place2dframe" nodegroup="math">
-    <input name="texcoord" type="vector2" value="0.0, 0.0" />
-    <input name="pivot" type="vector2" value="0.0,0.0" />
-    <input name="scale" type="vector2" value="1.0,1.0" />
-    <input name="rotate" type="float" value="0.0" unittype="angle" unit="degree" />
-    <input name="offset" type="vector2" value="0.0,0.0" />
-    <output name="out" type="vector2" defaultinput="texcoord" />
-  </nodedef>
+  <nodedef name="ND_place2dframe_vector2" node="place2dframe" inherit="ND_place2d_vector2" />
 
   <!--
     Node: <arrayappend>
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -2333,7 +2333,11 @@
   <!--
     Node: <place2d> Supplemental Node
     Transform incoming UV texture coordinates from one 2D frame of reference to another.
-    The order of transformation is: -pivot, scale, rotate, translate, pivot.
+    operationorder (integer enum): the order in which to perform the transform operations.
+    "0" or "SRT" performs -pivot, scale, rotate, translate, +pivot as per the original
+    implementation matching the behavior of certain DCC packages, and "1" or "TRS" performs
+    -pivot, translate, rotate, scale, +pivot which does not introduce texture shear.
+    Default is 0 "SRT" for backward compatibility.
   -->
   <nodedef name="ND_place2d_vector2" node="place2d" nodegroup="math">
     <input name="texcoord" type="vector2" value="0.0, 0.0" />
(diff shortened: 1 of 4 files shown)


francis-wangfr (approved):
LGTM

ashwinbhat:
Hi @kwokcb please let us know if you see any issues with this new addition of `place2dframe` node.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -2333,6 +2333,7 @@
   <!--
     Node: <place2d> Supplemental Node
     Transform incoming UV texture coordinates for 2D texture placement.
+    Sequence of transformation (in the order of application): -pivot, scale, rotate, translate, pivot.
```
kwokcb, line 2336: Perhaps wording could be more like this ?

> Transform incoming UV texture coordinates from one 2D frame of
> reference to another. The order of transformation is:   -pivot, scale, rotate, offset, -pivot.

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1046,6 +1046,7 @@
   <!--
     Node: <place2d>
     Transform incoming UV texture coordinates for 2D texture placement.
+    Sequence of transformation (in the order of application): -pivot, scale, rotate, translate, pivot.
   -->
   <nodegraph name="NG_place2d_vector2" nodedef="ND_place2d_vector2">
```
kwokcb, line 1051: Instead of continuing to use the `nodedef=` association on the nodegraph, I'd suggest to use the newer "implementation<nodgraph, nodedef>" association introduced for 1.39. Standard surface has an example of this.
This is supported by USD already if there are any concerns on that front.

resources/Materials/TestSuite/stdlib/math/transform.mtlx
```diff
@@ -178,4 +178,18 @@
     </place2d>
     <output name="out" type="vector2" nodename="place2d1_1" />
   </nodegraph>
+  <!-- place2dframe -->
```
kwokcb, line 181: This doesn't really show much so it might be better to add this in as a texture test so the rendered results are more obvious. For now can you please add to the check-in comment the expected output color value ? Thanks.

kwokcb (commented):
This looks good. I have some minor comments.

libraries/stdlib/stdlib_ng.mtlx
francis-wangfr: FYI @vernalchen , I think th implementation example mentioned by Bernard is here: https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/standard_surface.mtlx#L106

resources/Materials/TestSuite/stdlib/math/transform.mtlx
vernalchen: The output color for place2dframe is not constant. Can I attach place2d_vector2_out_glsl.png and place2dframe_vector2_out_glsl.png in the comment?

vernalchen: Or I can add a new test \resources\Materials\TestSuite\stdlib\texture\image_transform.mtlx which shows the difference between place2d & place2dframe

kwokcb (approved):
Looks good to me, if Jonathan is good with this.

francis-wangfr:
That sounds almost done @kwokcb, thank you！
Hi @jstone-lucasfilm, would you like to share your ideas and suggestions about this fix? Thank you very much!

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1071,6 +1073,36 @@
     <output name="out" type="vector2" nodename="N_addpivot" />
   </nodegraph>
 
+  <!--
+    Node: <place2dframe>
+    Transform incoming UV texture coordinates from one 2D frame of reference to another.
+    The order of transformation is: -pivot, translate, rotate, scale, pivot.
+  -->
+  <implementation name="IMPL_place2dframe_vector2" nodedef="ND_place2dframe_vector2" nodegraph="NG_place2dframe_vector2" />
```
jstone-lucasfilm, line 1081: Although this new implementation-based pattern for assigning nodegraphs to nodedefs is supported in 1.38, I wonder if it would be better to align with the approach used by other graph definitions in this file.  Notice that in nodes such as "contrast" below, we're simply assigning the nodegraph to its nodedef with a single attribute, and no intermediate implementation element is required.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -2343,6 +2344,13 @@
     <output name="out" type="vector2" defaultinput="texcoord" />
   </nodedef>
 
+  <!--
+    Node: <place2dframe> Supplemental Node
```
jstone-lucasfilm, line 2348: CC'ing @dbsmythe to get his thoughts on the proposed node name "place2dframe".

jstone-lucasfilm (commented):
This looks good to me, thanks @vernalchen, and I've posted two minor questions to resolve.

libraries/stdlib/stdlib_ng.mtlx
vernalchen: I switched to this implementation pattern as suggested by @kwokcb, https://github.com/AcademySoftwareFoundation/MaterialX/pull/1027#discussion_r919596826
Do you mean that it's better to revert my latest change for assigning nodegraph NG_place2d_vector2 and NG_place2dframe_vector2 to their nodedefs in [e356b76](https://github.com/AcademySoftwareFoundation/MaterialX/commit/e356b7674aa7a413daf1b6daf0036235c6a67341)?

francis-wangfr: If we don't need implementation node It seems we don't need the node definition `ND_place2dframe_vector2`, just use original `ND_place2d_vector2` for 2 node graphs as below: `<nodegraph name="NG_place2dframe_vector2" nodedef="ND_place2d_vector2">`

jstone-lucasfilm: @vernalchen Yes, that's the change I would recommend, where you associate the nodegraphs with their nodedefs directly through attributes on the nodegraph elements:

```
<nodegraph name="NG_place2dframe_vector2" nodedef="ND_place2dframe_vector2">
```

@francis-wangfr I believe we'll still need two separate nodedef elements for `place2d` and `place2dframe`, so that they are declared as two separate nodes in the library.  Let me know, though, if I'm missing a detail in this regard.

jstone-lucasfilm: That sounds like a good suggestion to me, @dbsmythe, and it would eliminate the need for two distinct nodes with very similar behaviors in common cases.

dbsmythe: I will mock up a suggested update to the node description in the v1.39 Supplemental Notes document and we can hash out details there.

dbsmythe: Here is the initial description for the added input, feel free to suggest alternate wording:
`operationorder` (integer enum): the order in which to perform the transform operations. "0" or "SRT" performs -pivot scale rotate translate +pivot as per the original implementation matching the behavior of certain DCC packages, and "1" or "TRS" performs -pivot translate rotate scale +pivot which does not introduce texture shear.  Default is 0 "SRT" for backward compatibility.

kwokcb: I will add my 2-cents for counter arguments :).
- There is already an existing 1:1 mapping to other APIs like USD which assume one behaviour and now there this node adds new possible behaviour based on a toggle. Basically overloading the existing behaviour so downstream implementations may need to add additional branching logic. 
- This adds another case were a "value change" can result in the requirement for new code generation. If done at the graph level we need to introduce an if statement which is also not good for performance as both paths needs to be emitted.
- @jstone-lucasfilm , as for using implementation, I suggested this to allow migration to the "new" way of association.
I would assume the nodedef inside a nodegraph would be deprecated.

jstone-lucasfilm: @kwokcb For the usage of an implementation element, though, wouldn't we want to make this change globally in our data libraries, rather than piecemeal as new node definitions are added?  I can't see an advantage in creating a mixture of both techniques in the data libraries, when we're currently using one approach consistently.

dbsmythe: Re: @kwokcb 's concerns:
1) I think that even with this proposal, there's still a 1:1 mapping, it's just with the new input value hard-coded to "0" (or "1", depending on which behavior you want to match).
2) I think that in the *vast* majority of cases, the `operationorder` input would be set to a static value at the time the node is generated or certainly by the time initial code generation is done.  And in this situation, I'd expect the code generation to include both branches if the "if" or "switch", but one side of the branch would be optimized out due to realizing the `operationorder` value is constant.
3) Agreed with @jstone-lucasfilm that any switch to using implementations for nodedef-nodegraph connections would be done all at once, *if* we agree that this is a desirable thing to do.  (it's one thing to say this is supported, and a different thing to say it's preferred.)

vernalchen: So we have reached an agreement that we should add an `operationorder` input to` place2d` to switch between two behaviors, right?

francis-wangfr: Yes, @vernalchen, it looks we need a enum switch input `operationorder ` in the place2d node that has default value as original order SRT (0), and we set different order RTS(1) to get what we want. Thanks for your guide and discussion, @kwokcb @jstone-lucasfilm @dbsmythe , let us know if there is not the final solution.

vernalchen: Hi @jstone-lucasfilm @dbsmythe @ashwinbhat @francis-wangfr , please help review the latest changed committed recently, thanks!

resources/Materials/TestSuite/stdlib/math/transform.mtlx
```diff
@@ -175,6 +175,7 @@
       <input name="scale" type="vector2" value="2.0,2.0" />
       <input name="rotate" type="float" value="1.0" unittype="angle" unit="radian" />
       <input name="offset" type="vector2" value="0.2,0.2" />
+      <input name="operationorder" type="integer" value="0" />
```
jstone-lucasfilm, line 178: Since this binding is to the default value, I'd recommend that we omit this added line and just leverage the default.

jstone-lucasfilm (commented):
This looks good to me, @vernalchen, though I'd like to get feedback from @dbsmythe on the proposed language for the new `operationorder` input.  Once the language has been worked out, we'll likely want to copy it directly into the 1.39 specification.

dbsmythe (commented):
Looks good to me!
Leaving to @jstone-lucasfilm to officially approve the changes.

jstone-lucasfilm (approved):
This looks good to me, thanks @vernalchen!

PR #1028, opened 2022-07-13
Add in proper path generation for glTF geometry loader

kwokcb:
## Pathing Fixes

1. Add in proper path generation when loading in glTF files. For now no "/" root is ever added and mesh names are
not part of the path, unless the parent transform path is empty. Previously incorrect paths could be produced, especially
as transform names were "generated". 

2. As part of this, add in valid name checking per string in a path, as transform names that contain spaces currently cause
geometry name matching for `materialassign`s to incorrectly match (this is a separate issue that can be raised be addressed.  e.g. "/ blah" will match all geometry with any "/" in it, which could be everything. )

3. Update default shaderball to remove a top level transform called "input.obj". No other changes were made to the file. The change makes both meshes reside at the root of the scene, which in turns matches the path specifiers used for the example materials file which use "look" material assignments (std surface brass and wood).

## Testing
Tested on desktop and Web viewer.

## Details
The original glb file file converted to ascii has the following scene setup:
```
"nodes": [
    {
      "name": "input.obj",
      "children": [
        1,
        2,
        3,
        4
      ]
    },
    {
      "name": "default"
    },
    {
      "name": "Preview_Mesh",
      "mesh": 0
    },
    {
      "name": "default_node"
    },
    {
      "name": "Calibration_Mesh",
      "mesh": 1
    }
  ],
  "scenes": [
    {
      "nodes": [
        0
      ]
    }
  ],
  "scene": 0
  ```
  The modified version looks like this:
  ```
  "nodes": [
    {
      "name": "Preview_Mesh",
      "mesh": 0
    },
    {
      "name": "Calibration_Mesh",
      "mesh": 1
    }
  ],
  "scenes": [
    {
      "nodes": [
        0, 1
      ]
    }
  ],
  "scene": 0
  ``` 
  The Khronos converter was used to convert from/to glb/glTF-ascii/glb.

### CODE REVIEW

File: source/MaterialXRender/CgltfLoader.cpp [hunk 1]
```diff
@@ -402,7 +456,10 @@ bool CgltfLoader::load(const FilePath& filePath, MeshList& meshList, bool texcoo
         }
     }
 
-    cgltf_free(data);
+    if (data)
```
jstone-lucasfilm, line 459: This is minor, but static analysis packages will flag this change with a warning, since we dereference `data` in the logic above, without checking for nullptr.  I'd recommend leaving this call to `cgltf_free` unguarded, since it doesn't appear possible to reach this point with a null pointer.

jstone-lucasfilm (commented):
Looks good, and I had just one comment in the new code.

source/MaterialXRender/CgltfLoader.cpp
kwokcb: Yes. Your correct. Code change has been reverted.

jstone-lucasfilm (approved):

PR #1034, opened 2022-07-18
Improve wording of Standard Surface inputs

pablode:
The Standard Surface defines inputs that take the "geometric" normal and the "geometric" tangent - I believe that this is incorrect, and that the shading normal is meant.


kwokcb:
Hi @pablode, 
As a quick note, you'll want to check with maintainers of the spec at Autodesk for this, and also that the reference is kept in sync, https://autodesk.github.io/standard-surface/.

pablode:
Thanks for the hint, @kwokcb! I think it makes more sense to discuss the issue over there and sync the changes back if/once they get merged.

PR #1039, opened 2022-08-01
Shader generation support for new OSL closures in MaterialXGenOsl

niklasharrysson:
This change list adds a build option to select between supporting either the old legacy OSL closures or the new OSL closures in development. By default the old closures are selected for backwards compatibility.

To select the new closures one needs to disable the build option MATERIALX_OSL_LEGACY_CLOSURES and use a build of OSL that has the new closures supported in testrender. Also make sure to remove the ifdef MATERIALX_CLOSURE guards in stdosl.h to make the new closures visible when compiling the generated shaders.


libraries/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
         DESTINATION "${MATERIALX_INSTALL_STDLIB_PATH}" MESSAGE_NEVER
-        PATTERN "CMakeLists.txt" EXCLUDE)
+        PATTERN "CMakeLists.txt" EXCLUDE
+        PATTERN "pbrlib_genosl_impl.legacy" EXCLUDE)
+
+if (MATERIALX_OSL_LEGACY_CLOSURES)
+    install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/pbrlib/genosl/pbrlib_genosl_impl.legacy"
+        DESTINATION "${MATERIALX_INSTALL_STDLIB_PATH}/pbrlib/genosl/" RENAME pbrlib_genosl_impl.mtlx)
+endif()
```
niklasharrysson, line 9: Some of the code snippets are different when using the legacy OSL closures. This will install a separate mtlx file with implementation elements pointing to the right code.

libraries/pbrlib/genosl/legacy/mx_anisotropic_vdf.osl
```diff
@@ -0,0 +1,5 @@
+void mx_anisotropic_vdf(vector absorption, vector scattering, float anisotropy, output VDF vdf)
```
niklasharrysson, line 1: Old code snippets are move into a "legacy" directory.

libraries/pbrlib/genosl/mx_subsurface_bsdf.osl
```diff
@@ -1,6 +1,5 @@
 void mx_subsurface_bsdf(float weight, color _color, vector radius, float anisotropy, normal N, output BSDF bsdf)
 {
     // TODO: Subsurface closure is not supported by vanilla OSL.
-    bsdf.response = _color * weight * diffuse(N);
-    bsdf.throughput = color(0.0);
+    bsdf = _color * weight * diffuse(N);
```
fpsunflower, line 4: `testrender` will recognize the subsurface closure (even if it happens to use diffuse in the implementation right now). So it would be safe to use it.

libraries/pbrlib/genosl/mx_dielectric_bsdf.osl
```diff
@@ -1,36 +1,15 @@
...
-    if (scatter_mode == "R")
+    else if (scatter_mode == "T")
     {
-        bsdf.response = tint * weight * comp * microfacet(distribution, N, U, safeAlpha.x, safeAlpha.y, ior, 0);
+        bsdf = weight * dielectric_bsdf(N, U, color(0.0), tint, roughness.x, roughness.y, ior, distribution);
     }
     else
     {
-        bsdf.response = tint * weight * comp * microfacet(distribution, N, U, safeAlpha.x, safeAlpha.y, ior, 2);
+        bsdf = weight * dielectric_bsdf(N, U, tint, tint, roughness.x, roughness.y, ior, distribution);
```
fpsunflower, line 13: For refraction to occur this has to be paired with a medium closure. Is it expected that the user will have a medium node in the graph if they want refraction?

fpsunflower (commented):
Great to see this!

libraries/pbrlib/genosl/mx_dielectric_bsdf.osl
niklasharrysson: If refraction with absorption/scattering is requested then it is expected that the user sets a `medium_vdf` below the `dielectric_bsdf`, using the `layer` operator, in the MaterialX graph. 

But it would be good if the `dielectric_bsdf` could support basic refraction without having the `medium_vdf` present. In that case it would use the same ior for both the reflection and transmission parts. This would ensure backwards compatibility as this is the current behavior before we add the `medium_vdf` to MaterialX, and it's the behavior used in MDL for example.

I'm imagining the following behavior could be used for the two dielectric BSDF's.

- `dielectric_bsdf` If transmission is enabled but no medium is present use the one and only ior for both reflection and transmission.
- `generalized_schlick_bsdf` If transmission is enabled but no medium is present use ior=1.0 for transmission and hence get straight transmission without any refraction.

niklasharrysson: To my mind it would be better to just have ior=1.0 and get no refraction for `generalized_schlick_bsdf` if no medium is present. But this is mainly because I find it so unintuitive that f0 will control how light is refracted. It seems very hard to control for an artist and especially since refraction has such a drastic effect on the look. But this is my personal preference and I'm open to other suggestions.

jstone-lucasfilm: Perhaps we should start with the current MaterialX behavior, where IOR is derived from F0 in `generalized_schlick_bsdf`, but discuss this further in the ASWF Slack channel for this topic.  Ideally, I'd like to bring Naty Hoffman and Andre Mazzone from Lucasfilm/ILM into that conversation, since they've given that question additional thought in a production context.

niklasharrysson: Sounds good. Thanks for the input @jstone-lucasfilm.

jstone-lucasfilm (approved):
This looks like a great foundation to build upon, thanks @niklasharrysson, and let's go ahead with the merge to main.

PR #1046, opened 2022-08-05
Extend <extract> to allow extraction of all channels

kwokcb:
## Extension to \<extract\>

* Allow for all channels for an N channel input (color3, color4, vector2, vector3, or vector4) to be extractable, instead of
just one channel. 
* This basically mirrors what `<combine>` currently which is take N channel input and output a single output.

## Usage
Depending on the variant desired the basic usage syntax is as follows:
```
  <extract name="[extract_node_name]" type="multioutput" >
    <input name="in" type="[type]">
    <input name="index0" type="integer" value="[channel number]" />
    ...
    <input name="index[N]" type="integer" value="[channel number]" />
    <output name="out0" type="float" />
    ...
    <output name="out[N]" type="float" />
  </extract>
  ```
 Where 
 * [extract_node_name] is the name of the node instance
 * [type] is the type of input = {  "color3", "color4", "vector2",  "vector3", or "vector4"  }.
 * [N] is the up to the number of channels on the input
 * [channel number] is <= [N-1], and
  
@jstone-lucasfilm, @proog128 please have a look. Thanks.

### CODE REVIEW

**Comment by jstone-lucasfilm** [comment]
One alternative suggestion here:

What if you were to instead use "channels" attributes in the definition of the glTF PBR, as in the following?

```
<output name="out_occlusion" type="float" nodename="image_orm" channels="r"/>
<output name="out_roughness" type="float" nodename="image_orm" channels="g"/>
<output name="out_metallic" type="float" nodename="image_orm" channels="b"/>
```

(For completeness, I should note that there is a proposal in MaterialX 1.39 to update the syntax of channels attributes, but this shouldn't impact the 1.38 usage described above.)

**Comment by kwokcb** [comment]
Hi Jonathan,
This could work but I'm looking to reduce the complexity of any upstream graph and also create a self contained definition which would allow an input driven way to specify the channels. Thus -- whether you think the extract definition can be part of stdlib ? 

The graph in this document just happens to be one way to build the upstream pattern so isn't really part of the definition.

**Comment by jstone-lucasfilm** [comment]
@kwokcb Wouldn't the channels attribute approach be even simpler than this proposal, in that no new nodes would be required at all, and only a single attribute would be use to select which channel ought to be associated with each output?  Additionally, the channels attribute approach isn't limited to <tiledimage> nodes, and can be used to associate outputs with individual channels of arbitrary pattern nodes as well.

**Comment by kwokcb** [comment]
Yes, this is simpler, 

Off the top my concerns are:

1. You cannot take an existing node, or graph and add extraction to it without modifying it unless you add in new output nodes.
1a. You'd need create a nodegraph to encapsulate the existing node, and if it's an upstream graph, then it won't be supported
as nest nodegraphs aren't supported. 
1b. A top level output is generally not supported except as an AOV output.
2. The channels attribute assumes the the upstream connection has the given channel and hence you have a 
connection from an unknown upstream type to a float in this case. I don't think there are any editors that handle this
currently.
3. You cannot drive the output chanel chosen based on a user input.
4. The `<extract>` node as I recall is the explicit representation when "channels" are not supported / preferred so should be more generally supported.

I think a "channels" scheme would work well if the and "input" can have a "channels" parameter. This has been asked for before.
but currently isn't supported. 

All that said :). 

I think I have a easy addition for this which can be handle any upstream N-channel input.

Add or extend the current "extract" definition so that it can extract out all channels and remove the tiledimage input.
This would exactly mirror what the `combine` nodes currently do. 
The implementation would look something like this, from which new nodedefs could be created.
Any upstream node can be connected to it so image / procedural extraction can be additive by hooking up one of these.
```
  <nodegraph name="NG_extract_all_channels" >
    <input name="in" type="color3" value="0, 0, 0" />
    <input name="index1" type="integer" value="1" />
    <input name="index2" type="integer" value="2" />
    <extract name="extract" type="float" >
      <input name="in" type="color3" interfacename="in" value="0, 0, 0" />
      <input name="index" type="integer" uniform="true" interfacename="index" value="0" />
    </extract>
    <extract name="extract1" type="float" >
      <input name="in" type="color3" interfacename="in" value="0, 0, 0" />
      <input name="index" type="integer" uniform="true" interfacename="index1" value="1" />
    </extract>
    <extract name="extract2" type="float" >
      <input name="in" type="color3" interfacename="in" value="0, 0, 0" />
      <input name="index" type="integer" uniform="true" interfacename="index2" value="2" />
    </extract>
    <output name="out" type="float" nodename="extract" />
    <output name="out1" type="float" nodename="extract1" />
    <output name="out2" type="float" nodename="extract2" />
  </nodegraph>
```

dbsmythe:
The "NG_extract_all_channels" nodegraph proposed above appears to me to be exactly the same as the existing <separate3> supplemental channel node (there's also a <separate2> and <separate4>).  Does that do what you need?  To me, the idea of using the existing <tiledimage> node followed by the existing <separate3> node is a lot more appealing than adding a whole new node.

We could consider expanding the definition of the current <tiledimage> node to have single-channel outputs in addition to the existing (e.g.) color3 "out" output, but I would worry though about this creeping into many other (if not all) nodes which seems incredibly excessive.

kwokcb:
HI @dbsmythe 
I guess either `separate` can be extended to allow channel picking or `extract` extended to support multiple channel output. 

* Extending `separate` can be done by adding to the input signature with optional "index" numbers to choose the output routing. But the cost to evaluate the existing nodes would be a bit more, or new variants could be made.

* Extending `extract` add new variants with multiple outputs. 

I did the latter for now.

dbsmythe:
Is there a situation where you'd want to change or reorder the channels that the outputs are connected to?  E.g. would you ever want "out1" to get channel 0 or channel 2 instead of channel 1?  I guess I'm not seeing why the "indexN" inputs need to be exposed, as opposed to simply changing what outputs of this extract/separate operation (with 0-0, 1-1, 2-2 channel mapping) the downstream node connects to.

If you really do need to change/reorder the channels, then I would propose exposing "indexN" inputs on the <separateN> nodes, so they gain the functionality of 2/3/4-channel "shuffle" nodes (but would still be called "separate").  I don't want to overload the standard <extract> node to make it more complicated, and I'd rather not create any new node types when existing supplemental nodes can be trivially extended with new functionality without making them much more complex or break any usage of the existing form.

kwokcb:
Hi @dbsmythe,
Did a offline chat with @jstone-lucasfilm, and for now we'll park this and look at generic combine and extract for images
a bit later including seeing how this would fit in shader translation and texture baking. 
I think I have good idea of what **not** to add for now :).

Issue #1047, opened 2022-08-09
"channels" attribute ignored for code generation for upstream outputs


kwokcb:
## Summary
The "channels" attribute on an upstream output is being ignore when emitting code
for a surface shader. If just emitting the code for the upstream output then the "channels" attribute is considered.
This could be a something that is long-standing as the test suite does not test the case of downstream connections:.

## Example
This file will fail to generate correct code for all 3 generators (glsl, osl, and mdl). It's a simple constant in an upstream graph
which is output with "channels" to extract out one channel. That output is connected to a downstream shader.
`unlit_surface`, `pbr_gltf` and `standard_surface` all have the same issue so assume it's a common one. 

This are the upstream graphs tested
```
      <nodegraph name="channel_graph">
		<constant name="constant" type="vector3">
			<input name="value" type="vector3" value="0, 0, 1" />
		</constant>
		<output name="out" type="float" nodename="constant" channels="z"/>
	</nodegraph>

	<nodegraph name="non_channel_graph">
		<constant name="constant" type="float">
			<input name="value" type="float" value="1" />
		</constant>
		<output name="out" type="float" nodename="constant"/>
	</nodegraph>
```
The downstream connection is this. `non_channel_graph` works, but `channel_graph` does not.
```
      <standard_surface name="stdsurf" type="surfaceshader">
		<input name="base" type="float" nodegraph="channel_graph" output="out" />
	</standard_surface>
	
	<surfacematerial name="failed_material" type="material">
		<input name="surfaceshader" type="surfaceshader" nodename="stdsurf" />
	</surfacematerial>
```
If you dump the code, the issue can be seen that the type declared is correct but the evaluation to convert from
input to output types is not performed. So assuming the "channels" attribute is just not being checked.
e.g. declarations:
```
// glsl
uniform float stdsurf_base = 0.000000, 0.000000, 1.000000;
// osl
float stdsurf_base = 0, 0, 1
// mdl
float stdsurf_base = 0, 0, 1,
```

## Workaround
* Use explicit channel extraction via nodes such as `extract` node or `separate`.
* Specify the "channels" attribute on the downstream input.
```
	<nodegraph name="channel_graph_in">
		<tiledimage name="tiledimage" type="vector3">
			<input name="file" type="filename" value="resources/Images/grid.png" />
		</tiledimage>
		<constant name="constant" type="vector3">
			<input name="value" type="vector3" value="0, 0, 1" />
		</constant>
[...]

PR #1048, opened 2022-08-09
Update boombox example

kwokcb:
* Change boombox example to use channel attributes vs extract nodes
for greter simplicity.
* Use "channels" attributes on inputs instead of outputs as this does not work with ouptuts. See issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/1047


kwokcb:
@jstone-lucasfilm, as discussed this is the best supported simplification possible for now.

jstone-lucasfilm (approved):
This looks like a good simplification, thanks @kwokcb!

Issue #1035, opened 2022-07-26
Issues with Thin-Film


proog128:
I am trying to add support for [KHR_materials_iridescence](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_iridescence) to the glTF PBR node graph and stumbled upon a couple of issues.

Reproducer: [gltf_pbr.mtlx](https://github.com/AcademySoftwareFoundation/MaterialX/files/9192898/gltf_pbr.mtlx.txt)

### GLSL code generation produces invalid code

The following pattern in the graph seems to produce invalid GLSL code:

```
    <!-- Metal -->

    <generalized_schlick_bsdf name="metal_bsdf" type="BSDF">
      <input name="color0" type="color3" interfacename="base_color" />
      <input name="color90" type="color3" value="1, 1, 1" />
      <input name="roughness" type="vector2" nodename="roughness_uv" />
      <input name="normal" type="vector3" interfacename="normal" />
    </generalized_schlick_bsdf>

    <!-- Thin Film -->
    <thin_film_bsdf name="thinfilm_bsdf" type="BSDF">
      <input name="thickness" type="float" interfacename="iridescence_thickness" />
      <input name="ior" type="float" interfacename="iridescence_ior" />
    </thin_film_bsdf>

    <layer name="iridescent_metal_bsdf" type="BSDF">
      <input name="top" type="BSDF" nodename="thinfilm_bsdf" />
      <input name="base" type="BSDF" nodename="metal_bsdf" />
    </layer>

    <mix name="mix_iridescent_metal_bsdf" type="BSDF">
      <input name="bg" type="BSDF" nodename="metal_bsdf" />
      <input name="fg" type="BSDF" nodename="iridescent_metal_bsdf" />
      <input name="mix" type="float" interfacename="iridescence" />
    </mix>
```

Generated GLSL code (excerpt):

```
BSDF iridescent_metal_bsdf_out = BSDF(vec3(0.0),vec3(1.0), 0.0, 0.0);
iridescent_metal_bsdf_out.ior = iridescence_ior;
iridescent_metal_bsdf_out.thickness = iridescence_thickness;
mx_generalized_schlick_bsdf_reflection(L, V, P, occlusion, 1.000000, base_color, vec3(1.000000, 1.000000, 1.000000), 5.000000, roughness_uv_out, normal, geomprop_Tworld_out, 0, 0, iridescent_metal_bsdf_out);
[...]

PR #1055, opened 2022-08-20
Improvements to thin-film BSDF

proog128:
This PR improves several aspects of the thin film BSDF:
* Adds support for wavelength-dependent complex refraction indices (relevant for metallic base layer)
* Adds (limited) support for thin film effects to `generalized_schlick_bsdf`
* Adds iridescence to glTF PBR

## Wavelength-dependent complex IOR

The current implementation of `thin_film_bsdf` does not support colored base layers, like for example metals. The new implementation supports this case, as shown in the following image.

colored_base_layer

Left: Conductor BSDF with colored reflection.

Center: Thin film on top of the conductor BSDF, old implementation. The yellow tint of the base layer is gone. Only the thin film is visible.

Right: Thin film on top of the conductor BSDF, new implementation. The yellow tint is preserved.

The new implementation now closely follows the Mitsuba reference implementation given in [Belcour and Barla (2017): A Practical Extension to Microfacet Theory for the Modeling of Varying Iridescence](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html). The old implementation was copied from their GLSL shader for BRDF Explorer, which lacked a wavelength-dependent conductor Fresnel term. Note that the authors do not explicitly mention the license of their code. Given the fact that it seemed to be fine to copy their GLSL code, I hope it's fine to copy (parts of) their C++/Mitsuba implementation too.

## Thin film for `generalized_schlick_bsdf`

As mentioned in #1035, the Fresnel term used in `generalized_schlick_bsdf` does not support thin film yet. This changes with this PR.

Given a dielectric base layer, the results between `dielectric_bsdf` and `generalized_schlick_bsdf` are quite close. The following image shows a comparison between thin film over `dielectric_bsdf` (left) and `generalized_schlick_bsdf` (right) BSDFs for a thin film with IOR = 1.33 and thickness = 500 nm over a base layer with IOR = 1.59.

dielectric_exact_schlick

However, the implementation has some limitations. It mainly comes from the fact that the Schlick base layer does not consider IOR of the thin film layer. This limitation results in an increase of error if the base layer IOR becomes equal to the thin film IOR. When evaluating the Schlick Fresnel term of the base layer, the incident IOR is assumed to be 1. The following image shows the error by varying the thin film IORs on top of a base layer with IOR = 1.5:

dielectric_ior_range

A similar problem occurs if the base layer is metallic. Here we additionally cannot take the (unknown) complex IOR of the base layer into account for computing the phase shift, thus the error is even worse. Comparison (left: `conductor_bsdf`, right: `generalized_schlick_bsdf`):

metal_exact_schlick

Nevertheless, although it has limitations, it conveyes the visual impression of a thin film effect.

## Iridescence for glTF PBR

Recently, the glTF PBR was extended by an iridescence effect ([KHR_materials_iridescence](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_iridescence)). Since the MaterialX implementation of glTF PBR is based on `generalized_schlick_bsdf`, it requires the thin film effect for the Schlick Fresnel. Now that this effect is available, it's straightforward to extend the material by thin film. Straightforward except for the [workaround mentioned in #1035](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1035#issuecomment-1196541449), which is required because of limitations in GLSL code generation when referencing the same BSDF multiple times with and without thin-film. Unfortunately, this leads to some duplicated code.


niklasharrysson:
This looks like a great improvement, thanks @proog128! 

We're planning to take a closer look in the next couple of days. There seems to be some test failures we should investigate.

libraries/bxdf/gltf_pbr.mtlx
```diff
@@ -13,6 +13,9 @@
     <input name="alpha" type="float" value="1" uimin="0" uimax="1" uiname="Alpha" uifolder="Alpha" />
     <input name="alpha_mode" uniform="true" type="integer" enum="OPAQUE, MASK, BLEND" enumvalues="0, 1, 2" value="0" uiname="Alpha Mode" uifolder="Alpha" />
     <input name="alpha_cutoff" uniform="true" type="float" value="0.5" uimin="0" uimax="1" uiname="Alpha Cutoff" uifolder="Alpha" />
+    <input name="iridescence" type="float" value="0" uimin="0" uimax="1" uiname="Iridescence" uifolder="Iridescence" />
+    <input name="iridescence_ior" uniform="true" type="float" value="1.3" uimin="1" uisoftmax="3" uiname="Iridescence Index of Refraction" uifolder="Iridescence" />
+    <input name="iridescence_thickness" type="float" value="100" uimin="0" uisoftmin="100" uisoftmax="400" uiname="Iridescence Thickness" uifolder="Iridescence" />
```
kwokcb, line 18: Hi @proog128 , 

Is the assumption here that thickness min and max needs to be handled upstream if there is a thickness texture defined as defined [here](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_iridescence/README.md).

If so then a supplementation node definition for a "gltf_iridescence_image" can be added to allow min, max and image mapping to occur. 

If not, then perhaps it would be a closer match to add in: iridescence multiplier, and and thickness min, max and ratio input ? If there is an image it would map into multiplier and ratio, and min, and max are unmapped values.

kwokcb: I've added in thickness lookup as a new utility definition in the pattern PR so no need to handle it here :).

jstone-lucasfilm:
@proog128 @niklasharrysson Indeed, this looks like a great improvement, and I'll just highlight one of the issues caught by the test suite.  This looks to be an ESSL restriction that we'd need to take into account in the new code:

```
--- Error log:  D:\a\MaterialX\MaterialX\resources\Materials\Examples\StandardSurfaceTiled_Brass.essl.frag

ERROR: 0:4[82](https://github.com/AcademySoftwareFoundation/MaterialX/runs/7931852759?check_suite_focus=true#step:14:83): '*' :  wrong operand types: no operation '*' exists that takes a left-hand operand of type ' const int' and a right operand of type ' in mediump 3-component vector of float' (or there is no acceptable conversion)
```

libraries/pbrlib/genglsl/mx_generalized_schlick_bsdf.glsl
```diff
@@ -20,7 +20,7 @@ void mx_generalized_schlick_bsdf_reflection(vec3 L, vec3 V, vec3 P, float occlus
     float avgAlpha = mx_average_alpha(safeAlpha);
     vec3 Ht = vec3(dot(H, X), dot(H, Y), dot(H, N));
 
-    FresnelData fd = mx_init_fresnel_schlick(color0, color90, exponent);
+    FresnelData fd = mx_init_fresnel_schlick_airy(color0, color90, exponent, bsdf.thickness, bsdf.ior);
```
niklasharrysson, line 23: Just a nit-picky comment on code structure: In the dielectric case we check for the presence of thin-film here, and call the right fresnel init function accordingly. In the generalized schlick case now the airy fresnel version is always called and the check for thin-film is done inside the init instead. I don't have a strong opinion on which approach is better, but for consistency and readability we could use the same approach in all BSDF's.

Very minor, and we can fix this in another change list.

niklasharrysson (approved):
This looks great @proog128 , and I only had a minor note above. We can fix that later.

jstone-lucasfilm:
Thanks for putting this proposal together, @proog128, and I've started testing this proposal with existing material examples.  One issue that I'm encountering is that Standard Surface materials seem to lose their thin-film rendering effects with this pull request:

Standard Surface Thin Film (main branch):
StandardSurfaceThinFilm_Main

Standard Surface Thin Film (PR 1055)
StandardSurfaceThinFilm_PR1055

proog128:
@jstone-lucasfilm @niklasharrysson Sorry for the delay. Standard Surface Thin Film is fixed now, it was caused by a typo in the latest refactorings. In addition, I fixed another bug that increased overall brightness of the thin-film Fresnel.

jstone-lucasfilm:
This looks really promising, thanks @proog128, and I'm now seeing a good visual match with the main branch.

One remaining concern that I have, though, is that the performance for rendering Standard Surface thin-film effects seems to be _reduced_ with respect with the main branch.

To quantify this, I'm attaching screenshots of the standard_surface_thin_film.mtlx material in MaterialXView, with the refresh count set to zero ("--refresh 0"), and with Fraps used to display the frames per second in each build:

Standard Surface Thin Film (main branch):
MaterialXView_ThinFilm_Main

Standard Surface Thin Film (PR 1055)
MaterialXView_ThinFilm_PR1055

proog128:
The code is quite generic (i.e., it calls the conductor Fresnel function with k = 0 for dielectrics) which is nice but unfortunately problematic for performance. I added a few special cases to improve performance, but it's still not as fast as before. However, given that the BSDF is now more accurate and thus more complicated to compute, I expect it to be slower, at least slightly.

I am measuring the following numbers on my machine:

| | main branch (52b5339) | new impl without optimizations (684e916) | new impl with optimizations (81e8e3b) |
|-|-|-|-|
| Standard Surface Thin Film | 51 fps | 23 fps | 44 fps |
| Thin Film Test 7 (new) | 239 fps | 86 fps | 134 fps |

Thin Film Test 7 is a new test to check the case of `thin_film_bsdf` over `conductor_bsdf`. It is interesting because it cannot take any shortcuts in the implementation.

jstone-lucasfilm (approved):
Great work, thanks @proog128!  The small performance difference I'm now seeing is restricted to surfaces with thin-film effects, and seem easily justified by the improved accuracy.

PR #1064, opened 2022-08-31
Standard Surface to glTF PBR Translation Graph

ashwinbhat:
This PR introduces a simple nodegraph for mapping standard_surface.mtlx material documents to gltf_pbr.mtlx material documents. It was implemented as part of my Summer '22 internship with Autodesk as part of a larger Standard Surface => glTF export workflow.

The parameters supported by this translation are simply passed through, or multiplied in the case of weighted colors.

image

**Translation**
We map properties as follows - items in the glTF PBR column are outputs, items in the Standard Surface column are inputs.

<html>
<body>

glTF PBR | Standard Surface
-- | --
base_color | multiply(base_color, base)
metallic | dot(metallic)
roughness | dot(specular_roughness)
transmission | dot(transmission)
specular | dot(specular)
specular_color | dot(specular_color)
ior | dot(specular_ior)
sheen_color | multiply(sheen_color, sheen)
sheen_roughness | dot(sheen_roughness)
clearcoat | dot(coat)
clearcoat_roughness | dot(coat_roughness)
emissive | dot(emission_color)
emissive_strength | dot(emission)

</body>
</html>

### Transmission
Our mapping of transmission ([Standard Surface specular transmission](https://autodesk.github.io/standard-surface/#closures/speculartransmission)) to transmission (glTF KHR_materials_transmission) covers only the transmission weight. Better approximation of Standard Surface's transmission behavior should be possible via the [KHR_materials_volume properties](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_volume/README.md#extending-materials) which are supported by the gltf_pbr.mtlx [node definition](https://github.com/AcademySoftwareFoundation/MaterialX/blob/f2e3fdb1436b1ee443e36284c62fbe87282a72c6/libraries/bxdf/gltf_pbr.mtlx#L23).

### Coat
We map Standard Surface's [coating](https://autodesk.github.io/standard-surface/#closures/coating) to [KHR_materials_clearcoat](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md). KHR clearcoat does not support coat color; this behavior can be approximated by mixing incoming base color and coat color, but was not sufficiently explored in the course of this work and is omitted on this PR.

### Thin film
Standard Surface's [specular transmission](https://autodesk.github.io/standard-surface/#closures/speculartransmission) supports a thin film behavior that may be mappable to [KHR_materials_iridescence](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_iridescence#extending-materials), but this extension is not supported by the current implementation of the gltf_pbr node definition.

### Other behaviors
Any behaviors not listed on this PR were not explored and are not supported by this implementation of the translation nodegraph.

### Other work
Other work implemented for this workflow includes pipelines for translating test materials to native *.gltf and rendering test scenes, there is some discussion on ASWF slack in materialx-gltf about where it may live.

image


jstone-lucasfilm:
This proposal looks promising, thanks @roblesch and @ashwinbhat.

My sense is that we'd want this translation pathway to reach the same level of visual parity as the existing graph from Standard Surface to UsdPreviewSurface before moving forward.  Running a few local tests with materials such as standard_surface_plastic.mtlx and standard_surface_copper.mtlx, it doesn't look like we're quite hitting that level of parity yet.

As one example, there isn't yet any handling for coat attenuation in the new translation graph, so the copper material loses all of its metal coloration when translated to gltf_pbr.  Additionally, there seems to be an issue with the handling of dielectric specular that shows up as a loss of highlights when translating the plastic material to gltf_pbr.

roblesch:
Hi @jstone-lucasfilm, thank you for the feedback and for your suggestions on further improvements! Here's the current results with the most recent push - Standard surface on the left, glTF pbr on the right.

image

jstone-lucasfilm:
This looks excellent, thanks @roblesch!

One additional change that I would recommend is to add _default values_ for each input in the new `standard_surface_to_gltf_pbr` node.  This will guarantee that Standard Surface materials without bindings for these inputs will correctly inherit their correct defaults during shader translation.

A good example to test this with is `standard_surface_brick_procedural.mtlx`, which only binds three of the Standard Surface inputs, leaving all others at their default values.

roblesch:
I've added default input values. Thank you for the continued input @jstone-lucasfilm!

jstone-lucasfilm (approved):
This looks great to me, and thanks for this contribution, @roblesch!

PR #1072, opened 2022-09-08
Make shared libraries relocatable on MacOS and Linux

pmolodo:
[MR805](https://github.com/AcademySoftwareFoundation/MaterialX/pull/805) ensured a relocatable cmake config file, but the `.so` shared libs built on Linux are not themselves relocatable, as they have absolute paths baked into their `RPATH`/`RUNPATH`.

This fixes this issue by making the libs use a relative `RPATH`/`RUNPATH`.


pmolodo:
It seems this is failing the tests because the python libs are being copied into the python folder's site-packages dir, ie:

```
copying build/lib/MaterialX/PyMaterialXGenGlsl.cpython-37m-x86_64-linux-gnu.so.1.38.6 -> /opt/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/MaterialX
```

...and this is then how the modules are found (as opposed to modifying the PYTHONPATH).  I suppose to deal with this scenario, and for maximum backward compatibility, I can also use the absolute build path as a last fallback...?

CMakeLists.txt
```diff
@@ -165,7 +165,26 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
...
+
+# We want the RPATHS to be relative, so the libraries can be relocated; however,
+# we also include the absolute RPATH as a last option because:
+# - the github tests copy the python libs to the python site packages directory;
+#   from that location, the python libs can't find the MaterialX libs at their
+#   "standard" relative location
+# - for maximum backward compatibility - in case other end users are relying on
+#   the absolute paths (like the github tests are)
+# For linking to libraries in the same dir - ie, things in "MATX/lib" to each other
+set(MATERIALX_SAME_DIR_RPATH "${RPATH_RELATIVE_SYMBOL};${CMAKE_INSTALL_PREFIX}/${MATERIALX_INSTALL_LIB_PATH}")
```
jstone-lucasfilm, line 183: Is this line correct?  My sense of symmetry leads me to expect that the first path should be `${RPATH_RELATIVE_SYMBOL}/${MATERIALX_INSTALL_LIB_PATH}`, so that it matches the pattern in the other two `RPATH` variables.  But it's very possible that my sense of symmetry is misguided in this case, and that the current logic is correct.

jstone-lucasfilm (commented):
This looks like a good proposal to me, and I had just one question about the three RPATH variables defined below.

CMakeLists.txt
pmolodo: Yes, I believe this is correct - the name of this var is "MATERIALX_SAME_DIR_RPATH" - so it's intended to be used on targets thar should search for their dependencies in the same directory as they are located - ie, it will generally be used on libraries that are in the lib dir, and need to find other libraries sitting beside them in the same lib dir.

It's also tacked into the end of the other MATERIALX_..._RPATH dirs, so will also check beside themselves for binaries in the same dir.

jstone-lucasfilm (approved):
This looks good to me, thanks @pmolodo!

PR #1074, opened 2022-09-13
Web viewer profiling and optimizations

kwokcb:
## Updates
- Fix to avoid running 2 instance of the viewer as `HtmlWebpackPlugin` will inject code already to run an instance. 
Loading the main code again results in 2 competing instances being executed affecting all load / update time. (`index.ejs`)
- Avoid translating 4 channel IBLs again on load. Remapping is only required if is 3 channel of which the packaged IBL is not. (`helpers.js`)
- Avoid unnecessary geometry transform updates on every frame. (`viweer.js`)
- Restore texture uniform asynchronous load. This still appears to be the most costly in terms of material setup performance.
- Add some perf logging.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -215,6 +215,7 @@ jobs:
     - name: Shader Validation Tests
       if: matrix.test_shaders == 'ON' && runner.os == 'Windows'
       run: |
+        vcpkg/vcpkg remove glslang
         vcpkg/vcpkg install glslang --triplet=x64-windows
         glslangValidator.exe -v
         python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_look_brass_tiled.mtlx --path . --target glsl --validator glslangValidator.exe --vulkanGlsl True --validatorArgs="-V --aml"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,7 +155,7 @@ jobs:
         vcpkg/bootstrap-vcpkg.bat -disableMetrics
         echo $PWD/build/installed/bin | Out-File -FilePath $env:GITHUB_PATH -Append
         echo $PWD/vcpkg/installed/x64-windows/bin | Out-File -FilePath $env:GITHUB_PATH -Append
-        echo $PWD/vcpkg/installed/x64-windows/tools | Out-File -FilePath $env:GITHUB_PATH -Append
+        echo $PWD/vcpkg/installed/x64-windows/tools/glslang | Out-File -FilePath $env:GITHUB_PATH -Append
 
     - name: Install Python ${{ matrix.python }}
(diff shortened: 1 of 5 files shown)


javascript/MaterialXView/index.ejs
```diff
@@ -37,6 +37,5 @@
       <canvas id="webglcanvas" tabindex="1" style="outline: none;"></canvas>
     </div>
     <script src="JsMaterialXGenShader.js"></script>
-    <script src="main.js"></script>
```
kwokcb, line 40: This should never have been there as it invokes 2 instances to run causing all evaluation and load to occur twice.

jstone-lucasfilm: Good catch, @kwokcb!

.github/workflows/main.yml
```diff
@@ -215,6 +215,7 @@ jobs:
     - name: Shader Validation Tests
       if: matrix.test_shaders == 'ON' && runner.os == 'Windows'
       run: |
+        vcpkg/vcpkg remove glslang
```
jstone-lucasfilm, line 218: Now that the glslang test is disabled, I believe we can remove this workaround from recent pull requests.

javascript/MaterialXView/source/helper.js
```diff
@@ -38,34 +38,41 @@ export function prepareEnvTexture(texture, capabilities)
...
-        case THREE.FloatType:
-            rgbaData = new Float32Array(length);
-            break;
-        case THREE.HalfFloatType:
-            rgbaData = new Uint16Array(length);
-            break;
-        default:
-          break;
-    }
+            case THREE.FloatType:
```
jstone-lucasfilm, line 54: Is this change intentional, or just an accidental whitespace edit?

javascript/MaterialXView/source/helper.js
```diff
@@ -38,34 +38,41 @@ export function prepareEnvTexture(texture, capabilities)
...
 
     if (rgbaData)
     {
         for (let i = 0; i < length / 4; i++)
         {
-            rgbaData[(i * 4) + 0] = rgbData[(i * 3) + 0];
-            rgbaData[(i * 4) + 1] = rgbData[(i * 3) + 1];
-            rgbaData[(i * 4) + 2] = rgbData[(i * 3) + 2];
-            rgbaData[(i * 4) + 3] = 1.0;
+                rgbaData[(i * 4) + 0] = rgbData[(i * 3) + 0];
```
jstone-lucasfilm, line 68: As above, is this an intentional change to whitespace?

jstone-lucasfilm (commented):
These optimizations and timing tests look really promising, thanks @kwokcb, and I have just a few thoughts posted below.

jstone-lucasfilm (approved):
Thanks for these optimizations and improvements, @kwokcb!

PR #1076, opened 2022-09-14
OIIO update to v2.3.17 for CI

kwokcb:
* Update for more recent OIIO build. ( [2.3.17 - July 2022 ](https://github.com/OpenImageIO/oiio/releases/tag/v2.3.17.0)) available as download via vcpkg)
  * Need to include OpenImageUI_Util library as part of build so change discovery logic.
  * Minor update to ignore a few new type cast warnings.

### DIFF
--- a/source/MaterialXRender/External/OpenImageIO/FindOpenImageIO.cmake
+++ b/source/MaterialXRender/External/OpenImageIO/FindOpenImageIO.cmake
@@ -45,6 +45,11 @@ find_library ( OPENIMAGEIO_LIBRARY
                HINTS ${OPENIMAGEIO_ROOT_DIR}/lib
                PATH_SUFFIXES lib64 lib
                PATHS "${OPENIMAGEIO_ROOT_DIR}/lib" )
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:
Before we make improvements to our own FindOpenImageIO.cmake, I'd like to take a closer look at the new approach recommended by the OpenImageIO project for finding its libraries:

https://github.com/OpenImageIO/oiio/pull/3098

jstone-lucasfilm:
One option going forward would be to call find_package in "config" mode instead of "module" mode, which should correctly find recent versions of OpenImageIO that export a config file:

https://cmake.org/cmake/help/latest/command/find_package.html

To support earlier versions of OpenImageIO, we could fall back to the existing find_package behavior, leaving the original copy of FindOpenImageIO.cmake unchanged.

jstone-lucasfilm:
Since it sounds like we should move forward with a "config" mode approach to finding OpenImageIO, I'll close out this proposal to edit the original FindOpenImageIO.cmake file.

PR #1079, opened 2022-09-15
Add support for transmission_extra_roughness in Standard Surface

niklasharrysson:
This change list adds support for parameter `transmission_extra_roughness`, that was previously unused in the MaterialX standard_surface implementation.

### DIFF
--- a/libraries/bxdf/standard_surface.mtlx
+++ b/libraries/bxdf/standard_surface.mtlx
@@ -112,6 +112,7 @@
   <nodegraph name="NG_standard_surface_surfaceshader_100">
 
     <!-- Roughness influence by coat-->
+    <!-- Calculate main specular roughness -->
     <multiply name="coat_affect_roughness_multiply1" type="float">
       <input name="in1" type="float" interfacename="coat_affect_roughness" />
       <input name="in2" type="float" interfacename="coat" />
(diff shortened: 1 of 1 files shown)


niklasharrysson:
specular_roughness = 0.15, transmission_extra_roughness = 0.0
specular_0 15_transmission_0 00

specular_roughness = 0.15, transmission_extra_roughness = -0.12 (removing most of the roughness from transmission)
specular_0 15_transmission_-0 12

specular_roughness = 0.15, transmission_extra_roughness = 0.05 (adding additional roughness to transmission)
specular_0 15_transmission_0 05

jstone-lucasfilm (approved):

dbsmythe:
I am thinking it might be better if this was implemented as a multiplier for the original roughness value, rather than being subtracted.  In particular, if the original roughness value is texture mapped and contains a mix of very small and much larger values, then you couldn't subtract very much without roughness going negative.  If implemented as a multipler this can never happen.  I think that artistically, globally doubling or halving roughness on a texture-varying surface would look more natural as well.

niklasharrysson:
Thanks for the feedback @dbsmythe.
I see your point, but the way it is implemented now, being additive/subtractive,  is how it is defined in the Standard Surface whitepaper: https://autodesk.github.io/standard-surface/
It's how it has been implemented in Arnold and other Autodesk products. So if we want to change that behavior I think that would have to be discussed with Autodesk and the authors of the whitepaper.

Minianikum:
I also see your point @dbsmythe, but I think that a multiplier would move away from the point of "transmission_extra_roughness".
The light passing through the material is "blurred" by the surface roughness, then it might also be additionally blurred by the medium itself - that is not related to how rough the surface is, just added to it. If we were to use a multiplier the "blurriness" of the medium would always be proportional to the roughness of the surface, and there shouldn't be such a relation, they need to be separate things that are added together.
(they need to be capped at 0-1 of course)

PR #1085, opened 2022-09-23
Add a simple Bump node

crydalch:
This adds a MtlX Bump node. It simplifies adding bump to shaders by wrapping up the necessary nodes, which helps artists using MaterialX. Having a node called 'bump' also makes it easier to discover.

mtlxbump

Thanks!

### DIFF
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -1105,10 +1105,12 @@
     Offset the surface normal by a scalar value.
   -->
   <nodedef name="ND_bump_vector3" node="bump" nodegroup="geometric">
-    <input name="height" type="float" uiname="Height" uisoftmin="0.0" uisoftmax="1.0" value="0" />
-    <input name="scale" type="float" uiname="Scale" uisoftmin="0.0" uisoftmax="1.0" value="1" />
-    <input name="normal" type="vector3" uiname="Normal" defaultgeomprop="Nworld" />
-    <output name="out" type="vector3" />
+    <input name="height" type="float" uiname="Height" uisoftmin="0.0" uisoftmax="1.0" value="0" doc="Amount to offset the surface normal." />
+    <input name="scale" type="float" uiname="Scale" uisoftmin="0.0" uisoftmax="1.0" value="1" doc="Scalar to adjust the height amount." />
+    <input name="space" type="string" value="tangent" enum="tangent, object" uniform="true" doc="The space to transform the normal from ('tangent' or 'object'); defaults to 'tangent'" />
+    <input name="normal" type="vector3" uiname="Normal" defaultgeomprop="Nworld" doc="Surface normal; defaults to the current world-space normal." />
+    <input name="tangent" type="vector3" uiname="Tangent" defaultgeomprop="Tworld" doc="Surface tangent vector, defaults to the current world-space tangent vector." />
+    <output name="out" type="vector3" doc="Offset surface normal; connect this to a shader's 'normal' input." />
   </nodedef>
 
   <!-- ======================================================================== -->
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -1104,10 +1104,10 @@
     Node: <bump>
     Offset the surface normal by a scalar value.
   -->
-  <nodedef name="ND_bump_vector3" node="bump" nodegroup="houdini">
+  <nodedef name="ND_bump_vector3" node="bump" nodegroup="geometric">
     <input name="height" type="float" uiname="Height" uisoftmin="0.0" uisoftmax="1.0" value="0" />
     <input name="scale" type="float" uiname="Scale" uisoftmin="0.0" uisoftmax="1.0" value="1" />
-    <input name="normal" type="vector3" uiname="Normal" value="0, 0, 0" />
+    <input name="normal" type="vector3" uiname="Normal" defaultgeomprop="Nworld" />
     <output name="out" type="vector3" />
   </nodedef>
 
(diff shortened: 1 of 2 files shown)


libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -1100,6 +1100,17 @@
     <output name="out" type="vector4" default="0.0, 0.0, 0.0, 0.0" />
   </nodedef>
 
+  <!--
+    Node: <bump>
+    Offset the surface normal by a scalar value.
+  -->
+  <nodedef name="ND_bump_vector3" node="bump" nodegroup="houdini">
```
jstone-lucasfilm, line 1107: Maybe nodegroup "math" or "convolution2d" here?

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -1100,6 +1100,17 @@
...
+    <input name="height" type="float" uiname="Height" uisoftmin="0.0" uisoftmax="1.0" value="0" />
+    <input name="scale" type="float" uiname="Scale" uisoftmin="0.0" uisoftmax="1.0" value="1" />
+    <input name="normal" type="vector3" uiname="Normal" value="0, 0, 0" />
```
jstone-lucasfilm, line 1110: Ideally the default value should be the `normal` geometric property, rather than a constant value.

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1031,6 +1031,21 @@
   <!-- Geometric nodes                                                          -->
   <!-- ======================================================================== -->
 
+  <nodegraph name="NG_bump_vector3" nodedef="ND_bump_vector3">
+    <normalmap name="N_normalmap" type="vector3">
+      <input name="in" type="vector3" nodename="N_heighttonormal" value="0.5, 0.5, 1" />
```
jstone-lucasfilm, line 1036: These constant values won't have any effect, since they're overridden by a node connection, so we should likely remove them for clarity.

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1031,6 +1031,21 @@
...
+      <input name="normal" type="vector3" interfacename="normal" value="0, 0, 0" />
+      <input name="space" type="string" value="tangent" />
+      <input name="scale" type="float" value="1" />
+      <input name="tangent" type="vector3" value="0, 0, 0" />
```
jstone-lucasfilm, line 1040: Perhaps this assignment of a constant value to the `tangent` input should be removed, so that the default geometric tangent is used instead?

libraries/stdlib/stdlib_defs.mtlx
crydalch: Thanks, yes will fix this oversight (along with the "houdini" group you pointed out 🤦 )

libraries/stdlib/stdlib_defs.mtlx
crydalch: @jstone-lucasfilm  this was actually meant to be "geometric", since it sort of lives with the other geometry-related nodes; but I can change it to math or convolution2d if those would be better.

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1031,6 +1031,21 @@
...
+  <nodegraph name="NG_bump_vector3" nodedef="ND_bump_vector3">
+    <heighttonormal name="N_heighttonormal" type="vector3">
+      <input name="in" type="float" interfacename="height" />
+      <input name="scale" type="float" interfacename="scale" />
+    </heighttonormal>
+    <normalmap name="N_normalmap" type="vector3">
+      <input name="in" type="vector3" nodename="N_heighttonormal" />
+      <input name="normal" type="vector3" interfacename="normal" />
+      <input name="space" type="string" value="tangent" />
+      <input name="scale" type="float" value="1" />
```
kwokcb, line 1043: A few small notes:
* It seems it would be useful to expose normalmap `scale` and `space` as controllable by the user.
* Also the current scale is not documented as to what it's scaling :). 
* Curious about the nodegroup as `geometric` as this has been used solely for geometric stream so far. Would `texture2d` be a possible alternative. heighttonormal is `convolution2d` and normalmap is `math` (?).

crydalch: @kwokcb  thanks, good nodes; for the last note about the group, I'd be happy to pick one, but I find it super strange from an artist standpoint to see those nodes in either of those categories. To be fair, it's not as user-facing as most tools like Houdini's tab menu are probably searched by name. But I find it strange that both are grouped in that way; texture2d also doesn't seem applicable in this case.

crydalch: @kwokcb  actually that first note, wouldn't adding a second scale be redundant? The scale input is a multiplier on the amount of input bump. We should probably plug the bump node's scale into the normalmap's scale, rather than heighttonormal.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -1100,6 +1100,19 @@
...
   </nodedef>
 
+  <!--
+    Node: <bump>
+    Offset the surface normal by a scalar value.
+  -->
+  <nodedef name="ND_bump_vector3" node="bump" nodegroup="geometric">
+    <input name="height" type="float" uiname="Height" uisoftmin="0.0" uisoftmax="1.0" value="0" doc="Amount to offset the surface normal." />
+    <input name="scale" type="float" uiname="Scale" uisoftmin="0.0" uisoftmax="1.0" value="1" doc="Scalar to adjust the height amount." />
+    <input name="space" type="string" value="tangent" enum="tangent, object" uniform="true" doc="The space to transform the normal from ('tangent' or 'object'); defaults to 'tangent'" />
```
jstone-lucasfilm, line 1110: I believe we can remove this `space` input, and assume the normal map is in tangent space, since the `heighttonormal` node is designed to generate a map in this space.  In the latest draft of MaterialX 1.39, the `space` input has been removed from the `normalmap` node, so removing the input should make the `bump` node more forward-looking.

jstone-lucasfilm (approved):
This looks ready to merge, thanks @crydalch!

PR #1087, opened 2022-09-27
Adds a Unified Noise node

crydalch:
Adds a unified noise node, for artists to have a single noise node that unifies shared controls (such as frequency and offset).  A common complaint we hear from artists, is the MaterialX noises are too bare, they are missing the very basics of frequency/offset, and even some post-processing, which requires creating all sorts of nodes. Rather than modify all of the existing noises, it seemed a slightly higher-level node to encompass shared features made more sense (and would be easier to adopt).

Jitter is treated as a common control; it behaves like jitter does on the standard worley noise, and for the non-Worley noises it scrambles the input. The fractal controls are unique to that particular noise type, so an artist tool would probably hide those parameters when the noise type is not set to fractal.

Thanks!

unifiednoise_nodes

### DIFF
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -983,7 +983,7 @@
     
     The noise types are:
     
-       0 - Noise (noise2d/noise3d)
+       0 - Perlin (noise2d/noise3d)
        1 - Cell (cellnoise2d/cellnoise3d)
        2 - Worley (worley2d/worley3d)
        3 - Fractal (fractal3d)
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -991,18 +991,17 @@
     The output is 1 channel, with controls to adjust the output range.
   -->
   <nodedef name="ND_unifiednoise2d_float" node="unifiednoise2d" nodegroup="procedural2d">
-    <input name="texcoord" type="vector2" uifolder="Common" value="0, 0" />
-    <input name="freq" type="vector2" uiname="Frequency" uifolder="Common" value="1, 1" />
-    <input name="offset" type="vector2" uiname="Offset" uifolder="Common" value="0, 0" />
-    <input name="jitter" type="float" uiname="Jitter" uifolder="Common" uisoftmin="0.0" uisoftmax="1.0" value="1" />
-    <input name="outmin" type="float" uiname="Output Min" uifolder="Post Process" value="0" />
-    <input name="outmax" type="float" uiname="Output Max" uifolder="Post Process" value="1" />
-    <input name="outgamma" type="float" uiname="Output Gamma" uifolder="Post Process" uisoftmin="0.0" uisoftmax="3.0" value="1" />
-    <input name="clampoutput" type="boolean" uiname="Clamp Output" uifolder="Post Process" value="true" />
-    <input name="octaves" type="integer" uiname="Octaves" uifolder="Fractal" value="3" />
-    <input name="lacunarity" type="float" uiname="Lacunarity" uifolder="Fractal" value="2" />
-    <input name="diminish" type="float" uiname="Diminish" uifolder="Fractal" uisoftmin="0.0" uisoftmax="1.0" value="0.5" />
-    <input name="type" type="integer" uiname="Noise Type" uifolder="Common" uisoftmin="0" uisoftmax="3" value="0" enum="Noise,Cell,Worley,Fractal" enumvalues="0,1,2,3"/>
+    <input name="texcoord" type="vector2" uifolder="Common" defaultgeomprop="UV0" doc="The input 2d space. Default is the first texture coordinates." />
+    <input name="freq" type="vector2" uiname="Frequency" uifolder="Common" value="1, 1" doc="Adjusts the noise frequency, with higher values producing smaller noise shapes. Default is (1,1)." />
+    <input name="offset" type="vector2" uiname="Offset" uifolder="Common" value="0, 0" doc="Shift the noise in 2d space. Default is (0,0)." />
+    <input name="jitter" type="float" uiname="Jitter" uifolder="Common" uisoftmin="0.0" uisoftmax="1.0" value="1" doc="Adjust uniformity of Worley noise; for other noise types jitters the results." />
+    <input name="outmin" type="float" uiname="Output Min" uifolder="Post Process" value="0" doc="The lowest values fit to the noise. Default is 0.0." />
+    <input name="outmax" type="float" uiname="Output Max" uifolder="Post Process" value="1" doc="The highest values fit to the noise. Default is 1.0." />
+    <input name="clampoutput" type="boolean" uiname="Clamp Output" uifolder="Post Process" value="true" doc="Clamp the output to the min and max output values." />
+    <input name="octaves" type="integer" uiname="Octaves" uifolder="Fractal" value="3" doc="The number of octaves of Fractal noise to be generated. Default is 3." />
+    <input name="lacunarity" type="float" uiname="Lacunarity" uifolder="Fractal" value="2" doc="The exponential scale between successive octaves of Fractal noise. Default is 2.0." />
+    <input name="diminish" type="float" uiname="Diminish" uifolder="Fractal" uisoftmin="0.0" uisoftmax="1.0" value="0.5" doc="The rate at which noise amplitude is diminished for each octave of Fractal noise. Default is 0.5." />
+    <input name="type" type="integer" uiname="Noise Type" uifolder="Common" uisoftmin="0" uisoftmax="3" value="0" enum="Perlin,Cell,Worley,Fractal" enumvalues="0,1,2,3" doc="Menu to select the type of noise: Perlin, Cell, Worley, or Fractal. Default is Perlin." />
     <output name="out" type="float" />
   </nodedef>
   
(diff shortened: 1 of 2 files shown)


libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1026,6 +1026,143 @@
...
+    <noise2d name="N_perlin_noise2d" type="float">
+      <input name="texcoord" type="vector2" nodename="N_apply_cell_jitter" value="0, 0" />
+      <input name="amplitude" type="float" value="0.5" />
+      <input name="pivot" type="float" value="0.5" />
+    </noise2d>
+    <cellnoise2d name="N_cellnoise2d" type="float">
+      <input name="texcoord" type="vector2" nodename="N_apply_cell_jitter" value="0, 0" />
+    </cellnoise2d>
+    <worleynoise2d name="N_worleynoise2d" type="float">
+      <input name="texcoord" type="vector2" nodename="N_apply_offset" value="0, 0" />
```
jstone-lucasfilm, line 1062: Although these uniform default values won't impact the rendered results, we should likely remove them for clarity, as they're immediately overridden by node connections.  Would it be helpful if we emitted a warning for this situation in Element::validate, so that it's straightforward to address all such cases at once?

crydalch: Thanks. That's probably not a bad idea, to have a warning emitted. But I'll also make a note to update our script that generated the initial output, so node or interface connections don't also get authored values.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -973,6 +973,59 @@
...
+    
+       0 - Noise (noise2d/noise3d)
+       1 - Cell (cellnoise2d/cellnoise3d)
+       2 - Worley (worley2d/worley3d)
+       3 - Fractal (fractal3d)
+    
+    The output is 1 channel, with controls to adjust the output range.
+  -->
+  <nodedef name="ND_unifiednoise2d_float" node="unifiednoise2d" nodegroup="procedural2d">
+    <input name="texcoord" type="vector2" uifolder="Common" value="0, 0" />
```
kwokcb, line 994: It would be useful to have a default stream binding (e.g. `defaultgeomprop=UV0`) to avoid possibly having to cause users to create an explicit `texcoord` node to feed this.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -973,6 +973,59 @@
...
+    <input name="offset" type="vector3" uiname="Offset" uifolder="Common" value="0, 0, 0" />
+    <input name="jitter" type="float" uiname="Jitter" uifolder="Common" uisoftmin="0.0" uisoftmax="1.0" value="1" />
+    <input name="outmin" type="float" uiname="Output Min" uifolder="Post Process" value="0" />
+    <input name="outmax" type="float" uiname="Output Max" uifolder="Post Process" value="1" />
+    <input name="outgamma" type="float" uiname="Output Gamma" uifolder="Post Process" uisoftmin="0.0" uisoftmax="3.0" value="1" />
+    <input name="clampoutput" type="boolean" uiname="Clamp Output" uifolder="Post Process" value="true" />
+    <input name="octaves" type="integer" uiname="Octaves" uifolder="Fractal" value="3" />
+    <input name="lacunarity" type="float" uiname="Lacunarity" uifolder="Fractal" value="2" />
+    <input name="diminish" type="float" uiname="Diminish" uifolder="Fractal" uisoftmin="0.0" uisoftmax="1.0" value="0.5" />
+    <input name="type" type="integer" uiname="Noise Type" uifolder="Common" uisoftmin="0" uisoftmax="3" value="0" enum="Noise,Cell,Worley,Fractal" enumvalues="0,1,2,3"/>
```
kwokcb, line 1025: Would it be better to use "Perlin" vs "Noise" as the first option -- as it's all "noise" :).

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -973,6 +973,59 @@
...
+    The output is 1 channel, with controls to adjust the output range.
+  -->
+  <nodedef name="ND_unifiednoise2d_float" node="unifiednoise2d" nodegroup="procedural2d">
+    <input name="texcoord" type="vector2" uifolder="Common" value="0, 0" />
+    <input name="freq" type="vector2" uiname="Frequency" uifolder="Common" value="1, 1" />
+    <input name="offset" type="vector2" uiname="Offset" uifolder="Common" value="0, 0" />
+    <input name="jitter" type="float" uiname="Jitter" uifolder="Common" uisoftmin="0.0" uisoftmax="1.0" value="1" />
+    <input name="outmin" type="float" uiname="Output Min" uifolder="Post Process" value="0" />
+    <input name="outmax" type="float" uiname="Output Max" uifolder="Post Process" value="1" />
+    <input name="outgamma" type="float" uiname="Output Gamma" uifolder="Post Process" uisoftmin="0.0" uisoftmax="3.0" value="1" />
```
kwokcb, line 1000: Feels strange to have an output gamma as part of this definition. This can be done separately if needed.

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1026,6 +1026,141 @@
     </mix>
     <output name="out" type="vector4" nodename="N_mix_vector4" />
   </nodegraph>
+  
+  <!--
+    <unifiednoise2d>
+    Combined 2d noises for artists.
+  -->
+  <nodegraph name="NG_unifiednoise2d_float" nodedef="ND_unifiednoise2d_float">
+    <range name="N_range" type="float">
```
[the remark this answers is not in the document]
crydalch: @kwokcb  Sorry for the delay following up; aren't we supposed to split the nodedef and nodegraphs into separate files? I thought that's the pattern followed by other nodegraph-defined nodes; but I'd be happy to merge them in one place (FWIW, I also followed this practice for the other two nodes, so if you and @jstone-lucasfilm are okay moving the nodedefs into the nodegraph document, I'll do it for all 3).

libraries/stdlib/stdlib_defs.mtlx
crydalch: @kwokcb yes I agree! That was the original label, but I chose "noise" since the "noise2d" also omitted "perlin" from it's name was named. But yes I will change Noise to Perlin, thanks!

libraries/stdlib/stdlib_defs.mtlx
crydalch: Yes it can be done separately. I only included it b/c the `<range>`node has it, and if folks are accustomed to its presence on the range node, they might like it on this sort of a convenience node to go along with remapping the output range.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -973,6 +973,59 @@
     <output name="out" type="vector3" default="0.0, 0.0, 0.0" />
   </nodedef>
 
+  <!--
+    Node: <unifiednoise2d>
+    Combines the available noises into an artist-friendly node, adding controls
```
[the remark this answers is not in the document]
crydalch: @kwokcb  yes I need to add the docstrings to this and the other PR's as well. Thank you for the reminder!

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1026,6 +1026,141 @@
...
+    <rotate2d name="N_apply_cell_jitter" type="vector2">
+      <input name="in" type="vector2" nodename="N_apply_offset" />
+      <input name="amount" type="float" nodename="N_cell_jitter_mult" />
+    </rotate2d>
+    <add name="N_apply_offset" type="vector2">
+      <input name="in1" type="vector2" nodename="N_apply_freq" />
+      <input name="in2" type="vector2" interfacename="offset" />
+    </add>
+    <combine3 name="N_combine_with_jitter" type="vector3">
+      <input name="in1" type="float" nodename="N_separate" output="outx" />
```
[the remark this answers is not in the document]
crydalch: @kwokcb I'll try this out; if this doesn't work with UsdMtlX I'll leave it in.

libraries/stdlib/stdlib_ng.mtlx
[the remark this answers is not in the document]
kwokcb: Sorry for any confusion. What I meant was to reflect the inputs on the nodedef into the nodegraph. Not to change where they reside.

It will fit the 1.39 change to have nodedefs referenced via an implementaion easier, and 
It also makes the current  implementaion logic do less work when evaluating.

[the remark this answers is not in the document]
crydalch: Ah! Okay I think I understand, thanks for the explanation @kwokcb 

Do I need to include docs on the inputs in both places, or just the nodedef document?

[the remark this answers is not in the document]
jstone-lucasfilm: Documentation in the nodedef element is fine, and the documentation above looks good to me!

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -973,6 +973,57 @@
...
+    to adjust frequency, offset, jitter, etc... Where possible controls have
+    been unified between noises; any noise-specific controls should live in a
+    dedicated folder. These folders should be hidden by artist tools, unless their
+    corresponding noise type is selected.
+    
+    The noise types are:
+    
+       0 - Noise (noise2d/noise3d)
```
jstone-lucasfilm, line 986: This is a minor note, but the documentation needs an update from "Noise" to "Perlin" to match the latest set of enums.

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -1026,6 +1026,161 @@
...
+      <input name="in2" type="float" value="1" />
+    </subtract>
+  </nodegraph>
+  
+  <!--
+    <unifiednoise3d>
+    Combined 3d noises for artists.
+  -->
+  <nodegraph name="NG_unifiednoise3d_float" nodedef="ND_unifiednoise3d_float">
+    <input name="texcoord" type="vector2" />
```
jstone-lucasfilm, line 1116: Shouldn't this input be a vector3 named "position"?

jstone-lucasfilm (commented):
This is looking really good, and I have just two additional notes to resolve.

libraries/stdlib/stdlib_ng.mtlx
crydalch: Indeed, another copy-pasta oversight. Thank you

jstone-lucasfilm (approved):
This looks great to me, thanks @crydalch!

PR #1088, opened 2022-09-29
Add guards for negative colors in Standard Surface

jstone-lucasfilm:
This changelist adds guards for negative color inputs to the Standard Surface graph, addressing edge cases where out-of-gamut color inputs would trigger undefined shader behavior.

### DIFF
--- a/libraries/bxdf/standard_surface.mtlx
+++ b/libraries/bxdf/standard_surface.mtlx
@@ -200,12 +200,20 @@
       <input name="in1" type="float" nodename="coat_gamma_multiply" />
       <input name="in2" type="float" value="1.0" />
     </add>
-    <power name="coat_affected_diffuse_color" type="color3">
+    <max name="base_color_nonnegative" type="color3">
       <input name="in1" type="color3" interfacename="base_color" />
+      <input name="in2" type="float" value="0.0" />
+    </max>
+    <power name="coat_affected_diffuse_color" type="color3">
+      <input name="in1" type="color3" nodename="base_color_nonnegative" />
       <input name="in2" type="float" nodename="coat_gamma" />
     </power>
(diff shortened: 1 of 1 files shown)


niklasharrysson:
The change itself looks good.

But I wonder if this is the right place for input validation. To my mind it's better if validation happens outside, and only in cases where it's really needed. That is, the shading model only states what the expected value range is for it's inputs, and the user or DCC is responsible to make sure the values are in the correct range. We then avoid this extra processing to always occur.

If the shading model is to handle this instead, maybe _all_ inputs should be validated and not just a few selected? 

That are your thoughts?

jstone-lucasfilm:
These are good questions, @niklasharrysson, and I can provide additional clarity on this proposal.

This change doesn't aim to _prevent_ negative input colors, since these are nearly unavoidable in practice, and will frequently occur when an `acescg` texture is converted to the `lin_rec709` color space.  As an example, the color `0.104704, 0.241883, 0.818` in `acescg` will have a slightly negative red channel when it is transformed to the smaller gamut of `lin_rec709`.

Instead, this change is focused on preventing undefined shader behavior, which would occur when a shader computes pow(x, y) when the value of x is negative.  In GLSL this computation produces NaN values in practice, but technically the shader could return _any_ value, since the operation is undefined by the language.

For the Standard Surface graph, there are only two instances where pow(x, y) was computed on an unguarded input color, and it's these two specific computations where I've added guards against negative inputs.  For all other computations, it seems ideal to allow negative values to proceed as usual through the graph, since they don't trigger any undefined behaviors in shading languages.

niklasharrysson (approved):
I see, the guard is for input to the pow function specifically. That makes sense. LGMT :)

PR #1089, opened 2022-09-30
Remove operators 'add' and 'multiply'  for shader types

niklasharrysson:
This change list removes the `add` and `multiply` operator nodes for shader semantic types, as agreed on at the last TSC meeting.


jstone-lucasfilm (approved):

PR #1090, opened 2022-09-30
Add missing keywords to list of reserved words for OSL code generation

niklasharrysson:
This change list add in all OSL standard library function names as reserved words for OSL code generation, including the new closure names. This fixes a regression with name clashes for generated identifier names.


jstone-lucasfilm (approved):

PR #1094, opened 2022-10-05
Change glTF PBR version from 0.0.1 to 2.0.1

pablode:
Since the glTF PBR is pretty much complete in its current state and is publicly used, it seems appropriate to change the version number to 1.0. (For comparison: UsdPreviewSurface is at 2.3 and Standard Surface at 1.0.1.) It could also be kept in sync with the version of the [glTF spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html), in which it is described in detail (that would be 2.0.1.)

I'm not sure if this change should be seen as an "upgrade" and implemented using a separate NodeDef - please let me know your thoughts. Unlike `gltf_colorimage`, which has not been part of a release yet, `gltf_pbr` has been both part of 1.38.4 and 1.38.5.

### DIFF
--- a/libraries/bxdf/gltf_pbr.mtlx
+++ b/libraries/bxdf/gltf_pbr.mtlx
@@ -758,7 +758,7 @@
       <input name="pivot" type="vector2" interfacename="pivot" value="0, 1" />
       <input name="scale" type="vector2" nodename="invert_scale" value="1, 1" />
       <input name="rotate" type="float" nodename="negate_rotate" />
-      <input name="offset" type="vector2" nodename ="negate_offset" />
+      <input name="offset" type="vector2" nodename="negate_offset" />
       <input name="operationorder" type="integer" interfacename="operationorder" value="0" />
     </place2d>
     <output name="out" type="vector3" nodename="normalmap" />
--- a/libraries/bxdf/gltf_pbr.mtlx
+++ b/libraries/bxdf/gltf_pbr.mtlx
@@ -30,6 +30,7 @@
   </nodedef>
 
   <nodegraph name="IMPL_gltf_pbr_surfaceshader" nodedef="ND_gltf_pbr_surfaceshader">
+
     <!-- Volume -->
 
     <convert name="attenuation_color_vec" type="vector3">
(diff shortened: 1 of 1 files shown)


kwokcb:
Hi Pablo, 

Thanks for bringing up this, as I had let this fall off the radar.
I had a chat with Leonard Daly on this. This is the response:

> glTF is V2.0.x, where x is incremented with minor text changes that do not change the meaning of the specification. Extensions are not part of the glTF specification. Each has it's own specification. A new extension does not change the version # of glTF. The official glTF specification is at https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html. There are no plans (either active or passive) to go to glTF V3. The same for glTF V2.x, x > 0; except there is recognition that it might need to happen at a future date.
> 
> Please do not refer to glTF V1.x as it is not compatible with V2.0.
> 
> If there are no other conventions or requirements external extension versioning numbers should start at 1.0 for the initial release. Khronos extension numbers start at 1.0. These are for initial release packages. During development, we try to follow Semantic Versioning. Versions 0.x are not for general use.
> 

Based on this, I recommend we go with "2.0.1" for tagging of all gltf related definitions including the translation graph.
We should not use "1.0".

There is only one version of this definition and changing the number before the 1.39 release will not affect compatibility.
We should not create 2 nodedef versions since that will adversely affect USD as well.

How does this sound to you? Thanks.

pablode:
Thank you Bernard for this insightful comment! It certainly seems more appropriate to go with the 2.0.1 version for the glTF PBR and I've updated the PR accordingly.

However, I'm not sure if this versioning should also apply to the glTF helper nodes - as they are a creation of this MaterialX project and not specified by glTF in any way (unlike e.g. UsdUVTexture, which is described by a spec). They evolve separately based on needs of MaterialX stakeholders, and I believe it would be best to version them separately.

jstone-lucasfilm (approved):
This change looks good to me, thanks @pablode!

kwokcb:
> Thank you Bernard for this insightful comment! It certainly seems more appropriate to go with the 2.0.1 version for the glTF PBR and I've updated the PR accordingly.
> 
> However, I'm not sure if this versioning should also apply to the glTF helper nodes - as they are a creation of this MaterialX project and not specified by glTF in any way (unlike e.g. UsdUVTexture, which is described by a spec). They evolve separately based on needs of MaterialX stakeholders, and I believe it would be best to version them separately. 

Actually, the support nodes are based on the specification, so I still think we should match the version number. Will take this offline. Thanks.

PR #1104, opened 2022-10-14
Codegen improvements to surfaceshader mix operator

niklasharrysson:
This change list improves the handling of surfaceshader mix operator in code generation. One notable change is for OSL shader generation, which now implements the surfaceshader type as a struct of separate BSDF, EDF and Opacity components, in order to simplify operations like mixing.

Also adds new logic to HW transparency detection to support generations of GLSL shaders with transparency when mixing surface shaders.

Attached are some results of mixing surface shaders. There may be some additional changes need on GLSL side to match transparency results better with OSL when mixing shaders. This is left for a future change list to keep this reasonably small.

[mix_shader.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/9786982/mix_shader.pdf)


source/MaterialXGenOsl/Nodes/MaterialNodeOsl.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+    const static string FUNCTION_DEFINITION =
+        "MATERIAL mx_surfacematerial(surfaceshader surface, displacementshader disp)\n"
+        "{\n"
+        "    float opacity_weight = clamp(surface.opacity, 0.0, 1.0);\n"
+        "    return (surface.bsdf + surface.edf) * opacity_weight + transparent() * (1.0 - opacity_weight);\n"
+        "}\n\n";
+
+    const ShaderGenerator& shadergen = context.getShaderGenerator();
+    shadergen.emitBlock(FUNCTION_DEFINITION, FilePath(), context, stage);
```
niklasharrysson, line 28: Material node is now responsible for composing the `surfaceshader` components into a single closure.

source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -277,7 +277,37 @@ ShaderPtr OslShaderGenerator::generate(const string& name, ElementPtr element, G
...
-    emitShaderOutputs(outputs, stage);
+    const ShaderPort* singleOutput = outputs.size() == 1 ? outputs[0] : NULL;
+
+    const bool isSurfaceShaderOutput = singleOutput && singleOutput->getType() == Type::SURFACESHADER;
+
+#ifdef MATERIALX_OSL_LEGACY_CLOSURES
+    const bool isBsdfOutput = singleOutput && singleOutput->getType() == Type::BSDF;
+#endif
+
+    if (isSurfaceShaderOutput)
```
niklasharrysson, line 288: This logic added here is a bit hacky with hard coded handling of `surfaceshader` and `BSDF` as output types. The reason for this is to keep supporting output of nodegraphs / documents that does not have a material node as the final output terminal (there are many exampels of this in out test suite).

For now this is hard coded here for `surfaceshader` and `BSDF` types. I have a plan for how to improve and generalize this, but it includes change to our `Syntax` and `TypeSyntax` classes that I think is too intrusive for 1.38.6. So I'll hold on this until 1.39.

source/MaterialXGenShader/Util.cpp
```diff
@@ -88,12 +98,22 @@ namespace
             return false;
         }
 
-        // Inputs on a surface shader which are checked for transparency
-        const OpaqueTestPairList inputPairList = { {"opacity", 1.0f},
-                                                   {"existence", 1.0f},
-                                                   {"alpha", 1.0f},
-                                                   {"transmission", 0.0f} };
-
+        if (node->getCategory() == MIX_CATEGORY)
```
niklasharrysson, line 101: New logic for transparency detection of mixed surface shaders.

niklasharrysson:
Results for running the whole render testsuite:
[testsuite_compressed.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/9787149/testsuite_compressed.pdf)

jstone-lucasfilm (approved):
This looks like a really worthwhile change, thanks @niklasharrysson!

PR #1113, opened 2022-10-21
Fix parent/child handling in getGeometryBindings

### LINKED ISSUE #1109: Material assignments to child objects
Spotted by @pablode in https://github.com/AcademySoftwareFoundation/MaterialX/pull/1098#issuecomment-1282935116

When two objects have a parent/child relationship, `<materialassign ... geom="...">` cannot distinguish between them correctly.  In the linked PR, there is a chess set pawn object, and it has a "top" child object with a different material.

```
    <materialassign name="Pawn_Body_B1" geom="/Pawn_Body_B1" material="M_Pawn_Body_B" />

    <materialassign name="Pawn_Top_B1" geom="/Pawn_Body_B1/Pawn_Top_B1" material="M_Pawn_Top_B" />
```

The first line above should assign the pawn's body material to the base of the pawn, which is the parent object.  The next line should assign a separate material to the top of the pawn, which is a child of the base.

In MaterialXView, I see the top material get assigned to both parent and child.

kwokcb:
Fix #1109 

## Change 
The matching logic (string) was "reversed" for membership so that parent assignments would  override child assignments. That is the logic should be to check if the path of the assignment is completely contained within the path to the geometry being tested.

e.g.
* The assignment string "/a/b" is contained in path to the geometry: "/a/b/c" so the material can be assigned to "/a/b/c"
* The assignment string "/a/b" is contained in path to the geometry: "/a/b" so the material can be assigned to "/a/b"
* The assignment string "/a/b/c" is not contained in path to the geometry: "/a/b" so the child's material cannot be assigned to the parent geometry.

## Desktop
parent_child_assignment2

## Web
parent_child_assignment

### DIFF
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -628,7 +628,6 @@ export class Material
                                 shader = shaders[0];
                             }
                         }
-                        
                         let collection = materialAssign.getCollection();
                         let geom = materialAssign.getGeom();
                         let newAssignment;
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -605,7 +605,7 @@ export class Material
         // and assign to the associatged geometry. If there are no looks
         // then the first material is found and assignment to all the
         // geometry. 
-        while (this._materials.length) { this._materials.pop(); } 
+        this._materials.length = 0;
         this._defaultMaterial = null;
         var looks = doc.getLooks();
         if (looks.length)
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -605,7 +605,7 @@ export class Material
         // and assign to the associatged geometry. If there are no looks
         // then the first material is found and assignment to all the
         // geometry. 
-        this._materials = [];
+        while (this._materials.length) { this._materials.pop(); } 
         this._defaultMaterial = null;
         var looks = doc.getLooks();
         if (looks.length)
@@ -629,7 +629,15 @@ export class Material
                             }
                         }
                         let collection = materialAssign.getCollection();
+                        if (collection && collection && collection.charAt(0) == "/")
+                        {
+                            collection = collection.slice(1);
+                        }
                         let geom = materialAssign.getGeom();
+                        if (geom && geom && geom.charAt(0) == "/")
+                        {
+                            geom = geom.slice(1);
+                        }
                         let newAssignment;
                         if (collection || geom)
                         {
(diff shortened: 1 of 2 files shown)


source/MaterialXCore/Look.cpp
```diff
@@ -29,7 +29,7 @@ vector<MaterialAssignPtr> getGeometryBindings(ConstNodePtr materialNode, const s
         {
             if (matAssign->getReferencedMaterial() == materialNode)
             {
-                if (geomStringsMatch(geom, matAssign->getActiveGeom()))
+                if (geomStringsMatch(matAssign->getActiveGeom(), geom, true))
```
kwokcb, line 32: The previous test a geometry path against the geometry string for an assignment so
a parent could pick up a child's assignment (e.g. geom: "/A" and could be assigned the material in assigned to: "/A/B").

This has now be reversed to test that the assignment is completely contained within the test geometry so an assignment to "/A/B" will not be assigned to a geometry "/A".

javascript/MaterialXView/source/viewer.js
```diff
@@ -605,7 +605,7 @@ export class Material
         // and assign to the associatged geometry. If there are no looks
         // then the first material is found and assignment to all the
         // geometry. 
-        this._materials = [];
+        this._materials.length = 0;
```
kwokcb, line 608: This was not being cleared properly when searching for new assignments so an old assignment (sucha as to "*") could be left behind as the last assignment.

javascript/MaterialXView/source/viewer.js
```diff
@@ -629,7 +629,15 @@ export class Material
                             }
                         }
                         let collection = materialAssign.getCollection();
+                        if (collection && collection && collection.charAt(0) == "/")
```
kwokcb, line 632: Path comparison needs to strip out the "root" = "/" for proper comparison.

jstone-lucasfilm, line 632: What is the motivation for checking the existence of `collection` twice in this expression?

javascript/MaterialXView/source/viewer.js
```diff
@@ -629,7 +629,15 @@ export class Material
...
+                        {
+                            collection = collection.slice(1);
+                        }
                         let geom = materialAssign.getGeom();
+                        if (geom && geom && geom.charAt(0) == "/")
```
jstone-lucasfilm, line 637: What is the motivation for checking the existence of `geom` twice in this expression?

source/MaterialXCore/Look.cpp
jstone-lucasfilm: Good catch!

javascript/MaterialXView/source/viewer.js
kwokcb: None -- it's a cut and paste typo :). I'll fix it.

jstone-lucasfilm (approved):

Issue #1110, opened 2022-10-20
Handle naming conflicts in MaterialX source includes for OSL 


kwokcb:
Currently there is an `mx_funcs.h` as well as other include files with some custom identifiers and functions.

When included in code their names could cause generated code to conflict with it.
An example is logged here (https://github.com/AcademySoftwareFoundation/MaterialX/issues/1106) for instance.

The current proposal is to at least rename possible conflicts to start with `mx_` consistently, though the side-effects of this
are unknown. This is a similar naming approach taken in general for "MaterialX" source code additions.

Additionally these names need to be added to the [list of restricted keywords in for OSL](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenOsl/OslSyntax.cpp)

This causes a unique name to be generated. Note that if this is for reference implementation generation and an 
a definition's argument identifier conflicts it will be renamed, and hence will no longer match the nodedef's input name.
This is a orthogonal issue brought up to have control over this naming but conflicts must still be avoided.

Note that some of these would not be required and should be removed if they become intrinsics (color4 seems to be in the plan)
to avoid conflict with intrinsics.

jstone-lucasfilm:
Thanks for this report, @kwokcb, and this issue has been addressed as part of #1184.

PR #1117, opened 2022-10-23
Rename OSL helper method to mx_contrast

jstone-lucasfilm:
This changelist renames an OSL helper method from 'contrast' to 'mx_contrast', removing a source of name conflicts in custom nodes.

### DIFF
--- a/libraries/stdlib/genosl/include/mx_funcs.h
+++ b/libraries/stdlib/genosl/include/mx_funcs.h
@@ -263,9 +263,9 @@ vector4 fgamma(vector4 in, float g)
 
 
 //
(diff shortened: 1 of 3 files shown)


jstone-lucasfilm:
Attaching render tests for this changelist, using the latest build of OSL with MaterialX closures enabled:

[MaterialXRenderTests_10_23_2022.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/9847450/MaterialXRenderTests_10_23_2022.pdf)

jstone-lucasfilm:
This changelist is a first step towards addressing issue #1110, where the same approach can be applied to other OSL helper methods such as `noise` and `remap`.

PR #1126, opened 2022-10-30
Fix "unbind geometry" MaterialXView OpenGL errors on MacOS

pablode:
In `GlslProgram::unbindGeometry`, glDisableVertexAttribArray is called. According to the [specification](http://registry.khronos.org/OpenGL-Refpages/gl4/html/glEnableVertexAttribArray.xhtml), this requires a vertex array to be bound, which is not the case. On MacOS, the driver seems to be less lenient and returns an error which is then printed as "OpenGL error after program unbind geometry: 1282".

Since we destroy the vertex array a few lines later, there's no need to disable the attributes.

Removing the glDisableVertexAttribArray calls fixes the errors.


jstone-lucasfilm (approved):
Looks good to me.  Thanks for the fix, @pablode!

PR #1130, opened 2022-10-31
Make "reduced interface" a render option in MaterialXView

kwokcb:
- Update to set the correct option on all context options so all languages are affected.
- Call to re-generate and rebuild the property editor.
- Move option from load to render.
-


jstone-lucasfilm (approved):

PR #1132, opened 2022-11-04
MaterialXView startup improvements

kwokcb:
1. Simplify startup pathing to make install and dev builds behave the same.
2. Disable display of window frame (on Windows at least) for offscreen rendering from command line.

Note for 1: This was partially written to help address ##1058, but it seems that this works regardless of this change, but this
change should improve startup library searching never-the-less for different invocation points.


source/MaterialXView/Main.cpp
```diff
@@ -300,19 +296,26 @@ int main(int argc, char* const argv[])
         viewer->setBakeHeight(bakeHeight);
         viewer->setBakeFilename(bakeFilename);
         viewer->initialize();
-        if (!bakeFilename.empty()) 
+
+        bool drawOffscreen = !bakeFilename.empty() || !captureFilename.empty();
```
jstone-lucasfilm, line 300: The pathing change above looks great, but I'm less clear on the motivation for this change.  If we're baking textures to disk, why would the viewer need to draw at all?  And why would we need to explicitly set the visibility of the viewer to false, when it defaults to that setting?

kwokcb: I've fixed this up. Only the capture needs a forced redraw as the viewport is not set to be visible before performing the capture.

jstone-lucasfilm (approved):

PR #1137, opened 2022-11-08
Include sample Python scripts as part of build

kwokcb:
The Python utility examples in python/Scripts have been mentioned as references so
seems useful to add these to the distribution. Also integrations that use installs can call these as desired for things
like texture baking, code generation etc.


jstone-lucasfilm (approved):

PR #971, opened 2022-05-29
Fix incorrect glTF PBR default input values

pablode:
Hi! I've noticed that some of the default input values used in the glTF PBR implementation do not match the glTF specification.

The following table compares current default values to what I think they should be:

| glTF PBR input       | MaterialX default value | Spec value | Spec section
| ---                  | ---                     | ---        | ---
| base_color           | 0.8, 0.8, 0.8           | 1, 1, 1    | [5.22](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-material-pbrmetallicroughness)
| metallic             | 0                       | 1          | [5.22](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-material-pbrmetallicroughness)
| roughness            | 0                       | 1          | [5.22](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-material-pbrmetallicroughness)
| normal               | Nworld                  | -
| occlusion            | 0                       | 1?         | [3.9.3](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#additional-textures)
| transmission         | 0                       | 0          | [KHR_materials_transmission](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_transmission#properties)
| specular             | 1                       | 1          | [KHR_materials_specular](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_specular#overview)
| specular_color       | 1, 1, 1                 | 1, 1, 1    | [KHR_materials_specular](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_specular#overview)
| ior                  | 1.5                     | 1.5        | [KHR_materials_ior](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_ior#extending-materials)
| alpha                | 0                       | 1          | [5.22](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-material-pbrmetallicroughness)
| alpha_mode           | OPAQUE                  | OPAQUE     | [5.19](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-material)
| alpha_cutoff         | 0.5                     | 0.5        | [5.19](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-material)
[...]

PR #1138, opened 2022-11-08
Fix values for default gltf_pbr example

kwokcb:
### Change
Fix to make default gltf_pbr example shader to match the change defaults values updated as part of this [PR](
https://github.com/AcademySoftwareFoundation/MaterialX/pull/971)

### Comparison
| Correct render | Previous Render |
| ---------------  | ------------------ |
|  |  |

The main difference is that roughness and metallic have changed from 0 values to 1 values.
 
 ### Validation
 Tested this via conversion from MTLX to glTF. The cgltf write logic does not write out default values so should be a shader without any value overrides (as shown)
 ```JSON
 {
  "asset": {
    "generator": "MaterialX 1.38.71.38.4 to glTF generator",
    "version": "1.38.7",
    "min_version": "1.38.7"
  },
  "materials": [
    {
      "name": "SR_default",
      "pbrMetallicRoughness": {
      }
    }
  ]
}
 ```

### DIFF
--- a/resources/Materials/Examples/GltfPbr/gltf_pbr_default.mtlx
+++ b/resources/Materials/Examples/GltfPbr/gltf_pbr_default.mtlx
@@ -1,9 +1,9 @@
 <?xml version="1.0"?>
 <materialx version="1.38" colorspace="lin_rec709">
   <gltf_pbr name="SR_default" type="surfaceshader">
-    <input name="base_color" type="color3" value="0.8, 0.8, 0.8" />
-    <input name="metallic" type="float" value="0" />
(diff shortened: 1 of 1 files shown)


kwokcb:
@jstone-lucasfilm : As a side thought, I think these "default" instances should be generated from the the nodedef to avoid accidentally writing a wrong default value via copy-and-paste. There is already a method to create a node instance's inputs from a nodedef (`addInputsFromNodeDef()`) which might be good to add to a high level "create node" interface :).

jstone-lucasfilm (approved):
Looks good, @kwokcb, and thanks for the fix!

Issue #1140, opened 2022-11-09
Texture baking output access for multiple material documents


kwokcb:
## Issue
Currently `TextureBaker` always writes out a new document per material found. 

1. It is very hard to find the output documents. 
The only indication as to where these files are is output to a given stream like std out and no way to access them in-memory.
2. It is very hard to merge these documents together afterwards as unique naming is per document so things like the upstream bake nodegraph are duplicated (`NG_graph`) and / or `geompropvalues`

## Suggestions
1. Add an option to write to a single output document as an option.
2. Add an option to write to disk separately and have access to in-memory document(s).
(It would be useful to not always write the image cache to disk as well and have access to that).

jstone-lucasfilm:
Thanks for this original report, @kwokcb, and I'll close this out now that #1141 has been merged.

Issue #1139, opened 2022-11-09
MaterialXView: baking can miss image size detection when using include files


kwokcb:
** Issue
Sometimes images are not detected since they are in xinclude files. If no images are found then it can appear
like baking has failed since 4x4 images are used which appear pretty much like constant values.

** Cause
The code to  `ImageHandler::getReferencedImages()` always skips elements which are not from the main document
based on source URI. This will skip library files for instance but can also skip images which are part of shaders used for baking.
The code is like this:
```
ImageVec ImageHandler::getReferencedImages(DocumentPtr doc)
{
    ImageVec imageVec;
    for (ElementPtr elem : doc->traverseTree())
    {
        if (elem->getActiveSourceUri() != doc->getSourceUri())
        {
            // Always skips here <<<<<<<<<<<<<<<<<
            continue;
        }
 ```
 This means files like `standard_surface_look_brass_tiled.mtlx which has only 2 xincludes will never find any of the images
 used for the shaders to bake. This resutls in an erroneous value if {0,0} as the bake size -- which is clamped to {4,4}.
 
 As a result the images when rendered appear pretty much like constant values which to a user might indicate some
 failure has occurred in baking when in fact the baked images are just too low resolution.
 
 Example looks like this:
 ```
 <materialx version="1.38" xmlns:xi="http://www.w3.org/2001/XInclude">
  <!-- The images are in the include files which have different URI's from the main document ******* --->
  <xi:include href="standard_surface_brass_tiled.mtlx" />
  <xi:include href="standard_surface_greysphere_calibration.mtlx" />
  
  <look name="Brass_Look">
    <materialassign name="preview" geom="Preview_Mesh" material="Tiled_Brass" />
    <materialassign name="calibration" geom="Calibration_Mesh" material="Greysphere_Calibration" />
  </look>
</materialx>
```
| Result using look (containing calibration shader) file |  Result using file directly containing same shader |
| ------------ | ---------------|
| ![image](https://user-images.githubusercontent.com/49369885/200918154-780d088d-a550-4ee7-a231-60e49f57c1c5.png) | ![image](https://user-images.githubusercontent.com/49369885/200917930-0c75ef83-4f4c-4b42-a35d-1bf4257d685c.png) |

This was found by comparing with the `texturebaker` script which sets a default of {1024,1024}.

## Suggestions
* Easiest is to allow a higher default value / or a consistent value with `texturebaker`.
[...]

PR #1141, opened 2022-11-09
Texture baker updates

kwokcb:
## Issue
* Writing documents separately means that each documents elements are uniquely named but when combined it is possible to get clashes. As there is no API to merge and resolve duplicates, a new option is added to just create a single document and use the built in texture baking element uniqueness resolve things. This also means that texture instances that used by more than one shader are not baked many times (e.g. chess set).
* Update #1139, #1140 

## Changes

* Add in option to write output to single document: `writeDocumentPerMaterial`
  * Add `texturebaker` utility option
  * Add `translateshader` utility option

* Update MTLXView to have a minimum bake size set to 1K to match `texturebaker` and allow for single document option:

# Results

* Example translating shader to N (default)
```
python translateshader.py resources\Materials\Examples\StandardSurface\standard_surface_look_brass_tiled.mtlx x.mtlx gltf_pbr --writeDocumentPerMaterial True
Processing material: Tiled_Brass
Wrote baked image: Tiled_Brass_gltf_pbr_base_color.png
Wrote baked image: Tiled_Brass_gltf_pbr_roughness.png
Wrote baked image: Tiled_Brass_gltf_pbr_clearcoat.png
Wrote baked image: Tiled_Brass_gltf_pbr_clearcoat_roughness.png

Processing material: Greysphere_Calibration
Wrote baked document: x_SR_brass1.mtlx
Wrote baked document: x_SR_Greysphere_Calibration.mtlx
```
vs 1 document:
```
python translateshader.py resources\Materials\Examples\StandardSurface\standard_surface_look_brass_tiled.mtlx single_doc.mtlx gltf_pbr --writeDocumentPerMaterial False
Processing material: Tiled_Brass
Wrote baked image: Tiled_Brass_gltf_pbr_base_color.png
Wrote baked image: Tiled_Brass_gltf_pbr_roughness.png
Wrote baked image: Tiled_Brass_gltf_pbr_clearcoat.png
Wrote baked image: Tiled_Brass_gltf_pbr_clearcoat_roughness.png

Processing material: Greysphere_Calibration
Wrote baked document: single_doc.mtlx
```
with the following result:
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <nodegraph name="NG_baked" colorspace="srgb_texture">
    <image name="base_color_baked" type="color3">
      <input name="file" type="filename" value="Tiled_Brass_gltf_pbr_base_color.png" />
    </image>
    <output name="base_color_output" type="color3" nodename="base_color_baked" />
    <image name="roughness_baked" type="float">
      <input name="file" type="filename" value="Tiled_Brass_gltf_pbr_roughness.png" />
    </image>
    <output name="roughness_output" type="float" nodename="roughness_baked" />
    <image name="clearcoat_baked" type="float">
      <input name="file" type="filename" value="Tiled_Brass_gltf_pbr_clearcoat.png" />
    </image>
    <output name="clearcoat_output" type="float" nodename="clearcoat_baked" />
    <image name="clearcoat_roughness_baked" type="float">
      <input name="file" type="filename" value="Tiled_Brass_gltf_pbr_clearcoat_roughness.png" />
    </image>
    <output name="clearcoat_roughness_output" type="float" nodename="clearcoat_roughness_baked" />
  </nodegraph>
  <gltf_pbr name="SR_brass1_baked" type="surfaceshader">
    <input name="base_color" type="color3" output="base_color_output" nodegraph="NG_baked" />
    <input name="roughness" type="float" output="roughness_output" nodegraph="NG_baked" />
    <input name="clearcoat" type="float" output="clearcoat_output" nodegraph="NG_baked" />
    <input name="clearcoat_roughness" type="float" output="clearcoat_roughness_output" nodegraph="NG_baked" />
  </gltf_pbr>
  <surfacematerial name="Tiled_Brass_baked" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_brass1_baked" />
  </surfacematerial>
  <gltf_pbr name="SR_Greysphere_Calibration_baked" type="surfaceshader">
    <input name="base_color" type="color3" value="0.478431, 0.47451, 0.47451" colorspace="srgb_texture" />
    <input name="roughness" type="float" value="0.698039" />
    <input name="clearcoat_roughness" type="float" value="0.0980392" />
  </gltf_pbr>
  <surfacematerial name="Greysphere_Calibration_baked" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_Greysphere_Calibration_baked" />
  </surfacematerial>
</materialx>
```
* Note that an invalid arg will now throw an exception `error: argument --writeDocumentPerMaterial: Boolean value expected.`
which is performed in a new `stringToBoolean()` utility added to MaterialX `datatypes.py`

* Chess set bake result:
[standard_surface_chess_set_baked.mtlx.txt](https://github.com/AcademySoftwareFoundation/MaterialX/files/10716534/standard_surface_chess_set_baked.mtlx.txt)


jstone-lucasfilm (commented):
This looks really promising, thanks @kwokcb.

I would recommend that we harmonize on a single name for the flag that controls this new behavior, rather than inverting the name between our Python scripts and C++ classes (e.g. "writeSingleDocument" vs "writeSeparateDocuments").

One suggestion for a unified naming convention would be "writeDocumentPerMaterial", which seems clear enough for new readers to follow.  We would then provide this same flag name in both Python scripts and C++ classes, and follow your existing convention of defaulting its value to "true" for backwards compatibility.

kwokcb:
> This looks really promising, thanks @kwokcb.
> 
> I would recommend that we harmonize on a single name for the flag that controls this new behavior, rather than inverting the name between our Python scripts and C++ classes (e.g. "writeSingleDocument" vs "writeSeparateDocuments").
> 
> One suggestion for a unified naming convention would be "writeDocumentPerMaterial", which seems clear enough for new readers to follow. We would then provide this same flag name in both Python scripts and C++ classes, and follow your existing convention of defaulting its value to "true" for backwards compatibility.

Sorry, I missed this. I will clean up the naming issues. Thanks.

kwokcb:
Fixed to use consistent option naming and added in example.

python/Scripts/baketextures.py
```diff
@@ -10,6 +10,16 @@
 from MaterialX import PyMaterialXRender as mx_render
 from MaterialX import PyMaterialXRenderGlsl as mx_render_glsl
 
+def tobool(v):
```
jstone-lucasfilm, line 13: Ah, I see that there isn't a universally-available `strtobool` method in Python right now:

https://stackoverflow.com/questions/42248342/yes-no-prompt-in-python3-using-strtobool

Could we instead add a `stringToBoolean` method to `datatypes.py` in MaterialX Python, allowing our example scripts to access `mx::stringToBoolean` in a consistent way?

jstone-lucasfilm: With a `stringToBoolean` method available, we may then be able to directly reference this as the `type` of the argument in our parser, as described in this post:

https://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse

kwokcb: Good idea. 

I've done this but I needed to add `argpase` as a dependency to throw a proper exception. The "simpler"
options which just return `false` if not parsed are not correct if the default value is not `false` (which happens in the current case). Thus we throw an exception as shown in the sample code when a bool cannot be parsed. I removed the 'y' and 'n' options as most just parse 't', 'f', '0', '1', 'True', 'False', True, False.

source/MaterialXView/Viewer.cpp
```diff
@@ -2124,15 +2147,15 @@ void Viewer::bakeTextures()
...
         mx::ImageVec imageVec = _imageHandler->getReferencedImages(doc);
         auto maxImageSize = mx::getMaxDimensions(imageVec);
-        unsigned int bakeWidth = std::max(maxImageSize.first, (unsigned int) 4);
-        unsigned int bakeHeight = std::max(maxImageSize.second, (unsigned int) 4);
+        unsigned int bakeWidth = std::max(maxImageSize.first, _minimumBakeDimension);
+        unsigned int bakeHeight = std::max(maxImageSize.second, _minimumBakeDimension);
         if (_bakeWidth)
         {
-            bakeWidth = std::max(_bakeWidth, (unsigned int) 4);
+            bakeWidth = std::max(_bakeWidth, _minimumBakeDimension);
```
jstone-lucasfilm, line 2154: What is the additional value provided by this `minimumBakeDimension` input?

As things currently stand, we have `bakeWidth` and `bakeHeight` values that the user can set, and I believe we would want to use these literal values when they have been provided (e.g. 256x256), rather than clipping them to another minimum (e.g. 1024).

In the alternative case where the user _doesn't_ set `bakeWidth` or `bakeHeight`, then we currently use the maximum size of all input images, falling back to rendering 4x4 images if no input images are present.  In that alternative case, we similarly don't want to clip to an additional minimum size (e.g. 1024), as the user would have provided an explicit `width` and `height` if one was intended.

python/MaterialX/datatype.py
```diff
@@ -86,3 +86,13 @@ def isColorType(t):
...
+
+def stringToBoolean(value):
+    "Return boolean value found in a string. Throws and exception if a boolean value could not be parsed"
+    if isinstance(value, bool):
+        return value
+    if value.lower() in ('yes', 'true', 't', '1'):
+        return True
+    elif value.lower() in ('no', 'false', 'f', '0'):
+        return False
+    raise argparse.ArgumentTypeError('Boolean value expected.')
```
jstone-lucasfilm, line 98: Since this helper function is now generic, rather than being linked to argument parsing, could we simply raise a `TypeError` in this case?

source/MaterialXView/Viewer.cpp
kwokcb: It's been a long while but I think what I encountered is that the `bakeWidth` and `bakeHeight` are not setting from the UI so can end up with very small textures if there are no images mapped as this is the overriding heuristic to "guess" the output image size. 

For example, if you want to bake the marble example you end up with something that is pretty well "useless". 

Perhaps a resolution for this is if `bakeWidth` or `bakeHeight` are explicitly specified, don't use this minimum value and don't use the heuristics value as that is also "deceiving" in that if there are textures mapped then they unknowingly supersede the explicit width and height passed in ?

The alternative is to expose `bakeWidth` and `bakeHeight` which would allow editing of `_bakeWidth` and `_bakeHeight` ? This seems the easist way and adds no new logic.

jstone-lucasfilm: We could indeed add `bakeWidth` and `bakeHeight` as options in the MaterialX viewer, though for production texture baking we usually recommend the use of Python scripts, which already have access to these features.

It's a fine line for the MaterialX viewer, where we want to enable easy testing of features such as texture baking and shader translation, but we don't necessarily want to expose the entire production-level feature set in the viewer interface.

kwokcb: I'll leave it out. It's not directly related to the main change anways.

python/MaterialX/datatype.py
kwokcb: This works, but for some reason it does not take the argument when used from the argparser.
I've made the change.

jstone-lucasfilm (approved):
This looks good to me, thanks @kwokcb!  I'd be interested in thoughts from @lfl-eholthouser as well, since she has done development work in this area.

lfl-eholthouser (approved):
I think this look great @bernardkwok. Thanks!

kwokcb:
Merge is done. The only change is to add the Python wrapper for Metal version of the TextureBaker interface.

PR #1150, opened 2022-11-19
Fix issue with removing constant nodes in codegen optimization.

niklasharrysson:
This change list improves on the removal of constant nodes during codegen graph optimization. Constant nodes are now always removed, bypassing them in the graph. Also fixes a bug related to node removal.

The problem was triggered by the example given by Frieder Erdmann in this Slack thread: https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1668096291597659

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -1015,15 +1015,10 @@ void ShaderGraph::optimize(GenContext& context)
     {
         if (node->hasClassification(ShaderNode::Classification::CONSTANT))
         {
-            // Constant nodes can be removed by assigning their value downstream
-            // But don't remove it if it's connected upstream, i.e. it's value
-            // input is published.
-            ShaderInput* valueInput = node->getInput(0);
-            if (!valueInput->getConnection())
-            {
-                bypass(context, node, 0);
-                ++numEdits;
-            }
+            // Constant nodes can be removed by moving their value 
+            // or connection downstream.
+            bypass(context, node, 0);
+            ++numEdits;
         }
         else if (node->hasClassification(ShaderNode::Classification::IFELSE))
         {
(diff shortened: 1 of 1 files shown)


source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1015,15 +1015,10 @@ void ShaderGraph::optimize(GenContext& context)
     {
         if (node->hasClassification(ShaderNode::Classification::CONSTANT))
         {
-            // Constant nodes can be removed by assigning their value downstream
-            // But don't remove it if it's connected upstream, i.e. it's value
-            // input is published.
-            ShaderInput* valueInput = node->getInput(0);
-            if (!valueInput->getConnection())
```
niklasharrysson, line 1022: Previously constant nodes where not removed if connected upstream. I can't find a good reason for this restriction. 

A constant node results in an extra local variable declaration in generated code, which In GLSL triggered an issue if the input is of filename type (a texture) since GLSL don't allow local variables of sampler type.

By always removing (bypassing) constant nodes we avoid the extra local variable, and avoid the GLSL issue.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1069,9 +1064,11 @@ void ShaderGraph::optimize(GenContext& context)
         // Traverse the graph to find nodes still in use
         for (ShaderGraphOutputSocket* outputSocket : getOutputSockets())
         {
-            if (outputSocket->getConnection())
+            // Make sure to not include connections to the graph itself.
+            ShaderOutput* upstreamPort = outputSocket->getConnection();
+            if (upstreamPort && upstreamPort->getNode() != this)
```
niklasharrysson, line 1069: Fixing a bug where we didn't check if the connection upstream was to the graph itself (the graph interface). So the graph itself could potentially be added as a node in the graph here.

jstone-lucasfilm (approved):
This looks excellent, thanks @niklasharrysson!

PR #1151, opened 2022-11-22
Add uisoftmin attribute to IOR inputs

rasmusbonnedal:
The usual lower bound of ior values is 1.0 but can be lower for relative ior or for turning off fresnel completely. Having uisoftmin set allows user interfaces to have a slider with this range.


jstone-lucasfilm (approved):
This looks like a good improvement to me, thanks @rasmusbonnedal!

PR #1152, opened 2022-11-22
Elide dot nodes to avoid local sampler creation

kwokcb:
### Fix
* A dot node is equivalent to a pass-through so bypass it during shader optimization. This fixes the issue of `dot` nodes
creating intermediate samplers within functions for GLSL which is disallowed.
* Uses same logic as for https://github.com/AcademySoftwareFoundation/MaterialX/pull/1150 and constant nodes.

### Implementation
* Adds "dot" node classification which can be checked at optimization time to bypass.

### Example
This will generate properly now
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <nodedef name="nd_px_mx_test" node="px_mx_test">
    <input name="SD_CustomTexture" type="filename" value="some/texture/path.dds" />
  <output name="Color" type="color3" />
  </nodedef>
  <nodegraph name="ng_px_mx_test" nodedef="nd_px_mx_test">
    <dot name="hlsl_Sampler2D_v2_274" type="filename">
      <input name="in" type="filename" interfacename="SD_CustomTexture" />
    </dot>
    <image name="hlsl_SampleTexture2_294" type="color3">
      <input name="file" type="filename" nodename="hlsl_Sampler2D_v2_274" />
    </image>
    <output name="Color" type="color3" nodename="hlsl_SampleTexture2_294" />
  </nodegraph>
  <px_mx_test name="n_px_mx_test" type="color3">
    <input name="SD_CustomTexture" type="filename" value="some/texture/path.dds" />
  </px_mx_test>
  <standard_surface name="n_standard_surface" type="surfaceshader">
    <input name="base_color" type="color3" nodename="n_px_mx_test" />
  </standard_surface>
  <surfacematerial name="m_px_mx_test" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="n_standard_surface" />
  </surfacematerial>
</materialx>

```


kwokcb:
@niklasharrysson, I added in `dot` category recognition and add these nodes as part of the bypass logic since they are pass-through nodes. I didn't reuse constant categorization since it's checked in specific places for other logic branching.

jstone-lucasfilm (approved):

PR #1155, opened 2022-11-25
Add utility document to visualize geometric nodes

pablode:
This PR proposes a document that contains materials which visualize geometric properties as colors.

It can be used 1) to inspect and verify an asset's geometry (f.i. using MaterialXView) and 2) to compare the handling of geometric nodes across different MaterialX-supporting renderers (e.g. tangents in HdStorm).

The shader compilation error is fixed by commit [`3c9996b769`](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1156/commits/3c9996b769acdc31eb8e706d6d57844f3b2ae3d2) (or PR #1156).


kwokcb:
Hi @pablode ,
I'd suggest moving / merging this with what's in the unit test area here: https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/resources/Materials/TestSuite/stdlib/geometric
as test files are kept there.

jstone-lucasfilm:
Agreed that it would be valuable to merge these new tests with the existing `streams.mtlx` example document, which has a similar goal of providing debug visualization of geometry nodes:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/stdlib/geometric/streams.mtlx

On the other hand, our existing examples are challenging for new developers to find, and perhaps there's a better home for them within the MaterialX resources folder structure?  All ideas on this front are welcome.

jstone-lucasfilm:
@pablode I'll close out this issue for now, but feel free to propose improvements to the coverage and accessibility of `streams.mtlx` as needed in the future.

PR #1157, opened 2022-11-27
Add tangent input to glTF PBR

pablode:
This PR adds a _tangent_ input to the glTF PBR, similar to how it's done in the [standard surface](https://github.com/AcademySoftwareFoundation/MaterialX/blob/ef1b933b5c5192403aa4ae2d493a2b450670c934/libraries/bxdf/standard_surface.mtlx#L97-L98).


pablode:
Some background info: this allows the use of tangents that are not provided by the runtime via the 'Tworld' geomprop or `<tangent>` node. For example, tangents could be read direcly from a mesh stream (or USD primvar) using a `<geompropvalue>` node, and connected to the proposed input for correct shading.

jstone-lucasfilm:
@pablode This looks like a good proposal to me, and I'd have no reservations about integrating it to main.  Is there a representative on the glTF side who might want to weigh in here, making sure that this approach for connecting tangent streams is aligned with their vision as well?

proog128:
Looks good to me. As there is no anisotropy yet in glTF PBR, this change probably does not have an effect yet. However, as far as I can tell it is in line with the coming KHR_materials_anisotropy extension.

PR #1158, opened 2022-11-29
Fix syntax error in color transform MDL implementation.

niklasharrysson:


jstone-lucasfilm (approved):

PR #1159, opened 2022-11-29
Fix inconsistent use of "uniform" flag on gltf nodes.

niklasharrysson:
This change list adds a missing uniform flag to interface inputs which are connected to uniform inputs on internal nodes. An inconsistent use of the uniform flag will cause compilation of generated MDL code to failed, since MDL is strict about the use of the uniform keyword in source code.


jstone-lucasfilm (approved):

PR #1177, opened 2022-12-28
Fix missing basis orthogonalizations in GLSL backend

pablode:
I've encountered shading artifacts in normal mapped MaterialX materials, and upon closer inspection I believe that these are caused by the missing basis orthogonalizations @Tellusim brought to attention in PR #1049.

However, instead of using a second `normalize(cross(..))` call, I've applied the Gram-Schmidt algorithm.

Here are some comparisons from the soon-to-be-released test suite for my glTF to USD+MaterialX converter:

Name | Old | New | Diff
---  | --- | --- | ---
NormalTangentTest | NormalTangentTest_ref | NormalTangentTest_test | NormalTangentTest_diff
DamagedHelmet | DamagedHelmet_ref | DamagedHelmet_test | DamagedHelmet_diff
IridescentDishWithOlives | IridescentDishWithOlives_ref | IridescentDishWithOlives_test | IridescentDishWithOlives_diff
ClearCoatTest | ClearCoatTest_ref | ClearCoatTest_test | ClearCoatTest_diff

The changes can be tested using following model: 
[NormalTangentMirrorTest.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/10313740/NormalTangentMirrorTest.zip)

It requires a recent build of MaterialXView as it makes use of the support for explicit bitangents introduced in PR #1156.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
@@ -13,8 +13,8 @@ float mx_latlong_compute_lod(vec3 dir, float pdf, float maxMipLevel, int envSamp
 vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 alpha, int distribution, FresnelData fd)
 {
     // Generate tangent frame.
-    vec3 Y = normalize(cross(N, X));
-    X = cross(Y, N);
+    X = normalize(X - dot(X, N) * N);
+    vec3 Y = cross(N, X);
     mat3 tangentToWorld = mat3(X, Y, N);
 
     // Transform the view vector to tangent space.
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm:
Thanks @pablode, this looks like a great improvement to me.

Refining it one step further, perhaps we should wrap the new, repeated logic in an `mx_gram_schmidt_orthogonalize` helper function?  Its return value could be the new Y vector, and its input arguments could be the N vector and `inout` X vector.

One potential home for this new helper function would be:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/a345d134b865a6cdaf63e10cfc870df49ed841be/libraries/pbrlib/genglsl/lib/mx_microfacet.glsl

jstone-lucasfilm:
@pablode Let's plan to make that additional refinement in a future pull request, and I believe your existing change is worthwhile as it stands.  I'll run some visual tests in a local build to confirm your results, and then I believe we can merge this pull request to main.

jstone-lucasfilm (approved):
Thanks for the fixes, @pablode!

PR #1185, opened 2023-01-03
Simplify OSL node implementations

jstone-lucasfilm:
This changelist simplifies the OSL implementations of common nodes such as 'if', 'add', and 'subtract', removing duplicate code and harmonizing inline implementations where possible.


jstone-lucasfilm:
Updated render comparisons between GLSL and OSL:

[MaterialXRenderTests_01_02_2023.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/10333852/MaterialXRenderTests_01_02_2023.pdf)

libraries/stdlib/genosl/include/mx_funcs.h
```diff
@@ -5,49 +5,25 @@
...
 #include "color4.h"
 #include "vector2.h"
 #include "vector4.h"
 #include "matrix33.h"
 
+//
+// Support functions for OSL implementations of the MaterialX nodes.
+//
+
+float mx_ternary(int expr, float v1, float v2) { if (expr) return v1; else return v2; }
```
lgritz, line 18: Is this not more easily expressed with the ternary operator?

lgritz, line 18: N.B. You can say `expr ? v1 : v2` in OSL.

jstone-lucasfilm: @lgritz I took that approach at first, but it looks like OSL's built-in ternary operator doesn't yet handle values that are structures, which is the case when `v1` and `v2` are of type `color4`, `vector4`, `vector2`, or `matrix33`.  Let me know if I should write this up as an issue on the OSL GitHub, or if perhaps I'm missing a detail that would allow me to use this approach across all types.

lgritz: OH. Hmmm. Yeah, please file an issue. You can 'return' a struct, and you can assign a struct, so probably there's just some little missing bit of logic that prevents a struct from being the operands to a ternary operator.

Of course, 50/50 chance I simply don't remember that I looked at this ages ago and bailed when it turned out to be disproportionately difficult. Sometimes you turn over a rock, and when you see what's under there, you just put the rock back and slowly back away. :-)

PR #1186, opened 2023-01-04
Fix opacity edge case in UsdPreviewSurface

jstone-lucasfilm:
This changelist fixes an edge case for the opacity computation in UsdPreviewSurface, making surfaces visible when their non-zero opacity is exactly equal to the opacity threshold.

This aligns the MaterialX graph for UsdPreviewSurface with its specification (https://graphics.pixar.com/usd/release/spec_usdpreviewsurface.html), which states:

"... when opacityThreshold is greater than zero, the opacity values less than the opacityThreshold will not be rendered, and the opacity values greater than or equal to the opacityThreshold will be fully visible."

### DIFF
--- a/libraries/bxdf/usd_preview_surface.mtlx
+++ b/libraries/bxdf/usd_preview_surface.mtlx
@@ -272,12 +272,12 @@
       <input name="low" type="float" value="0.00001" />
       <input name="high" type="float" value="1.0" />
     </clamp>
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks to @klucknav and @spiffmon for noting this issue, and I've CC'ed them both for validation of this proposed change.

PR #1187, opened 2023-01-04
Fix metallic F90 in UsdPreviewSurface

jstone-lucasfilm:
This changelist fixes the computation of F90 in the metallic workflow of UsdPreviewSurface.

This aligns the MaterialX graph for UsdPreviewSurface with its specification (https://graphics.pixar.com/usd/release/spec_usdpreviewsurface.html), which states:

"If metallic is 1, then both F0 (reflectivity at 0 degree incidence) and edge F90 reflectivity will simply be the Albedo. If metallic is 0, then Albedo is ignored in the calculation of F0 and F90; F0 is derived from ior via ((1-ior)/(1+ior))^2 and F90 is white. In between, we interpolate."


jstone-lucasfilm:
Thanks to @klucknav and @spiffmon for noting this issue, and I'm additionally CC'ing @pablode for his thoughts on this proposed change.

Issue #1059, opened 2022-08-24
Incorrect usage of find_first_of in replacing {{in}} pattern


tomas-davidovic:
I am writing a very hacked together ShaderGen and found when I have `sourcecode="{0}"` where in the output I want to see literally `{0}`, MaterialX fails.

The problem is in the SourceCodeNode::emitFunctionCall method, which uses `_functionSource.find_first_of("{{")` to find the first occurrence of the `{{`. But that's not what `find_first_of` actually does. It finds the first occurrence of any of the characters in the provided string, so it finds even single `{`. The code should be using `find` instead, which finds the exact match. Same goes for the postfix search a bit later.

jstone-lucasfilm:
This is a great catch, @tomas-davidovic, and I believe this issue goes back to our very first implementation of SourceCodeNode::emitFunctionCall.  Would you have the bandwidth to put together a pull request addressing this, where you can validate the results with your local ShaderGen implementation?

tomas-davidovic:
I can try (my very first PR on github ever).

jstone-lucasfilm:
Thanks for this report, @tomas-davidovic, and this has been addressed in #1191.

PR #1191, opened 2023-01-07
Fix inline variable parsing

jstone-lucasfilm:
This changelist fixes the parsing of inline source code variables, where a single bracket would incorrectly be interpreted as a prefix/suffix marker.


jstone-lucasfilm:
This addresses issue #1059, reported by @tomas-davidovic.

PR #1195, opened 2023-01-15
Upgrade NanoGUI to fix Apple M1 builds

jdcole:
In previous version of NanoGUI M1 architecture wasn't detected properly (used CMAKE_OSX_ARCHITECTURES vs. CMAKE_SYSTEM_ARCH...) 

see https://github.com/mitsuba-renderer/nanogui/commit/c66b89e55025d0920ee904870a1cd678e780a63c


jstone-lucasfilm (approved):
Good catch, @jdcole, and this is an upgrade that we've been meaning to make for a while.

PR #1200, opened 2023-01-18
Add missing volumetric parameters to MDL surfacematerial

niklasharrysson:
This change list adds a missing transfer of volumetric parameters from surfaceshader node to material node in MDL implementation of <surfacematerial>.


jstone-lucasfilm (approved):
This looks good to me, @niklasharrysson, and thanks for the fix!

PR #1202, opened 2023-01-20
Fix Python wrappers for function callbacks

kwokcb:
**Fix** 
Need to include `<pybind11/functional.h>` for function callbacks (std::function refs) to be passable.
Otherwise the types are considered to be incompatible between a Python function and C++ function pointer.

**Test**
Add in new test into `MaterialXTest/main.py `to show write predicate working to skip library `nodedefs` on write.


jstone-lucasfilm (commented):
Good catch on the added include for `PyMaterialX.h`.  Instead of adding a new test script for this specific case, would it be possible to extend the existing unit tests in `main.py` to include a filtered read operation as well?

kwokcb:
> 

Yes. I forgot about the fact that tests were in `main.py`. 
New test function called `test_WriteXml()` has been added.

python/MaterialXTest/main.py
```diff
@@ -481,6 +481,44 @@ def test_ReadXml(self):
         mx.readFromXmlFile(doc, filename, _searchPath)
         self.assertTrue(doc.validate()[0])
 
+    def test_WriteXml(self):
```
jstone-lucasfilm, line 484: Since we're already testing XML serialization in the `test_ReadXml` command above, and we already have access to the full set of data libraries in that context, perhaps it makes more sense to add a handful of new code blocks to the existing test, rather than creating a brand new one?

If it would add clarity, I think it would be fine to rename the existing command `test_XmlIo`, to reinforce the idea that it covers XML serialization as well.

kwokcb: Should be simplified now to a single additional code block & did the test rename.

jstone-lucasfilm (approved):
Thanks @kwokcb, this looks great to me!

Issue #1203, opened 2023-01-23
Unlit h/w code optimization does not work with definitions


kwokcb:
There is an optimization to avoid emitting lighting fragments for h/w code generation

* if an "unlit" surfaceshader is defected does not work if the shader is inside a definition.
It will always omit lighting.
* Also the environment lighting information never performs any check so is always included
though unused.
* Finally for unit tests which empty out the direct lighting list, this results in a strange situation where
you emit the direct lighting code but not the maximum number of lights uniform.

Seems in part that the GLSL code has a requiresLighting() logic that is only used in one place but not others.

Example:
```
<?xml version="1.0"?>
<materialx version="1.38">
<!-- This does not optimize -->
  <nodedef name="ND_convert_color3_shader" node="convert">
    <input name="in" type="color3" value="0, 0, 0" />
    <output name="out" type="surfaceshader" />
  </nodedef>
  <nodegraph name="NG_convert_color3_shader" nodedef="ND_convert_color3_shader">
    <surface_unlit name="surface" type="surfaceshader" >
      <input name="emission_color" type="color3" interfacename="in" />
    </surface_unlit>
    <output name="out" type="surfaceshader" nodename="surface" />
  </nodegraph>

  <convert name="surface_in_def" type="surfaceshader">
    <input name="in" type="color3" value="1, 0, 0" />
  </convert>
  <surfacematerial name="material_add_color3_out" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="surface_in_def" />
  </surfacematerial>

<!-- This optimizes -->
   <surface_unlit name="surface_outside_def" type="surfaceshader" >
      <input name="emission_color" type="color3" value ="0, 1, 0" />
    </surface_unlit>
  <surfacematerial name="mat2" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="surface_outside_def" />
  </surfacematerial>

</materialx>
```

PR #1205, opened 2023-01-24
Add convert nodes for surface shader types

kwokcb:
1. Add `<convert>` variants for:
   * float
   * vec2, vec3, vec4
   * color3, color4
   * bool
   * integer
  
    to surface shader. This allows for users to more easily set up networks to sample upstream nodes. BSD and EDF to shader is not included at this time as code generation has issues with generating appropriate code. 

2. The output of a small subset to test the possible output types (including multitoutput) has been added to a new `TestSuite` `convert` area.  Results for GLSL and OSL: [convert_tests.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/10768658/convert_tests.pdf)

3. Code generation for GLSL, OSL and MDL has been tested with render tests for GLSL, OSL and reference implementation checking for the new variants in OSL included as part of the existing unit tests. Render tests results are attached for GLSL + OSL.
 (Note that every variant and every output on every nodedef in the core libraries was tested: [nodedef_tests.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/10485843/nodedef_tests.pdf))
 The script to create these tests is [genMaterialsFromDefs.py.txt](https://github.com/AcademySoftwareFoundation/MaterialX/files/10851635/genMaterialsFromDefs.py.txt)

4. There are issues with direct lighting being in an inconsistent state for unit tests has been patched. That is lights are set but there are no light inputs, resulting in only part of the lighting code to be emitted for unlit shaders in definitions -- see issue #1203.

@jstone-lucasfilm , @niklasharrysson : These are the definitions which seem to work without minimal issues with the caveat of no BSDF and EDF to shader nodedef and unlit shader code output.


source/MaterialXTest/MaterialXRender/RenderUtil.cpp
```diff
@@ -191,6 +191,10 @@ bool ShaderRenderTester::validate(const mx::FilePath optionsFilePath)
 
     setupTime.endTimer();
 
+    if (!options.enableDirectLighting)
```
kwokcb, line 194: @niklasharrysson . If this is not done then we end up in a state where the direct lighgin code is emitted but the definition of the uniforms for direct lighting are not for unlit shaders. There is some inconsistency it how things are emitted I believe due to the classification of the shader graph relying on checking if the root node is explicit a "ND_surface_unit" but it's the implementation perhaps needs to be detected?

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -3846,6 +3846,38 @@
     <input name="in" type="integer" value="0" />
     <output name="out" type="float" default="0.0" />
   </nodedef>
+  <nodedef name="ND_convert_color3_shader" node="convert" version="1.0" isdefaultversion="true" nodegroup="shader" doc="Convert color3 to shader">
```
niklasharrysson, line 3849: Perhaps the naming convention could be ND_xxx_surfaceshader, to clarify that output type is "surfaceshader", and avoid conflicts in the future with other shader types?

python/Scripts/definitionMaterials.py
```diff
@@ -0,0 +1,216 @@
+#!/usr/bin/env python
+'''
+Create a document per node definition such that an instance of that 
+definition is routed to a "material" root. Makes use of "convert"
+nodes for various output types. A material root is created for each
+output on a nodedef, if multiple exist.
+'''
```
niklasharrysson, line 7: Could this be considered a test script that can stay local and not be included in the main public repo? Or is it generally useful to be included here? Just a thought.

source/MaterialXTest/MaterialXRenderOsl/GenReference.cpp
```diff
@@ -63,7 +63,7 @@ TEST_CASE("GenReference: OSL Reference", "[genreference]")
     // Ignore the following nodes:
     const mx::StringSet ignoreNodeList = { "surfacematerial", "volumematerial",
                                            "constant_filename", "arrayappend",
-                                           "dot_filename"};
+                                           "dot_filename", "convert_BSDF_shader", "convert_EDF_shader", };
```
niklasharrysson, line 66: Was this left here by mistake? As far as I can see there are no convert nodes for BSDF or EDF yet.

resources/Materials/TestSuite/nodedefs/add_color3.mtlx
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0"?>
+<materialx version="1.38">
+  <add name="add_color3" type="color3">
+    <input name="in1" type="color3" value="0.0, 0.0, 0.0" />
+    <input name="in2" type="color3" value="0.0, 0.0, 0.0" />
+  </add>
+  <convert name="shader_add_color3_out" type="surfaceshader">
```
niklasharrysson, line 7: There is some overlap between the new tests for add_xxx and convert_xxx below. Could these be unified, or maybe the add tests excluded? Just thinking we should avoid testing the same thing several times to reduce running cost.

libraries/stdlib/stdlib_defs.mtlx
kwokcb: Sounds good. I'll make the change.

python/Scripts/definitionMaterials.py
kwokcb: I think generating these graphs creates a smaller subset of graphs for test than what was hand-done.
So was thinking that this could be done as part of the test setup. 
Also maybe it can be moved to MaterialXTest for now as scripts in that folder are not part of the distribution.

source/MaterialXTest/MaterialXRenderOsl/GenReference.cpp
kwokcb: I'll remove this. I was trying to get BSDF and EDF convert to work but there are issues and was removed.

resources/Materials/TestSuite/nodedefs/add_color3.mtlx
kwokcb: Let me take a look. These are generated so if there is overlap then they could be excluded.

niklasharrysson: I was just thinking that the add nodes where tested elsewhere and the convert nodes are tested by the other tests you added. But maybe it would be good to have generated tests for _all_ nodedefs, to have better test coverage. Would be cool if we could autogenerate these test files as part of the build or as part of running the testsuite, so they don't have to be checked into the repo as static mtlx documents. But that's probably for another PR :)

kwokcb: Yes. For now they are excluded from the test suite as Jonathan made the change to only include certain folders. I think a lot of the stdlib 'trivial' tests and a number of `pbrlib` tests can be removed as part of this. Shader / material "Examples" and should remain as it's hard to create useful tests for this automatically.

python/Scripts/definitionMaterials.py
kwokcb: I've revamped the script so that it can handle generating material graphs from any file which has definitions.
So the TestSuite/stdlib/surface.mtlx which has a checker definition can have a graph generated. See updated check-in comment.

niklasharrysson (approved):

source/MaterialXCore/Interface.cpp
```diff
@@ -337,7 +337,7 @@ bool Input::validate(string* message) const
     if (parent->isA<Node>())
     {
         bool hasValueBinding = hasValue();
-        bool hasConnection = hasNodeName() || hasNodeGraphString() || hasOutputString() || hasInterfaceName();
+        bool hasConnection = hasNodeName() || hasNodeGraphString() || hasOutputString() || hasInterfaceName() || hasDefaultGeomPropString();
```
jstone-lucasfilm, line 340: I'm not sure this change makes sense, since node inputs don't support `defaultgeomprop` strings, so the presence of one wouldn't constitute a meaningful value or connection for the node input.

Is there a case that I'm missing here, where `defaultgeomprop` might be meaningful on a node input, as opposed to its more traditional usage on a nodedef input?

libraries/bxdf/gltf_pbr.mtlx
```diff
@@ -25,7 +25,7 @@
     <input name="emissive" type="color3" value="0, 0, 0" uimin="0, 0, 0" uimax="1, 1, 1" uiname="Emissive" uifolder="Emission" />
     <input name="emissive_strength" uniform="true" type="float" value="1" uimin="0" uiname="Emissive Strength" uifolder="Emission" />
     <input name="thickness" uniform="false" type="float" value="0" uimin="0" uiname="Thickness" uifolder="Volume" />
-    <input name="attenuation_distance" uniform="true" type="float" uimin="0" uiname="Attenuation Distance" uifolder="Volume" />
+    <input name="attenuation_distance" uniform="true" type="float" value="1e+36" uimin="0" uiname="Attenuation Distance" uifolder="Volume" />
```
jstone-lucasfilm, line 28: Perhaps this collection of data library fixes should become its own pull request?  Specifically for this one, I believe we should bring Khronos into the process of validating it.

resources/Materials/TestSuite/nodedefs/convert_boolean_surfaceshader.mtlx
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+<materialx version="1.38">
+  <convert name="convert_boolean_surfaceshader" type="surfaceshader" version="1.0">
```
jstone-lucasfilm, line 3: Would these new tests make more sense as part of the existing `convert.mtlx` file in the test suite?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/stdlib/channel/convert.mtlx

jstone-lucasfilm (commented):
This looks like a good start, and I've written up a few thoughts on clarifying this pull request further.

source/MaterialXCore/Interface.cpp
kwokcb: This is currently producing erroneous warning, so I guess an additional check for if it's a `nodedef` input should be added.
Currently tests fail since we skip on validation warnings.

libraries/bxdf/gltf_pbr.mtlx
kwokcb: Yes. I can move this out.

resources/Materials/TestSuite/nodedefs/convert_boolean_surfaceshader.mtlx
kwokcb: These were generated so I left them as is, but good suggestion to group them together. I'll make the change.

source/MaterialXCore/Interface.cpp
jstone-lucasfilm: I'm not seeing any warnings or errors when I run the render test suite for MaterialX, and I'd imagine it would represent a real error if one appeared in a future build, so I'd recommend that we fix the error rather than suppressing it here.

python/MaterialXTest/genMaterialsFromDefs.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python
+'''
+Utility to create a document per node definition such that an instance of that
```
jstone-lucasfilm, line 3: This seems more like a pipeline-specific utility script, rather than a component of our standard Python test suite.  Maybe this should be omitted from the pull request, and we can consider individual elements of this for inclusion further in the future?

kwokcb: True enough. Script has been removed.

kwokcb:
[Sug](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1266) to simplify the number of unit tests by generating them using something like the generation script mentioned here has been added for discussion.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1347,7 +1347,8 @@ string ShaderGraph::populateColorTransformMap(ColorManagementSystemPtr colorMana
     }
 
     const string& sourceColorSpace = input->getActiveColorSpace();
-    if (shaderPort && !sourceColorSpace.empty())
+    // Note that "none" means no transform should be applied.
+    if (shaderPort && !sourceColorSpace.empty() && sourceColorSpace != "none")
```
jstone-lucasfilm, line 1351: This is a good proposal for the future, but I'd recommend keeping this separate from the `convert` node improvement, since this change has the potential to alter the visuals associated with existing materials.

Additionally, based on conversations with Naty Hoffman and other experts in this area, we may end up removing the "none" markings from pseudo-colors in PBR nodes, and that change ought to be made before activating the render-time effect of the marking.

source/MaterialXCore/Interface.cpp
```diff
@@ -334,7 +334,7 @@ bool Input::validate(string* message) const
     {
         validateRequire(getDefaultGeomProp() != nullptr, res, message, "Invalid defaultgeomprop string");
     }
-    if (parent->isA<Node>())
+    else if (parent->isA<Node>())
```
jstone-lucasfilm, line 337: Wouldn't this change introduce a subtle bug, where a node input with a `defaultgeomprop` string but no binding would now be considered valid?

My thinking here is that only `nodedef` inputs support `defaultgeomprop` attributes in MaterialX, so the presence of this attribute on a _node_ input should have no effect,.

Here's the section of the specification that discusses `defaultgeomprop` attributes in the context of `nodedef` inputs, and let me know if I'm missing a section that might discuss additional use cases:

https://materialx.org/assets/MaterialX.v1.38.Spec.pdf#page=51

dbsmythe: Yes, that is the correct intended interpretation: defaults, including "defaultgeomprop"s, can only be defined on nodedefs, not in node instances.

source/MaterialXCore/Interface.cpp
```diff
@@ -332,6 +332,10 @@ bool Input::validate(string* message) const
 
     if (hasDefaultGeomPropString())
     {
+        if (!parent->isA<NodeDef>())
```
kwokcb, line 335: @jstone-lucasfilm , @dbsmythe. This should fix the defaultgeomprop test to be compliant.

source/MaterialXCore/Interface.cpp
kwokcb: This should be fixed now. Please have a look. I added in a extra check that the parent must be a `<nodedef>` for an input to have a `defaultgeomprop`. Thanks.

source/MaterialXCore/Interface.cpp
```diff
@@ -332,6 +332,7 @@ bool Input::validate(string* message) const
 
     if (hasDefaultGeomPropString())
     {
+        validateRequire(parent->isA<NodeDef>(), res, message, "Invalid defaultgeomprop on non-definition input");
```
kwokcb, line 335: This should catch invalid `defaultgeomprop` properly now. Fixed a unit test.

source/MaterialXCore/Interface.cpp
kwokcb: I've added in what should be the proper test now and fix a failing unit test.

jstone-lucasfilm (approved):

Issue #1211, opened 2023-02-01
MaterialXEditor : Can crash on invalid file load / lack of filtering 


kwokcb:
Exceptions don't seem to be caught on file load so will crash the editor.
It would be good to add in this check here, and maybe other places where an exception can be thrown.

As a sug, it would be useful to add in file type filtering as I tried to load in an image as a mtlx file.
e.g. add something like this in on load and save:
```
    mx::StringVec mtlxFilter;
    mtlxFilter.push_back(".mtlx");

    _fileDialog.SetTypeFilters(mtlxFilter);
    _fileDialogSave.SetTypeFilters(mtlxFilter);
```

jstone-lucasfilm:
Thanks for addressing these issues in #1218, @kwokcb, and I'll mark the original report as closed.

PR #1218, opened 2023-02-01
Graph editor file checking and filtering

kwokcb:
Update #1211
Add some basic filtering and exception handling for file load.
Note: There is no "logging" function or feedback panel to route messages to so they go to stderr as with current messages.


jstone-lucasfilm (approved):
Thanks for the fixes, @kwokcb, and this change looks good to me!

PR #1219, opened 2023-02-02
Add UiNode null checks

lfl-eholthouser:
This pull request is attempting the solve the issue with the Graph Editor where it fails when a node is clicked on MacOs


jstone-lucasfilm (approved):
Looks good to me, thanks @lfl-eholthouser!

PR #1221, opened 2023-02-02
Merge RenderView loading methods

jstone-lucasfilm:
This changelist merges shared logic between the RenderView::loadDocument and RenderView::updateMaterials methods, clarifying the role of each.

Additionally, the RenderView::_showAllInputs variable has been removed, as this property is now stored for each UiNode.


lfl-eholthouser (approved):
Everything looks great to me!

PR #1225, opened 2023-02-05
Fix build issue with Win32 APIs when using UNICODE

walbourn:
Per the *UTF-8 Everywhere Manifesto*, it's best to not rely on the legacy ANSI .vs UNICODE macros for Win32 functions that take strings. Since you explicitly use a ``string`` for all your Win32 method calls, it's best to use the "A" variant directly instead of relying on the macro.

> This causes problems when trying to build your library using vcpkg manager when the setup adds the recommended UNICODE _UNICODE defines.


jstone-lucasfilm:
This looks like a good fix to me, @walbourn, and we'd be happy to merge the change once CLA issues are resolved.  Let us know if you run into issues with CLA approval, and the Linux Foundation is usually very responsive if you submit a support ticket as noted above.

jstone-lucasfilm (approved):
Thanks for the fix, @walbourn!

PR #1226, opened 2023-02-06
Refinements to UiNode class

jstone-lucasfilm:
- Replace custom UiNode operators with explicit logic in Graph::layoutPosition, since these custom operators can potentially be invoked by different compilers in unexpected ways.
- Merge two similar UiNode constructors, reducing the number of unique code paths to maintain.
- Pass constant method arguments by const reference.


lfl-eholthouser (approved):
Looks great to me!

Issue #1228, opened 2023-02-06
MaterialXEditor: Missing various input and output interface types "nodes"


kwokcb:
Currently the types of inputs and outputs is fixed and is a subset of the available types (as specified in stdlib_defs.mtlx).
Thus they are missing on load and cannot be created interactively.

This prevents users from creation using graphs with these types of inputs / outputs.
In particular PBR and all shader types are missing, matrix, arrays, geomname, lights and materials.

jstone-lucasfilm:
Thanks for this report, @kwokcb, and thanks to @niklasharrysson for the fix!

kwokcb:
Thanks @niklasharrysson for fixing this. I had one more thing that I think would be good to add here.

@jstone-lucasfilm , the 'none' type is no longer used anywhere so I think it should be excluded from being able to be created. I believe it was nixed when `backdrop` stopped being a node, but I think the `typedef` is still required to perform upgrades of older files. Suggest to filter it out so you can't create these?

PR #1229, opened 2023-02-07
Support for creating inputs and outputs of all types in the Graph Editor 

niklasharrysson:
This change list addresses issue #1228, where the menu for creating input and output nodes is now filled according to all typedefs in the loaded document.

Also addresses a crash issue when deleting empty nodegraphs or group nodes.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -2655,59 +2655,63 @@ void Graph::deleteNode(UiNodePtr node)
             }
         }
     }
-    // update downNode info
-    std::vector<Pin> outputConnections = node->outputPins.front().getConnections();
 
-    for (Pin pin : outputConnections)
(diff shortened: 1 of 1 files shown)


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2655,59 +2660,63 @@ void Graph::deleteNode(UiNodePtr node)
             }
         }
     }
-    // update downNode info
-    std::vector<Pin> outputConnections = node->outputPins.front().getConnections();
 
-    for (Pin pin : outputConnections)
+    if (node->outputPins.size() > 0)
```
niklasharrysson, line 2664: This diff looks messy.. But the only change is this one line guard (and added indentation).

jstone-lucasfilm (approved):
Thanks for these improvements, @niklasharrysson!

Issue #1230, opened 2023-02-07
MaterialXEditor: shader and material nodes cannot be created inside nodegraphs


kwokcb:
Seems these are always created at the top level even when trying to create them inside a nodegraph.
So you can't create new surface shader graphs and materials in nodegraphs and it's confusing since the UI
shows them in the graph but they are saved in the Document outside the graph.

Seems like a simple fix to remove the code which forces them to be created at the top level.

jstone-lucasfilm:
Closing this issue as it's been addressed in #1231.

PR #1231, opened 2023-02-07
Allow shader and material creation in nodegraphs

kwokcb:
Update #1230 

Fix so that shaders and materials are not place at the top level while the UI shows them being created in the proper place. Minor name cleanup to create name first to use at creation time.

Also added small firewall check for a possible crash when deleting nodegraphs.


jstone-lucasfilm (approved):

PR #1233, opened 2023-02-08
Fix installation issue with OSL legacy closures

rasmusbonnedal:
When building with MATERIALX_OSL_LEGACY_CLOSURES=ON CMake would install the non-legacy pbrlib_genosl_impl.mtlx first and then it would find that it is already up-to-date and not install the correct legacy on top of it. This commit fixes this by excluding both the legacy and the non-legacy file from the glob install and then installing the correct one explicitly.

Note that if the user changes the value of MATERIALX_OSL_LEGACY_CLOSURES and then installs again the correct file will NOT be installed.

### DIFF
--- a/libraries/CMakeLists.txt
+++ b/libraries/CMakeLists.txt
@@ -1,9 +1,13 @@
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/
         DESTINATION "${MATERIALX_INSTALL_STDLIB_PATH}" MESSAGE_NEVER
         PATTERN "CMakeLists.txt" EXCLUDE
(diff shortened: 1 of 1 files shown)


rasmusbonnedal:
I have tested install with `MATERIALX_OSL_LEGACY_CLOSURES` both ON and OFF to make sure the correct file is installed in each case. Tested on VS2019 and CMake 3.25.2.

niklasharrysson (approved):
Thanks for the fix @rasmusbonnedal, this looks good to me!

jstone-lucasfilm (approved):
Looks good to me, thanks @rasmusbonnedal!

PR #1234, opened 2023-02-08
Add normal to gltf_pbr translation graph

rasmusbonnedal:


libraries/bxdf/translation/standard_surface_to_gltf_pbr.mtlx
```diff
@@ -17,6 +17,7 @@
     <input name="coat_roughness" type="float" value="0.1" />
     <input name="emission" type="float" value="0" />
     <input name="emission_color" type="color3" value="1, 1, 1" />
+    <input name="normal" type="vector3" value="0.5, 0.5, 1.0" />
```
pablode, line 20: The value seems to come from the Standard Surface to UsdPreviewSurface translation graph, and I don't think it is correct in this context. We should use `defaultgeomprop="Nworld"` instead.

pablode:
While we're at it, I think we should add support for the _tangent_ input in a similar way!

jstone-lucasfilm:
@rasmusbonnedal This looks like a step in the right direction, but as @pablode notes, we'll need to handle the novel situation that both Standard Surface and glTF PBR take _world-space_ normal inputs rather than tangent-space values.

In addition to the default value change that Pablo suggests above, I believe we'd need to refine the logic that optimizes away inputs at their default values during texture baking, adding handling for the `defaultgeomprop` case as well:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXRender/TextureBaker.inl#L289

PR #1238, opened 2023-02-10
Improvements to editor main loop

jstone-lucasfilm:
- Clarify the scope and ownership of the editor context.
- Create the editor configuration and context before the main loop.
- Poll the cursor position before graph layout to minimize input latency.


lfl-eholthouser (approved):
Looks good to me!

PR #1239, opened 2023-02-12
Simplifications to editor headers

jstone-lucasfilm:
This changelist simplifies the header files in MaterialXGraphEditor, with no changes to editor logic or behavior.

- Remove a bidirectional dependency between Graph.h and UiNode.h, moving the UiEdge class class from Graph.h to UiNode.h.
- Move include directives from headers to source code where possible.
- Remove unused include directives where possible.


lfl-eholthouser (commented):
Looks great to me!

PR #1241, opened 2023-02-13
Add Python bindings for TargetDef and AttributeDef

kwokcb:
There are no new C++ APIs. They are just missing from the Python API

## Changes

* Add in `AttributeDef` Python interface
* Add in `TargetDef` Python interface
* Add in missing Python  interfaces on `Document` and `ShaderGenerator` using these interface and metadata registration

## Test
* Update `genshader` Python test to test `TargetDef` to find targets and match against the sahder generator
* Update `gensahder` test to add custom `AttributeDef` and test that they make it through OSL code gen. Resulting code should have custom string data `node_name` which is associated with the std surface nodedef:
```c++
surface standardSurface1
[[
    string mtlx_category = "standard_surface",
    string mtlx_name = "standardSurface1",
    string help = "Autodesk standard surface shader",
    string node_name = "Standard_Surface_Number_1"
]]
```


jstone-lucasfilm (approved):
Thanks for putting this change together, @kwokcb, and it looks good to me!

PR #1242, opened 2023-02-13
Use ImGui constants in editor

jstone-lucasfilm:
This changelist switches from GLFW to ImGui key constants in the editor, removing the need to include GLFW headers for input handling.


lfl-eholthouser (commented):
Looks great!

PR #1244, opened 2023-02-14
Update to UiPinPtr in editor

jstone-lucasfilm:
This changelist updates the graph editor to use UiPinPtr (std::shared_ptr<UiPin>) for storing and referencing individual pins, removing some unneeded copies from common editor operations.


lfl-eholthouser (commented):
Looks good to me!

Issue #1213, opened 2023-02-01
MaterialXEditor Sug: Would be useful to handle high dpi


kwokcb:
The doesn't seem to be any high dpi handling so the UI, esp fonts can appear very small.
It seems imgui is not well set up for this, but assume folks have something out there.

Here is a snapshot (4K monitor) between the editor and a [site](https://floooh.github.io/sokol-html5/imgui-highdpi-sapp.html) 
that seems to do scaling properly.

![image](https://user-images.githubusercontent.com/49369885/216139697-3badd9c1-7319-476f-84ab-0e6133c105de.png)

Seems `ScaleAllSizes()` could be the thing to call after finding the DPI scaling. Is mentioned [here](https://github.com/ocornut/imgui/issues/5452) for instance.

Would be nice to look at for the sake of my bad eyes but maybe others as well :).

jstone-lucasfilm:
Thanks for this report, @kwokcb, as well as for the fix in #1248!

PR #1248, opened 2023-02-15
Support DPI scaling in Graph Editor

kwokcb:
Update #1213

## Fix
* Detect DPI scale for monitor. For now the primary monitor will be used as there is no good solution for multi-monitor support. Sets ImGui global padding / spacing values. Font's must be manually handled but since they are already being scaled just update to use a "DPI scaling factor". 

## Test
* Tested by changing the DPI on monitor with a comparison with editor and other app ImGui app which handles scaling. This is with a scaling of 2.0. The scaling now matches (though the font used differs).
image


lfl-eholthouser (commented):
This looks great to me! Thanks for adding this.

jstone-lucasfilm (approved):
This looks great to me as well, thanks @kwokcb!

PR #1249, opened 2023-02-16
Fix node linking issue in editor

lfl-eholthouser:
This pull request addresses and issue where edges were not being properly deleted when a node was connected to the same node through multiple different input pins. The remove edge function now relies on both the upstream node id as well as the downstream node pin in order to identify and remove the edge.


jstone-lucasfilm (approved):
This looks great to me, thanks @lfl-eholthouser!

PR #1250, opened 2023-02-17
Fix elision of dot nodes

jstone-lucasfilm:
This changelist fixes a bug in the elision of dot nodes (#1152), where dot nodes with channels attributes would be incorrectly bypassed, causing the shader generator to produce incorrect code.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -1020,13 +1020,21 @@ void ShaderGraph::optimize(GenContext& context)
     size_t numEdits = 0;
     for (ShaderNode* node : getNodes())
     {
-        if (node->hasClassification(ShaderNode::Classification::CONSTANT) ||
-            node->hasClassification(ShaderNode::Classification::DOT))
+        if (node->hasClassification(ShaderNode::Classification::CONSTANT))
         {
-            // Constant and dot nodes can be removed by moving their value
-            // or connection downstream.
+            // Constant nodes can be elided by moving their value downstream.
             bypass(context, node, 0);
             ++numEdits;
+        }
+        else if (node->hasClassification(ShaderNode::Classification::DOT))
+        {
+            // Dot nodes without modifiers can be elided by moving their connection downstream.
+            ShaderInput* in = node->getInput("in");
+            if (in->getChannels().empty())
+            {
+                bypass(context, node, 0);
+                ++numEdits;
+            }
         }
         else if (node->hasClassification(ShaderNode::Classification::IFELSE))
         {


jstone-lucasfilm:
CC'ing @kwokcb to make sure this change looks reasonable!

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1020,14 +1020,22 @@ void ShaderGraph::optimize(GenContext& context)
...
-            node->hasClassification(ShaderNode::Classification::DOT))
+        if (node->hasClassification(ShaderNode::Classification::CONSTANT))
         {
-            // Constant and dot nodes can be removed by moving their value
-            // or connection downstream.
+            // Constant nodes can be elided by moving their value downstream.
             bypass(context, node, 0);
             ++numEdits;
         }
+        else if (node->hasClassification(ShaderNode::Classification::DOT))
```
kwokcb, line 1029: I know that according to the spec you can put "channels" on any input but I had always
assumed `dot `means "pass-through", but this is "pass-through with extraction".

Would it be better to not allow this and have folks add convert nodes as to me this is an edge case.
Especially as it's not something that the node editor can support currently.
(There is an issue with "channels" not being removed on disconnection which I'll log an issue for).

I assume this is the use case?
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <nodegraph name="test">
    <dot name="dot_float" type="float" xpos="5.572464" ypos="-1.284483">
      <input name="in" type="float" nodename="source_color3" channels="r" />
      <input name="note" type="string" value="test" />
    </dot>
    <multiply name="source_color3" type="color3" xpos="4.130435" ypos="-0.896552">
      <input name="in1" type="color3" value="0.893023, 0.0581504, 0.0581504" />
      <input name="in2" type="color3" value="0.00441322, 0.303115, 0.948837" />
    </multiply>
    <output name="output_float" type="float" nodename="dot_float" xpos="6.768116" ypos="-1.181034" />
  </nodegraph>
  <surfacematerial name="surfacematerial" type="material" xpos="4.594203" ypos="-1.146552">
    <input name="surfaceshader" type="surfaceshader" nodename="standard_surface_surfaceshader" />
  </surfacematerial>
  <standard_surface name="standard_surface_surfaceshader" type="surfaceshader" xpos="1.862319" ypos="-0.732759">
    <input name="base" type="float" output="output_float" nodegraph="test" />
  </standard_surface>
</materialx>
```

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1020,14 +1020,22 @@ void ShaderGraph::optimize(GenContext& context)
...
+        {
+            // Dot nodes without modifiers can be elided by moving their connection downstream.
+            ShaderInput* in = node->getInput("in");
+            if (in->getChannels().empty())
```
kwokcb, line 1033: BTW, if we add this, then the type should be checked as validation fails on this, but this logic does not.
e.g. if it's a string. then compilation will fail.

kwokcb:
Hi @jstone-lucasfilm , I put more details in the code area, but I think it's worthwhile to not allow "channels" on dot nodes and have users do things explicitly as they could never create this in the node editor (from example I added). 

image

If they could then I think along the thoughts of moving towards a code gen friendly graph, then these could be pre-converted
from implicit to explicit.

source/MaterialXGenShader/ShaderGraph.cpp
jstone-lucasfilm: Agreed that it's not a great practice to set the `channels` string on a `dot` node, but it's supported in both the specification and codebase for MaterialX, and I see this pattern in production assets.

For now, I think we need to maintain support for this feature in 1.38.7, since it has been present in all previous versions of our shader generators, though I'd be fine with encouraging better practices in the future.

source/MaterialXGenShader/ShaderGraph.cpp
jstone-lucasfilm: Can you clarify what you mean here?

jstone-lucasfilm:
@kwokcb Although it's not yet possible to create connections via `channels` strings in the editor, that's just a temporary limitation, rather than a deprecation of the `channels` string feature.  In the future, the goal is to support `channels` string connections in the editor, and this ought to work for `dot` nodes as well, since there is no rule in the specification or codebase to discourage this.

source/MaterialXGenShader/ShaderGraph.cpp
dbsmythe: Personally, I would lean toward changing the spec (1.39 and forward) so Dot nodes *cannot* set `channels`.  The intent of a <dot> node is to be a pure pass-through for nodegraph cleanliness (to help keep connection "noodles" from crossing over everywhere) and possibly documentation (since dot nodes can have "doc" strings).  If you want to extract channels, that should be done on the input of the node the chain of <dot>s feed into, or use an explicit <extract> or <convert> node.

jstone-lucasfilm: That's a reasonable proposal, @dbsmythe, though I wonder if we want to open the door to these node-specific rules for `channels` attributes, since none are currently present in MaterialX.  Would the same rule apply to other nodes such as `constant`, for example?  And what about custom nodes with similar behaviors to `dot` and `constant`?

dbsmythe: For "Constant", if you connected its lone "value" input to something, I think it's reasonable to allow a `channels` attribute, or at least I don't see a compelling argument to _disallow_ `channels`, precisely for the "slippery slope" argument I believe you're making.  The "dot" node is special, in that it is intended to just be a nodegraph shaping element and specifically not do any processing to data.

jstone-lucasfilm: That's a good argument, and I think you're right about adding this rule to MaterialX 1.39.  For now, I believe we'll need to deal with existing `dot` nodes with `channels` strings in 1.38, so we'll still need to support it in code generation for the moment, and we can auto-upgrade this usage to a better pattern in 1.39.

source/MaterialXGenShader/ShaderGraph.cpp
jstone-lucasfilm: Maybe you're referring to the presence of an empty `channels` string on an `input` element in MaterialX?  That would indeed emit a validation warning, but this object is actually a `ShaderNode` within the shader generation system, which uses an empty `channels` string to mean no channels attribute is present.

PR #1251, opened 2023-02-17
UI Enhancements for MaterialX Graph Editor

kwokcb:
UI enhancement proposals:

* Add in a file menu with hotkeys
  * Move the exisitng buttons to menus, with CTRL-O, N, S for open, new and save
    

* Add in a help menu.
  * Add in all the hot-keys, and mouse short-cuts.
  * Added `L` for layout since it has none.
  * Could add a menu for graph for "hidden" functionality.
    

* Add in Viewer menu.
  * For now ad in geometry load as is possible from command line

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -825,10 +825,14 @@ void Graph::updateMaterials(mx::InputPtr input, mx::ValuePtr value)
     }
 }
 // set the value of the selected node constants in the node property editor
-void Graph::setConstant(UiNodePtr node, mx::InputPtr& input)
+void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIProperties& uiProperties)
 {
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm:
This looks excellent so far, thanks @kwokcb!

One initial thought:

Since `Auto Layout` has the potential to undo an artist's choices with respect the layout of a graph, maybe we shouldn't assign this command to a hot key?

lfl-eholthouser (commented):
This looks great. Thanks @bernardkwok! One minor note for the help window is that Ctrl-F is used instead of Ctrl-S to find a node by its name

kwokcb:
Thanks for the feedback. I'll not add any new keys for now and fix the Ctrl-F. If it's okay, I'm thinking to move "Automatic Layout" under a "Graph" menu so existing hot-key items as well as new items can reside there.

jstone-lucasfilm (approved):
Great work, thanks @kwokcb!

PR #1254, opened 2023-02-21
Font scale updates for High DPI

lfl-eholthouser:
This pull requests address issues where a larger dpiScale would increase the font size but not the node size and therefore the text was larger than the node. It also adjust the spacing for the layout so the nodes no longer overlap.


jstone-lucasfilm (approved):
Thanks for the fix, @lfl-eholthouser!

PR #1255, opened 2023-02-23
Fix binding of geomprops to functional graphs in GLSL

niklasharrysson:
This change list adds binding of geomprop data to functional graphs that has inputs with `defaultgeomprop` set. 

As a result we can now render such functional graphs directly in the MaterialX Viewer and Graph Editor. Examples includes the implementation graph for Standard Surface, UsdPreviewSurface, etc. Previously they would render black if visualized directly, since data like normals and tangents were not bound to the shader.

This changes applies to HW targets only. It's already supported in OSL and MDL where this data is assigned directly as default values on the shader inputs.


jstone-lucasfilm (approved):

PR #1256, opened 2023-02-24
Added Native File Browser for Graph Editor

dgovil:
This PR adds support for using a platform native file browser on macOS to the MaterialXGraphEditor.

It introduces a new class that matches the Imgui File Dialog API, though limited to just the calls used by the GraphEditor.
Being able to use the native browser allows for quicker navigation to locations the user may have saved, and allows handling of some cases like cloud synced files.


dgovil:
I went through the CLA registration process and got a confirmation. Not sure if there's a lag time between that and the test updating here

source/MaterialXGraphEditor/Graph.h
```diff
@@ -210,10 +215,17 @@ class Graph
     bool _delete;
 
     // file dialog information
+#if defined(__APPLE__)
```
kwokcb, line 218: Hi @dgovil , I'm wondering if it's better to wrap this up in a platform agnostic class and put platform specific code there. Also would be be more consistent to use `zenity` as is done with [NanoGui](https://github.com/mitsuba-renderer/nanogui/blob/f5020e2f3e5114d517642e67afbb21cb88cf04c0/src/common.cpp#L406) ? I'm not sure of the pros / cons of using this but could be used for Mac and Linux, with Windows using specific code.

kwokcb (commented):
Would be great to get rid of the built in browser on desktop. I "think" it would still be useful as a runtime option however as believe that all UI drawn using ImGui constructs so maintains consistency with the rest of the UI which can have customizable attributes set, and maybe also is more cross-platform (non-desktop) ?

jstone-lucasfilm:
In a Slack discussion with @dgovil, he suggested the idea of importing the existing cross-platform file browser from NanoGUI into MaterialXGraphEditor, effectively replacing the dependency of the graph editor on ImGuiFileBrowser.  This sounds like a good approach to me, and further in the future we might consider moving the imported file browser code from MaterialXGraphEditor into MaterialXRender, making it available for other example applications to use.  This future step would require either raising the minimum supported platform for MaterialXRender to C++17, or replacing the `std::filesystem` references in the file browser code with equivalent functions from MaterialXFormat.

dgovil:
@jstone-lucasfilm and @kwokcb , I've updated this PR to include the code from the NanoGUI end so in theory it should work on macOS, Windows and Linux. I've only tested this PR on macOS since I don't have a Linux or Windows machine handy.

Would someone else be able to test+modify the PR for those platforms? I think I captured all the includes and CMake configurations that might be needed, but I am not sure of course.

Thanks!

kwokcb:
Hi @dgovil, 
I've checked out Windows. Looks good from my testing.
I don't have access to modify your branch but here's the diff to fix windows builds and clear the file filter properly. 
[patch.diff.txt](https://github.com/AcademySoftwareFoundation/MaterialX/files/11091833/patch.diff.txt)
 
Sorry I don't have Linux. Add in @lfl-eholthouser, to see if she has it, and for review.

lfl-eholthouser:
> Hi @dgovil, I've checked out Windows. Looks good from my testing. I don't have access to modify your branch but here's the diff to fix windows builds and clear the file filter properly. [patch.diff.txt](https://github.com/AcademySoftwareFoundation/MaterialX/files/11091833/patch.diff.txt)
> 
> Sorry I don't have Linux. Add in @lfl-eholthouser, to see if she has it, and for review.

I tried it out and its working for me on Linux. I'm getting a Linux error caused by IM_ifgreater_vector3_genglsl  that's affecting the rendering so nothing shows up in the render view. I think this is related to a different change so I'm going to keep investigating that.

jstone-lucasfilm (approved):
Thanks for putting this proposal together, @dgovil, and this looks good to me!

PR #1257, opened 2023-02-24
Set MaterialX Graph Editor config path directory

dgovil:
The Imgui library used by the MaterialXGraphEditor will leave an `imgui.ini` file in the current working directory.
This causes a lot of files littered around the place , unless the user always launches with the same cwd.

This PR changes that so it will store the config in a specified directory.

I've opted to only put this in the GraphEditor library for now, since it's the only thing that would need to use it.
However, if other parts of MaterialX need to use a config directory, it's an easy enough function to just hoist out at a later time.

The logic for finding a config directory is as follows:

1. Start with respecting the users `XDG_CONFIG_HOME` variable if set. This is Linux convention but applies to most Posix based systems
2. If it's not set, it'll default to the fallback values for `XDG_CONFIG_HOME` which is `~/Library/Preferences` on macOS and `~/.config` on Linux.
3. If either of the above conditions fail, it just fallsback to the current behaviour of using the current working directory.

I do not have a windows machine to test adding windows specific behaviour here.

### DIFF
--- a/source/MaterialXGraphEditor/Main.cpp
+++ b/source/MaterialXGraphEditor/Main.cpp
@@ -5,6 +5,8 @@
 
 #include <MaterialXGraphEditor/Graph.h>
 #include <MaterialXFormat/Environ.h>
+#include <MaterialXFormat/File.h>
+
 
 #include <imgui_impl_glfw.h>
 #include <imgui_impl_opengl3.h>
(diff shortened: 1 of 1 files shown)


dgovil:
Updated the PR to use `mx::getEnviron()` so that it doesn't trigger the windows side warning-as-error for use of getenv

dgovil:
ah, sorry, forgot to cleanup the non Apple path, when I updated to use the mx::getEnviron call. Apologies, updated now with that.

dgovil:
Fixed an issue where MSVC doesn't like conversion of aliased types  `cannot convert from 'const std::filesystem::path::value_type *' to 'const char *' ` 

At least for clang, value_type is an alias for char. Not sure what it is for msvc.

(I should really get a Windows VM setup)

source/MaterialXGraphEditor/Main.cpp
```diff
@@ -39,6 +40,38 @@ mx::FileSearchPath getDefaultSearchPath()
...
 }
 
+std::filesystem::path getConfigPath()
+{
+    std::filesystem::path config_dir;
+    auto xdg_config_home = mx::getEnviron("XDG_CONFIG_HOME");
+    auto home_directory = mx::getEnviron("HOME");
+    if (!xdg_config_home.empty())
+    {
+        config_dir = std::filesystem::path(xdg_config_home) / "MaterialX";
```
jstone-lucasfilm, line 50: This looks like a good proposal, but could it potentially be written in terms of the `FilePath` class in MaterialXFormat, rather than the `filesystem::path` class in C++17?

For now, we only _require_ C++11 in the MaterialX core libraries, though we support all C++ versions, because there are studios that still need to build MaterialX with GCC 5/6.

Since this new `getConfigPath` function has the potential to move to MaterialX core libraries in the future, it would be ideal to write it without adding any new dependencies.

dgovil: Yeah, that makes sense. It'll potentially be a little clunkier (since filesystem does a few more things than mkdir does) but that should be fine for this simple a case.

I'll get an update to the PR later this week

dgovil: @jstone-lucasfilm Updated with the use of FilePath so it should be portable to other parts of MaterialX now

jstone-lucasfilm (approved):
This looks good to me, thanks @dgovil!

kwokcb:
@dgovil looks like the [issue ](https://github.com/ocornut/imgui/issues/4294) you reference has some sample code to use `SHGetKnownFolderPath` on Windows.

PR #1259, opened 2023-02-24
Add nullptr check in getTypeSyntax

dgovil:
I hit a segfault when I accidentally created a malformed shader, and found this raw pointer that needed a nullptr check.
Ultimately the function will still crash the program since it throws an exception either way, but this way it's not hitting a sigabrt when trying to construct the exception message.


jstone-lucasfilm (approved):
Good catch, @dgovil, and thanks for the fix!

PR #1260, opened 2023-02-24
Fix a nullptr check in pasteNode handling

dgovil:
Just a quick fix for another crash scenario that I was hitting while working with the graph. 

If you delete the default surface shader node in the graph (Ctrl+X), and then create another after that, you'll encounter a case where the pointers on the pasteNode are nullptrs.

This is probably obfuscating an issue further down the pipeline of clipboard handling, but it at least fixes the crash for now.


jstone-lucasfilm (approved):
Thanks for the fix, @dgovil!

Issue #1057, opened 2022-08-22
Coat is partially unaffected by Emission in MaterialXView


DADA246c:
[SurfaceNodeGlsl::emitFunctionCall](https://github.com/AcademySoftwareFoundation/MaterialX/blob/322bf9bd9dc5be6265d7ead677737f36d294458e/source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.cpp#L177) adds Emission after direct lighting calculation, so Coat is not taking partial effect of Emission.

At Standard Surface, Emission is affected by Coat.
![image](https://user-images.githubusercontent.com/65879515/185866921-7b6458a3-5fd4-45bf-97be-bd6a010662b1.png)

I think the current implementation is shown below.
![image](https://user-images.githubusercontent.com/65879515/185867001-7844ea1f-c8d8-4355-825d-c642a4c86bdc.png)

MaterialXView
Indirect Lighting OFF
Render Environment OFF
Specular 0

Coat 0 , Base 0.5 , Emission 0
![image](https://user-images.githubusercontent.com/65879515/185867247-da0ea81d-7793-42eb-ae66-8ed52bf3e33c.png)

Coat 1 , Base 0.5 , Emission 0 
Coat darkens the mesh outlines.
![image](https://user-images.githubusercontent.com/65879515/185867324-14f25cdf-5ee0-449e-bdeb-0d00cc6bdaaa.png)

Coat 0 , Emission 0.5 , Base 0
![image](https://user-images.githubusercontent.com/65879515/185867533-85204069-114e-46ed-a19c-95716b4e3dc1.png)

Coat 1 , Emission 0.5 , Base 0
The mesh outlines are not darkened.
![image](https://user-images.githubusercontent.com/65879515/185867589-66e5e2aa-100a-42cd-8e4d-e4fd8f46fd01.png)

DADA246c:
Thanks for the reply.
I understood the reasons.

jstone-lucasfilm:
Thanks for this original report, @DADA246c, and this has been addressed by @niklasharrysson in #1262.

PR #1262, opened 2023-02-27
Add node for directionally varying emission.

niklasharrysson:
This change list adds a new node to pbrlib: `generalized_schlick_edf`.

```
  <!--
    Node: <generalized_schlick_edf>
    Modifies an EDF with a directional factor. Attenuates the emission distribution of the base EDF according to
    a generalized Schlick fresnel function.
  -->
  <nodedef name="ND_generalized_schlick_edf" node="generalized_schlick_edf" nodegroup="pbr" doc="Modifies an EDF with a directional factor.">
    <input name="color0" type="color3" value="1.0, 1.0, 1.0" />
    <input name="color90" type="color3" value="1.0, 1.0, 1.0" />
    <input name="exponent" type="float" value="5.0" />
    <input name="base" type="EDF" value="" />
    <output name="out" type="EDF" />
  </nodedef>
```

This node can be used to create a directionally varying / Fresnel-like modulation to emission, using the generalized Schlick Fresnel curve. The main idea is to use this to approximate how emission is attenuated when placed under a coating, which is supported in shading models like Standard Surface.

Using the new node the graph implementation of Standard Surface is updated, to address this as reported in #1057.

Note that this is a simple solution that only account for the Fresnel effect. The roughness of the coat, or multiple scattering in the coat layer, is not considered. We are limited to what can be represented in all the shading languages that MaterialX needs to support. Directionally varying emission was not supported in MDL util recently, and this new node matches what can be expressed in MDL (from version 1.7).

Simple example to show the node result, here modulating a uniform white emission with yellow->red for facing->gracing angles:
Screenshot 2023-02-27 19 15 49

Example to show the improved implementation of Standard Surface (top = old implementation, bottom = new implementation):
Screenshot 2023-02-27 20 07 20
Screenshot 2023-02-27 20 07 31


niklasharrysson:
Comparing GLSL (top) with OSL (bottom):

generalized_schlick_edf_test_out_glsl
generalized_schlick_edf_test_out_osl

jstone-lucasfilm (approved):

PR #1265, opened 2023-02-28
Improve envmap in OSL render tests

jstone-lucasfilm:
This changelist improves the envmap function in OSL render tests, basing it on the built-in environment function, which handles edge cases more robustly.

Additionally, this changelist removes an unused file (emitter.osl) from the OSL render test utilities.


jstone-lucasfilm:
Updated render comparisons using these latest changes:

[MaterialXRenderTests_02_27_2023.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/10845670/MaterialXRenderTests_02_27_2023.pdf)

PR #1267, opened 2023-02-28
Add raytype background to OSL tests

jstone-lucasfilm:
This changelist adds a raytype-dependent background to OSL render tests, appearing as a constant color to camera rays and matching the environment color in secondary rays.


jstone-lucasfilm:
Updated render comparison of GLSL and OSL:

[MaterialXRenderTests_02_28_2023.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/10855597/MaterialXRenderTests_02_28_2023.pdf)

source/MaterialXTest/MaterialXRenderOsl/Utilities/raytype_background.osl
```diff
@@ -0,0 +1,9 @@
+surface raytype_background(color Cin = 1, string envmap_filename = "")
+{
+    color bgColor = Cin;
+    if (!raytype("camera"))
+    {
+        bgColor = environment(envmap_filename, I);
+    }
+    Ci = bgColor * emission();
```
fpsunflower, line 8: environment shaders should use the `background()` closure (no practical difference, just a convention)

fpsunflower:
LGTM otherwise!

source/MaterialXTest/MaterialXRenderOsl/Utilities/raytype_background.osl
fpsunflower: After more discussion over slack -- I missed this was assigned to a quad.

But a simpler solution is proposed here: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1649

that should avoid the need for creating an extra quad entirely.

jstone-lucasfilm: Agreed with @fpsunflower that removing the quad will provide a more robust solution, though we may need to take this in stages, as I encounter some stretching artifacts when viewing the environment through transmissive surfaces with that approach.

Issue #1268, opened 2023-03-01
MaterialXGenShader(ShaderGraph) : Enumeration values do not remap to integers


MasayaNikaido:
When ShaderGraph::create() calls with an element of type **Node**, enumeration input values do not remap to integers.
Is remapEnumeration() necessary here?
https://github.com/AcademySoftwareFoundation/MaterialX/blob/b24e9dd16624fe2d09aa167f03a8154b8589b24e/source/MaterialXGenShader/ShaderGraph.cpp#L730-L737

MasayaNikaido:
Additionally, addDefaultGeomNode() is required here, but inputSocket->makeConnection() is not required in this case.

https://github.com/AcademySoftwareFoundation/MaterialX/blob/b24e9dd16624fe2d09aa167f03a8154b8589b24e/source/MaterialXGenShader/ShaderGraph.cpp#L759-L763

niklasharrysson:
Thank you @MasayaNikaido for reporting these issues. I can reproduce the problem and I'm working on a fix.

MasayaNikaido:
Thank you for your quick improvement. It's working as expected.

jstone-lucasfilm:
Thanks to @MasayaNikaido for this report, and to @niklasharrysson for the fix!

PR #1269, opened 2023-03-02
Improve geomprop handling and fix enum value remapping

niklasharrysson:
This changelist extends the support for default geomprops in GLSL to all graphs and not only functional graphs. 

It also fixes an issue where remapping of enum values were not done properly when running codegen on single nodes.

These changes should resolve logged issue #1268.


jstone-lucasfilm (approved):

PR #1270, opened 2023-03-02
Add two-sided refraction option in GLSL

jstone-lucasfilm:
This changelist adds an option for two-sided refraction in the GLSL approximation of transmission, increasing the visual parity with OSL testrender.

Two-sided refraction approximates the effect of transmitted light rays passing through a double-sided sphere geometry, where the tint of the surface is effectively applied twice.

This new option is disabled by default, and it's enabled manually through LightHandler::setRefractionTwoSided in render comparisons between GLSL and OSL.

Two legacy options that were used only in render comparisons, LightHandler::setRefractionEnv and LightHandler::setRefractionColor, have been removed for clarity.


jstone-lucasfilm:
Here are the associated render comparisons between GLSL and OSL, where materials with tinted transmission now match more closely:

[MaterialXRenderTests_03_02_2023.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/10875349/MaterialXRenderTests_03_02_2023.pdf)

niklasharrysson (approved):
Looks great @jstone-lucasfilm! 

As discussed offline, let's revisit this area once we get further on the thin-walled materials support, and let the thin-walled setting control transmission rendering.

PR #1271, opened 2023-03-05
Improvements to findRenderable functions

jstone-lucasfilm:
- Simplify the interfaces of findRenderableElements and findRenderableMaterialNodes, using return values instead of references to output arguments.  This enables function composability and copy elision optimizations, and improves the consistency of the API across languages.
- Add deprecated wrappers for the original interfaces of these functions, allowing clients to proceed without code changes.
- Mark BSDF, EDF, and VDF data types as unsupported in findRenderableElements, as these cannot yet be directly rendered across all languages.


jstone-lucasfilm:
Here are the associated render comparisons between GLSL and OSL, where unsupported data types are no longer included:

[MaterialXRenderTests_03_04_2023.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/10890515/MaterialXRenderTests_03_04_2023.pdf)

source/MaterialXGenShader/Util.cpp
```diff
@@ -363,78 +363,62 @@ bool elementRequiresShading(ConstTypedElementPtr element)
...
+    return renderableNodes;
 }
 
-void findRenderableElements(ConstDocumentPtr doc, vector<TypedElementPtr>& elements, bool includeReferencedGraphs)
+vector<TypedElementPtr> findRenderableElements(ConstDocumentPtr doc)
 {
-    std::unordered_set<ElementPtr> processedSources;
-    findRenderableMaterialNodes(doc, elements, includeReferencedGraphs, processedSources);
-    if (!elements.empty())
+    vector<TypedElementPtr> renderableElements = findRenderableMaterialNodes(doc);
```
kwokcb, line 381: This is missing finding shader nodes not connected to materials at the top level, but I believe this was "lost" in a previous change.

source/MaterialXGenShader/Util.cpp
```diff
@@ -363,78 +363,62 @@ bool elementRequiresShading(ConstTypedElementPtr element)
...
         }
-    }
 
-    // Add renderable graph outputs to the return vector.
-    for (OutputPtr output : graphOutputs)
-    {
-        NodePtr node = output->getConnectedNode();
-        if (node && node->getType() != LIGHT_SHADER_TYPE_STRING)
+        // Filter out unconnected outputs and unsupported data types.
+        const StringSet UNSUPPORTED_TYPES =
```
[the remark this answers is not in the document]
jstone-lucasfilm: @kwokcb This is a good suggestion, though ideally I'd like to wait a user to request this flexibility before adding the feature, to make sure we choose the right level of abstraction in our implementation (e.g. passing a set of type names, passing a filter predicate, etc.).

source/MaterialXGenShader/Util.cpp
jstone-lucasfilm: Good point, though it looks like we've lacked this feature since the transition from 1.37 to 1.38, so I'd like to make sure it's still required for adding the functionality back.

jstone-lucasfilm:
Based on a side conversation with @kwokcb and @niklasharrysson, I'll plan to move forward with this change, and we can continue refining the findRenderable functions in upcoming pull requests.

PR #1277, opened 2023-03-13
Updated MaterialXTest to improve MDL support

krohmerNV:
- changed the MDL test renderer to be used with Example DXR (shipped with the MDL SDK)
- with that, added a gltf sphere and a rotated hdri to get a closely matching scene setup
- extended the tests_to_html script for a three-way-comparison
- added a preprocessing step in the test to allow mdlgen to use flattenFilenames for application side resource handling
- replaced element->getActiveSourceUri() by doc->getSourceUri() because of failures with standard_surface_look_* tests


python/MaterialXTest/tests_to_html.bat
```diff
@@ -1 +1 @@
-python tests_to_html.py -i ../../build
+python tests_to_html.py -i ../../build --destlang2 mdl
```
kwokcb, line 1: Probably shouldn't hardcode this to mdl :).

resources/Geometry/sphere.gltf
kwokcb, line 1: I assume this should be the same as sphere.obj? If not it could be a bit confusing to have 2 different spheres.

kwokcb (commented):
Thanks for all the improvements and fixes. 👍  Just a few small comments.

python/MaterialXTest/tests_to_html.bat
krohmerNV: depends, I did not change the default in which GLSL is compared against OSL. MDL is the only option that makes sense I guess. But it's your decision. It's only a suggestion. What do you prefer as default:
* GLSL vs. OSL as before my changes
* GLSL vs. OSL vs. MDL when passing no arguments to the python script I think that makes two-way comparison odd because you would need to disable the third
* or my suggestion which is sort of a mixture of both

resources/Geometry/sphere.gltf
krohmerNV: it's the same sphere but with camera position and field of view

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.cpp
```diff
@@ -764,7 +767,7 @@ void ShaderGeneratorTester::validate(const mx::GenOptions& generateOptions, cons
                     {
                         const std::string elementNameSuffix(bindingContextUsed ? LAYOUT_SUFFIX : mx::EMPTY_STRING);
 
-                        mx::FilePath path = element->getActiveSourceUri();
```
krohmerNV, line 767: @kwokcb would you have a close look at this. It's about the material test that end with `_look`

python/MaterialXTest/tests_to_html.bat
kwokcb: Actually, I'd be more than happy to just delete this file. 
It's using a hard-coded location and is not part of any tests nor in CI.
(Also, we have Metal coming so it's option 4 :)).

resources/Geometry/sphere.gltf
kwokcb: Interesting, Are you using this for camera matching for unit testing ? If so then we'd need to keep this in sync with the unit test camera setup.

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.cpp
kwokcb: Hi @krohmerNV,
What you have should be correct. The look example has `xincludes`, so the "active" variant is probably grabbing the `URI `of the child included files instead of the the parent working document (which is what you have now).
Thanks for the fix.

python/MaterialXTest/tests_to_html.bat
jstone-lucasfilm: @kwokcb I believe this script is still the main path for render comparisons, and it's what Niklas and I still use to validate new work.

kwokcb: I don't think the Window batch file is needed though is it? I just run the python script directly.

jstone-lucasfilm: I know that I still use it!  :)

kwokcb: Ah ok. Sorry about that. I think defaulting to 2 way comparison of glsl vs osl then. Kai, this allows for vulcan, essl, metal to be a possible third option. Does this sound okay?

krohmerNV: From my point, you can drop `--destlang2 mdl` in the bat. I don't mind. Since Jonathan uses the script i would make it his decision.

resources/Geometry/sphere.gltf
krohmerNV: Yes, only for the unit tests actually. I guess we could do some string replacements as you did with the scene template used for OSL?

kwokcb: Maybe a bit of overkill. Perhaps move this to the `TestSuite` area. Then it's not a generic file that may be accidently changed or used ?

python/MaterialXTest/tests_to_html.py
```diff
@@ -14,27 +14,42 @@
...
     parser.add_argument('-ht', '--imageheight', type=int, dest='imageheight', action='store', help='Set image display height', default=256)
     parser.add_argument('-cp', '--cellpadding', type=int, dest='cellpadding', action='store', help='Set table cell padding', default=0)
     parser.add_argument('-tb', '--tableborder', type=int, dest='tableborder', action='store', help='Table border width. 0 means no border', default=3)
-    parser.add_argument('-sl', '--sourcelang', dest='sourcelang', action='store', help='Source language. Default is source', default="glsl")
-    parser.add_argument('-dl', '--destlang', dest='destlang', action='store', help='Destination language. Default is dest', default="osl")
+    parser.add_argument('-sl', '--sourcelang', dest='sourcelang', action='store', help='Source language. Default is glsl', default="glsl")
+    parser.add_argument('-dl', '--destlang', dest='destlang', action='store', help='Destination language. Default is osl', default="osl")
+    parser.add_argument('-dl2', '--destlang2', dest='destlang2', action='store', help='Second destination language. Default is empty', default="")
```
jstone-lucasfilm, line 52: For consistency with the "inputdir" options above, perhaps we should name these three arguments "lang1", "lang2", and "lang3"?  To my mind, that seems clearer than having a single "source" language and two "destination" languages, and the slight reduction in backwards compatibility seems acceptable for a script used in render testing.

resources/Geometry/sphere.gltf
jstone-lucasfilm: For now, I'd recommend that we omit these derived geometry and environment light files (e.g. `sphere.gltf` and `san_giuseppe_bridge_rot90.hdr`), making it a future goal to generate these elements on the fly as needed for render testing.

My thinking here is that we want to make it straightforward to maintain these geometry and lighting assets over time, without needing to propagate changes from authoritative versions to derived versions, in addition to wanting to keep the disk footprint of the MaterialX project as small as possible.

krohmerNV:
I believe I addressed all issues with this PR.
* removed the additional test scene and the rotated environment
* renamed parameters and variables in the python script for creating comparison images

jstone-lucasfilm:
Thanks for those updates, @krohmerNV, and this changelist looks ready to go.  I'd just wanted to run some local tests before merging, but otherwise this looks good to me.

jstone-lucasfilm (approved):

PR #1280, opened 2023-03-15
Fix GlslRenderer to unbind mesh.

ashwinbhat:
Ensure we unbind existing geometry.

### DIFF
--- a/source/MaterialXRenderGlsl/GlslProgram.cpp
+++ b/source/MaterialXRenderGlsl/GlslProgram.cpp
@@ -335,6 +335,9 @@ void GlslProgram::bindMesh(MeshPtr mesh)
         throw ExceptionRenderError("No mesh to bind");
     }
 
+    // Unbind existing geometry
+    unbindGeometry();
+
     GlslProgram::InputMap foundList;
     const GlslProgram::InputMap& attributeList = getAttributesList();
 


source/MaterialXRenderGlsl/GlslRenderer.cpp
```diff
@@ -202,6 +202,7 @@ void GlslRenderer::render()
                         MeshIndexBuffer& indexData = part->getIndices();
                         glDrawElements(GL_TRIANGLES, (GLsizei)indexData.size(), GL_UNSIGNED_INT, (void*)0);
                     }
+                    _program->unbindGeometry();
```
kwokcb, line 205: I don't think you need this inside the mesh `for` loop as `bindMesh()` should unbind the previous geometry. Suggest to move outside the loop to unbind the last geometry.

ashwinbhat: @kwokcb I went back and checked `bindMesh()` I don't see it unbinding the previous geometry
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXRenderGlsl/GlslProgram.cpp#L325
Am I missing something?

ashwinbhat: However, it looks like the GlslMaterial has that logic:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXRenderGlsl/GlslMaterial.cpp#L108-L121

ashwinbhat: @kwokcb We can update GlslProgram::bindMesh to unbindGeometry(), I don't have any preference.

kwokcb: Your right. I misread where the logic was. Sounds like a state variable for the currently bound mesh can to be added here and to unbind when the new mesh differs from the currently bound one. 

Seems to be the "safest" route without changing `GlslProgram` behaviour to avoid any possible side-effects.

How does that sound?

kwokcb: Your right. I misread where the logic was. Sounds like a state variable for the currently bound mesh can to be added here and to unbind when the new mesh differs from the currently bound one. 

Seems to be the "safest" route without changing `GlslProgram` behaviour to avoid any possible side-effects.

How does that sound?

ashwinbhat: @kwokcb PTAL at https://github.com/AcademySoftwareFoundation/MaterialX/pull/1299

PR #1281, opened 2023-03-17
Adds missing TextureBaker.inl to installation package.

Morteeza:
Fixes the issue that installation package was incomplete due to missing TextureBaker.inl file.


jstone-lucasfilm (approved):
Thanks for the fix, @Morteeza!

Issue #1038, opened 2022-07-31
color3 to vector3 type decay issue with MDL backend


pablode:
I'm having trouble using the UsdPreviewSurface node together with the UsdUVTexture node to achieve normal mapping with the MDL backend.

UsdPreviewSurface exposes the _normal_ input (of type `vector3`)
https://github.com/AcademySoftwareFoundation/MaterialX/blob/a255e89372e768b806c63104336dd5e67edf24fc/libraries/bxdf/usd_preview_surface.mtlx#L20-L22
which is then connected to the _rgb_ output of the UsdUVTexture node (type `color3`).
https://github.com/AcademySoftwareFoundation/MaterialX/blob/a255e89372e768b806c63104336dd5e67edf24fc/libraries/bxdf/usd_preview_surface.mtlx#L48-L49

This is not a problem with GLSL, as `color3` is essentially a typedef for `vec3`:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/a255e89372e768b806c63104336dd5e67edf24fc/source/MaterialXGenGlsl/GlslSyntax.cpp#L213-L223

However, in the MDL backend, `color3` is translated to the native `color` type
https://github.com/AcademySoftwareFoundation/MaterialX/blob/a255e89372e768b806c63104336dd5e67edf24fc/source/MaterialXGenMdl/mdl/materialx/core.mdl#L47-L49
which has no implicit conversions defined (MDL spec 6.12.2):
https://raytracing-docs.nvidia.com/mdl/specification/MDL_spec_1.7.2_17Jan2022.pdf.

This means the generated code results in a compilation error.

There are multiple ways to solve this. I think an elegant solution would be to parameterize the UsdUVTexture node with an output type similar to how it's done for the native MaterialX image node.

For reference, the UsdPreviewSurface specification defines the output type of the UsdUVTexture node as `float3`, which is then implicitly converted to `normal3f` and `color3f`:
https://graphics.pixar.com/usd/release/spec_usdpreviewsurface.html#texture-reader.

krohmerNV:
changing the type of `rgb` to from `color3` to `vector3` makes the MDL code gen produce correct results.
Giving the USD spec, I would consider this to be a good solution.

PR #1282, opened 2023-03-17
Improve MDL thin-film support

krohmerNV:
- push down the thin film parameters to the nodes that can use them, i.e., dielectric, conductor and generalized schlick
- removed additional inputs, that transport thin film parameters, from the public material interface by making them non-editable
- make sure thin_film will not result in undefined behavior when passing zeros as parameters (0 thickness disables thin_film but 0 ior is invalid)

- Vertical Layering, push base deeper than one level, needs review, contains TODOs for mix, add, and multiply (If not 100% sure this is correct, we can drop that commit and postpone.


pablode:
Hello! I'm seeing a regression when I use these changes with my MaterialX/MDL-based renderer:

| Before | After|
|--- |--- |
| latest_master_1 | new_pr_1 |

I've prepared a glTF+mtlx version of this asset in hope that it can be reproduced with the MDL SDK DXR example:
[maneki_gltf.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/11006837/maneki_gltf.zip)

krohmerNV:
I can reproduce it. And it showed another problem as well:
When reading the the normal texture, the output is color and connecting that to a float3 in MDL requires and explicit type construction (because color could be spectral). 
But I'm a bit surprised about the missing gold here. I'll investigate and keep you posted.
 
image

krohmerNV:
Looking into the regression and it revealed an MDL bug which will be fixed with MDL 1.8. Meanwhile, I added a workaround to detect the problematic case and handle it correctly.

kwokcb:
If we want the vec3 version of USDUVTexture how about we bring it up in the MTLX/USD channel and see what folks have to say.
IMO, it's an implementation variant and not a spec change but would need to discuss as it still produces 2 implementation signature variants for USD to consume.

krohmerNV:
@jstone-lucasfilm and @kwokcb is there something outstanding for this PR? I believe the type of the usd texture return rgb is a different topic discussed in https://github.com/AcademySoftwareFoundation/MaterialX/issues/1038

jstone-lucasfilm:
@krohmerNV As long as the original regression noted by @pablode has been addressed, I'm not aware of any outstanding issues, and I'll plan to take a closer look at these changes soon.

jstone-lucasfilm (approved):
Thanks for this improvement, @krohmerNV, and this looks good to me!

PR #1284, opened 2023-03-17
Fixes MaterialXView GLFW error triggering after baking a material

Morteeza:


jstone-lucasfilm (approved):
Thanks for the quick fix, @Morteeza, and to @kwokcb for the catch!

PR #1290, opened 2023-03-20
Fix MSL render test when no lights are present

Morteeza:


jstone-lucasfilm (approved):

PR #1293, opened 2023-03-20
Recursively search for library paths for Viewer and Editor

### LINKED ISSUE #1292: Definition Libraries cannot be found when running MaterialXView from build area (Mac)
This error occurs:
(The module path statement is a debug print added in)
```
Module path: /Users/bernardkwok/work/bernard_materialx/build/bin
Could not find standard data libraries on the given search path: /Users/bernardkwok/work/bernard_materialx/build
Failed to generate shadow shader: Could not find a nodedef for node 'node1'
Failed to generate shadow blur shader: Could not find a nodedef for node 'blur'
```

The reason is that only the parent directories of the module path are added in, and fails to find the `libraries` folder under /bin since `loadLibraries()` only checks for direct children.  Add an additional `--path` does not work for some reason but should not be required.

Seems to be a regression between releases, or was there but did not show up for me as I changed to using clang vs gcc build and the build folders may have changed ? 

Easy fix is to add in the module path as part of the search path.

kwokcb:
Fix #1292 

* Add utility to iteratively search upwards for a `libraries` folder from a start path.
* Use the module path as the start path for both the Viewer and Node Editor.

This fixes when the binary is in the same folder as the `libraries` folder. 

Note that this does not affect MaterialX test since it already searches in the same folder as the test binary.


jstone-lucasfilm:
@kwokcb This might deserve some deeper investigation, as it's not an issue that's been reported on MacOS before.  Does this only apply to developer builds, or does it apply to the shipping builds of the viewer in releases?  If we can narrow this down further, then we should be able to implement a more targeted solution.

kwokcb:
Hi @jstone-lucasfilm, 
This does not occur with the installed folder as MaterialXView is in a subfolder below `libraries`.  For builds,
with my current configuration MaterialXView is at the same level as `libraries` so always fails.

If anyone sets up MaterialXView at the same level as `libraries` then it will fail, so I think this is a suitable patch.
It also feels strange to always only look for libraries in the parent folder of the binary.

ashwinbhat:
Hi @kwokcb 
What does the output layout look like on Ventura? Are you creating application bundle?

kwokcb:
For builds which only create one build area, there is no subfolder for the build type (e.g. `Release/RelWithDebugInfo/Debug`).
Thus the executable is in the same folder as `libraries`. I was using `-GXcode` which created the separate build areas and am no longer doing so (I'm basically trying to build exactly as CI does as the multiple build case does not create Python executables with the correct architecture for M1/M2 etc.). 

@ashwinbhat, neither the build area or installs create bundles so it's always possible to move files around and have libraries not found. I guess this would be the best way to handle this, but it's not currently being done.

kwokcb:
@jstone-lucasfilm , this sounds like simple logic to add. As the Graph Editor also follows the same logic it will also need to be updated as it is also failing when in the same folder as `libraries`. I have created a utility which starts from any root to do the searching.

source/MaterialXFormat/Util.cpp
```diff
@@ -222,4 +222,21 @@ FileSearchPath getSourceSearchPath(ConstDocumentPtr doc)
     return searchPath;
 }
 
+FileSearchPath getDefaultLibraryPath(const FilePath& startPath)
```
jstone-lucasfilm, line 225: This looks like a great direction!  To make the parallels between C++ and Python more clear, perhaps we should call this method `getDefaultDataSearchPath` and omit the input argument?  What are your thoughts?

kwokcb: I think we can do this:  Call this `getDefaultDataSearchPath()`  with a default argument of an empty path.
If empty then the  module path would be used.

Unlike Python where we always know where to look, I think we'd still want to allow users to specify where to start from instead of assuming the libraries reside within the same folder hierarchy as the binary as the user can create their binaries anywhere.

source/MaterialXFormat/Util.cpp
```diff
@@ -222,4 +223,26 @@ FileSearchPath getSourceSearchPath(ConstDocumentPtr doc)
     return searchPath;
 }
 
+FileSearchPath getDefaultDataSearchPath(const FilePath& startPath)
```
jstone-lucasfilm, line 226: @kwokcb This looks really close to the mark, though I'd recommend omitting the unused `startPath` argument from this helper function, as it's much easier to add new arguments if needed later on than it is to remove ones that may never have been used.  We may ultimately find that some other degree of freedom is needed by clients, e.g. controlling the input library folder name rather than the starting folder, and it will be difficult to guess these needs before they arise.

kwokcb: Sounds good. I'll do the cleanup

jstone-lucasfilm (approved):

PR #1296, opened 2023-03-20
Support swizzles for user types in the MDL generator

krohmerNV:
Support swizzles for user types in the MDL generator.

This seems to be required for USD Preview Surface Texture Readers.


jstone-lucasfilm:
@krohmerNV I was waiting on the resolution of some unrelated build issues before commenting, but it looks like this PR triggers undefined behavior warnings when run with dynamic analysis enabled, as seen in the Clang 14 build on GitHub Actions:

```
44: Test command: /home/runner/work/MaterialX/MaterialX/build/bin/MaterialXTest "GenShader: Track Dependencies"
44: Working Directory: /home/runner/work/MaterialX/MaterialX/build/bin
44: Test timeout computed to be: 10000000
44: /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/shared_ptr_base.h:1350:9: runtime error: reference binding to null pointer of type 'std::__shared_ptr_access<MaterialX_v1_38_7::ShaderNodeImpl, __gnu_cxx::_S_atomic, false, false>::element_type' (aka 'MaterialX_v1_38_7::ShaderNodeImpl')
44: SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/shared_ptr_base.h:1350:9 in 
44/71 Test #44: MaterialXTest_GenShader_Track_Dependencies ...................***Failed    0.51 sec
/usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/shared_ptr_base.h:1350:9: runtime error: reference binding to null pointer of type 'std::__shared_ptr_access<MaterialX_v1_38_7::ShaderNodeImpl, __gnu_cxx::_S_atomic, false, false>::element_type' (aka 'MaterialX_v1_38_7::ShaderNodeImpl')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/shared_ptr_base.h:1350:9 in 
```

krohmerNV:
Thanks for checking @jstone-lucasfilm. I hope to continue on the PRs this week.

krohmerNV:
Added an update that hopefully fixes the error reported by clang (added another null check).
Also reduced a bit of code duplication by changing the core class. I believe this is cleaner but up to you to decide.

My local tests still fail for at least one test. Not sure how that passed before because I think it's not related:
```
E:\github\MaterialX\build\bin\resources\Materials\TestSuite\stdlib\definition\definition_reduced_interface\NG_pattern_shader_out.genmdl.mdl(181,61): Error C292: strict relative file path contains too many directory names '..'
```

pablode:
@krohmerNV you are correct in assuming that the line with the dynamic cast is the problem. It turns out that `ShaderNode::getImplementation()` internally dereferences a pointer, which is null in this case.

lldb reveals that this occurs for a node with the name "NG_separate4_color4", and I suspect that no implementation exists because this node is implemented as a [functional node graph](https://github.com/AcademySoftwareFoundation/MaterialX/blob/6864251c8a903bd252cfd2c5d87514606d293e32/libraries/stdlib/stdlib_ng.mtlx#L2401-L2422).

An additional skip condition fixes the issue:
https://github.com/pablode/MaterialX/commit/1d1a5a17c6c801788204cc12b1a34e1019aad8cc

@jstone-lucasfilm  is there a chance this PR can make it into 1.38.7?

niklasharrysson:
Thanks for finding the issue @pablode, and the fix looks good. 

We've never had a case that triggered this before. But I think we should look into if we can make the API safer here. At a minimum we should document that `getImplementation()` cannot be called on a `ShaderGraph` since it has no implementation pointer. But this we can do in another PR.

krohmerNV:
fix is in, my local rendering tests look good as well

jstone-lucasfilm (approved):

PR #1299, opened 2023-03-20
Unbind existing meshes before binding new.

ashwinbhat:
Minor improvement to GlslProgram to unbind existing meshes before binding new meshes.

### DIFF
--- a/source/MaterialXRenderGlsl/GlslProgram.cpp
+++ b/source/MaterialXRenderGlsl/GlslProgram.cpp
@@ -335,7 +335,7 @@ void GlslProgram::bindMesh(MeshPtr mesh)
         throw ExceptionRenderError("No mesh to bind");
     }
 
-    if (_boundMesh && mesh->getName() != _boundMesh->getName())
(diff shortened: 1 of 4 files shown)


source/MaterialXRenderGlsl/GlslProgram.cpp
```diff
@@ -335,6 +335,12 @@ void GlslProgram::bindMesh(MeshPtr mesh)
         throw ExceptionRenderError("No mesh to bind");
     }
 
+    if (_boundMesh && mesh->getName() != _boundMesh->getName())
```
jstone-lucasfilm, line 338: @ashwinbhat In this test, perhaps we should really compare the two _shared pointers_, rather than comparing the names of the meshes?  Because shared pointer equality is based on the objects they point to, this should correctly handle cases that a name comparison cannot, e.g. switching between geometries with missing or duplicated name strings.

jstone-lucasfilm (approved):
This looks great to me, thanks @ashwinbhat!

PR #1300, opened 2023-03-21
Mip mapping for image samplers can be disabled per material pso, in p…

### LINKED ISSUE #1294: Minor: Seam in bakcground environment display for MaterialXView (Mac/Metal)
Seems to be a trivial issue with displaying the environment on Mac with the change to Metal.
There is a seam coming up so guess it's some kind of filtering issue with background display? 
Need to turn on background env display option to see this. It does not show up on the the model.

( This is on an M1, running Ventura, with a clang build. )

Morteeza:
Fixes #1294


jstone-lucasfilm (approved):

PR #1303, opened 2023-03-23
Definition and NodeGraph Publishing Logic Updates

kwokcb:
## API Logic Updates

This s a set of fixes to the existing "publishing" / interface API to fix logic errors and conform properly with the current specification. The current interface `addNodeDefFromGraph()` has been deprecated and replaced with a new `createDefinition()` interface which takes one argument which contains all possible options.

### `<nodedef>` Updates

* The logic for creating `<nodedefs>` has logic errors in it so adding this PR to fix it up and to make sure it matches the current specification.  Of note was that inputs were not being removed from functional graphs, inputs were not being published, and attributes which are disallowed on functional graphs or nodedefs were not being filtered out.

### `<nodegraph>` Updates
* Publishing interfaces is missing compound `<nodegraph>` (vs functional) support so add it in so input interfaces can be "published" from nodes for both compound and functional graphs. 

### General
* Undesired attributes on are now filtered out properly on create. his includes handling positional information (`xpos`, `ypos`) which is generated by the `MaterialXNodeEditor` and `sourceURI`.
* Fix so that add / remove interface transfers values from interfaces from and to interior inputs properly when adding and removing `interfacename`.

## Integration
* This has been hooked up into "create definition" workflow in node editor. 
  * For now only the node `category` and a compound node graph is required with all other arguments being optional. 
  * A single document on disk is created in a default "user library" location, unless an explicit file location is specified. 
  * The "user library" will be loaded by default on startup if it's exists. 
  * New definitions can be manually reloaded within a work session as desired. 
  * Users can still add in additional library paths from the command line as before.

https://user-images.githubusercontent.com/49369885/236569295-9d27f667-b098-42f0-9b94-694ca84ad127.mp4

* A new Python script `createdefinition.py` has been added which will use the current logic if the version is 1.38.8, otherwise 
contains the logic fixes as part of the script. Example syntax:
```
createdefinition.py --category mymarble --nodegroup texture2d --version 1.0 --defaultversion 1 --namespace myspace --documentation "This is a custom marble definition" resources\Materials\Examples\StandardSurface\standard_surface_marble_solid.mtlx --comment "Custom Marble" --separateDocuments True
```

## Unit Test Update
The publishing and definition creation test has been refactored to test the logic changes. 

### Input Data
* New graph created in the `MaterialXNodeEditor`
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <nodegraph name="test_colorcorrect">
    <multiply name="AlphaGain" type="float" xpos="-17.978260" ypos="1.137931">
      <input name="in1" type="float" nodename="inputAlpha" />
    </multiply>
    <add name="AlphaOffset" type="float" xpos="-13.043478" ypos="0.991379">
      <input name="in1" type="float" nodename="AlphaGain" />
    </add>
    <multiply name="ColorGain" type="color3" xpos="-18.094202" ypos="-0.396552">
      <input name="in1" type="color3" nodename="inputColor" />
    </multiply>
    <add name="ColorOffset" type="color3" xpos="-13.043478" ypos="-0.327586">
      <input name="in1" type="color3" nodename="ColorGain" />
    </add>
    <constant name="inputColor" type="color3" xpos="-21.094202" ypos="-0.517241">
      <input name="value" type="color3" value="0.5, 0.5, 0.5" />
    </constant>
    <constant name="inputAlpha" type="float" xpos="-21.094202" ypos="0.827586">
      <input name="value" type="float" value="1" />
    </constant>
    <output name="out" type="color3" nodename="ColorOffset" xpos="-3.355072" ypos="-0.663793" />
    <output name="out1" type="float" nodename="AlphaOffset" xpos="-3.355072" ypos="0.672414" />
  </nodegraph>
</materialx>
```

### Example 1 : Definition from Nodegraph
* Input interfaces are first added to the compound nodegraph and then when creating the functional graph, transferred.
* All "valid" attributes are copied over for the interface inputs and outputs
  * Positional information is kept for the functional graph nodes
  * UI tags on input interfaces and nodedef is now tested.
```xml
<?xml version="1.0"?>
<materialx version="1.38" xmlns:xi="http://www.w3.org/2001/XInclude">
  <nodegraph name="NG_test_colorcorrect" nodedef="ND_test_colorcorrect">
    <multiply name="AlphaGain" type="float" xpos="-17.746376" ypos="1.379310">
      <input name="in1" type="float" nodename="inputAlpha" />
      <input name="in2" type="float" interfacename="AlphaGain_in2" />
    </multiply>
    <add name="AlphaOffset" type="float" xpos="-13.065217" ypos="1.689655">
      <input name="in1" type="float" nodename="AlphaGain" />
      <input name="in2" type="float" interfacename="AlphaOffset_in2" />
    </add>
    <multiply name="ColorGain" type="color3" xpos="-17.514492" ypos="-0.801724">
      <input name="in1" type="color3" nodename="inputColor" />
      <input name="in2" type="color3" interfacename="ColorGain_in2" />
    </multiply>
    <add name="ColorOffset" type="color3" xpos="-13.043478" ypos="-0.318966">
      <input name="in1" type="color3" nodename="ColorGain" />
      <input name="in2" type="color3" interfacename="ColorOffset_in2" />
    </add>
    <constant name="inputColor" type="color3" xpos="-21.094202" ypos="-0.508621">
      <input name="value" type="color3" interfacename="inputColor_value" />
    </constant>
    <constant name="inputAlpha" type="float" xpos="-21.094202" ypos="0.818965">
      <input name="value" type="float" interfacename="inputAlpha_value" />
    </constant>
    <output name="out" type="color3" nodename="ColorOffset" xpos="-3.347826" ypos="-0.655172" />
    <output name="out1" type="float" nodename="AlphaOffset" xpos="-3.347826" ypos="0.672414" />
  </nodegraph>
  <nodedef name="ND_test_colorcorrect" node="test_colorcorrect" nodegroup="adjustment" version="1.0" uiname="test_colorcorrect Version: 1.0" doc="This is version 1 of the definition for the graph: NG_test_colorcorrect">
    <input name="AlphaGain_in2" type="float" value="0.8" uiname="AlphaGain in2" uifolder="Common" />
    <input name="AlphaOffset_in2" type="float" value="1" uiname="AlphaOffset in2" uifolder="Common" />
    <input name="ColorGain_in2" type="color3" value="0.9, 0.9, 0.9" uiname="ColorGain in2" uifolder="Common" />
    <input name="ColorOffset_in2" type="color3" value="0.379147, 0.0341412, 0.0341412" uiname="ColorOffset in2" uifolder="Common" />
    <input name="inputColor_value" type="color3" value="0.5, 0.5, 0.5" uiname="inputColor value" uifolder="Common" />
    <input name="inputAlpha_value" type="float" value="1" uiname="inputAlpha value" uifolder="Common" />
    <output name="out1" type="float" />
  </nodedef>
  <nodegraph name="NG_test_colorcorrect_2" nodedef="ND_test_colorcorrect_22">
    <multiply name="AlphaGain" type="float" xpos="-17.746376" ypos="1.379310">
      <input name="in1" type="float" nodename="inputAlpha" />
      <input name="in2" type="float" interfacename="AlphaGain_in2" />
    </multiply>
    <add name="AlphaOffset" type="float" xpos="-13.065217" ypos="1.689655">
      <input name="in1" type="float" nodename="AlphaGain" />
      <input name="in2" type="float" interfacename="AlphaOffset_in2" />
    </add>
    <multiply name="ColorGain" type="color3" xpos="-17.514492" ypos="-0.801724">
      <input name="in1" type="color3" nodename="inputColor" />
      <input name="in2" type="color3" interfacename="ColorGain_in2" />
    </multiply>
    <add name="ColorOffset" type="color3" xpos="-13.043478" ypos="-0.318966">
      <input name="in1" type="color3" nodename="ColorGain" />
      <input name="in2" type="color3" interfacename="ColorOffset_in2" />
    </add>
    <constant name="inputColor" type="color3" xpos="-21.094202" ypos="-0.508621">
      <input name="value" type="color3" interfacename="inputColor_value" />
    </constant>
    <constant name="inputAlpha" type="float" xpos="-21.094202" ypos="0.818965">
      <input name="value" type="float" interfacename="inputAlpha_value" />
    </constant>
    <output name="out" type="color3" nodename="ColorOffset" xpos="-3.347826" ypos="-0.655172" />
    <output name="out1" type="float" nodename="AlphaOffset" xpos="-3.347826" ypos="0.672414" />
  </nodegraph>
  <nodedef name="ND_test_colorcorrect_22" node="test_colorcorrect" nodegroup="adjustment" version="2.0" isdefaultversion="true" uiname="test_colorcorrect Version: 2.0" doc="This is version 2 of the definition for the graph: NG_test_colorcorrect_2">
    <input name="AlphaGain_in2" type="float" value="0.8" uiname="AlphaGain in2" uifolder="Common" />
    <input name="AlphaOffset_in2" type="float" value="1" uiname="AlphaOffset in2" uifolder="Common" />
    <input name="ColorGain_in2" type="color3" value="0.9, 0.9, 0.9" uiname="ColorGain in2" uifolder="Common" />
    <input name="ColorOffset_in2" type="color3" value="0.379147, 0.0341412, 0.0341412" uiname="ColorOffset in2" uifolder="Common" />
    <input name="inputColor_value" type="color3" value="0.5, 0.5, 0.5" uiname="inputColor value" uifolder="Common" />
    <input name="inputAlpha_value" type="float" value="1" uiname="inputAlpha value" uifolder="Common" />
    <output name="out1" type="float" />
  </nodedef>
</materialx>
```
### Example 2 : Exposing functional graph inputs 
* The functional graph node inputs are "published" as nodedef inputs. 
* This is pretty when the existing test  to ensure no regressions for functional graph interface publishing.
```xml
<?xml version="1.0"?>
<materialx version="1.38" xmlns:xi="http://www.w3.org/2001/XInclude">
  <xi:include href="D:\Work\materialx\bernard_MaterialX\build\bin\libraries\stdlib\stdlib_defs.mtlx" />
  <xi:include href="D:\Work\materialx\bernard_MaterialX\build\bin\libraries\stdlib\stdlib_ng.mtlx" />
  <nodegraph name="NG_test_colorcorrect_3" nodedef="ND_test_colorcorrect_32">
    <multiply name="AlphaGain" type="float" xpos="-17.746376" ypos="1.379310">
      <input name="in1" type="float" nodename="inputAlpha" />
      <input name="in2" type="float" interfacename="NG_test_colorcorrect_3_AlphaGain_in2_renamed" />
    </multiply>
    <add name="AlphaOffset" type="float" xpos="-13.065217" ypos="1.689655">
      <input name="in1" type="float" nodename="AlphaGain" />
      <input name="in2" type="float" interfacename="NG_test_colorcorrect_3_AlphaOffset_in2_renamed" />
    </add>
    <multiply name="ColorGain" type="color3" xpos="-17.514492" ypos="-0.801724">
      <input name="in1" type="color3" nodename="inputColor" />
      <input name="in2" type="color3" interfacename="NG_test_colorcorrect_3_ColorGain_in2_renamed" />
    </multiply>
    <add name="ColorOffset" type="color3" xpos="-13.043478" ypos="-0.318966">
      <input name="in1" type="color3" nodename="ColorGain" />
      <input name="in2" type="color3" interfacename="NG_test_colorcorrect_3_ColorOffset_in2_renamed" />
    </add>
    <constant name="inputColor" type="color3" xpos="-21.094202" ypos="-0.508621">
      <input name="value" type="color3" interfacename="NG_test_colorcorrect_3_inputColor_value_renamed" />
    </constant>
    <constant name="inputAlpha" type="float" xpos="-21.094202" ypos="0.818965">
      <input name="value" type="float" interfacename="NG_test_colorcorrect_3_inputAlpha_value_renamed" />
    </constant>
    <output name="out" type="color3" nodename="ColorOffset" xpos="-3.347826" ypos="-0.655172" />
    <output name="out1" type="float" nodename="AlphaOffset" xpos="-3.347826" ypos="0.672414" />
  </nodegraph>
  <nodedef name="ND_test_colorcorrect_32" node="test_colorcorrect" nodegroup="adjustment" version="3.0">
    <output name="out1" type="float" />
    <input name="NG_test_colorcorrect_3_AlphaGain_in2_renamed" type="float" value="0.8" />
    <input name="NG_test_colorcorrect_3_AlphaOffset_in2_renamed" type="float" value="1" />
    <input name="NG_test_colorcorrect_3_ColorGain_in2_renamed" type="color3" value="0.9, 0.9, 0.9" />
    <input name="NG_test_colorcorrect_3_ColorOffset_in2_renamed" type="color3" value="0.379147, 0.0341412, 0.0341412" />
    <input name="NG_test_colorcorrect_3_inputColor_value_renamed" type="color3" value="0.5, 0.5, 0.5" />
    <input name="NG_test_colorcorrect_3_inputAlpha_value_renamed" type="float" value="1" />
  </nodedef>
</materialx>
```


kwokcb:
@jstone-lucasfilm , I was trying to add / remove interfaces and publish nodedefs and found that the current code isn't robust and will result in validation errors. Also this fixes up all the attribute transfer to / from interfaces on nodedefs / functional graphs / compound graphs as far as I can see. There is no place which codifies the "rules" except for here AFAIK.

source/MaterialXCore/Document.cpp
```diff
@@ -208,9 +208,49 @@ NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const str
         }
     }
 
-    for (auto output : graph->getOutputs())
+    // Expose any existing interface.
+    // Any connection attributes on the existing interface should be removed from the definition.
+    // as well as any source URI
+
+    // Attributes which should not be copied over
```
kwokcb, line 215: Logic to copy over inputs properly, and strip out things like connection attributes for both inputs and outputs.

source/MaterialXCore/Node.cpp
```diff
@@ -652,34 +652,33 @@ void Node::addInputsFromNodeDef()
     }
 }
 
-void NodeGraph::addInterfaceName(const string& inputPath, const string& interfaceName)
+InputPtr NodeGraph::addInterfaceName(const string& inputPath, const string& interfaceName)
 {
     NodeDefPtr nodeDef = getNodeDef();
-    if (!nodeDef)
+    InterfaceElementPtr interfaceElement = ( nodeDef ? nodeDef->asA<InterfaceElement>() : getSelf()->asA<InterfaceElement>() );
```
kwokcb, line 658: Fix to work properly on compound graphs now.

source/MaterialXCore/Node.cpp
```diff
@@ -689,24 +688,44 @@ void NodeGraph::removeInterfaceName(const string& inputPath)
...
+        {
+            NodeDefPtr nodeDef = getNodeDef();
+            InterfaceElementPtr interface = ( nodeDef ? nodeDef->asA<InterfaceElement>() : getSelf()->asA<InterfaceElement>() );
+            ElementPtr interfaceElement = interface->getChild(interfaceName);
+            if (interfaceElement)
+            {
+                InputPtr interfaceInput = interfaceElement ? interfaceElement->asA<Input>() : nullptr;
+                if (interfaceInput && interfaceInput->hasValue())
+                {
+                    input->setValueString(interfaceInput->getValueString());
```
kwokcb, line 701: Transfer value from interface properly.

source/MaterialXTest/MaterialXCore/Node.cpp
```diff
@@ -609,97 +609,106 @@ TEST_CASE("Node Definition Creation", "[nodedef]")
...
+        for (mx::InputPtr origInput : graph->getInputs())
         {
-            temp = nullptr;
-            temp = doc->addNodeDefFromGraph(graph, newNodeDefName, NODENAME, VERSION1, isDefaultVersion, GROUP, newGraphName);
+            mx::InputPtr nodeDefInput = nodeDef->getInput(origInput->getName());
+            REQUIRE(nodeDefInput);
+            REQUIRE(*origInput == *nodeDefInput);
         }
-        catch (mx::Exception&)
+        mx::StringSet connectionAttributes = { mx::PortElement::NODE_GRAPH_ATTRIBUTE,
```
kwokcb, line 655: Attribute transfer test. Checking that undesired attribute are not transfered.

source/MaterialXTest/MaterialXCore/Node.cpp
```diff
@@ -720,8 +729,82 @@ TEST_CASE("Node Definition Creation", "[nodedef]")
...
         REQUIRE(findDefault);
+
+        doc->removeChild(graph->getName());
+        mx::writeToXmlFile(doc, "definition_from_nodegraph_out.mtlx");
     }
+}
 
-    REQUIRE(doc->validate());
-    mx::writeToXmlFile(doc, "definition_from_nodegraph_out.mtlx");
+TEST_CASE("Functional Graph Interface Change", "[nodedef]")
```
kwokcb, line 738: Separate into a "clean" test which just tests functional nodegraph interface publishing, removal and modification.

source/MaterialXCore/Document.h
```diff
@@ -339,12 +339,15 @@ class MX_CORE_API Document : public GraphElement
...
     /// @param version Version for the new declaration
     /// @param isDefaultVersion If a version is specified is thie definition the default version
+    /// @param nodeGroup Optional node group for the new declaration. The Default value is an emptry string.
     /// @param newGraphName Make a copy of this NodeGraph with the given name if a non-empty name is provided. Otherwise
     ///        modify the existing NodeGraph. Default value is an empty string.
-    /// @param nodeGroup Optional node group for the new declaration. The Default value is an emptry string.
+    /// @param docString Optional documentation string. Default is an empty string.
+    /// @param namesSpace Optional namespace for the Nodedef and NodeGraph. Default is an empty string.
     /// @return New declaration if successful.
     NodeDefPtr addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const string& nodeDefName, const string& node, const string& version,
```
jstone-lucasfilm, line 348: I'm a bit wary about the complexity of this function signature, and I wonder if it could be streamlined in this changelist, rather than making it more all-encompassing.  For arguments such as `docString` and `nodeGroup`, could these instead be assigned outside of the function, since the caller has access to the returned `NodeDefPtr`?  If it's possible to remove all of the function arguments that can easily be set by the caller, then I think this would be a better approach, leaving the helper function to focus on the core logic of creating a NodeDef from a NodeGraph.

kwokcb:
@jstone-lucasfilm, I've added in the new interface `Document.createDefinition()` which takes a `DefinitionOptions` struct.
Graph editor and `createdefinition.py` script, unit tests have been updated as well.

source/MaterialXCore/Document.cpp
```diff
@@ -171,46 +171,127 @@ void Document::initialize()
...
+    options.useVersion = !version.empty();
+    options.versionString = version;
+
+    options.newNodeDefName = nodeDefName;
+    options.newNodeGraphName = newGraphName;
+
+    return createDefinition(options);
+}
+
+NodeDefPtr Document::createDefinition(DefinitionOptions& options)
```
jstone-lucasfilm, line 193: Ah, I think I was imagining exactly the opposite approach, where the new version of `Document::createDefinition` would have no inputs for features that can be set just as easily by the client (e.g. doc string, node group string, version string, default version).  I wasn't imagining that we'd bundle all of these inputs into a new `DefinitionOptions` structure, but that the inputs wouldn't exist at all!

What are you thinking of as the motivation for providing these inputs as arguments of `Document::createDefinition`, rather than just letting the client set them after they create their NodeDef from the graph?

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -13,6 +13,7 @@
 #include <widgets.h>
 
 #include <iostream>
+#include <algorithm>
```
jstone-lucasfilm, line 16: Isn't this standard library automatically included through the MaterialX headers?

jstone-lucasfilm:
@kwokcb From a high level, I think that this changelist may be combining too many independent ideas, making it challenging to bring all of them to the required level of polish in a single pull request.  What would you think about breaking this out into separate pull requests, perhaps starting with an improved, streamlined version of `Document::createDefinition` that you can build upon in future changes?

source/MaterialXCore/Document.cpp
kwokcb: > 

The main impetus is to:
1. Make this an atomic operation where you don't need, and don't want users to go back and change
the signature. Conversely avoid the inherent pitfalls of modifying a definition after it's been created without forethought.
2. Make it so that all relevant arguments are up-front to avoid not noticing important arguments which affect behaviour and signature. 

For example, category is required, nodegroup affects behaviour, namespace needs to be set on 2 elements and this and version should never really be changed after the fact. 

I think the only attribute which can be trivially changed is the doc string.

kwokcb:
> @kwokcb From a high level, I think that this changelist may be combining too many independent ideas, making it challenging to bring all of them to the required level of polish in a single pull request. What would you think about breaking this out into separate pull requests, perhaps starting with an improved, streamlined version of `Document::createDefinition` that you can build upon in future changes?

Actually, you had requested that there be a workflow for definition creation :) So two logical places were a script and interactive in the node editor. I think this has helped to refine the requirements and signature. I don't think there is much additional in the node editor interface, except that it sets up a convention for naming of the nodegraph and nodedef i.e. ND_category_<rest of signature>, and NG_...

For the former point, making the arguments explicit and atomic allows for facilitating unique signature creation as mentioned in the interface comment above.

jstone-lucasfilm:
That's a very fair point, @kwokcb, and I think I was hoping that these concrete use cases would help to clarify the motivation for the proposed functionality in MaterialXCore, though so far I'm not quite seeing the benefits.  Let me take a closer look at these proposals, and I'll see if there are any specific refinements that might be worthwhile to suggest.

source/MaterialXTest/MaterialXCore/Node.cpp
```diff
@@ -609,97 +609,123 @@ TEST_CASE("Node Definition Creation", "[nodedef]")
...
+        options.namespaceString = mx::EMPTY_STRING;
+        options.nodeGroupString = GROUP;
+        options.useNamespace = false;
+        options.useVersion = true;
+        options.versionString = VERSION1;
+
+        options.newNodeDefName = newNodeDefName;
+        options.newNodeGraphName = newGraphName;
+
+        mx::NodeDefPtr nodeDef = doc->createDefinition(options);
```
jstone-lucasfilm, line 655: This seems like a great use case for the new `createDefinition` functionality, so let's focus on this one for now.  In this situation, I would imagine it's clearer and more straightforward for a client to call the following:

```
mx::NodeDefPtr nodeDef = doc->createDefinition(graph);
nodeDef->setVersion("1.0");
nodeDef->setIsDefaultVersion(true);
```

This makes it clear that nearly all of the required data (e.g. name, category) is being derived from the NodeGraph itself, with only the versioning data for the NodeDef being stated explicitly by the client.

What are the benefits that you see in the DefinitionOptions class for this use case?  I could see it being useful if a single DefinitionOptions were constructed once and then reused multiple times, but that doesn't seem like a common situation that clients would encounter in practice.

Very interested in your thoughts on this!

jstone-lucasfilm (commented):
I've written up some thoughts on the new use case for `createDefinition` in the graph editor, since that's a great example to focus on.

source/MaterialXTest/MaterialXCore/Node.cpp
kwokcb: > This seems like a great use case for the new `createDefinition` functionality, so let's focus on this one for now. In this situation, I would imagine it's clearer and more straightforward for a client to call the following:
> 
> ```
> mx::NodeDefPtr nodeDef = doc->createDefinition(graph);
> nodeDef->setVersion("1.0");
> nodeDef->setIsDefaultVersion(true);
> ```
> 
> This makes it clear that nearly all of the required data (e.g. name, category) is being derived from the NodeGraph itself, with only the versioning data for the NodeDef being stated explicitly by the client.
> 
> What are the benefits that you see in the DefinitionOptions class for this use case? I could see it being useful if a single DefinitionOptions were constructed once and then reused multiple times, but that doesn't seem like a common situation that clients would encounter in practice.
> 
> Very interested in your thoughts on this!

1. For the first example, the main worry here is what if someone does this and then changes the version. This also involves the unique signature semantics being used. i.e. the proposal is the the identifier includes version. so you'd end up with something like `ND/NG_foo_<version>_....etc`
IMO, the user should not need to know where the information resides (whether it's on the graph or definition), though I agree that the options should probably be nodedef specific and not include the graph so it would be more of:
```
createDefinition(graph, nodedef_options)
```
I wanted to change this anyways as there is possible reuse when dealing with source code implementations -- i.e. the nodedef options are still valid. 

2. An example for the second case is if you want to create variants on definition. e.g. if you had nodegraphs with float, color3, vector3 etc outputs. Then you'd want consistent version, nodegroup, namespace, etc but the signature would
differ slightly depending on the nodegraph. e.g. You could end up with:
```
myspace:ND_foo_v1_float
myspace:ND_foo_v1_vector3
myspace:ND_ foo_v1_color3
etc...
```

jstone-lucasfilm: @kwokcb I'm not quite sure what you mean by "someone does this and changes the version".

Since the desired version is being stated just _once_ by the client, with no coupling between the NodeGraph and NodeDef, I don't see the benefit in stating this version inside of a `DefinitionOptions` class instead of stating it explicitly through a `setVersion` call.

What is the situation you're imagining, where using the `DefinitionOptions` structure would be more robust to version changes?

jstone-lucasfilm: I see where you're coming from, but I'm not quite as convinced that the `DefinitionOptions` structure adds clarity or robustness from the user's perspective.  To my mind it's just _hiding_ the details of what happens inside of `createDefinition`, and it seems clearer and more robust for the client to see these explicit steps in their code.  My hope is that `createDefinition` would handle all of the details that are implied by the input NodeGraph itself (e.g. name, category), but not try to handle details that the client needs to explicitly define (e.g. versioning, documentation, and so forth).

kwokcb: Also my thinking is the emphasis of the API is on nodeef creation so you want the arguments available with a nodegraph is one source which affects the nodedef's characteristics, not so much that you create "some" nodedef with a nodegraph and then specify it's characteristics afterwards.

I think the flip-side question is how to create consistent definitions if the code logic is explicitly performed in separate steps. For example for "global" consistency the utility below exists. Having all args allows a way to take all the information to produce unique ids which are also used for Element existence checking. It would be difficult to promote this if examples show attributes being added after the definition creation. 
```
string generateIdentifier() const
    {
        if (!compoundGraph || categoryString.empty())
        {
            return EMPTY_STRING;
        }

        std::string identifier = categoryString;

        if (useVersion && !versionString.empty())
        {
            identifier += "_" + versionString;
        }

        for (OutputPtr output : compoundGraph->getOutputs())
        {
            const std::string& outputType = output->getType();
            identifier += "_" + outputType;
        }

        return identifier;
    }
```
For:

`namespace`: This needs to be set on the nodedef and nodegraph so the simple reasoning would be to just avoid mistakes. 

`nodegroup`: can be set after, but can change the definitions codegen behaviour, and as far as I've seen nobody thinks about this so will likely be forgotten. So folks can end up with unexpected shader gen results. For example
the houdini nodes are in the `houdini` group but if they were to add `pbr` nodes then the behaviour could be unknown.

category, `compound nodegraph`, & unique signatures for the nodedef and functional nodegraph should be consistent -- and this relies on the identifier generation.

`document` agreed can be done anytime, but again folks don't seem to think about adding it in, so most nodes are undocumented.

I just don't see anyone ever remembering this or they do how can you promote / enforce producing consistent definitions. 

BTW the existing `addNodeDefFromGraph()` already has most of the "desired" requirements. Perhaps we just keep this
with the logic fixes and add in the unique id generator as a util or embedded inside the API for id generation consistency :).  The only outlier is `namespace` which could be added for the aforementioned reason that I don't think
anyone will remember to set it on the graph and nodedef.  Which I think led to the struct creation...

jstone-lucasfilm: Omitting these changes to `addNodeDefFromGraph` sounds like a good strategy to me, and from a high level I would recommend that you make this pull request as minimal as it can possibly be.  That should definitely help me to provide a useful review with my limited GitHub development time, and hopefully it will make this change easier to maintain in the future as well.  :)

kwokcb: Hi @jstone-lucasfilm 
I've created 2 smaller PRs. The original one which was just to update the API logic (no signature change) is #1403.
The UI integration into the graph editor is here #1404.

I will close this one out once I copy over all the PR docs.

PR #1307, opened 2023-03-31
Enable MaterialXRenderGlsl to use shared OpenGL context

ashwinbhat:
Allow the MaterialXRenderGlsl to use an shared context by introducing an overloaded initialize method that accepts a shared OpenGL context.


ashwinbhat:
@bernardkwok @jstone-lucasfilm Please me know your thoughts about this additional API `setSharedContextHandle`

kwokcb:
Hi @ashwinbhat ,
I guess this is a OpenGL only thing and doesn't exist for other languages -- i.e. everything else is device level?

Seems to be okay as a "backdoor" but deviates from the base class interface so you could incorrectly call the wrong initializer. I guess not a big deal if you know your strictly dealing with OpenGL. 

BTW, how would you pass this in for Python ?

ashwinbhat:
> Hi @ashwinbhat , I guess this is a OpenGL only thing and doesn't exist for other languages -- i.e. everything else is device level?
> 
> Seems to be okay as a "backdoor" but deviates from the base class interface so you could incorrectly call the wrong initializer. I guess not a big deal if you know your strictly dealing with OpenGL.
> 
> BTW, how would you pass this in for Python ?

Hi @kwokcb thanks for looking into this. 
The base class initializer would continue to create an OpenGL context as before. This is indeed a customization for better integration with existing OpenGL applications. I ran into an issue with Hydra HGI GL which creates a GL context and we need to share it with MaterialX GLSL Renderer. 

I have a feeling that we might run into similar scenarios with other hardware API renderers such as Metal. 

For Python, I believe Python OpenGL bindings such as [PyOpenGL](https://pypi.python.org/pypi/PyOpenGL) have ability to get the current OpenGL Context.

ashwinbhat:
Hi @kwokcb @jstone-lucasfilm 
Wanted to check if you had any additional thoughts on this, with supporting OpenGL Context sharing workflows

kwokcb:
Still seems this is OpenGL only thing (Metal is via a device API AFAIK) so if you know about the method you can use it. No idea who may be using `PyOpenGL`, but I guess you could make the API appear the same by only exposing the interface with the shared context as a optional arg and hiding the existing one with no args.

source/MaterialXRenderGlsl/GlslRenderer.cpp
```diff
@@ -70,6 +70,11 @@ void GlslRenderer::initialize()
     }
 }
 
+void GlslRenderer::initialize()
+{
+    initialize(nullptr);
+}
+
```
jstone-lucasfilm, line 77: @ashwinbhat Instead of maintaining two separate `initialize` methods, perhaps this could be a single `initialize` method with a default value for its `sharedContextHandle` argument?  That pattern should extend to Python as well, as PyBind11 allows you to specify default values for arguments.

source/PyMaterialX/PyMaterialXRender/PyShaderRenderer.cpp
```diff
@@ -13,7 +13,7 @@ namespace mx = MaterialX;
 void bindPyShaderRenderer(py::module& mod)
 {
     py::class_<mx::ShaderRenderer, mx::ShaderRendererPtr>(mod, "ShaderRenderer")
-        .def("initialize", &mx::ShaderRenderer::initialize)
+        .def("initialize", static_cast<void (mx::ShaderRenderer::*)(void*)>(&mx::ShaderRenderer::initialize))
```
jstone-lucasfilm, line 16: Since there's only one `initialize` method for `ShaderRenderer`, I don't believe you need this static casting anymore, but you do need to declare the default value of the `sharedContext` argument, as in:

```
.def("initialize", &mx::ShaderRenderer::initialize, py::arg("sharedContext") = nullptr)
```

The same binding approach should work for each subclass of `ShaderRenderer` as well.

source/MaterialXRender/ShaderRenderer.h
```diff
@@ -18,6 +18,9 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
+/// Shared context handle 
+using SharedContextHandle = void*;
```
jstone-lucasfilm, line 22: Would `RenderContextHandle` be a better name for this type?  It's really a generalization of the notion of OpenGL Context, and this object isn't always meant to be shared.

source/MaterialXRenderGlsl/GlslRenderer.h
```diff
@@ -61,7 +61,7 @@ class MX_RENDERGLSL_API GlslRenderer : public ShaderRenderer
     /// required for program validation and rendering.
     /// An exception is thrown on failure.
     /// The exception will contain a list of initialization errors.
-    void initialize() override;
+    void initialize(SharedContextHandle sharedcontext = nullptr) override;
```
jstone-lucasfilm, line 64: This looks like a good approach for the C++ API, but I believe we'll still need to declare these new default values in our Python bindings, so that clients can continue to call the method without arguments:

```
.def("initialize", &mx::GlslRenderer::initialize, py::arg("sharedContext") = nullptr)
```

jstone-lucasfilm (approved):

PR #1310, opened 2023-04-04
Upgrade Catch library to version 2.13.9

Morteeza:
This needs to be done to fix MetalTests compilation errors on newer versions of Xcode.


jstone-lucasfilm (approved):
Thanks for putting this change together, @Morteeza, and this looks good to me!

PR #1314, opened 2023-04-10
Code Generation Implementation Inlining

kwokcb:
## Implementation Change:

Switch to using the "recommended" direct source code inlining vs inline file.

## Outcomes:
* Makes inline approaches the same for all "base" shading languages. (OSL, GLSL, and MSL now match MDL).
* Removes the source code file load time for inline files, and reduces number of files that need to shipped with the release. ( Has less affect on ESSL as `libraries` is pre-packaged into a single data file ).

## Testing
Generation and rendering tests pass for OSL, MSL and GLSL. There is no change to MDL.
[inline_tests_GLSL_OSL.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/11191403/inline_tests_GLSL_OSL.pdf)
[inline_tests_GLSL_MSL.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/11191885/inline_tests_GLSL_MSL.pdf)

@Morteeza, this has been discussed with @niklasharrysson and @jstone-lucasfilm and this is a start to lessen the dependence of MSL on GLSL source code. We can discuss outside this PR the possible next steps and eventual independence of GLSL from MSL (including removing target inheritance). 

---
Script used to inline inline files included here:
[mxinline.py.txt](https://github.com/AcademySoftwareFoundation/MaterialX/files/11191968/mxinline.py.txt)


documents/DeveloperGuide/ShaderGeneration.md
```diff
@@ -29,7 +29,7 @@ In the following sub-sections each of these methods are explained. For all metho
 Provided code generators support a very simple expression language for inlining code. This is useful for simple nodes where the operation can be expressed as a single line of code. Inlining will reduce the number of function calls and produce more compact code. The syntax to use is the same as the target shading language, with the addition of using the node’s input ports as variables wrapped in double curly brackets: `{{input}}`. The code generator will replace these variables with values assigned or connected to the respective inputs. Figure 2 gives an example.
 
 Connecting the expression to the nodedef is done using an `<implementation>` element as seen in
-Figure 2. The file extension is used to differentiate inline expressions from source code functions, using `filename.inline`.
+Figure 2. The first option is to keep inline code in a file. The file extension is used to differentiate inline expressions from source code functions, using `filename.inline`. The second option is to directly embed the inlined code using `sourcecode`. This is the recommended approach for inlining if there the logic can fit on one line of code.
```
kwokcb, line 32: @niklasharrysson , adding in note about `sourcecode` usage. Whatever you think is best to say here let me know, or replace it directly. Thanks.

libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
```diff
@@ -6,21 +6,16 @@
...
   <!-- ======================================================================== -->
   <!-- Shader nodes                                                             -->
   <!-- ======================================================================== -->
-
   <!-- <surfacematerial> -->
   <implementation name="IM_surfacematerial_genglsl" nodedef="ND_surfacematerial" target="genglsl" />
-
   <!-- <surface_unlit> -->
   <implementation name="IM_surface_unlit_genglsl" nodedef="ND_surface_unlit" target="genglsl" />
-
```
jstone-lucasfilm, line 19: This change looks really promising, thanks @kwokcb!

One question that comes to mind is whether the newlines in these original documents provided any valuable clarity, and if so, whether there's a practical way to maintain them when we apply automated updates to our data libraries over time.

kwokcb: I'm not sure how many folks beyond maintainers would read these file. There are section comments already, just not spacing. I did turn on "preserve comments" but these are not comment lines unfortunately. Maybe we could try to add additional lines to existing comments?

jstone-lucasfilm: @kwokcb My sense is that more experienced MaterialX developers do read through this file to improve their understanding of the nodes and their implementations, and I often send out links to individual lines in MaterialX discussions.  I don't yet have a solution in mind for preserving newlines in addition to comments, but it seems worthwhile to consider whether it's practically possible before we remove them from the documents.

kwokcb: Fair enough and can either manually restore blank lines, or I think better to insert empty comment lines in the interim as they do get preserved.
It will look like this using comments. 
```xml
<!---->
  <!-- <crossproduct> -->
  <implementation name="IM_crossproduct_vector3_genmsl" nodedef="ND_crossproduct_vector3" target="genmsl" sourcecode="cross({{in1}}, {{in2}})" />
  <!---->
  <!-- <transformpoint> -->
  <implementation name="IM_transformpoint_vector3_genmsl" nodedef="ND_transformpoint_vector3" target="genmsl" />
  <!---->
  <!-- <transformvector> -->
  <implementation name="IM_transformvector_vector3_genmsl" nodedef="ND_transformvector_vector3" target="genmsl" />
  <!---->
  <!-- <transformnormal> -->
  <implementation name="IM_transformnormal_vector3_genmsl" nodedef="ND_transformnormal_vector3" target="genmsl" />
  <!---->
```
Whether support for blank lines can be done can be investigated separately.

That said I'd hope that there be better formal documentation than reading through these files. I did extract out this information to create Markdown/html which can be formatted as desired as part of `mxdoc` script improvements. Then folks can reference docs vs source files.
e.g. This is what I have generated now for code impls for e.g.:
image

jstone-lucasfilm: That's an interesting idea, using empty comments instead of newlines, and I'd be open to that approach if it maintains readability.

With respect to more formally documenting node interfaces in HTML, that's definitely useful, but it's worthwhile to make the analogy with C++ source code, where the ground truth always remains the original source code on GitHub, and the C++ API documentation in HTML doesn't remove the need for this source code to remain clear and readable.

kwokcb: True enough, but I'd rather read Doxygen docs than C++ code :). 
I could not find any way to parse blank lines as white space is consumed and no DOM nodes are created. I will add in empty comments which do turn into DOM nodes so will work.

jstone-lucasfilm: Agreed, I really like Doxygen-generated API documentation as well, but it's not quite good enough that I'd want to compromise on newlines and whitespace in the C++ source!

jstone-lucasfilm: Thanks for putting together this version with empty comments instead of newlines, and it's definitely an interesting proposal, though I'm not sure this creates a very readable experience for users.

My own, very subjective ranking of legibility would be:

1. Separating sections by newlines (the original version)
2. No separators between sections
3. Separating sections by empty comments (the latest proposal)

What are your thoughts?

kwokcb: I do like that idea, but wouldn't want to do it here for risking destabilizing anything. I "think " we just need to look for
lines with only `<eol>` in them and create an new node type which is empty_line node. Then insert it as a child Element like comments.

We can go with 1, but I have one other variant which is split comments across multiple lines to create spacing.
So if you want spacing, add it inside the comment for now. Still a bit ugly.

jstone-lucasfilm: That's an interesting idea as well, though I suspect we'll get the finest control over layout using actual newlines, and our current formatting of library documents strikes a pretty good balance between readability and efficient parsing.  If we do add support for newline elements in some form, we would naturally leave this feature disabled by default, to maintain the efficiency of document parsing in existing tools.

kwokcb: Sounds good. I can leave an new issue for an empty line / white-space option with some info that I have so far, and will use option 1.

kwokcb:
@jstone-lucasfilm , Thanks for doing the line compare and fixes. I was going to do this, and have compared the impl files against what's in main. There should be no more newline diffs not intentionally added in to fix some formatting.

meshula:
Ah, gotcha. I was reacting to the deletion of 120+ files, I thought there'd be a "material" impact on upstream users. Maybe the only impact is a ton less files in an installation?

jstone-lucasfilm:
@meshula I think that's exactly right, and the main impact is just a smaller footprint for the data libraries in MaterialX, possibly correlating to slightly faster shader generation times.

jstone-lucasfilm (approved):
This change looks good to me, thanks @kwokcb, and I think it's important enough to incorporate into the upcoming 1.38.7 release.

PR #1317, opened 2023-04-10
Generate Python wheels in CI

### LINKED ISSUE #570: Add MaterialX Python to PyPi
Add the MaterialX Python library to the PyPi registry, making it available from the command line via "pip install" on Windows, Linux, and MacOS.  Some starting references may be found below:

https://pypi.org/
https://github.com/pybind/python_example

JeanChristopheMorinPerso:
Fixes #570.

This PR adds a new GitHub Actions workflow that generates Python wheels and an sdist that all can be uploaded to PyPI.

It generates wheels for Python 3.7, 3.8, 3.9 and 3.10 for Windows, macOS and Linux.

I used https://github.com/scikit-build/scikit-build-core as the build backend. It's a new tool that is quite simple and allows to build Python extensions (C, C++, etc) without setuptools. Since it's a build backend, everything can be built with `pip install .` or https://github.com/pypa/build (`python -m build`).

Marked as draft because I'll take some more time to document. But at least we have something working for users to test. I also need to create the workflow that will publish to PyPI. And as a reminder, we have reserved the `MaterialX` name on PyPI, see https://pypi.org/project/MaterialX/.


.github/workflows/main.yml
```diff
@@ -9,6 +9,10 @@ on:
       - '**.md'
   workflow_dispatch:
 
+concurrency:
+  cancel-in-progress: true
```
JeanChristopheMorinPerso, line 13: This will cancel in-progress builds on the same branch when new commits are pushed. I generally like this because it avoids wasting resources when someone pushes two commits in a short time. But I don't want to force it, so let me know if you would prefer to remove that.

python/CMakeLists.txt
```diff
@@ -1,18 +1,25 @@
-set(SETUP_PY_IN "${CMAKE_CURRENT_SOURCE_DIR}/setup.py.in")
-set(SETUP_PY "${CMAKE_INSTALL_PREFIX}/python/setup.py")
-
-configure_file(${SETUP_PY_IN} ${SETUP_PY})
+if(NOT SKBUILD)
+  install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/MaterialX" DESTINATION "python" MESSAGE_NEVER)
```
JeanChristopheMorinPerso, line 2: This will be automatically be copied into the package by scikit-build-core. Because we've set `wheel.packages = ["python/MaterialX"]`.

I guess I should add a comment in the file?

python/CMakeLists.txt
```diff
@@ -1,18 +1,25 @@
...
-  install(DIRECTORY "${MATERIALX_PYTHON_OCIO_DIR}/" DESTINATION "python/MaterialX/config/" MESSAGE_NEVER)
+  install(DIRECTORY "${MATERIALX_PYTHON_OCIO_DIR}/" DESTINATION "${MATERIALX_PYTHON_FOLDER_NAME}/config/" MESSAGE_NEVER)
 endif()
 
-if(MATERIALX_INSTALL_PYTHON AND PYTHON_EXECUTABLE)
-  install(CODE "execute_process(COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} install clean --all)" MESSAGE_NEVER)
+if(MATERIALX_INSTALL_PYTHON AND PYTHON_EXECUTABLE AND NOT SKBUILD)
+  set(SETUP_PY "${CMAKE_INSTALL_PREFIX}/python/setup.py")
+  configure_file("${CMAKE_CURRENT_SOURCE_DIR}/setup.py.in" "${SETUP_PY}")
+  install(CODE "execute_process(COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} install clean --all)")
```
JeanChristopheMorinPerso, line 24: I didn't have time to look into this, but I would like to get rid of the `setup.py` here. Maybe in a second PR.

source/MaterialXCore/CMakeLists.txt
```diff
@@ -27,14 +27,16 @@ target_include_directories(MaterialXCore
...
-    ARCHIVE DESTINATION ${MATERIALX_INSTALL_LIB_PATH}
-    LIBRARY DESTINATION ${MATERIALX_INSTALL_LIB_PATH}
-    RUNTIME DESTINATION bin)
-
-install(FILES ${materialx_headers}
-    DESTINATION ${MATERIALX_INSTALL_INCLUDE_PATH}/MaterialXCore/)
-
-install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/MaterialXCore.pdb"
-    DESTINATION "${MATERIALX_INSTALL_LIB_PATH}/" OPTIONAL)
+if(NOT SKBUILD)
```
JeanChristopheMorinPerso, line 30: Because we statically link, we don't need and want to install the libraries. They would end up in the package which is useless and adds weight.

pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+    "MANIFEST.in"
+
+]
+
+[tool.scikit-build.cmake.define]
+MATERIALX_BUILD_SHARED_LIBS = 'OFF' # Be explicit
+MATERIALX_BUILD_PYTHON = 'ON'
+MATERIALX_TEST_RENDER = 'OFF'
+MATERIALX_WARNINGS_AS_ERRORS = 'ON'
+MATERIALX_BUILD_RENDER = 'OFF'
```
JeanChristopheMorinPerso, line 76: This requires X11 which I'm not sure we necessarily want in a Python package. Let me know if including `MaterialXRender` is absolutely required and I'll see what I can do.

.github/workflows/main.yml
jstone-lucasfilm: I like this idea, @JeanChristopheMorinPerso, and I'd have no objection to including this change as part of the PR.

JeanChristopheMorinPerso (commented):
I left some comments for the reviewers

.github/workflows/python.yml
```diff
@@ -0,0 +1,126 @@
...
+      with:
+        name: wheels
+        path: wheels
+
+    - name: Install wheel
+      run: python -m pip install MaterialX --find-links wheels --no-index
+
+    - name: Python tests
+      run: |
+        python python/MaterialXTest/main.py
```
kwokcb, line 117: Not a requirement for this PR, but @jstone-lucasfilm ,
seems like it's worthwhile to batch these up so that a single script can be used to run all of these for all CI workflows and for local devs.

python/CMakeLists.txt
JeanChristopheMorinPerso: It's far from silly! TO be able to answer your question, I'll need more context on the history of this workflow and know more about how it's currently being used. Then I'll be able to see if we can improve the builds by removing the `setup.py` or at least improving it.

But I'll leave that as a separate task that can be dealt with later.

.github/workflows/python.yml
JeanChristopheMorinPerso: From what I see, these could all be run from the `main.py` file as "unit tests".

pyproject.toml
```diff
@@ -0,0 +1,79 @@
+[build-system]
+requires = ["scikit-build-core"]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "MaterialX"
+version = "0.0.0.dev1"  # TODO: Change version
```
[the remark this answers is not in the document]
jstone-lucasfilm: @JeanChristopheMorinPerso Should this string match the current version of the MaterialX codebase?  If so, then we could potentially define this as a `pyproject.toml.in` file, and allow CMake to create the `pyproject.toml` file at generation time.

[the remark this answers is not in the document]
jstone-lucasfilm: Got it, and that makes sense, @JeanChristopheMorinPerso.  For now, should the string simply read "1.38.7", in preparation for potentially uploading data to PyPI?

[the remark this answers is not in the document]
JeanChristopheMorinPerso: Yes it should. I'll update it before undrafting the PR. I also want to add a workflow to release the wheels and I prefer to keep the version like this (at least for now) to avoid mistake from my part.

jstone-lucasfilm:
This work looks really promising, and I'd recommend that we continue refining it in the upcoming 1.38.8 version of MaterialX, rather than trying to rush it into the 1.38.7 release.

JeanChristopheMorinPerso:
@jstone-lucasfilm Yep, it makes sense to not rush it. Anyway I'll probably have limited time to work on this in the next 2 weeks.

JeanChristopheMorinPerso:
Rebased with main branch.

pyproject.toml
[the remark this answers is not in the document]
JeanChristopheMorinPerso: @jstone-lucasfilm The version is now fetched from `CMakeLists.txt`

JeanChristopheMorinPerso:
THe wheel generation is failing with my last commit. I'll need to investigate why it's failing.

mtx_skbuild_plugin.py
```diff
@@ -0,0 +1,90 @@
+"""
+This is a custom scikit-build-core plugin that will
+fetch the MaterialX version from the CMake project.
+"""
```
jstone-lucasfilm, line 4: @JeanChristopheMorinPerso Can these new root-scope files be moved into the `python` folder for clarity, or are there scikit limitations that would prevent this?  If there's a way to move them into the `python` folder, I think it would be a worthwhile refactor, as root-scope files tend to obscure the organization of a GitHub project as they accumulate.

JeanChristopheMorinPerso: This file can be moved into a sub directory, but the `pyproject.toml` AFAIK. We could move it but it would break the source distribution (sdist) since installers and build tools expect the root of the sdist to contain a pyproject.toml (or a setup.py file). `scikit-build-core` doesn't provide a way to modify the sdist.

jstone-lucasfilm: Got it, and thanks for those additional details, @JeanChristopheMorinPerso.

Let's go ahead and move `mtx_skbuild_plugin.py` into the `python` folder for consistency, and it seems like a fair tradeoff for `pyproject.toml` to live in the root folder, given the major benefits that this new functionality will bring.

JeanChristopheMorinPerso: Done!

source/PyMaterialX/PyMaterialXCore/CMakeLists.txt
```diff
@@ -13,9 +13,8 @@ set_target_properties(
     OUTPUT_NAME PyMaterialXCore
     COMPILE_FLAGS "${EXTERNAL_COMPILE_FLAGS}"
     LINK_FLAGS "${EXTERNAL_LINK_FLAGS}"
+    # This is suspicious...
```
jstone-lucasfilm, line 16: What are your thoughts on this, @JeanChristopheMorinPerso?  This was added by @pmolodo to enable shared libraries on MacOS and Linux, and I want to make sure it's aligned with these latest updates to Python distribution.

JeanChristopheMorinPerso: If I remember correctly, I wrote this because it makes the wheel a little bit less "self-contained".

```
[jcmorin@arch01 asd]$ readelf -d MaterialX/PyMaterialXCore.cpython-311-x86_64-linux-gnu.so

Dynamic section at offset 0x238a58 contains 30 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000e (SONAME)             Library soname: [PyMaterialXCore.cpython-311-x86_64-linux-gnu.so]
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../../lib:$ORIGIN:/tmp/tmp5iogfed3/build/installed/lib]
```
This is is coming from one of the Linux wheels. Preferably, only `$ORIGIN` would be there. We don't want anything outside the installed wheel to interfere with it. I'll see if I can conditionally add the RPATH for non wheel builds.

JeanChristopheMorinPerso: Alright, I pushed a commit that should fix this.

PR #1318, opened 2023-04-11
Improve transparent shading order in GLSL render tests

Morteeza:
Applying same strategy we took for MaterialX rendering tests in Metal solution, when rendering an object first shade back faces, and then front faces. This guarantees correct order of blending.

image

left: glsl, middle: msl, right: diff


jstone-lucasfilm (approved):
This change looks good to me, thanks @Morteeza, and it should help to improve visual parity across languages.

PR #1319, opened 2023-04-11
Disable stb_image warning for Clang on Mac

kwokcb:
Buitling with warning-as-errors on Clang14 on Mac detects a deprecated function usage in stb_image
For now patch this by disabling this warning when building this file.
 
The current rev of this code still has this issue as noted [in this issue](https://github.com/nothings/stb/issues/1446).


jstone-lucasfilm (approved):
Thanks for the fix, @kwokcb!

PR #1321, opened 2023-04-12
MDL 1.7 dielectric and generalized schlick improvements

krohmerNV:
* Change the `Dielectric` MDL implementation to use custom curve instead of Fresnel because of issues with diffuse transmission. This comes with limited thin_film support, which will be restored by MDL 1.8
* Improve `Generalized Schlick` MDL implementation so Reflect, Transmit and ReflectTransmit are more plausible. The OSL backend produces very similar results.


jstone-lucasfilm (approved):
Thanks for these improvements, @krohmerNV!

PR #1323, opened 2023-04-14
Minor improvements to Mesh binding

ashwinbhat:
Improve mesh binding in metal renderer to support multiple mesh geometry and prevent repeated binding.


jstone-lucasfilm (approved):
Thanks for the fix, @ashwinbhat!

PR #1327, opened 2023-04-17
[Metal MaterialXTests] Adds support for interpreting string values and setting them in uniform Buffer

Morteeza:
Fixes `ng1_color_lin_rec709` test output. Problem was that the base colo constant type was string and set to "0.5, 0.0, 0.0", and string value setting in uniform buffer in MSL layer was not supported.

GLSL output:
ng1_color_lin_rec709_output_glsl

Previous MSL output:
ng1_color_lin_rec709_output_msl

Current MSL output:
ng1_color_lin_rec709_output_msl

### DIFF
--- a/source/MaterialXRenderMsl/MslPipelineStateObject.mm
+++ b/source/MaterialXRenderMsl/MslPipelineStateObject.mm
@@ -1235,7 +1235,9 @@ void MslProgram::bindUniformBuffers(id<MTLRenderCommandEncoder> renderCmdEncoder
         return false;
     };
 
-    auto setValue = [](MaterialX::ValuePtr value, std::vector<unsigned char>& data, size_t offset)
+    std::function<void(MaterialX::MslProgram::InputPtr, MaterialX::ValuePtr, std::vector<unsigned char>&, size_t)> setValue;
+    
+    setValue = [&setValue](MaterialX::MslProgram::InputPtr input,  MaterialX::ValuePtr value, std::vector<unsigned char>& data, size_t offset)
     {
         if (value->getTypeString() == "float")
         {
@@ -1293,6 +1295,11 @@ void MslProgram::bindUniformBuffers(id<MTLRenderCommandEncoder> renderCmdEncoder
         else if (value->getTypeString() == "string")
         {
             // Bound differently. Ignored here!
+            if(input)
+            {
+                ValuePtr v = Value::createValueFromStrings(value->getValueString(), input->typeString);
+                setValue(nullptr, v, data, offset);
+            }
         }
         else
         {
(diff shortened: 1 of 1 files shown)


kwokcb:
Hi @Morteeza, seems like this patches thing but it seems like it's better to fix `base color` in the implementation to not be a string as this is an incorrect type. Which definition is being used ?

Morteeza:
Hi @kwokcb

This is happening with `resources/Materials/TestSuite/stdlib/color_management/color_management.mtlx` and with the following input:

```
<standard_surface name="color_lin_rec709_standard_surface" type="surfaceshader">
      <input name="base_color" type="color3" nodename="color_lin_rec709" value="0.5, 0.0, 0.0" channels="rgb" /></standard_surface>
```

I am not sure why the value is not becoming `color3` but it comes as `string`. I changed the solution, so it looks less like patching.

kwokcb:
Hi @Morteeza ,
Thanks for the information. This is indeed very strange. I could be that all color4 -> color3 conversions are incorrect, or a combination of a no-op color transform with a type transform is causing this. I'll take a quick look as this works in GLSL but strangely not in MSL. I don't see any specific MSL nodes / code to handle color transforms from a cursory look.

kwokcb:
This is what I found so far.

0. The generated code has the correct type, so it's only a binding issue. 
1. This has nothing to do with `colorspace`
2. The "channels" value conversion appears to be doing the proper thing (not creating a string) when performed in ShaderGraph code.
```
            const string& channels = downstream->getChannels();
            if (!channels.empty())
            {
                downstream->setValue(context.getShaderGenerator().getSyntax().getSwizzledValue(input->getValue(),
                                                                                               input->getType(),
                                                                                               channels,
                                                                                               downstream->getType()));
                downstream->setType(downstream->getType());
                downstream->setChannels(EMPTY_STRING);
            } 
```

My hunch is that there is somewhere a copy or setting of a value along then way using `setValue()` which when not supplied a type, will change the type to a string!  :(  I'm not sure where.  

It appears that the stage `getUniformBlocks()` is being queried in `MslProgram::updateUniformsList()`  but unsure if it's incorrect in the block or it occurs afterwards. I see this line which creates inputs ?
```
              InputPtr inputPtr = std::make_shared<Input>(arg.index, member.dataType, arg.bufferDataSize, EMPTY_STRING);
 ```

This is the minimal test file:
```
<?xml version="1.0"?>
  <nodegraph name="ng1">
    <constant name="color4_constant" type="color4">
      <input name="value" type="color4" value="0.5, 0.0, 0.0, 1.0"  />
    </constant>
    <add name="add1" type="color3">
      <input name="in1" type="color3" nodename="color4_constant" channels="rgb" />      
    </add>
    <output name="outc" type="color3" nodename="add1" />
  </nodegraph>
</materialx>
```

kwokcb:
@Morteeza , I think I've found a fix for the issue within the base shader generation code. I need to discuss with @niklasharrysson , but it should be a small fix so that none of the changes here are required. i.e. you should never get back a string Value.

jstone-lucasfilm:
Thanks for the investigation, @kwokcb, and let's hold off on this PR until you and @niklasharrysson have a chance to discuss.

Morteeza:
Thanks @kwokcb for spending time, finding the original stem of the issue. I appreciate it.

PR #1328, opened 2023-04-17
Add a 2d checkerboard pattern node

crydalch:
Add a 2d checkerboard pattern node.

checkers2d

### DIFF
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -1025,10 +1025,10 @@
   </nodedef>
 
   <!--
-    Node: <checkers2d>
+    Node: <checkerboard>
     A 2D checkerboard pattern.
   -->
-  <nodedef name="ND_checkers2d_float" node="checkers2d" nodegroup="procedural2d">
+  <nodedef name="ND_checkerboard_float" node="checkerboard" nodegroup="procedural2d">
     <input name="freq" type="vector2" uiname="Frequency" value="8, 8" doc="The frequency of checkers, with higher values producing smaller squares. Default is (8, 8)." />
     <input name="offset" type="vector2" uiname="Offset" value="0, 0"  doc="Shift the pattern in 2d space. Default is (0, 0)." />
     <input name="texcoord" type="vector2" uiname="Texture Coordinates" defaultgeomprop="UV0" doc="The input 2d space. Default is the first texture coordinates." />
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -1024,6 +1024,17 @@
     <output name="out" type="float" />
   </nodedef>
 
+  <!--
+    Node: <checkers2d>
+    A 2D checkerboard pattern.
+  -->
+  <nodedef name="ND_checkers2d_float" node="checkers2d" nodegroup="procedural2d">
+    <input name="freq" type="vector2" uiname="Frequency" value="8, 8" doc="The frequency of checkers, with higher values producing smaller squares. Default is (8, 8)." />
+    <input name="offset" type="vector2" uiname="Offset" value="0, 0"  doc="Shift the pattern in 2d space. Default is (0, 0)." />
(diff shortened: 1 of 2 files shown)


libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -1024,6 +1024,17 @@
     <output name="out" type="float" />
   </nodedef>
 
+  <!--
+    Node: <checkers2d>
```
jstone-lucasfilm, line 1028: Would `checkerboard` perhaps be the more canonical name for this node?  I can see the advantage in leaving MaterialX open to similar 3D patterns, but after all a checkerboard is definitively 2D.  :)

jstone-lucasfilm (commented):
Thanks for this proposal, @crydalch, and it seems like a very natural node to include in MaterialX.  I had just one suggestion below on the name of the node.

libraries/stdlib/stdlib_defs.mtlx
crydalch: That sounds like a good idea to me, thanks @jstone-lucasfilm

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -1024,6 +1024,17 @@
     <output name="out" type="float" />
   </nodedef>
 
+  <!--
+    Node: <checkerboard>
+    A 2D checkerboard pattern.
+  -->
+  <nodedef name="ND_checkerboard_float" node="checkerboard" nodegroup="procedural2d">
```
jstone-lucasfilm, line 1031: @crydalch The design of this node seems very reasonable to me, though it seems worthwhile to consider the idea of returning a `color3` value instead of a float, with the two colors of the checkerboard specified as `color1` and `color2` inputs (defaulting to white and black).

I don't have a strong opinion on which design is better for MaterialX, but for reference here are a few examples of `checkerboard` nodes that use the color1/color2 approach:

https://docs.unity3d.com/Packages/com.unity.shadergraph@6.9/manual/Checkerboard-Node.html
https://learn.foundry.com/mari/Content/reference_guide/nodes/Checkerboard.html
https://help.autodesk.com/view/ARNOL/ENU/?guid=arnold_user_guide_ac_texture_shaders_ac_texture_checkerboard_html

kwokcb: It would be nice to cover a reasonable "maximum" possible flexibility (which seems to be Arnold) and anything that supports less can restrict what inputs are modifiable / exposed. e.g. frequency in u and v allows for "stripe" textures to be created. A "customized" version of this would sync u and v values. Does this sound reasonable?

crydalch: I like this idea @jstone-lucasfilm

An early version of the node did have color inputs/outputs, but I don't recall why I went away from that... but regardless, I'll make that change. Thank you for the suggestion and references.

jstone-lucasfilm (approved):
This looks good to me, thanks @crydalch!  Before we merge this to main, what are your thoughts on the proposed interface for this node, @dbsmythe?

dbsmythe:
The color1/color2 output approach and parameterization looks good to me.  Just thinking aloud, I am wondering if there is any use for the originally-proposed "float" output variant, which would output exactly a 0 or a 1, which the color1/color2 variant would call?  I'm generally a fan of breaking larger more full-featured nodes into atomic pieces when some of the sub-pieces could be useful outside the full-featured node, and it feels to me like the "generate a checkerboard pattern" portion could fall into that category.  But if the only reasonable use of a 0/1 float output checkerboard would be to drive the color1/color2 color type node, then this is fine as-is.

jstone-lucasfilm:
@dbsmythe Looking at similar nodes in DCC tools, my thinking is that the current color-based approach is the most general-purpose, and artists can extract a single channel from the output color when they need scalar results.

Here are some of the DCC examples that we were looking at earlier in the thread:

https://docs.unity3d.com/Packages/com.unity.shadergraph@6.9/manual/Checkerboard-Node.html
https://learn.foundry.com/mari/Content/reference_guide/nodes/Checkerboard.html
https://help.autodesk.com/view/ARNOL/ENU/?guid=arnold_user_guide_ac_texture_shaders_ac_texture_checkerboard_html

PR #1329, opened 2023-04-17
Add Shapes2D Pattern node.

crydalch:
Adds Circles2D Pattern for generating circle or ring patterns.

circles2d


crydalch:
Thanks @kwokcb yes some stagger would be useful! @Cinifreak and I had some offline chats, and this circles/rings approach may be absorbed by a more generic nodegraph for generating 2d shapes.

jstone-lucasfilm:
@crydalch @Cinifreak Now that the `randomfloat` and `randomcolor` nodes are merged to MaterialX main, would you be ready to move forward with the generic 2D shape node that you're envisioning?

PR #1341, opened 2023-04-24
Add Herringbone Tiling node

Cinifreak:
Adding Herringbone Tilling node to procedural nodes set. (Including ping-pong Node).


Cinifreak:
Example for Herringbone Tilling node output:
Screenshot 2023-04-24 022650

Cinifreak:
I pushed an updated version including Triangle wave node. Kindly check it and let me know if you have any comments!

jstone-lucasfilm:
@Cinifreak This looks promising, though I believe you need to resolve merge conflicts between your change and previous ones, using the "Resolve Conflicts" button above.

jstone-lucasfilm:
@Cinifreak I wonder if it's worthwhile to consider the alternative approach to Herringbone Tiling shown in this GitHub Issue?

https://github.com/AcademySoftwareFoundation/MaterialX/issues/1415

In this alternative version, it looks like Herringbone Tiling is implemented in terms of the newly-proposed `line` node, which would add support for rounded corners, and might lead to a simpler definition.

What are your thoughts?

Cinifreak:
Is there a PR for `line` node so I can get the idea ?

Cinifreak:
about rounded edge, We can add this feature if there is a request for that to be implemented.

jstone-lucasfilm:
@Cinifreak The `line` node is part of this pull request for 2D shapes:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1411

Although the rounded corners could be a nice side benefit, I'm imagining the larger potential benefit being a simpler graph implementation, since we're aiming to keep our graph definitions as minimal as possible for efficiency in shader generation and application start-up overhead.

I'm open to a discussion about which implementation would be ideal for MaterialX, and I primarily want to make sure we're fully considering all options before committing to an approach.

jstone-lucasfilm:
@Cinifreak The recent work in #1411 is now merged, and I think the best path forward would be to implement the `herringbone` pattern in terms of the new `line` node if possible, so that there's consistency with other line patterns such as `grid` and `crosshatch`.  Let us know what your thoughts are!

jstone-lucasfilm:
I'll close out this original proposal, since the new implementation of `herringbone` is likely to use a different approach, and it deserves its own pull request.  Thanks for putting this together and getting the discussion started!

PR #1346, opened 2023-04-28
Graph Editor Property Panel Formatting

kwokcb:
Add some table formatting for  property panel in graph editor.
* Shows 2 columns with either the connection type (as before) or the value (as before). Second column is aligned right and first column aligned left.
* Table has vertical scroll-bar.

## Sample New Formatting
* Marble node graph
image
* Marble std surf
image
* Image node (showing browser)
image

## Vs old formatting
image
image
image

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -818,7 +818,8 @@ void Graph::updateMaterials(mx::InputPtr input, mx::ValuePtr value)
 void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIProperties& uiProperties)
 {
     std::string inName = !uiProperties.uiName.empty()? uiProperties.uiName : input->getName();
-    float labelWidth = ImGui::CalcTextSize(inName.c_str()).x;
+    //float labelWidth = ImGui::CalcTextSize(inName.c_str()).x;
+    ImGui::PushItemWidth(-1);
 
     mx::ValuePtr minVal = uiProperties.uiMin;
     mx::ValuePtr maxVal = uiProperties.uiMax;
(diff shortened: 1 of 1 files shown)


kwokcb:
@lfl-eholthouser, @jstone-lucasfilm . I have pulled out the formatting UI work into this separate PR.
The biggest change is to not use estimated indentation and switch to using tables which provides column support and algnment.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -818,7 +818,7 @@ void Graph::updateMaterials(mx::InputPtr input, mx::ValuePtr value)
 void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIProperties& uiProperties)
 {
     std::string inName = !uiProperties.uiName.empty()? uiProperties.uiName : input->getName();
-    float labelWidth = ImGui::CalcTextSize(inName.c_str()).x;
+    ImGui::PushItemWidth(-1);
```
kwokcb, line 821: Right alignment is "easy" by using -1.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -854,13 +851,10 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
         if (val && val->isA<int>())
         {
             int prev = val->asA<int>(), temp = val->asA<int>();
-            ImGui::SameLine();
```
kwokcb, line 857: All of these "same lines" are not needed in general since items are added into table cells.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3134,63 +3110,94 @@ void Graph::propertyEditor()
             }
         }
 
+        const float TEXT_BASE_HEIGHT = ImGui::GetTextLineHeightWithSpacing() * 1.3f;
+        const int SCROLL_LINE_COUNT= 20;
```
kwokcb, line 3114: This can be tweaked as needed. About 20 items seems reasonable now for a scroll area, and the table has no borders. I assume when UI grouping support is added these can be collapsible areas and scroll bars may not be needed as much since each group has less items. For another PR.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3274,7 +3320,9 @@ void Graph::propertyEditor()
 
         if (ImGui::BeginPopup("docstring"))
         {
+            ImGui::SetWindowFontScale(_fontScale);
```
kwokcb, line 3323: Minor update to make the font size more readable.

jstone-lucasfilm:
This looks great to me, @kwokcb, and I'd be interested in thoughts from @lfl-eholthouser as well.

lfl-eholthouser:
This looks great to me as well! Thanks! @kwokcb

Issue #1348, opened 2023-05-03
Downstream traversal does not traverse through nodegraphs / interface


kwokcb:
### The issue

`getMatchingPorts()` currently only checks the `node` attribute on a port to compare against a given upstream name. 
Thus connections specified using `nodegraph` or `interfacename` attributes are missed.

### Sample workflows affected:
* A workflow example is trying to find the downstream renderable for a given node in the node editor.
It will always stop early and never find a downstream shader node connected to a nodegraph if starting from
within a node in the nodegraph. As a number of example graphs use this type of configuration  it's quite easy to not switch
to the correct material when selecting nodes inside a nodegraph. (e.g. the chess set graphs don't provide correct switching)

* I am unsure if this make some cases for definition flattening fail but it is possible since connections can be missed.
`nodegraph`->`nodegraph` and `node`->`nodegraph` connections are candidates for issues with flattening. 

* Any type of "dirty propagation" scheme which depends on updating downstream nodes will stop early.

### Proposed Fix
* The fix appears to be to simply update the "port name map" cache to include the appropriate check for attributes which are not `node`. A proposed "better" long term solution is to remove the variance in connection names and only use one name.
* A secondary fix which can be deferred for now would be to modify the cache to keep paths instead of names as names are only unique within the context of the parent so an it seems that incorrect results can occur with 2 nodes named the same within a different context (e.g. one inside a nodegraph and one outside)

### Test 
* This is a sample test with marble graph, starting with `noise`.
* It will halt before the nodegraph output, when it should reach through to the surfaceshader and material nodes.

```python
def getDownstreamPorts(nodeName):
    downstreamPorts = []
    for port in doc.getMatchingPorts(nodeName):
        downstreamPorts.append(port)
    return downstreamPorts

def getDownstreamNodes(node, foundPorts, foundNodes, renderablePort):
    testPaths = set()
    testPaths.add(node.getNamePath())

    while testPaths:
        nextPaths = set()
        for path in testPaths:
            testNode = doc.getDescendant(path)
            ports = getDownstreamPorts(testNode.getName())
            for port in ports:
                downNode = port.getParent()
                downNodePath = downNode.getNamePath()
                if downNode and downNodePath not in nextPaths: #and downNode.isA(mx.Node):
                    foundPorts.append(port.getNamePath())
                    if port.isA(mx.Output):
                        renderablePort.append(port.getNamePath())
                    nodedef = downNode.getNodeDef()
[...]

PR #1349, opened 2023-05-04
Fix downstream traversal in graph editor

kwokcb:
## Issue
Update #1348 

## Fixes
- Includes nodegraph references in port connection cache 
- Adds NodeGraph::getDownstreamPorts() so downstream traversal is possible from graphs
- Update GraphEditor logic to use downstream traversal which will now propery traverse through nodegraphs.

## Test Document
- Has 2 sets of triplets: { nodegraph, surfacesshader, material }
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <nodegraph name="nodegraph1">
    <output name="output_color3" type="color3" xpos="6.615942" ypos="-2.181035" nodename="multiply_color3" />
    <input name="input_color3" type="color3" value="0.0115494, 0.337408, 0.119903" xpos="2.217391" ypos="-1.922414" />
    <multiply name="multiply_color3" type="color3" xpos="4.166667" ypos="-2.629310">
      <input name="in1" type="color3" interfacename="input_color3" />
    </multiply>
  </nodegraph>
  <surface_unlit name="surface_unlit" type="surfaceshader" xpos="5.652174" ypos="-4.793103">
    <input name="emission_color" type="color3" output="output_color3" nodegraph="nodegraph1" />
  </surface_unlit>
  <surfacematerial name="green_material" type="material" xpos="8.086957" ypos="-4.258621">
    <input name="surfaceshader" type="surfaceshader" nodename="surface_unlit" />
  </surfacematerial>
  <surface_unlit name="surface_unlit2" type="surfaceshader" xpos="5.884058" ypos="-2.724138">
    <input name="emission_color" type="color3" output="output_color3" nodegraph="nodegraph2" />
  </surface_unlit>
  <nodegraph name="nodegraph2">
    <output name="output_color3" type="color3" xpos="6.615942" ypos="-2.181035" nodename="multiply_color3" />
    <input name="input_color3" type="color3" value="0.94132, 0.346111, 0.0989652" xpos="2.217391" ypos="-1.922414" />
    <multiply name="multiply_color3" type="color3" xpos="4.166667" ypos="-2.629310">
      <input name="in1" type="color3" interfacename="input_color3" value="0.94132, 0.346111, 0.0989652" />
    </multiply>
  </nodegraph>
  <surfacematerial name="orange_material" type="material" xpos="8.115942" ypos="-2.594828">
    <input name="surfaceshader" type="surfaceshader" nodename="surface_unlit2" />
  </surfacematerial>
</materialx>
```

## Test Workflow
- Each of the triplet's elements is selected alternately to reflect new logic of finding surface shaders or materials directly or via downstream node / nodegraph.

https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/ea07c97c-ac00-46cf-86c6-6f0495667279


source/MaterialXCore/Document.cpp
```diff
@@ -97,6 +97,18 @@ class Document::Cache
                         portElementMap.emplace(portElem->getQualifiedName(nodeName), portElem);
                     }
                 }
+                else
```
kwokcb, line 100: Includes `nodegraph` references so can check if downstream ports reference a nodegraph.

kwokcb: Note that this is outside the scope of this change but there is a lot of update here and could be prune by not updating for std lib nodes / graphs which should be "static" once loaded in. e.g. There are about 200 odd graphs to cache here on any change to the document.

source/MaterialXCore/Node.cpp
```diff
@@ -733,6 +733,33 @@ InterfaceElementPtr NodeGraph::getImplementation() const
     return nodedef ? nodedef->getImplementation() : InterfaceElementPtr();
 }
 
+vector<PortElementPtr> NodeGraph::getDownstreamPorts() const
```
kwokcb, line 736: New method to check for downstream nodegraph references on ports.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -694,92 +699,124 @@ void Graph::setRenderMaterial(UiNodePtr node)
     }
     else
     {
-        // continue downstream using output connections until a material node is found
-        std::vector<UiNodePtr> outNodes = node->getOutputConnections();
-        if (outNodes.size() > 0)
+        // Traverse downstream looking for the first renderable item found.
```
kwokcb, line 702: New logic to perform traversal using Mtlx traversal. Other stopping criteria could be added or some other subset nodes could be added say for things like swatch rendering. 

If this becomes useful for general use a utility method can be added later one.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -694,92 +699,124 @@ void Graph::setRenderMaterial(UiNodePtr node)
...
 void Graph::updateMaterials(mx::InputPtr input, mx::ValuePtr value)
 {
     std::string renderablePath;
-    mx::TypedElementPtr renderableElem;
-    std::vector<mx::TypedElementPtr> elems = mx::findRenderableElements(_graphDoc);
-
-    size_t num = 0;
-    int num2 = 0;
-    for (mx::TypedElementPtr elem : elems)
+    if (_currRenderNode)
```
kwokcb, line 811: Simplify this logic as the current node is already found and is already known to be renderable so no need to do a search for renderables and then match.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3619,10 +3656,7 @@ void Graph::drawGraph(ImVec2 mousePos)
                     // update render material if needed
                     if (_currUiNode->getNode())
                     {
-                        if (_currUiNode->getNode()->getType() == mx::SURFACE_SHADER_TYPE_STRING || _currUiNode->getNode()->getType() == mx::MATERIAL_TYPE_STRING)
-                        {
-                            setRenderMaterial(_currUiNode);
-                        }
+                        setRenderMaterial(_currUiNode);
```
kwokcb, line 3659: Allow for search for renderable.

kwokcb:
Hi @fl-eholthouser, @jstone-lucasfilm 
Just curious if you had any feedback on this one? It should give much more predictable results when finding nodes for preview rendering. Thanks.

source/MaterialXCore/Document.cpp
```diff
@@ -97,6 +97,18 @@ class Document::Cache
...
+                {
+                    const string& nodegraphName = elem->getAttribute(PortElement::NODE_GRAPH_ATTRIBUTE);
```
jstone-lucasfilm, line 102: For readability, I would recommend moving this call to the top of the `for` loop, right after the storage of `nodeDefString`, so that the parallel between these two calls is clear.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -701,8 +701,13 @@ void Graph::selectMaterial(UiNodePtr uiNode)
 // set the node to display in render veiw based off the selected node or nodegraph
 void Graph::setRenderMaterial(UiNodePtr node)
 {
+    // For now surface shaders and materials are considered renderable.
+    // This can be adjusted as desired to includ being able to use outputs,
+    // and / a sub-graph in the nodegraph.
+    mx::StringSet renderableTypes = { "material", "surfaceshader" };
```
jstone-lucasfilm, line 707: For clarity, I'd recommend making this a constant StringSet named RENDERABLE_TYPES.

source/MaterialXGraphEditor/RenderView.cpp
```diff
@@ -452,7 +452,7 @@ void RenderView::updateMaterials(mx::TypedElementPtr typedElem)
         if (typedElem)
         {
             mx::NodePtr node = typedElem->asA<mx::Node>();
-            materialNode = node && node->getType() == mx::MATERIAL_TYPE_STRING ? node : nullptr;
+            materialNode = node;
```
jstone-lucasfilm, line 455: Could we take this a step further and remove the `materialNode` alias for `node`?

jstone-lucasfilm (commented):
This looks like a good proposal, thanks @kwokcb, and I've written up a couple of suggestions for edits.

kwokcb:
For demonstration, I've added a test file and test vid + updated the minor src changes.

jstone-lucasfilm (commented):
This looks very promising, thanks @kwokcb!

I'm running through initial tests with the pull request, though, and I'm encountering a regression where clicking on any node within the `standard_surface` graph of the initial marble example will trigger a shader compilation error:

MaterialXGraphEditor_ShaderCompileError

I have a build from the main branch in a separate folder, and I don't see the same issue when running the graph editor from that build.

kwokcb:
Thanks for finding this. I'll check out what's going on.

kwokcb:
Hmm. Seems like code generation is creating invalid code when trying to render the surface shader inside of the std surface definition graph. As, this was never supported, I'll just ignore trying to do this (same as before). 

There is no logic which supports traversing from inside a functional nodegraph  to where it's used as a node instance to downstream nodes. If this is desirable then new functionality needs to be added which is beyond the scope of this change.

@niklasharrysson, leaving a ping as the `mx_luminance_color3` has a a signature of  
```
void mx_luminance_color3(vec3 _in, vec3 lumacoeffs, out vec3 result)
```
and is being called with this call:
```
mx_luminance_color3(opacity, opacity_luminance_lumacoeffs, opacity_luminance_out);
```
where `opacity` is float,  `opacity_luminance_lumacoeffs` is vec3, and `opacity_luminance_out` is float.
Hence the implicit cast error.

kwokcb:
Actually if you copy the std surface graph to the top level it exhibits the same behaviour. This just seems to be a pre-existing issue which was ignored before.

kwokcb:
Update logic to avoid trying to traverse downstream through a functional nodegraph which was not supported before.

kwokcb:
Hi @niklasharrysson,
Yes, the situation you mentioned does work as I tried this in the node editor. 

Seems to be either when it's in a functional nodegraph or trying to evaluate the root `surface` node `shader_constructor`. If you grab the branch of code for this PR and revert my last change you can see it, but I'm not sure how much time and effort this is worth as it just seems like a edge case.

kwokcb:
Hi @jstone-lucasfilm, Not sure if you have to to re-look at this. There are some issues being raised by Marcel about missing evaluation / updates, and they could be fixed since this should do proper nodegraph downstream search now.

jstone-lucasfilm (commented):
One question about the logic in the new NodeGraph::getDownstreamPorts method:

source/MaterialXCore/Node.cpp
```diff
@@ -728,6 +728,33 @@ InterfaceElementPtr NodeGraph::getImplementation() const
...
+            node = node->getParent();
+            if (node->isA<GraphElement>())
+            {
+                graph = node->asA<GraphElement>();
+            }
+        }
+        if (graph && graph->getChild(getName()) == getSelf())
+        {
+            downstreamPorts.push_back(port);
+        }
```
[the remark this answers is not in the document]
kwokcb: This looks to be the same. I think it should still test if the parent is indeed a GraphElement as the code does this and checks for `node` child and not just "a child"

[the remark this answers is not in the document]
kwokcb: This should work to any depth, so if / when nested nodegraphs are supported it should still work.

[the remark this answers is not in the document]
jstone-lucasfilm: Just focusing on the current syntax in 1.38 for now, is the restriction to ports on nodes at the root scope of the document strictly correct, or are there other locations in which downstream ports might be found?  For example, would this new function handle examples like the following?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/stdlib/nodegraph_inputs/cascade_nodegraphs.mtlx

If the restriction looks correct to you, then I'd recommend simplifying this logic as far as possible, perhaps using the proposal above as a starting point.  That should help future readers to better understand the cases that we're supporting and not supporting.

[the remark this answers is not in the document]
kwokcb: I tried this out and it does work for this type of configuration. 
Here's a simplified version:
```
<?xml version="1.0"?>
<materialx version="1.38">
  <nodegraph name="nodegraph1">
    <input name="input_color3" type="color3" xpos="1.188406" ypos="-0.810345" output="output_color3" nodegraph="nodegraph2" />
    <surface_unlit name="surface_unlit" type="surfaceshader" xpos="3.246377" ypos="-1.232759">
      <input name="emission_color" type="color3" interfacename="input_color3" />
    </surface_unlit>
    <surfacematerial name="surfacematerial" type="material" xpos="5.884058" ypos="-2.060345">
      <input name="surfaceshader" type="surfaceshader" nodename="surface_unlit" />
    </surfacematerial>
  </nodegraph>
  <nodegraph name="nodegraph2">
    <output name="output_color3" type="color3" nodename="add_color3" xpos="4.992754" ypos="-1.344828" />
    <add name="add_color3" type="color3" xpos="3.246377" ypos="-1.482759">
      <input name="in2" type="color3" value="0.792176, 0.307071, 0.0600427" />
      <input name="in1" type="color3" value="0.0860289, 0.342787, 0.858191" />
    </add>
  </nodegraph>
</materialx>
```
A change on `add_color3` will go to the `output` up to the document and then to the `input_color3` on the other nodegraph (since it's in the downstream connection cache now) and onwards.

I will simplify the logic.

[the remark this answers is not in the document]
kwokcb: @jstone-lucasfilm 
I've added in new code and retested. Everything looks works as before. Thanks for the simplification.

jstone-lucasfilm (approved):

PR #1352, opened 2023-05-12
Define default color transforms as graphs

jstone-lucasfilm:
This changelist updates the definitions of the default color transforms from raw shader code to MaterialX graphs, removing the need for explicit code for each transform to be provided in each supported language.

Further in the future, this approach could enable direct communication of color transform graphs from OpenColorIO to MaterialX, allowing both standard ACES transforms and custom transforms to be communicated in a unified way.


jstone-lucasfilm:
Updated render tests using graph-based color transforms:

[MaterialXRenderTests_05_12_2023.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/11467839/MaterialXRenderTests_05_12_2023.pdf)

PR #1353, opened 2023-05-13
Standardize document formatting

kwokcb:
## mxupdate / mxformat
- Add `mxformat` Python script to format files. This merges in logic from `mxupdate` and removes that script.
 - The script preserves comments and newlines 
 - The script does not perform any version upgrades by default. If specified then it basically does what mxupdate currently does.
 - Xinclude file references are preserved (default).

### Notes
- Formatting within an node (Element) is not supported so Lama nodes and Autodesk Standard Surface "doc" indentation eeds to be manually reverted.

### Test
- Run on `libraries` and `resources` and `upgrade` forlders to update formatting.
- Run unit and render tests.
- Update CI workflow to use `mxformat` vs `mxupdate`.
- Found issue with MDL file duplicate naming. Fixed.


kwokcb:
@jstone-lucasfilm This is the follow-up to perform formatting on all existing `libraries` and `resources` files.
Added an `mxformat.py` utility for this. This can be run whenever desired -- and may be good to run on check-ins ?

libraries/bxdf/lama/lama_conductor.mtlx
```diff
@@ -1,34 +1,20 @@
 <?xml version="1.0"?>
 <materialx version="1.38" colorspace="acescg">
   <nodedef name="ND_lama_conductor" node="LamaConductor" nodegroup="pbr" doc="Lama conductor" version="1.0" isdefaultversion="true">
-    <input name="tint" type="color3" value="1, 1, 1" uiname="Tint" uifolder="Main"
-           doc="Overall color multiplier. It should be used with parcimony, as a non-white value breaks physicality. The prefered way to define the color of a conductor is through the Fresnel attributes right below." />
```
jstone-lucasfilm, line 5: @kwokcb I like the idea of this standardization pass on the data libraries and examples, and I have two recommendations for improvements:

1. In some situations, including the highlighted example text, we're using newlines before `doc` strings to enhance readability.  In these cases, I would recommend leaving the original formatting of these sections untouched, despite our not having a way to retain the formatting automatically.  By analogy, we take a similar approach when applying Clang Format to our C++ codebase, where certain formatting structures designed for readability (e.g. displaying 4x4 matrices as 4 rows of 4 numbers) are overwritten by the Clang Format process, and we need to restore them manually after applying our formatting step.
2. Since the new `mxformat.py` is substantially the same as our existing `mxupdate.py`, I would recommend merging these into one helper script.  I like your suggestion of naming this script `mxformat.py`, and perhaps it should simply have a command-line option that selects whether version upgrades are also applied to the selected set of MaterialX documents.

kwokcb: This sounds good. I'll make the changes. 
This is deprecate `mxupdate` so users will need to be notified of this.

python/Scripts/mxformat.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if not validDocs:
+        print('No MaterialX documents were found in "%s"' % (opts.inputFolder))
+        return
+
+    print('Found %s MaterialX files in "%s"' % (len(validDocs), opts.inputFolder))
+
+    mxVersion = mx.getVersionIntegers()
+
+    if not opts.yes:
```
kwokcb, line 56: This option merged in from mxupdate.

jstone-lucasfilm (approved):
This looks great to me, thanks @kwokcb!

PR #1355, opened 2023-05-16
Add vector2 variant of normalmap

mnikelsky:
The materialx specification defines the scale attribute of the normalmap to be either a float or a vector2 but the vector2 variant was missing from the implementation. This PR adds it.


jstone-lucasfilm:
Thanks for putting this change together, @mnikelsky!

To address the errors in unit testing, I believe we just need a single _usage_ of this node in our test suite, so that it will be run through shader generation and validation.

Here's one example where you could potentially update the `normalmap` node to use your new signature:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/1d67704c0b1989ede00932f3da1b3754fc0fe5d5/resources/Materials/TestSuite/pbrlib/surfaceshader/normalmapped_surfaceshader.mtlx#L15

mnikelsky:
Added a material example to the normalmapped_surfaceshader.mtlx

libraries/stdlib/genglsl/mx_normalmap.glsl
```diff
@@ -14,3 +14,20 @@ void mx_normalmap(vec3 value, int map_space, float normal_scale, vec3 N, vec3 T,
     // Normalize the result.
     result = normalize(value);
 }
+
+void mx_normalmap_vector2(vec3 value, int map_space, vec2 normal_scale, vec3 N, vec3 T,  out vec3 result)
```
jstone-lucasfilm, line 18: To my mind, your new generalized `vector2` implementation should now be used by _both_ code paths, with the original `float` version simply calling into this one with the same value in both channels.

This would remove any duplication of code between `vector2` and `float` paths, and make the node more straightforward to maintain over time.

Does that streamlined approach seem reasonable to you as well?

jstone-lucasfilm (commented):
This looks like a great start, thanks @mnikelsky, and I've written up one simplifying suggestion below.

libraries/stdlib/genglsl/mx_normalmap.glsl
mnikelsky: Sounds good, I adjusted the implementations accordingly.

jstone-lucasfilm (approved):
This looks good to me, thanks @mnikelsky!

PR #1362, opened 2023-05-23
Add integer signature to floor and ceil

crydalch:
Adds support to `ceil` and `floor` to output integers. There will be other tickets/discussions around improving integer support across MaterialX, but adding integer output to these nodes first would be a big help to some nodegraph contributions we'd like to contribute soon.

Thanks!


crydalch:
Locally it builds, but it's got some test failures because `implementationUseCount` is not matching `count`, in `GenShaderUtil`.

jstone-lucasfilm:
This looks really promising, thanks @crydalch, and I believe we just need to add examples of these new nodes to our unit tests, since they're new primitives with their own unique source code.  I should be able to help set this up, and I'll take a look when I have a moment.

jstone-lucasfilm (approved):
This looks good to me, thanks @crydalch!

Issue #1356, opened 2023-05-16
Add implementation of blackbody node for all targets


niklasharrysson:
MDL is the only target that has an implementation for the <blackbody> node. It should be implemented for GLSL and OSL as well.

nicolassavva-autodesk:
We are preparing an implementation, based on a cubic spline fit (https://en.wikipedia.org/wiki/Planckian_locus#Approximation)
and will open a PR addressing this shortly: 
https://github.com/autodesk-forks/MaterialX/compare/adsk_contrib/dev...autodesk-forks:MaterialX:pbr_blackbody?expand=1

ashwinbhat:
@niklasharrysson (fyi)

jstone-lucasfilm:
This is now addressed in #1367, thanks @nicolassavva-autodesk and @niklasharrysson!

PR #1367, opened 2023-05-25
Add blackbody PBR node implementations

nicolassavva-autodesk:
Address the missing targets for the PBR blackbody node implementation ( #1356 )
using the following approximation: [Wikipedia: Planckian Locus Approximation](https://en.wikipedia.org/wiki/Planckian_locus#Approximation)
Add a simple Standard Surface material example using the above blackbody node for emission.
Reenable blackbody node tests.


jstone-lucasfilm:
This looks really promising, thanks @nicolassavva-autodesk.  Let us know if you run into any issues with the Autodesk CLA, and the Linux Foundation is usually very responsive if you file a support ticket using the link above.

nicolassavva-autodesk:
Thanks Jonathan
I just re-authorized the Autodesk CLA with the account in question

libraries/pbrlib/genglsl/mx_blackbody.glsl
```diff
@@ -0,0 +1,49 @@
...
+{
+    float xc, yc;
+    float t, t2, t3, xc2, xc3;
+
+    t = 1000.0 / temperatureKelvin;
+    t2 = t * t;
+    t3 = t * t * t;
+
+    // if value outside valid range of approximation clamp to accepted temperature range
+    clamp(temperatureKelvin, 1667.0, 25000.0);
```
niklasharrysson, line 16: Should you not assign the result here? Like so:

`temperatureKelvin = clamp(temperatureKelvin, 1667.0, 25000.0);`

libraries/pbrlib/genosl/mx_blackbody.osl
```diff
@@ -0,0 +1,49 @@
...
+{
+    float xc, yc;
+    float t, t2, t3, xc2, xc3;
+
+    t = 1000.0 / temperature;
+    t2 = t * t;
+    t3 = t * t * t;
+
+    // if value outside valid range of approximation clamp to accepted temperature range
+    clamp(temperature, 1667.0, 25000.0);
```
niklasharrysson, line 11: Same here with assigning the result.

niklasharrysson:
Looks great @nicolassavva-autodesk , I just noted an issue with the clamps above.

nicolassavva-autodesk:
Good catch Niklas.
Added a quick fix for the missing clamping

jstone-lucasfilm (commented):
This looks great, thanks @nicolassavva-autodesk.  My one recommendation would be to move the example material from the Standard Surface examples folder to the test suite, where it would be a better fit:
https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/resources/Materials/TestSuite/pbrlib/bsdf

The example material could additionally be simplified to a simple nodegraph containing the `blackbody` node, rather than making it a Standard Surface material with pure emission.

jstone-lucasfilm (approved):
This looks good to me, thanks @nicolassavva-autodesk!

krohmerNV:
is there an MDL implementation missing now?

lgritz:
I hope I'm not opening a can of worms on a topic that has long since been settled, but is this function merely trying to return the color corresponding to a blackbody temperature, normalized to some standard luminance? Or is it supposed to be returning the actual emittance of the blackbody (W/m^2 or whatever)?

PR #1368, opened 2023-05-26
Add DisplayP3 colorspaces

dgovil:
This PR adds support for the [DisplayP3](https://www.color.org/chardata/rgb/DisplayP3.xalter) colorspace, a wider gamut range for computer displays that is standard in the Apple ecosystem and elsewhere. More information can be found in this [developer video](https://developer.apple.com/videos/play/wwdc2017/821/).

DisplayP3 uses the DCI-P3 gamut with a D65 white point and an sRGB transfer curve. In keeping with discussions with the OCIO group, we've opted to name it in the form `<transfer_curve>_<gamut>` for consistency.

Hence we add the following two colorspaces:

* srgb_displayp3
* lin_displayp3

The matrix3x3 values are derived from Nuke

image

### DIFF
--- a/libraries/cmlib/cmlib_defs.mtlx
+++ b/libraries/cmlib/cmlib_defs.mtlx
@@ -87,4 +87,24 @@
     <output name="out" type="color4" />
   </nodedef>
 
+  <nodedef name="ND_srgb_displayp3_to_lin_rec709_color3" node="srgb_displayp3_to_lin_rec709" nodegroup="colortransform">
+    <input name="in" type="color3" />
+    <output name="out" type="color3" />
+  </nodedef>
+
+  <nodedef name="ND_srgb_displayp3_to_lin_rec709_color4" node="srgb_displayp3_to_lin_rec709" nodegroup="colortransform">
+    <input name="in" type="color4" />
+    <output name="out" type="color4" />
+  </nodedef>
+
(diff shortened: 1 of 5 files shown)


libraries/cmlib/cmlib_defs.mtlx
```diff
@@ -87,4 +87,24 @@
     <output name="out" type="color4" />
   </nodedef>
 
+  <nodedef name="ND_srgb_displayp3_to_lin_rec709_color3" node="srgb_displayp3_to_lin_rec709" nodegroup="colortransform">
+    <input name="in" type="color3" />
```
kwokcb, line 91: Hi @dgovil . 

Minor thing: I have a fix so that default values are defined on the existing CM nodes [here](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1369). Can you add in default values on the inputs here as well. I also added a cm node test file but can be patched later. Thanks.

dgovil: Done!

jstone-lucasfilm (approved):
This change looks good to me, thanks @dgovil, and I'll reach out to the OpenColorIO team for their thoughts as well.

dgovil:
Thanks, @jstone-lucasfilm   

HP is working on a matching PR to the OCIO ACES folk and should have that up soon.

PR #1371, opened 2023-06-01
Specification Updates

dbsmythe:
Updates to the v1.39 markdown spec docs.

Main Specification:
- Removed the earlier "Shader Nodes" section and moved its important contents to a later existing also-named "Shader Nodes" section, cleaned up the text and node descriptions.
- Added clarification paragraph to "Output Nodes" section to say that "colorspace", "width", "height" and "bitdepth" are only applicable in applications that save output images, and do not affect values output to other nodes connected to this output (Slack discussion)
- Added descriptions of the new &lt;checkerboard>, &lt;round> and &lt;distance> nodes.  Also added note about integer-output variants of &lt;floor> and &lt;ceil>.
- Added note that geomcolor, geompropvalue and geompropvalueuniform can take a "colorspace" attribute for color3/4 values (Slack discussion)
- Removed notes on "add" and "multiply" about shader-semantic variants, because you can't actually add or multiply *shaders*, only BSDF's/EDF's/VDF's, and those are covered in the PBRSpec doc.

PBRSpec:
- Renamed "Shader Nodes" section to "PBR Shader Nodes" to avoid conflict with main specification heading
- Added links to the "displacement" and "surface_unlit" nodes in the main Spec doc

Supplement:
- Added "blend" input to the triplanar nodes, as this is a "no brainer" to be added officially.
- Some minor formatting fixes.

README:
- Updated to include new/updated nodes and a few other updates.

### DIFF
--- a/documents/Specification/inprog_v1.39/MaterialX.PBRSpec.md
+++ b/documents/Specification/inprog_v1.39/MaterialX.PBRSpec.md
@@ -37,7 +37,7 @@ This document describes a number of shader-semantic nodes implementing widely-us
  [BSDF Nodes](#bsdf-nodes)  
  [EDF Nodes](#edf-nodes)  
  [VDF Nodes](#vdf-nodes)  
- [Shader Nodes](#shader-nodes)  
+ [PBR Shader Nodes](#pbr-shader-nodes)  
  [Utility Nodes](#utility-nodes)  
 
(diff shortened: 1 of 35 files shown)


jstone-lucasfilm (approved):
This looks good to me, thanks @dbsmythe!

PR #1070, opened 2022-09-05
Improvements to UsdPreviewSurface and UsdUVTexture

kwokcb:
Update #1060

## Changes

* Add docs from Pixar USD spec page. Includes additional notation about `normal` input on `UsdPreviewSurface`.
* Add in vector3 version for `UsdUvTexture` along with new example to demonstrate how to map `normal`.
* Small fix to "mxdoc.py" documentation generator script.

### USD Brick Shader Example

Example brick shader using UsdPreviewSurface and UsdUvTexture (vector3 version) for normal mapping shown on right in MaterialXView. Original Standard surface version on the (using `tiledimage` and `normalmap`)
image

### Documentation (generated using mxdoc.py)

- *Nodedef*: ND_UsdPreviewSurface_surfaceshader
- *Type*: surfaceshader
- *Node Group*: pbr
- *Version*: 2.3. Is default: True
- *Doc*: USD preview surface shader

| Name | Type | Default Value | UI name | UI min | UI max | UI Soft Min | UI Soft Max | UI step | UI group | UI Advanced | Doc | Uniform |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
| **diffuseColor** | color3 | 0.18, 0.18, 0.18 |  | 0,0,0 | 1,1,1 |  |  |  |  |  | Diffuse color. When using metallic workflow this is interpreted as albedo. |  |
| **emissiveColor** | color3 | 0, 0, 0 |  | 0,0,0 |  |  | 1,1,1 |  |  |  | Emissive component. |  |
| **useSpecularWorkflow** | integer | 0 |  |  |  |  |  |  |  |  | This node can fundamentally operate in two modes: Specular workflow where you provide a texture/value to the specularColor input. Or, Metallic workflow where you provide a texture/value to the metallic input. |  |
| **specularColor** | color3 | 0, 0, 0 |  | 0,0,0 | 1,1,1 |  |  |  |  |  | Specular color to be used. Enabled when useSpecularWorkflow = 1. This is the color at 0 incidence. Edge color is assumed white. Transition between the two colors according to Schlick fresnel approximation. |  |
| **metallic** | float | 0.0 |  | 0.0 | 1.0 |  |  |  |  |  | Use 1 for metallic surfaces and 0 for non-metallic. Enabled when useSpecularWorkflow = 0. |  |
[...]

PR #1372, opened 2023-06-01
Add UsdUVTexture node vector3 specialization

pablode:
This PR adds a specialization to the UsdUVTexture node which exposes the _rgb_ output as `vector3` instead of `color3`.

This change has been discussed in #1038, #1282 and Slack. It is very similar to @kwokcb's proposal #1070, but less ambitious.

I've also ported over the procedural brick material, as in above PR, to showcase the change in action.

Remark: I've added a suffix to the existing NodeDef name for consistency, although this change may be debatable. If there are compatibility concerns, I can revert it.


jstone-lucasfilm:
@pablode One tricky issue with this proposal is that MaterialX can't currently distinguish between the `color3` and `vector3` variants of this node, since they only differ in the types of their multiple outputs (e.g. `r`, `g`, `b`, `rgb`).  If these nodes had single outputs, then MaterialX could distinguish between them in the usual way, but we don't currently have a mechanism to handle overloads by types of _multiple_ outputs.

I'm open to ideas on how to resolve this.  The simplest approach might be to make these nodes single-output, where artists would access individual channels of their outputs using standard `extract` and `convert` nodes in MaterialX.

pablode:
I was hoping that the runtime would be able to select one specialization based on the input connection type, but this does not seem to be the case and would perhaps be a challenging requirement for shader generation and implementations of the MaterialX standard.

I think requiring the use of `<extract>` nodes would come with a large authoring overhead, e.g. for roughness, metalness and other attributes. Yet another idea would be an `xyz` output as proposed in Slack, but then the implementation would deviate from the standard (although the change would only be additive and would not require additional non-USD nodes). Of course it would be different if the UsdPreviewSurface standard would be adjusted - which might make sense if we're dealing with a fundamental restriction of MaterialX, and not a technical limitation.

We could also leave the situation as it is, which would require the user to insert `<convert>` nodes for the normal input (which will be omitted in practice). Changing the type of the node to _vector3_ would fix the problems with MDL (because the type can be implicitly converted to _color3_), but documents would technically still be invalid.

kwokcb:
Hi @pablode ,
Nice to see this PR up.

1. For you question about renaming existing identifiers, as mentioned in Slack, I think you should not change this as the existing id may be referenced by the `info:id` Usd attrib.

2. You can make instance explicitly reference a definition by adding a `nodedef` attribute on the node instance (same way Usd explicitly associates defs). The other way is to specify the outputs explicitly on the instance explicitly though I think the former is more robust.

@jstone-lucasfilm , I kind of recall this and maybe a node instance which is `multioutput` but does not have enough inputs / outputs to distinguish the signature (or an explicit `nodedef` attrib) should be considered invalid ?

jstone-lucasfilm:
@kwokcb I'd probably recommend against requiring explicit `nodedef` attributes to get correct rendering behavior, as there aren't other nodes in MaterialX that require this, and it's likely that it won't be as robust and portable as the current approach.

@pablode @kwokcb What would you think about providing `UsdUVTexture` nodes with singular `color3`, `color4`, `vector3`, and `vector4` outputs?  These would handle the vast majority of use cases, and when artists need access to individual channels of these outputs they can either use the `channels` attribute in MaterialX or an `extract` or `separate` node.

This seems the most idiomatic to MaterialX, and it seems the most intuitive for an artist to understand as well.

kwokcb:
Hi @jstone-lucasfilm,
I have nothing against just have tuple versions it's just that you will need to add in an extra channel separate node which I believe means updating logic in `UsdMtlx` so that it maps MaterialX 2 nodes to 1 USDShade node.

BTW, I assume we need `vec4`, `color4` for v2.2 compatibility and can't just get away with just `vec3` and `color3` ?
@pablode, I assume you would have better insight on this than I.

pablode:
@kwokcb I'm faily certain that UsdMtlx just plainly interprets MaterialX nodes as UsdShade nodes, and does not perform any kind of reduction.

Regarding the `color4` and `vector4` overloads: yes, they are needed to retroactively fix the type mismatch for v2.2 of the standard.

@jstone-lucasfilm I like the idea if it means that we can single out individual channels using the `channels` attribute - not necessarily requiring additional nodes. However, is the `channels` attribute well-supported in user interfaces (e.g. LookdevX)?

Just another idea - if we had following setup:
```xml
<nodedef name="ND_UsdUVTexture_multioutput" node="UsdUVTexture">
  <output name="r" type="float" />
  <output name="g" type="float" />
  <output name="b" type="float" />
  <output name="a" type="float" />
</nodedef>
<nodedef name="ND_UsdUVTexture_color3" node="UsdUVTexture">
  <output name="rgb" type="color3" />
</nodedef>
<nodedef name="ND_UsdUVTexture_vector3" node="UsdUVTexture">
  <output name="rgb" type="vector3" />
</nodedef>
```
then we could differ between the NodeDefs using the `type` attribute (`multioutput`, `color3`, `vector3`).

kwokcb:
> @kwokcb I'm faily certain that UsdMtlx just plainly interprets MaterialX nodes as UsdShade nodes, and does not perform any kind of reduction.
> 
> Regarding the `color4` and `vector4` overloads: yes, they are needed to retroactively fix the type mismatch for v2.2 of the standard.
> 
> @jstone-lucasfilm I like the idea if it means that we can single out individual channels using the `channels` attribute - not necessarily requiring additional nodes. However, is the `channels` attribute well-supported in user interfaces (e.g. LookdevX)?
> 
> Just another idea - if we had following setup:
> 
> ```
> <nodedef name="ND_UsdUVTexture_multioutput" node="UsdUVTexture">
>   <output name="r" type="float" />
>   <output name="g" type="float" />
>   <output name="b" type="float" />
>   <output name="a" type="float" />
> </nodedef>
> <nodedef name="ND_UsdUVTexture_color3" node="UsdUVTexture">
>   <output name="rgb" type="color3" />
> </nodedef>
> <nodedef name="ND_UsdUVTexture_vector3" node="UsdUVTexture">
>   <output name="rgb" type="vector3" />
> </nodedef>
> ```
> 
> then we could differ between the NodeDefs using the `type` attribute (`multioutput`, `color3`, `vector3`).

I like this proposal as it's clear what signature your getting for each. As a suggestion maybe the first one could be called `ND_UsdUVTexture_rgba` ?

jstone-lucasfilm:
@pablode There's still more work to be done in providing good artist interfaces for the `channels` attribute, but I believe all MaterialX-supporting packages correctly handle the rendering of this attribute when it's present, and we have many instances of this technique in our supporting libraries and example materials.  So I believe it's safe to use this technique in practice, and we'll make an effort to integrate it well into the artist interfaces that we provide within the MaterialX project (e.g. the MaterialX Graph Editor).

jstone-lucasfilm:
@pablode @kwokcb I do like Pablo's proposal, where there are single-output nodes for each type variation of `UsdUVTexture`, plus a multioutput version for channel selection.

I wonder if existing functionality such as the `separate` node, `extract` node, and `channels` string might cover the multioutput cases just as effectively, though?  If you strongly believe that a dedicated multioutput node is still important in USD contexts, then I would support this idea, but it's worthwhile to check whether existing functionality can handle this today.

In either case, I believe the best path forward would be to convert the majority of the `UsdUVTexture` nodes into single-output nodes, so that they can easily be distinguished from each other in both MaterialX and USD.

pablode:
@jstone-lucasfilm  I think you're fully right that the channels approach covers the use cases just fine.

Adding a separate single-output node would only yield a minor usability improvement.

I'll adjust the PR to only include single-output nodes, and we can consider the other specialization if concerns arise in the future.

pablode:
@jstone-lucasfilm I've made changes to reflect what we have discussed, but something is not right yet. When viewing the procedural brick material, I'm getting following errors:
```
*** Validation warnings for usd_preview_surface_brick_procedural.mtlx ***
Node interface doesn't support this output type: <UsdUVTexture name="node_image_vector3_27" type="vector4">
Node interface doesn't support this output type: <UsdUVTexture name="node_image_float_22" type="vector4">
Node interface doesn't support this output type: <UsdUVTexture name="node_image_float_10" type="vector4">
Node interface doesn't support this output type: <UsdUVTexture name="node_image_float_7" type="vector4">
Node interface doesn't support this output type: <UsdUVTexture name="node_image_float_26" type="vector4">
Node interface doesn't support this output type: <UsdUVTexture name="node_image_float_24" type="vector4">
```

I'd appreciate if you could have a look, otherwise I'll investigate further.

jstone-lucasfilm (commented):
I really like this simplification of the UsdUVTexture nodes, thanks @pablode.  I can take a look at the remaining warnings when time permits, and for now I'd like to broaden the discussion to validate our overall approach.

I'm CC'ing @JGamache-autodesk, @kwokcb, and @crydalch for their thoughts on whether this change is safe for existing tools such as Maya and Houdini, or whether we would need to wait for a specification version update (e.g. 1.39), so that backwards compatibility logic can be added for legacy assets.

jstone-lucasfilm:
CC'ing @JGamache-autodesk for his thoughts on the topic above.

jstone-lucasfilm:
Thanks for the heads-up on that `channels` edge case, and we should definitely address that.  For the purposes of this proposal, I'd recommend using the `convert` node so that you're not blocked.

crydalch:
I'm still trying to fully grok the conversation/PR, my apologies. Adding a vector3 signature seems fine, but if the suggestion is to remove the multi-output nature from the UV Texture, that would break things, so waiting for 1.39 seems more appropriate.

UsdMaterialX is not aware of "channels", as far as I know; I'm not sure that UsdShade even supports it. And artists/users do very much appreciate the convenience of having the float outputs already on the node.

So yeah if we're removing outputs from USD UV Texture, that should probably wait for 1.39, and needs some upstream coordination with the USD project too?

jstone-lucasfilm:
Good points, @crydalch, and I'd support the idea of deferring this change to a future MaterialX version such as 1.39.

I believe UsdMtlx does support the `channels` string in MaterialX code generation, since it's such a commonly-used feature of MaterialX graph definitions (it's even used in Standard Surface), but you may be right that it's not a user-facing feature of USD itself.

Harmonizing the user-facing aspects of `channels` strings between MaterialX and USD would be valuable, and I believe there's already a proposal to simplify this feature for MaterialX 1.39, based on discussions with Autodesk, Pixar, and others in the community.

pablode:
> I'd support the idea of deferring this change to a future MaterialX version such as 1.39.

I agree, even if the proposed changes are purely additive and optional.

> UsdMaterialX is not aware of "channels", as far as I know; I'm not sure that UsdShade even supports it.

It seems that UsdMtlx does indeed [support channels](https://github.com/PixarAnimationStudios/OpenUSD/blob/6988a51c5da37f7904507ec7e48df98350893581/pxr/usd/usdMtlx/reader.cpp#L64), but this is not the case for Hydra shading networks: https://forum.aousd.org/t/passing-usdshadeinput-metadata-to-hydra/614

I'm aware that Pixar is currently working on supporting colorspaces, which are attributes in MaterialX, but I don't know if their changes extend to general attribute support in Hydra shading networks.

> And artists/users do very much appreciate the convenience of having the float outputs already on the node.

I can imagine - although my experience is limited. There's still the alternative presented in [this comment](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1372#issuecomment-1597530397).

jstone-lucasfilm:
@pablode @crydalch @dbsmythe

This side-bar discussion might be better suited for a future MaterialX TSC meeting, but I should note that the `channels` string in MaterialX was intended to represent the notion of "unpacking" a multi-channel output value in a graph editor, allowing the user to single out individual channels to connect with the inputs of other nodes.

You can see examples of this UI approach in Unreal Engine, Substance Designer, and many other material editing packages, and here's an example of how the associated UI looks in Unreal Engine:
image

Our thinking was that technique should be universally available for all multi-channel node outputs in MaterialX, and that's the reason that we describe this as a universal `channels` string rather than adding `r`, `g`, `b`, `a`, `rgb`, etc. outputs to every node in MaterialX with a color or vector output.

We'd love to have a discussion about whether our approach to this feature is the right one, and there's definitely room to revise this in future versions of MaterialX.  Right now, the proposal on the table for MaterialX 1.39 is to simplify this to a `channel` string that only allows individual channels (e.g. `r`, `g`, `b`, `a`) to be singled out, since this aligns more closely with the feature set of OSL.

Let us know whether you have ideas on improving this, and bringing in the experience of artists working with tools such as Houdini would clearly be valuable.

JGamache-autodesk:
We indeed see the same issue with Maya where we would like to be able to connect components on both outputs (which is supported via "channels") and inputs (which is uncharted MaterialX territory):
image
Since the "channels" attribute of a connection does not translate to UsdShade, and since there are no way to specify input channels, we instead must rely on adding regular MaterialX conversion nodes starting from regular `convert`, but sometimes descending into `swizzle`, `extract`, `separate`, and `combine` territory. Very robust, but it adds quite a few extra nodes in the view that a user might not expect.

In that case, having multiple single-output UsdUVTexture NodeDefs covering all image output types is the solution we prefer.

kwokcb:
I'll put in an idea for an alternative to both the new implicit single channel attribute and using additional nodes.

This is to add child channels as actual ports which can be connected. Thus a N channel input can have up to N child ports. 

These could added as needed and be directly connectable to child ports on outputs as needed. For serialization the appropriate link syntax would need to be extended.

So you'd have something like this:
```xml
<node name"foo" type="color4">
  <output name="out" type="color4">
    <channel name="a">
  </output>
</node>

<node name="blah">
  <input name="mycolor4" type="color4">
     <channel name="r" nodename="foo" output="out/a">
  </input>
</node>
```
Or a more USD / full-path like connection syntax like:
```xml
<node name="blah">
  <input name="mycolor4" type="color4">
     <channel name="r" connection="foo/out/a">
</node>
```

jstone-lucasfilm:
That's a great example from Maya, @JGamache-autodesk, and as you note it goes even further than MaterialX by allowing connections to individual channels of inputs.  So far we hadn't seen quite enough support for this feature in material graph editors to include it in the standard, but I can certainly see the power it would provide.

Your fallback approach of converting `channels` strings to explicit separate/extract/combine nodes is exactly what we had imagined would be used in applications that don't support direct `channels` connections, though for the USD/MaterialX case it would be ideal to harmonize on a set of per-channel connection features that both standards can embrace.

jstone-lucasfilm:
As an update from our last MaterialX TSC meeting, we're planning to move to a stricter approach to version upgrades, with one important benefit being that node signature upgrades will be easier to fast track and include in MaterialX 1.39:
https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1705437060523459

In that spirit, one compelling way to address this PR would be to move entirely to single-output signatures for `UsdUVTexture` in MaterialX 1.39, and to add an upgrade step that converts legacy references to the new signatures.

As an example of this pattern, Lee Kerley has recently proposed an update to the signature of the `atan2` node, including the appropriate code to upgrade legacy references:
https://github.com/AcademySoftwareFoundation/MaterialX/pull/1659

@pablode If you have the bandwidth, we'd be happy to provide guidance on writing a contribution to MaterialX 1.39 along these lines.  No problem if time doesn't allow, and we can keep this idea on GitHub Issues for the future.

pablode:
Sorry, I don't know when I'll come back to this PR. If I understand it correctly, the following work still needs to be done:
1. isolate and create issue for `channels` edge-case in [this comment](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1372#issuecomment-1710382123)
2. use a `<convert>` node in usd_preview_surface_brick_procedural.mtlx to circumvent the error
3. revert "Reintroduce multioutput NodeDefs for backwards compatibility" commit
4. implement upgrade logic from old to new NodeDefs

jstone-lucasfilm:
That sounds like a good summary to me, @pablode, and perhaps the best approach forward is to capture this plan in a new GitHub Issue, so that any developer with the bandwidth can take this on.  The existing PR can be referenced as a starting point for the work, but doesn't need to be directly revised and merged.

pablode:
Sounds good to me!

jstone-lucasfilm:
I've written up https://github.com/AcademySoftwareFoundation/MaterialX/issues/1673 to cover this proposed work, and I'll go ahead and close out this original pull request.  Thanks for the initial research and development on this issue, @pablode, and I believe we have a good path forward!

PR #1376, opened 2023-06-12
Deterministic code generation

SShikhali:
These changes make code generation deterministic by changing `std::set` to `std::vector` when removing unused nodes

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -1073,7 +1073,7 @@ void ShaderGraph::optimize(GenContext& context)
 
     if (numEdits > 0)
     {
-        std::set<ShaderNode*> usedNodes;
+        std::vector<ShaderNode*> usedNodes;
 
         // Traverse the graph to find nodes still in use
         for (ShaderGraphOutputSocket* outputSocket : getOutputSockets())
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks for this fix, @SShikhali, and let us know if you run into trouble with the EasyCLA process.  The Linux Foundation is usually very responsive if you submit a support ticket using the links above.

SShikhali:
@jstone-lucasfilm should be fixed now, thanks :)

jstone-lucasfilm (commented):
Thanks for putting this fix together, @SShikhali!

To make sure we don't lose any efficiency with respect to the original version, it may be better to maintain both a vector and a set, as in the following:

```
std::set<ShaderNode*> usedNodeSet;
std::vector<ShaderNode*> usedNodeVec;
```

This would allow us to efficiently maintain a vector of unique nodes with logic like the following:

```
ShaderNode* node = edge.upstream->getNode();
if (usedNodeSet.count(node) == 0)
{
    usedNodeSet.insert(node);
    usedNodeVec.push_back(node);
}
```

An at the end of the function, we could then simply replace the `resize` and `assign` calls with the following:

```
_nodeOrder = usedNodeVec;
```

SShikhali:
@jstone-lucasfilm yes, that's better. Pushed the improved version

jstone-lucasfilm (approved):
This looks good to me, thanks @SShikhali!

I'm CC'ing @niklasharrysson for his thoughts before we merge, but I don't have any concerns on my side.

niklasharrysson (approved):
Looks great! Thanks for the fix @SShikhali

Issue #1354, opened 2023-05-15
MDL specifies duplicate implementations for mix volumeshader / displacementshader


kwokcb:
The MDL file `stdlib_genmdl_impl.mtlx` has 2 implementations with the same Element name as used for surfacesahder mix.
When read in they will be not be read in. 
```
 <implementation name="IM_mix_surfaceshader_genmdl" nodedef="ND_mix_volumeshader" sourcecode="mx::stdlib::mx_mix_volumeshader({{fg}}, {{bg}}, {{mix}})" target="genmdl" />
 <implementation name="IM_mix_surfaceshader_genmdl" nodedef="ND_mix_displacementshader" sourcecode="mx::stdlib::mx_mix_displacementshader({{fg}}, {{bg}}, {{mix}})" target="genmdl" />
```

Changing the names to `IM_mix_volumeshader_genmdl` and `IM_mix_displacementshader` however results in
the side-effect of `genreference` test failing for some reason (throws an exception in this code:
```
 catch (mx::ExceptionRenderError& e)
        {
            logFile << "Error compiling OSL reference for '" << nodeName << "' : " << std::endl;
            logFile << e.what() << std::endl;
            for (const std::string& error : e.errorLog())
            {
                logFile << error << std::endl;
            }
            failedGeneration = true;
        }
```

jstone-lucasfilm:
Thanks for this report, @kwokcb, and to @niklasharrysson for the fix in #1380.

PR #1380, opened 2023-06-15
Assign unique names to MDL implementations

niklasharrysson:
This change list addresses issue #1354, renaming the implementation elements to unique names.


jstone-lucasfilm (approved):

Issue #878, opened 2022-03-18
Reading texture path from primvar


flord:
I am currently working on packaging USD assets for public release and I though it would be a good idea to use the MaterialX shaders instead of the Usd Preview Surface. I'm using Houdini 19 Solaris to assemble the asset and I'm testing with Karma and Arnold. 
I came accross an issue that prevents me from using MaterialX entirely. In our USD pipeline, we create a USD layer for the texture department. This layer adds primvars to the assets, and these primvars contain the path to the textures. Later, when the shading department creates a material, they simply read these primvars and use them to drive the texture path parameter of the textures nodes. This workflow works well with Karma and Arnold when using their respective shaders as well as when using the USD shaders. I realized that we can not drive the filename parameter of the `MtlX Image` and `MtlX Tiled Image` nodes with a `MtlX USD Primvar Reader` node. I saw in the documentation that MaterialX supports filename substitution in those nodes, and that it supports a \<geometry token\> to use 
> "The value of a specified token declared in a \<geominfo\> element for the current geometry." 

I tried different things but could not use this token to read a primvar on the geometry I'm using. 

Here is why we want to use this workflow. By having the texture paths in a seperate layer than the shading, it allows us to completely decouple the two pipeline steps. When there are new versions of texture, no one need to update the shading since it reads directly from the texture layer. It also allows us to create texture variants independently from shading variants. Setting the texture variantSet only changes the paths read by the shaders. We can also share the same material on multiple meshes but each one has a different texture. 

Just to be clear, I'm not necessarily after a solution to drive the path of a texture node from a shader, but mostly after a solution to use primvars to drive texture paths. 

Thanks!

Francois Lord
RodeoFX

kwokcb:
Hi Francois,

The current logic to add in "default" `geominfo` resolving is [here](
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Element.cpp#L422-L436)
When you create a string resolver you can optionally pass over a `geomInfo` name to use.

Based on the spec the `geomInfo` tokens are specified at that level. e.g. from spec doc:
```
<geominfo name="gi1" geom="/a/g1">
  <token name="txtid" type="string" value="Lengine"/>
</geominfo>

and then reference that token's value in a filename:

<image name="cc1" type="color3">
<input name="file" type="filename" value="txt/color/asset.color.<txtid>.tif"/>
</image>
```
[...]

PR #1381, opened 2023-06-17
Add udimextract and udimtexcoord nodes

crydalch:
These are two new nodes which let users commandeer UDIM support, to drive texture variation.

- `udimtexcoord` - provides the texture coordinates of the specified UDIM patch. The UDIM patch can be driven by geometry properties, letting users drive texture maps using primvars.
- `udimextract` - extracts the UDIM patch from the given texture coordinates.

MaterialXViewer and MaterialXGraphEditor both don't recognize image paths with the `<UDIM>` token, but here is a render from Karma. Each rubber toy is assigned a random integer primvar between 1001 and 1025.

udim_texcoord_example

### DIFF
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -394,6 +394,25 @@
     <output name="out" type="vector4" default="0.0, 0.0, 0.0, 0.0" />
   </nodedef>
 
+  <!--
+    Node: <udimtexcoord> Supplemental Node
+    Texture coordinates of the specified UDIM patch.
+  -->
+  <nodedef name="ND_udimtexcoord" node="udimtexcoord" nodegroup="texture2d">
+    <input name="patch" type="integer" uiname="UDIM Patch" value="1001" uimin="1001" uimax="1099" doc="The UDIM patch whose texture coordinates." />
+    <input name="texcoord" type="vector2" uiname="Texture Coordinates" defaultgeomprop="UV0" doc="The input 2d space. Default is the first texture coordinates." />
+    <output name="out" type="vector2" />
+  </nodedef>
+
+  <!--
+    Node: <udimextract> Supplemental Node
+    Extracts the UDIM patch from the given texture coordinates.
+  -->
+  <nodedef name="ND_udimextract" node="udimextract" nodegroup="texture2d">
+    <input name="texcoord" type="vector2" uiname="Texture Coordinates" defaultgeomprop="UV0" doc="The input 2d space. Default is the first texture coordinates." />
+    <output name="out" type="integer" />
+  </nodedef>
+
   <!-- ======================================================================== -->
   <!-- Procedural nodes                                                         -->
   <!-- ======================================================================== -->
--- a/libraries/stdlib/stdlib_ng.mtlx
+++ b/libraries/stdlib/stdlib_ng.mtlx
@@ -844,6 +844,96 @@
     <output name="out" type="vector4" nodename="N_add2_vector4" />
   </nodegraph>
 
+  <!--
+    Node: <udimtexcoord> Supplemental Node
+    Texture coordinates of the specified UDIM patch.
+  -->
+  <nodegraph name="NG_udimtexcoord" nodedef="ND_udimtexcoord">
+    <convert name="N_convertUdimPatch" type="float">
+      <input name="in" type="integer" interfacename="patch" />
+    </convert>
+    <clamp name="N_clampUdimRange" type="float">
+      <input name="in" type="float" nodename="N_convertUdimPatch" />
+      <input name="low" type="float" value="1001" />
+      <input name="high" type="float" value="1099" />
+    </clamp>
+    <modulo name="N_ensurePeriodicUVs" type="vector2">
+      <input name="in1" type="vector2" interfacename="texcoord" />
+      <input name="in2" type="vector2" value="1, 1" />
+    </modulo>
+    <modulo name="N_modIndex" type="float">
+      <input name="in1" type="float" nodename="N_clampUdimRange" />
+      <input name="in2" type="float" value="1001" />
+    </modulo>
+    <divide name="N_divideTen" type="float">
+      <input name="in1" type="float" nodename="N_modIndex" />
+      <input name="in2" type="float" value="10" />
+    </divide>
+    <separate2 name="N_splitUV" type="multioutput">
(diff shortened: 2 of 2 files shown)


crydalch:
While these nodes are related to the same artistic needs motivating issue #878, they are not intended to be a complete solution. What this does provide is a very simple way for artists to drive texture variation with primvars. Other than UDIM support, no additional work is needed on the part of the renderer or MaterialX.

Currently these nodes limit the UDIMs to 1001-1099 (inclusive), which is the range of UDIMs that USD supports. However if UDIM ever becomes an actual specification, we can adjust the limits accordingly.

jstone-lucasfilm:
@crydalch How does this proposal relate to the existing support for `<UDIM>` tokens in MaterialX and USD?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#filename-substitutions

These tokens currently work as expected in MaterialXView and MaterialXGraphEditor, for example, and I've successfully tested multi-UDIM MaterialX files in Houdini/Karma as well.  Houdini/Karma seems to understand the native UDIM support in MaterialX, and will correctly assign each UDIM texture to the associated geometric UDIM at render time.

crydalch:
Thanks @jstone-lucasfilm! Yeah this isn’t trying to “fix” UDIM support. These nodes allow users to leverage existing UDIM support, to drive texture variation using primvars. Full string/primvar token support (i.e. what’s been discussed in #878) is going to require more work on the part of MaterialX, UsdMaterialX, USD, and render delegates. While eventually that would likely be the more flexible approach, this seemed like a good alternative, which requires nothing beyond UDIM support. So even when primvar-substitution lands, (ab)using UDIM textures like this proves useful.

Maybe changing the names would help clarify things. Instead of `udimtexcoord` we could use `udimoffset`? Or `udimextract` could be `udimextractpatch`?

BTW, I actually haven’t been able to get a `<UDIM>` texture work in MaterialXViewer or MaterialXGraphEditor. I just get a message saying `Image file not found: /home/chrisr/Desktop/udimgrids/uvgrid.<UDIM>.jpg`

Screenshot_20230617_125445

Maybe something with my local build is the culprit? Or is there something messed up with this dataset? The bunny works in Houdini just fine. Does MaterialX have example geometry/texture resources for UDIMs?

Here is the .obj:

[bunny.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/11779813/bunny.zip)

Thanks!

jstone-lucasfilm:
In MaterialX, you just need to state the `udimset` for a multi-UDIM material, so that it can be rendered run through material processing in tools that don't have access to the geometry as well.

Here's a multi-UDIM version of the bunny, with the corresponding render in MaterialXView:

[MultiUdimBunny.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/11780073/MultiUdimBunny.zip)

MultiUdimBunny_MaterialXView

crydalch:
Ah - it was me :) Thank you for that example, I see it working now in MaterialX tools!

Screenshot_20230617_171431

Just to clarify the usage a bit better, I've attached image shows some usage examples. The shader has `udimtexcoord` connected to an `image` node. Images 1-3 are manually setting the UDIM patch on `udimtexcoord`, while the 4th image is reading the udim number from a primvar on each instance.

udimtexcoord_usage_examples

Maybe we can chat about this a the next TSC meeting? I'm still wrapping my head around the whole `udimset` concept, but it's possible that (ab)using these nodes to hijack UDIM support for primvar-driven texture variation can't work in a pure MaterialX environment.

jstone-lucasfilm:
@crydalch Sure, let's plan to pick up this topic in our upcoming MaterialX TSC meeting.

I think the difference you're encountering may not be between MaterialX and USD/MaterialX workflows (which have nearly identical interpretations of UDIMs), but between _real-time_ and _off-line_ renderers (e.g. Storm vs. Karma).

Real-time renderers traditionally require all textures referenced by an asset to be loaded in advance of shader execution, while off-line renderers can make this decision dynamically at shade time.  This makes it much more challenging for a real-time renderer to respond to shader-computed UDIM indices by loading a new set of textures from disk, and it's common for this category of renderer to disallow this approach.

kwokcb:
Adding some small notes based thoughts from the TSC. This is in the context that a texture atlas is used for real-time:

1. I think you still need to have the full udim set available otherwise the atlas is unbounded.
2. There are utilities to remap to a normalized atlas which pulled from Maya H/W rendering and are now found in core utilities and used by the h/w shader generator image node (tiled image can work if the atlas is normalized I think). 
But a normalized atlas still needs requires a udim set.
3. In Storm I'd assume the hierarchical tiling system also needs pre-knowledge of the udim set (range).
4. Adding the onus of something like the Unreal 2-pass system to id patches for resource loading sounds too complicated for something like the sample viewer.

So I think as long as the udim lookup does not go outside the udim set per image then it's something that can work theoretically if you have  normalized udim atlas (static or dynamic). BTW, I've always found the MaterialX udim set specification is a bit "odd" in that it's Document level however -- but these nodes are for a specific image.

jstone-lucasfilm:
Just to align our two discussions of this subject, I'll forward a high-level summary from today's MaterialX TSC meeting, where we had the chance to discuss this topic a bit:

> Three common approaches for handling UDIMs in rasterization engines are:
> 
> 1. Partition the geometry based on its texture coordinates, and assign each UDIM to its corresponding partition (e.g. MaterialXView)
> 2. Combine all possible UDIMs into a texture array, and select which UDIM is rendered at shade time (e.g. Storm)
> 3. Dynamically load textures into memory using Sparse Virtual Texturing (e.g. Mari, UE5)
> 
> Approaches (1) and (2) are relatively limited, and either place a cap on UDIM count or prevent dynamic UDIM offsets.
> Approach (3) potentially allows both arbitrary UDIM counts and dynamic UDIM offsets.

My sense is that approach (3) is the ideal, but a more practical alternative would be for MaterialX to support _both_ (1) and (2), allowing clients of the library to choose which one they prefer.  This would allow applications such as Storm to continue using pre-generated texture arrays, and would allow MaterialXView to switch between these two approaches dynamically as needed by users.

jstone-lucasfilm:
In a side discussion with @crydalch, we resolved that it makes more sense to work with the USD/MaterialX teams to resolve the core issue in https://github.com/AcademySoftwareFoundation/MaterialX/issues/878, so I'm going to close this pull request for now.

There's a USD/MaterialX working group within the ASWF that focuses on these alignment issues, and this subject would be a great one to bring up at their next meeting.

PR #1386, opened 2023-06-22
Add Python bindings for generation and rendering

kwokcb:
Most of these wrapper updates are related to exposing missing APIs in Python for code gen and rendering.
For now this is for GLSL rendering in Python this is sufficient ([driver is this example](https://kwokcb.github.io/MaterialX_Learn/pymaterialx/mtlx_render_notebook.html))


source/PyMaterialX/PyMaterialXRender/PyCgltfLoader.cpp
```diff
@@ -0,0 +1,18 @@
+//
```
kwokcb, line 1: GLTF loader was missing so can't load a number of sample geometry.

source/PyMaterialX/PyMaterialXCore/PyDefinition.cpp
```diff
@@ -43,6 +43,9 @@ void bindPyDefinition(py::module& mod)
         .def("getFunction", &mx::Implementation::getFunction)
         .def("setNodeDef", &mx::Implementation::setNodeDef)
         .def("getNodeDef", &mx::Implementation::getNodeDef)
+        .def("setNodeGraph", &mx::Implementation::setNodeGraph)
```
kwokcb, line 46: This is actually for definition creation but added in to avoid another PR.

source/PyMaterialX/PyMaterialXGenShader/PyGenOptions.cpp
```diff
@@ -27,12 +27,18 @@ void bindPyGenOptions(py::module& mod)
         .def_readwrite("shaderInterfaceType", &mx::GenOptions::shaderInterfaceType)
         .def_readwrite("fileTextureVerticalFlip", &mx::GenOptions::fileTextureVerticalFlip)
         .def_readwrite("targetColorSpaceOverride", &mx::GenOptions::targetColorSpaceOverride)
+        .def_readwrite("addUpstreamDependencies", &mx::GenOptions::addUpstreamDependencies)
```
kwokcb, line 30: This is required to be able to generate reference code for nodes.

jstone-lucasfilm (approved):
These changes look good to me, thanks @kwokcb!

PR #1387, opened 2023-06-22
Updating usability for Graph Editor Add Node

lfl-eholthouser:
Updating the add node popup to clear the input after it is closed and allow for spaces to be used instead of "_" when search for nodes.


jstone-lucasfilm (approved):
Looks good to me, thanks @lfl-eholthouser!

PR #1389, opened 2023-06-26
Auto layout fix for subgraphs

lfl-eholthouser:
This is a bug fix for auto layout for nodes like triplanarprojection. This pull request also includes formatting fixes.


jstone-lucasfilm (approved):
Thanks for the fix, @lfl-eholthouser!

PR #1391, opened 2023-06-29
Expose GLTF loader in Python

kwokcb:
Add GLTF loader to Python build so it's available.
> Note to @jstone-lucasfilm. Just adding file forgot to add to last PR for rendering API update #1386


jstone-lucasfilm (approved):
Looks good to me!

PR #1395, opened 2023-07-05
Fix volume mixes in MDL

krohmerNV:
Change the implementation of the volume mix functions in MDL.
The mix weight is now interpreted as probability to encounter a particle of one of the mixed media.


jstone-lucasfilm (approved):
Looks good to me, thanks @krohmerNV!

PR #1399, opened 2023-07-06
Fix bug in BSDF code gen for HW languages

niklasharrysson:
This change list fixes a bug in code gen for HW languages, where under certain conditions the function call for a BSDF node could be generated multiple times in the same scope.

For example, if a pure reflection BSDF node, like oren_nayar_bsdf, where referenced multiple times in a graph. Then during generation of the transmission section of a GLSL surface shader the node's output variable would be emitted multiple times, giving erroneous code.

A workaround was previously to duplicate the node in the graph, as for example done in the GlftPbr graph. With this fix duplication is no longer needed.

### DIFF
--- a/libraries/bxdf/gltf_pbr.mtlx
+++ b/libraries/bxdf/gltf_pbr.mtlx
@@ -148,11 +148,6 @@
     <!-- Thin-film + Dielectric
          Note: Due to limitations in codegen, the base layer BSDF is duplicated (#1035). -->
 
-    <oren_nayar_diffuse_bsdf name="tf_diffuse_bsdf" type="BSDF">
-      <input name="color" type="color3" interfacename="base_color" />
-      <input name="normal" type="vector3" interfacename="normal" />
-    </oren_nayar_diffuse_bsdf>
-
     <dielectric_bsdf name="tf_transmission_bsdf" type="BSDF">
       <input name="weight" type="float" value="1" />
       <input name="tint" type="color3" interfacename="base_color" />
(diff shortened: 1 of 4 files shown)


libraries/bxdf/gltf_pbr.mtlx
```diff
@@ -148,11 +148,6 @@
     <!-- Thin-film + Dielectric
          Note: Due to limitations in codegen, the base layer BSDF is duplicated (#1035). -->
```
jstone-lucasfilm, line 149: Can this comment now be removed as well?

jstone-lucasfilm (commented):
This looks like a great improvement, thanks @niklasharrysson, and I had just one question about a remaining comment in the graph definition.

libraries/bxdf/gltf_pbr.mtlx
niklasharrysson: Hey Jonathan! I think this comment is meant for the dielectric base layer under thin-film layering. The specular reflection and transmission BSDF's below are also duplicated, in order to have the dielectric layer both with and without thin-film, to blend between. That duplication is needed for another code gen limitation. You can't have the same BSDF node both with and without thin-film (even if it's possible to create such a graph).

However, I have another PR coming up after this one, with the change to thin-film handling that we discussed, where thin-film is set by parameters instead of an extra layer operator. That change will modify this graph as well, and I will remove the comment since it's then no longer valid.

jstone-lucasfilm: Thanks for the clarification, @niklasharrysson!

jstone-lucasfilm (approved):
Looks good to me!

PR #1400, opened 2023-07-07
Update image links to support Markdown preview

utzcoz:
Links start with "/" doesn't work for Markdown preview of CLion on Windows..


jstone-lucasfilm (approved):
Thanks for the fix, @utzcoz!

PR #1402, opened 2023-07-09
gitignore: Ignore .idea directory

utzcoz:
The Intellij IDEA family IDEs create .idea directory when opening the project with them. This directory is not used at many times, so this CL just ignore it with .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 build
+.idea


jstone-lucasfilm:
@utzcoz Thanks for this proposal!  So far, we've aimed to keep the main `gitignore` as minimal as possible, since there are so many development approaches and IDEs in the industry.  My initial sense is that we may want to omit this change, and continue to let developers locally modify the `gitignore` as needed.

utzcoz:
@jstone-lucasfilm Thanks for reviewing and pointing out team's thought about gitignore. I got it now, and will not send related PRs.

PR #1403, opened 2023-07-11
Improvements to publishing API

kwokcb:
## API Logic Updates

This s a set of fixes to the existing "publishing" / interface API to fix logic errors and conform properly with the current specification. 

### `<nodedef>` Updates

* The logic for creating `<nodedefs>` has logic errors in it so adding this PR to fix it up and to make sure it matches the current specification.  Of note was that inputs were not being removed from functional graphs, inputs were not being published, and attributes which are disallowed on functional graphs or nodedefs were not being filtered out.

### `<nodegraph>` Updates
* Publishing interfaces is missing compound `<nodegraph>` (vs functional) support so add it in so input interfaces can be "published" from nodes for both compound and functional graphs. 

### General
* Undesired attributes on are now filtered out properly on create. his includes handling positional information (`xpos`, `ypos`) which is generated by the `MaterialXNodeEditor` and `sourceURI`.
* Fix so that add / remove interface transfers values from interfaces from and to interior inputs properly when adding and removing `interfacename`.

### Test
* Unit test has been simplified and update
* Sample of possible nodedef creation with fixes:

```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <!--
	Node: <mymarble> 
	Custom Marble
  -->
  <nodedef name="ND_mymarble_1_0_color3" node="mymarble" nodegroup="texture2d" version="1.0" isdefaultversion="true" doc="This is a custom marble definition" namespace="myspace">
    <input name="base_color_1" type="color3" value="0.8, 0.8, 0.8" uiname="Color 1" uifolder="Marble Color" />
    <input name="base_color_2" type="color3" value="0.1, 0.1, 0.3" uiname="Color 2" uifolder="Marble Color" />
    <input name="noise_scale_1" type="float" value="6.0" uisoftmin="1.0" uisoftmax="10.0" uiname="Scale 1" uifolder="Marble Noise" />
    <input name="noise_scale_2" type="float" value="4.0" uisoftmin="1.0" uisoftmax="10.0" uiname="Scale 2" uifolder="Marble Noise" />
    <input name="noise_power" type="float" value="3.0" uisoftmin="1.0" uisoftmax="10.0" uiname="Power" uifolder="Marble Noise" />
    <input name="noise_octaves" type="integer" value="3" uisoftmin="1" uisoftmax="8" uiname="Octaves" uifolder="Marble Noise" />
  </nodedef>
</materialx>
```
and
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <!--
	Node: <mymarble> 
	Custom Marble
  -->
  <nodegraph name="NG_mymarble_1_0_color3" nodedef="ND_mymarble_1_0_color3" namespace="myspace">
    <position name="obj_pos" type="vector3" />
    <dotproduct name="add_xyz" type="float">
      <input name="in1" type="vector3" nodename="obj_pos" />
      <input name="in2" type="vector3" value="1, 1, 1" />
    </dotproduct>
    <multiply name="scale_xyz" type="float">
      <input name="in1" type="float" nodename="add_xyz" />
      <input name="in2" type="float" interfacename="noise_scale_1" />
    </multiply>
    <multiply name="scale_pos" type="vector3">
      <input name="in1" type="vector3" nodename="obj_pos" />
      <input name="in2" type="float" interfacename="noise_scale_2" />
    </multiply>
    <fractal3d name="noise" type="float">
      <input name="octaves" type="integer" interfacename="noise_octaves" />
      <input name="position" type="vector3" nodename="scale_pos" />
    </fractal3d>
    <multiply name="scale_noise" type="float">
      <input name="in1" type="float" nodename="noise" />
      <input name="in2" type="float" value="3.0" />
    </multiply>
    <add name="sum" type="float">
      <input name="in1" type="float" nodename="scale_xyz" />
      <input name="in2" type="float" nodename="scale_noise" />
    </add>
    <sin name="sin" type="float">
      <input name="in" type="float" nodename="sum" />
    </sin>
    <multiply name="scale" type="float">
      <input name="in1" type="float" nodename="sin" />
      <input name="in2" type="float" value="0.5" />
    </multiply>
    <add name="bias" type="float">
      <input name="in1" type="float" nodename="scale" />
      <input name="in2" type="float" value="0.5" />
    </add>
    <power name="power" type="float">
      <input name="in1" type="float" nodename="bias" />
      <input name="in2" type="float" interfacename="noise_power" />
    </power>
    <mix name="color_mix" type="color3">
      <input name="bg" type="color3" interfacename="base_color_1" />
      <input name="fg" type="color3" interfacename="base_color_2" />
      <input name="mix" type="float" nodename="power" />
    </mix>
    <output name="out" type="color3" nodename="color_mix" />
  </nodegraph>
</materialx>
```


source/MaterialXCore/Document.cpp
kwokcb, line 1: Jonathan, This is the original change from sometime in April / May. It patches to perform the transfer of input interfaces from the `nodegraph` to the `nodedef` and strips up undesired attributes and connections.

source/MaterialXCore/Node.cpp
```diff
@@ -657,34 +657,33 @@ void Node::addInputsFromNodeDef()
     }
 }
 
-void NodeGraph::addInterfaceName(const string& inputPath, const string& interfaceName)
+InputPtr NodeGraph::addInterfaceName(const string& inputPath, const string& interfaceName)
```
kwokcb, line 660: This fixes to support adding interfaces on nodegraphs or nodedefs for nodegraphs properly.

python/Scripts/createdefinition.py
kwokcb, line 1: Adds the wrapping logic to call the api properly and perform additional operations on the functional nodegraph or nodedef. Also extracts out these into another document for save.

python/Scripts/createdefinition.py
```diff
@@ -0,0 +1,205 @@
...
+                definition.setNamespace(namespace)
+        funcgraph = doc.getNodeGraph(nodegraphName)
+        if funcgraph and namespace:
+                funcgraph.setNamespace(namespace)
+
+        if not definition or not funcgraph:
+            print('Failed to create definition for nodegraph %s' % nodeGraph.getNamePath())
+            continue
+
+        if not use_1_38_8:
```
kwokcb, line 118: This will work pre the logic updates by patching after the fact here.

python/Scripts/createdefinition.py
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/env python
+'''
+Create definitions based on the nodegraphs in the specified document. Definitions are 
+saved to new documents with either the definition and functional graph in a single file or
+in separated into a file for definitions and a file for functional graphs.
+
+It is assumed that all nodegraphs have the same category, nodegroup, version and namespace
+as only one set of these input options can be provided. It is also assumed that each nodegraph
+has a different output signature. If the same signature if out then no definition will be created
+for that nodegraph.
```
jstone-lucasfilm, line 10: Perhaps we should omit this Python example, since it's relatively limited in its target user base?  If we have teams asking for this functionality in the future, we can propose it again, but for now it seems to be putting the cart before the horse!

kwokcb: Hi @jstone-lucasfilm,

With new nodes coming on line, and folks are hand editing them or using some custom nodedef generation that requires a DCC, it seems that utility is of use. How about I move this to a new PR, to get some feedback
and conversation as to what criteria and logic is of interest.

( Aside: I would actually prefer the previously brought up idea of have a Python "examples" or "utilities" library added.  The command line scripts are okay, but have overlap for some functionality which would be useful to pull out. )

What do you think ?

jstone-lucasfilm: @kwokcb I'm definitely interested in providing good examples of publishing workflows, which can at first be used directly by users in open-source tools like MaterialXGraphEditor, and eventually used as a template when proprietary tools implement their own publishing strategies.

So far, though, it's not clear that Python scripts will be a part of these workflows, and my initial sense is that we want to focus on the API level functionality (as the remainder of this pull request does), and then determine the best integration into MaterialXGraphEditor (as your counterpart pull request does).

In order to facilitate each of these steps, my recommendation would be to make each pull request as minimal as possible, so that we can hammer out the details of the required functionality that each next step will build upon.  In hammering out those details, I think it's likely that the direction of the steps that follow will be significantly altered, so there's a huge advantage in keeping them as simple as possible.

kwokcb: Sounds good. I've remove the sample script.

source/MaterialXCore/Document.h
```diff
@@ -339,9 +339,9 @@ class MX_CORE_API Document : public GraphElement
     /// @param node Node type for the new declaration
     /// @param version Version for the new declaration
     /// @param isDefaultVersion If a version is specified is thie definition the default version
+    /// @param nodeGroup Optional node group for the new declaration. The Default value is an emptry string.
     /// @param newGraphName Make a copy of this NodeGraph with the given name if a non-empty name is provided. Otherwise
     ///        modify the existing NodeGraph. Default value is an empty string.
```
jstone-lucasfilm, line 344: @kwokcb Maybe it's just me, but I'm still struggling with the idea that `addNodeDefFromGraph` modifies the graph itself, which seems very unexpected from a user perspective.  Could these two steps be decoupled, so that `addNodeDefFromGraph` treats the input graph as constant, with a separate non-const method that converts a compound graph into a functional graph?

As far as I know, there aren't yet any community examples of tools that leverage the original implementation of `addNodeDefFromGraph`, so I think we're in a good position to refactor and simplify it, and we can confirm on the MaterialX Slack before proceeding.

jstone-lucasfilm (commented):
Thanks for the updates, @kwokcb, and I had a few additional thoughts on the API that we're presenting to the user.

source/MaterialXCore/Document.h
kwokcb: Hmm. This feels wrong to me -- and I don't recall why this logic was added...
So as you say this method will always add in a new nodegraph. It will still fail if a graph of the same name exists and no unique name creation will occur (as is the case now).

Having a second version which converts an existing compound to a functional can be added later if needed as you suggest.

kwokcb: I've simplified the interface as much as I can see possible. The minimal is:
1. A compound graph
2. A new nodedef name and category
3. A new functional graph name
I removed version and group as these are not necessary.

jstone-lucasfilm (commented):
This proposal is coming along well, @kwokcb, and it looks like we just need to update JavaScript bindings to reflect the latest API signatures.

kwokcb:
Missed this. Put in a fix.

source/MaterialXCore/Document.cpp
```diff
@@ -182,47 +182,83 @@ void Document::initialize()
...
+
+    for (auto graphChild : graph->getChildren())
+    {
+        graphChild->removeAttribute("xpos");
+        graphChild->removeAttribute("ypos");
+    }        
     graph->setNodeDefString(nodeDefName);
 
+    // Create a new nodedef and set it's category
     NodeDefPtr nodeDef = addChild<NodeDef>(nodeDefName);
```
jstone-lucasfilm, line 216: It might be clearer to call `Document::addNodeDef` here.  If you don't want to set the `type` of the NodeDef, you can set this argument to the empty string, only passing the `node` string to the constructor.

source/MaterialXCore/Document.cpp
```diff
@@ -182,47 +182,83 @@ void Document::initialize()
...
-    {
-        nodeDef->setVersionString(version);
+    // Expose any existing interfaces from the graph.
+    // Any connection attributes ("nodegraph", "nodename", "channels", "interfacename") on the 
+    // existing interface should be removed from the definition as well as any source URI.
+
+    // Attributes which should not be copied over
+    StringSet filterAttributes = { PortElement::NODE_GRAPH_ATTRIBUTE, PortElement::NODE_NAME_ATTRIBUTE, 
+                                   PortElement::CHANNELS_ATTRIBUTE, PortElement::INTERFACE_NAME_ATTRIBUTE,
+                                   "xpos", "ypos" };
```
jstone-lucasfilm, line 226: It's probably time to add these two strings as static members of the `Element` class, so that we don't continue propagating the string literals throughout the codebase.  What are your thoughts?

source/MaterialXCore/Document.cpp
kwokcb: I'll make the change.

source/MaterialXCore/Document.cpp
kwokcb: Agreed. We should be using predefined constant strings especially if temp string allocation can affect performance. I'll add the constants and replace all existing instances.

source/MaterialXTest/MaterialXCore/Node.cpp
```diff
@@ -627,97 +627,116 @@ TEST_CASE("Node Definition Creation", "[nodedef]")
...
 
         // Add new version 
         const std::string VERSION2 = "2.0";
         newGraphName = mx::EMPTY_STRING;
-        nodeDef = doc->addNodeDefFromGraph(graph, newNodeDefName + "2", NODENAME, VERSION2, isDefaultVersion, GROUP, newGraphName);
+        newNodeDefName = doc->createValidChildName("ND_" + graph->getName() + "_2");
+        newGraphName = doc->createValidChildName("NG_" + graph->getName() + "_2");
+        nodeDef = doc->addNodeDefFromGraph(graph, newNodeDefName + "2", NODENAME, newGraphName);
+        nodeDef->setVersionString(VERSION2);
+        //nodeDef->setDefaultVersion(isDefaultVersion);
```
jstone-lucasfilm, line 734: Should this commented line of code now be removed?

jstone-lucasfilm (commented):
This looks nearly ready to merge, thanks @kwokcb, and I had just one question about a commented line of code.

source/MaterialXTest/MaterialXCore/Node.cpp
kwokcb: I've removed this. Also I put in node instance creation and proper default version checking when the default changes.

jstone-lucasfilm:
@kwokcb One additional detail is that we should follow our current rules for API compatibility, and should add deprecated wrappers for any function signatures that have been changed in this pull request.

Although the `addNodeDefFromGraph` method is not commonly used, I see a handful of references to this function in MaterialX-based projects on GitHub, and we should make sure these existing calls remain unchanged in their behavior.

Here's an example of a wrapper for deprecated C++ function signature in MaterialX, which will emit a compile-time warning when the legacy function is referenced:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/a043d9f3e19bbedb1db6d5ec21205515157762b3/source/MaterialXGenShader/Util.h#L93

jstone-lucasfilm:
@kwokcb Since we'd like to include these improvements in MaterialX 1.39, would you mind making this a pull request for the `dev_1.39` branch where active development on that version of MaterialX is ongoing?

kwokcb:
Moved to 1.39. 

@jstone-lucasfilm : The deprecation logic seems bit inconsistent based on the example you gave. 
- The C++ has no fallback to call the old method but the python wrappers does. 
- Also in the example it keeps the old interface but does not add the new one ? I put in the new one and the old one with a deprecation annotation.

jstone-lucasfilm:
@kwokcb For the deprecated example I gave above, here's the corresponding implementation, which just calls into the latest version of the same API:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/a043d9f3e19bbedb1db6d5ec21205515157762b3/source/MaterialXGenShader/Util.cpp#L591

kwokcb:
I think I've covered all logic paths for `C++`, `pybind11` and `emscripten` for deprecation.
@jstone-lucasfilm do you think it's worthwhile to document this somewhere ? Thx.

source/PyMaterialX/PyMaterialXCore/PyDocument.cpp
```diff
@@ -59,7 +71,9 @@ void bindPyDocument(py::module& mod)
         .def("removeTypeDef", &mx::Document::removeTypeDef)
         .def("addNodeDef", &mx::Document::addNodeDef,
             py::arg("name") = mx::EMPTY_STRING, py::arg("type") = mx::DEFAULT_TYPE_STRING, py::arg("node") = mx::EMPTY_STRING)
-        .def("addNodeDefFromGraph", &mx::Document::addNodeDefFromGraph)
```
jstone-lucasfilm, line 62: For the deprecated Python method, I'd instead recommend following the pattern here, which leverages the existing deprecation system in the Python language:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/python/MaterialX/main.py#L178

source/JsMaterialX/JsMaterialXCore/JsDocument.cpp
```diff
@@ -59,8 +60,17 @@ EMSCRIPTEN_BINDINGS(document)
         .function("getTypeDefs", &mx::Document::getTypeDefs)
         .function("removeTypeDef", &mx::Document::removeTypeDef)
         BIND_MEMBER_FUNC("addNodeDef", mx::Document, addNodeDef, 0, 3, stRef, stRef, stRef)
-        BIND_MEMBER_FUNC("addNodeDefFromGraph", mx::Document, addNodeDefFromGraph, 7, 7, const mx::NodeGraphPtr, 
-            stRef, stRef, stRef, bool, stRef, std::string)
+        BIND_MEMBER_FUNC("addNodeDefFromGraph", mx::Document, addNodeDefFromGraph, 4,4, const mx::NodeGraphPtr, const std::string,
```
jstone-lucasfilm, line 63: For now, I think it's fine to omit deprecated wrappers in JavaScript, since we don't yet have a base of MaterialX JavaScript clients.  Once we begin seeing significant usage of the JavaScript bindings, though, this looks like a good pattern that we can consider adopting.

jstone-lucasfilm (commented):
Thanks for getting things started on the deprecated wrappers, @kwokcb, and I've written up a few thoughts on specific approaches.

source/JsMaterialX/JsMaterialXCore/JsDocument.cpp
kwokcb: Done

source/PyMaterialX/PyMaterialXCore/PyDocument.cpp
jstone-lucasfilm: Since the original Python function for `addNodeDefFromGraph` required all seven arguments to be stated explicitly (without any support for default arguments), I believe you can just define a deprecated Python function named `addNodeDefFromGraph` with exactly seven arguments, where its implementation simply calls into the new API.

To give a concrete example of the industry usage that we're aiming to support, here's one from the QuiltiX project, and you can see that they're stating all seven arguments as we previously required:

https://github.com/Herizoran/QuiltiX/blob/d6466f954553c622aed1ef58c66911486c361e95/src/QuiltiX/quiltix.py#L814

jstone-lucasfilm:
@kwokcb Just bumping the thread above, which I believe is the last issue that needs to be resolved:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1403#discussion_r1566530272

source/PyMaterialX/PyMaterialXCore/PyDocument.cpp
kwokcb: Hi @jstone-lucasfilm ,

Yes this will work with what I have done currently I don't see any way to do it inside of `main.py`
The issue is that Python does not allow operator overloading to be done dynamically AFAIK.
e.g.
```python
class Test:
    def test(self, arg1):
        print(arg1)

instance = Test()
instance.test("Original method")    

def _test(self, arg1, arg2):
    print(arg1, arg2)

Test.test2 = _test
Test.test = _test # This just replaces the original and does not add a new version of test.

instance = Test()
instance.test("1 arg") # This will error out
instance.test("1 arg", "2 arg")
```

jstone-lucasfilm: Ah, I see what you mean now.  Ok, in that case, your workaround seems fine to me!

source/JsMaterialX/JsMaterialXCore/JsDocument.cpp
```diff
@@ -59,8 +60,8 @@ EMSCRIPTEN_BINDINGS(document)
...
```
jstone-lucasfilm, line 63: Should these arguments be const references to strings rather than const string values?

jstone-lucasfilm (commented):
This looks very close to the mark, and I had just one question about the JavaScript binding below.

source/JsMaterialX/JsMaterialXCore/JsDocument.cpp
jstone-lucasfilm: Fixed!

jstone-lucasfilm (approved):
This looks ready to merge, thanks @kwokcb!

Issue #1398, opened 2023-07-06
Graph Editor: Usability Improvements


kwokcb:
This groups a set of suggestions / proposals for improving how users interact with the Graph Editor.
The hope is to take the "best" ideas and designs and incorporate them if they will help interaction / usability.

Each item below is broken down into:
* Problem: description of issue
* Proposal: Proposal to address the issue
* Sub-issue: Link to sub-issue if applicable.

Folks can add / modify each item.
Please up / down vote an item.

@lfl-eholthouser  is the main stake-holder for this.

kwokcb:
## Problem
All UI is uniformly "grey" and aesthetically "dull"
## Proposal
 Add node color themes.
## Issue
https://github.com/AcademySoftwareFoundation/MaterialX/issues/1373

kwokcb:
## Problem
It is very hard to distinguish data flow paths as they show up as overlapping lines with uniform appearance
## Proposal
From Lee Kerley: Color graph edges to help to improve readability

kwokcb:
## Problem
It is very hard to find specific node types such as inputs, outputs, renderables, and conditions
## Proposal
Add in color coding based on node-group or category. Note that this has overlap with color theming.

kwokcb:
## Problem
Text labels/ fields in the property editor and node graph are hard to distinguish from each other (e.g. for importance, for categorization etc)
## Proposal
( Pulled from #1207 from @dbsmythe )
* I'd really like for nodes to indicate what their "nodecategory" is (e.g. add or over or fractal2d) right in the node itself, drawn in boldface at the top inside the node box, e.g. below the title with the node name and above the "out" output. Without this, it's hard to look at just the graph and see what mathematical operation it's doing.
* Similarly, in the Node Property Editor, it'd be nice if the nodecatagory value (e.g. the word "add" or "over" etc) was much more prominent: boldface, a different color, or something.

marwie:
## Problem

*Connected* Inputs and Outputs don't show current (receiving or outgoing) values making it harder to debug the flow / state of a graph at a certain point ("does this input actually receive what I think it does")

![image](https://github.com/AcademySoftwareFoundation/MaterialX/assets/5083203/b488131d-aac6-4cd5-8321-9669a8acc0b8)

## Proposal

Display readonly values next to `[type]`

marwie:
## Problem

[...]

PR #1405, opened 2023-07-11
Graph Editor : Add label coloring

kwokcb:
# Proposal

Add label coloring for nodes "of interest" and edge coloring. Group issue: #1398 

For lables this is:
- inputs and outputs : Allow to easily see where the interfaces are on graph elements.
- conditional nodes : Hilight conditional branching which affects code generation.

  image

- surfacematerials : This allows for easily seeing terminal renderable nodes.

  image

- `nodegraphs` and `backdrops` are already colored differently than other nodes.
- Helps a bit with finding nodes when there is visual "clutter"

  image

For edges. this is just using the pin type color as is used in Houdini and Maya for e.g.

image

## Modifications
- Add in a node label color dictionary to match the pin dictionary and used it for selecting node label colors.
- Minor refactor to cache `nodedef` associations as this is expensive to keep on querying, for property editor but also
to find `nodgroup` information if used for coloring like for conditionals. Otherwise cached ui-node category or other string key can be used to lookup into dict. Adding the dict allows for save and load if desired of color prefs.
- Add in a simple struct which has 2 flags to turn on/off coloring of nodes and/or edges.

For review by @lfl-eholthouser and others interested.


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -625,9 +625,30 @@ void Graph::layoutInputs()
...
+    {
+        labelColor = _nodeLabelColor[key];
+    }
+    return labelColor;
+}
+
+// Set up user colors for:
+// - Pins based on the type of the value of that pin
+// - Nodes are based UI node category or node group
+void Graph::setupUIColors()
```
kwokcb, line 642: Rename method to be a general UI color setup method and include node colors along with pin colors.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -625,9 +625,30 @@ void Graph::layoutInputs()
...
+{
+    // Set up node colors
+    _nodeLabelColor.insert(std::make_pair(mx::NodeDef::CONDITIONAL_NODE_GROUP, ImColor(255, 128, 0, 255)));
+    _nodeLabelColor.insert(std::make_pair(mx::Input::CATEGORY, ImColor(0, 120, 200, 255)));
+    _nodeLabelColor.insert(std::make_pair(mx::Output::CATEGORY, ImColor(0, 160, 120, 255)));
+    _nodeLabelColor.insert(std::make_pair(mx::SURFACE_MATERIAL_NODE_STRING, ImColor(0, 160, 120, 255)));
+    _nodeLabelColor.insert(std::make_pair(mx::Implementation::NODE_GRAPH_ATTRIBUTE, ImColor(35, 35, 35, 255)));
```
kwokcb, line 649: This is the custom color already used for nodegraphs.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1099,7 +1120,7 @@ void Graph::setUiNodeInfo(UiNodePtr node, const std::string& type, const std::st
     {
         if (node->getNode())
         {
-            mx::NodeDefPtr nodeDef = node->getNode()->getNodeDef(node->getNode()->getName());
+            mx::NodeDefPtr nodeDef = node->getNodeDef();
```
kwokcb, line 1123: See include file where the nodedef is cached on UInode initialization. 
I simplified this as the cost of querying the nodedef is non-trivial and shows up in performance profiles.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1212,7 +1233,7 @@ void Graph::buildUiBaseGraph(mx::DocumentPtr doc)
         std::string name = nodeGraph->getName();
         auto currNode = std::make_shared<UiNode>(name, _graphTotalSize);
         currNode->setNodeGraph(nodeGraph);
-        setUiNodeInfo(currNode, "", "nodegraph");
+        setUiNodeInfo(currNode, "", mx::Implementation::NODE_GRAPH_ATTRIBUTE);
```
kwokcb, line 1236: There a few hard-coded string replacements with constants to avoid string initialization overhead which is a niggling performance slowdown. I didn't do all of them :) ).

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2126,21 +2147,26 @@ std::vector<int> Graph::createNodes(bool nodegraph)
         }
         else
         {
-            // color for output pin
+            // Draw node
             std::string outputType;
             if (node->getNode() != nullptr)
             {
+                mx::NodeDefPtr nodeDef = node->getNodeDef();
```
kwokcb, line 2154: Custom node coloring.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2193,21 +2219,24 @@ std::vector<int> Graph::createNodes(bool nodegraph)
                     }
                 }
             }
+
+            // Draw input 
             else if (node->getInput() != nullptr)
             {
                 std::string longestInputLabel = node->getName();
+                ImColor labelColor = getNodeLabelColor(node->getInput()->getCategory());
```
kwokcb, line 2227: Custom input label coloring.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2262,21 +2291,24 @@ std::vector<int> Graph::createNodes(bool nodegraph)
                 }
                 drawOutputPins(node, longestInputLabel);
             }
+
+            // Draw output
             else if (node->getOutput() != nullptr)
             {
                 std::string longestInputLabel = node->getName();
+                ImColor labelColor = getNodeLabelColor(node->getOutput()->getCategory());
```
kwokcb, line 2299: Custom output label coloring.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2339,18 +2371,19 @@ std::vector<int> Graph::createNodes(bool nodegraph)
             else if (node->getNodeGraph() != nullptr)
             {
                 std::string longestInputLabel = node->getName();
+                ImColor labelColor = getNodeLabelColor(mx::Implementation::NODE_GRAPH_ATTRIBUTE);
```
kwokcb, line 2374: Existing custom nodegraph label coloring.

source/MaterialXGraphEditor/UiNode.h
kwokcb, line 1: Nodedef caching added for nodes and nodegraphs.

jstone-lucasfilm:
This looks really promising, thanks @kwokcb!  As you note above, review from @lfl-eholthouser would be a great first step, and we can take things from there.

kwokcb:
Yes, @lfl-eholthouser, when you have time please have a look to see if this fits with the direction you'd like to take things.

The background, I think just adding in a color table without any color changes could also be a first step if you agree. A possible subsequent step (not for this PR) would be to create "preferences" files with colors for pins, edges, and nodes to support the idea of custom user colors and "themes" also mentioned in the issue if this desirable.

lfl-eholthouser:
Hi @kwokcb ! This looks great! I really like how the different colors help highlight the individual nodes and I especially like the addition of highlighting the conditional nodes. I completely agree with your suggested next steps for custom colors. I think ImGui has a few native color themes that we could also consider using as a starting point.

kwokcb:
> Hi @kwokcb ! This looks great! I really like how the different colors help highlight the individual nodes and I especially like the addition of highlighting the conditional nodes. I completely agree with your suggested next steps for custom colors. I think ImGui has a few native color themes that we could also consider using as a starting point.

Thanks for taking a look. To keep it super simple, I added in a `GraphStyle` struct which just has 2 flags to turn on/off node lable and edge coloring. The latter was pretty easy to do so am including it here. Just grabs link type to look up the pin color. This makes it consistent with Maya and Houdini.

PR #1407, opened 2023-07-12
Remove modulo from randomfloat graph

BrianSharpe:
It did not achieve its intention of allowing for more consistent behavior over large ranges, giving no benefit over that of a simple scale.

The scale value of 4096 has been chosen arbitrarily.  Perhaps we could expose it as an input someday.


jstone-lucasfilm:
Thanks for this proposal, @BrianSharpe, and I'm CC'ing @code-monkey and @crydalch for their insights as well.

jstone-lucasfilm (approved):
Thanks for the simplification, @BrianSharpe!

PR #1411, opened 2023-07-14
Various Pattern nodes for AEC and MFG

ashwinbhat:
Collection of Pattern nodes for Circles, Hexagons, Cloverleafs, Grecian and Squares
Cloverleafs
cloverleafs
Grecian
grecian
Hexagon
hexagons
Squares
squares
Circles
circles


crydalch:
This is great, super awesome! Have you considered per-tile variation/randomization? That was the trickiest part of the shape nodes we were working to contribute, hehe, but is pretty important for artists. It's hard to "tack-on" because it's tied to how the tiles/shapes are generated, especially for randomizing individual shape features.

I'll link to a gist with our nodes, so they can be inspected for comparison.

jstone-lucasfilm:
@crydalch I like this suggestion, and perhaps random variation would make a great future contribution, building upon the core nodes defined here?  For example, one could imagine a future `circlesrandom` that builds upon the `circle` node in this changelist, with additional controls for random colors and placement, and this seems independent of the more structured `circles` node defined here.

zicher3d:
A modular approach to patterns is an interesting topic. Might require a bit of planning before throwing nodes at the problem. I'm all for proceduralism having worked on Substance in the past and knocked my head on other procedural node-based tools. But then we should have a plan for a "prclib" where changes can be more accepted than stdlib, and where a series of well planned foundational nodes can be used to build up a library of patterns.
There might be a feature or two missing now that would prevent full procedurals, but we can learn from existing solutions.

crydalch:
Procedural patterns is interesting, but is far beyond where my original comment about randomizing tiles was intended.

See our original PR for what I meant by per-tile randomization: https://github.com/AcademySoftwareFoundation/MaterialX/pull/1329#issue-1671811895

Apologies if my comment about randomization implied sophisticated procedural patterns/textures.

crydalch:
In the past, the number of individual node primitives in MaterialX has been raised as a big concern. So I think if we assume that we'll have _N_ shape nodes, then _N_ random-shape nodes, it's going to fly in the face of that guidance.

Even setting that aside, it makes sense that tiling and randomization should really be built-into tiling shape nodes...

jstone-lucasfilm:
@crydalch There's definitely a balance to strike between providing "ubernodes" that can accomplish everything, and providing very granular functionality that promotes reuse and recombination.  In this case, I think the `circle` node is a great low-level primitive that will find reuse in many contexts, and I can imagine a wider set of "tiling shape" nodes that would be built on top of it.  I like the idea of the current `circles` node as a very minimalistic member of that "tiling shape" set, but of course I'm open to ideas on other structures as well.

jstone-lucasfilm:
To clarify the difference between regular and random "tiling shape" nodes, what would you think of a naming convention such as `circlesregular` versus `circlesrandom`?  The `circlesregular` node would map to the current `circles` proposal, and the `circlesrandom` node would allow for a much more complex artistic interface with color ranges, offset ranges, and so forth.

zicher3d:
I still think that this PR is a set of basic nodes for common patterns, and it could fit as part of stdlib, but if we want to expand and create more complex basic nodes, with multiple parameters for tweaking and randomizing things, then it's better to separate them into a prclib where thing will grow and change frequently.
Note that the ten or so nodes I posted are a quarter of the size of the whole stdlib_ng.mtlx file. These things tends to get big quickly.
image

meshula:
I'm curious about the periodic aliasing in the Grecian and Squares samples. Is that a consequence of upload to github, or would it make sense to examine anti-aliasing? I imagine the same question applies to all the shapes, but for those two it's visible to my eye.

ashwinbhat:
> I'm curious about the periodic aliasing in the Grecian and Squares samples. Is that a consequence of upload to github, or would it make sense to examine anti-aliasing? I imagine the same question applies to all the shapes, but for those two it's visible to my eye.

Yes it appears to be so. 
I had run `MaterialXView.exe --material TestSuite\stdlib\procedural\squares.mtlx --captureFilename squares.png --screenWidth 800 --screenHeight 600 --mesh resources\Geometry\plane.obj`
In the viewer it appears to be ok.

ashwinbhat:
[grecian.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/12072674/grecian.zip)

meshula:
Thanks for uploading that, seeing the good pixels makes me happy :)

zicher3d:
Adding some hero images of the patterns in Maya/LookdevX
image
image
image
image
image

dbsmythe:
One general question for all of these nodes is what nodecategory these should be in, and thus how/where would they appear in the Specification.  Currently they are authored as "procedural2d" nodes, but I think it would be good to create an entirely new node category for these procedural shape/pattern generation nodes, and thus a new section in the Specification.  I would propose "**shape**" as the new nodecategory, with a corresponding **Shape Nodes** section in the Spec.  Thoughts?

zicher3d:
I'm ok with the changes you suggest.
I also experimented with returning a distance rather than a 0 or 1 value (actual SDF shapes), but could be a separate set of base nodes, as the output then needs further processing that for simple shapes is not necessary. 
I haven't looked at @crydalch randomizations. I think I missed them. Where are they?
I did play with randomizing Size/Center with noises on my nodes. You can get some interesting patterns.
I was instead thinking if a rotation option would be useful, and if it should be a base feature of a shape, or if we can just rotate the shape output somehow at the pattern level.

jstone-lucasfilm:
@zicher3d I've updated the node names to align with `tiledimage`, and I'm thinking that we can add `uvtiling` and `uvoffset` as a second step if this approach looks promising.

Let me know your thoughts, and of course I'm interested in ideas from others in the community as well.

jstone-lucasfilm:
@zicher3d I've now made the second proposed change as well, aligning the inputs of the tiling shape nodes with those of `tiledimage`.  In my local tests, this approach seems more artist-friendly and consistent with MaterialX conventions, but let me know how the new proposed interface looks to you.

zicher3d:
Ok with the tiling<shape> rename.

For the uvtiling/offset let me review it tomorrow. The reason of my "Size/Spacing" choice is because, if UVs are used in real-world mode (0-1 matches a scene unit, a technique used in our CAD apps and 3dsMax), those values will create a measured pattern.
If this change doesn't break it we are ok.

jstone-lucasfilm:
@zicher3d Let me know what you find in testing, and we can likely adjust the graph logic to maintain the consistency of real-world units if that's an important consideration.

zicher3d:
Checked the current status. I'm not in favor of the uvtiling change. We are losing the Spacing functionality.
I would think we could use both if we wire them properly. See the video.

https://github.com/AcademySoftwareFoundation/MaterialX/assets/7254958/625bf3a1-a8d6-46b6-a4eb-cae73e1811fd

zicher3d:
@jstone-lucasfilm I went back to my original graph to do this proof of concept. I could not figure out the graphs changes after all the optimizations and new uvtiling and offset.
Check the video to see the desired behavior, and then we can discuss how to implement it.

https://github.com/AcademySoftwareFoundation/MaterialX/assets/7254958/fef4450b-a016-4841-b1df-1e26d2814673

jstone-lucasfilm:
Following up on a discussion with @ashwinbhat and @zicher3d, I've split the original `grecian` node into two simpler line primitives, `grid` and `crosshatch`, allowing artists to recombine them in a wider variety of ways.

Here are a few renders from the latest set of example materials, giving a sense of how these line patterns look in practice:

Grid node:
Grid

Crosshatch node:
Crosshatch

Union of grid and crosshatch (just one of many possible combinations):
Union

Issue #1384, opened 2023-06-22
Proposal: Replace `thin_film_bsdf` with `tf_ior`/`tf_thickness` attributes


amsXYZ:
## Motivation
`thin_film_bsdf` is one of the `BSDF` nodes part of **PBR** spec of **MaterialX**. Differently to any other of these nodes, it has a lot of particularities that make it a bit unpredictable and hard to integrate in certain pipelines (i.e. realtime renderers).

Some of these issues are:
- As specified on spec, it's not supposed to represent a `BSDF` per-se, but a thin-layer that interferes with layers below by polarising light that traverses it. Because of this, the `thin_film_bsdf` node doesn't do anything on its own and needs to be paired with a `layer` element to have any effect on the output response.
  - This can be quite confusing for users, as any element that accepts a `BSDF` type as input (i.e. `mix`) will not output proper results when connected to a `thin_film_bsdf` element, being up to the user to know this beforehand.
- Using it can lead to graphs that, even when they're valid, present broken results (as already issued [here](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1197)) or have no perceivable change in the material's response (as seen below when layered on top of an `oren_nayar_diffuse_bsdf` element).
  - It only affects `dielectric`, `conductor` and `generalized_schlick` bsdfs, but can be layered on top of any `BSDF`. This is not obvious and can be quite confusing for any user without a strong PBR knowledge.
![thin_film](https://github.com/AcademySoftwareFoundation/MaterialX/assets/6959189/29d454f9-4c0d-4a75-93b1-4e75df5dcb3a)
- It breaks the assumption that every element encapsulates unique data/logic, since it introduces the concept of **modifiers**. Because of this, implementations cannot just compute a material's response following element edges from input to output, and are forced to recursively find all `BSDF` nodes below a `layer` with a `thin_film_bsdf` in its `top` attribute. Once these elements are found, the implementation has two "not-so-great" options:
  - Either modify the `ior` and `thickness` for all the found `BSDF` elements. This could lead to wrong results, if the output of any of these `BSDF` is used by another element before the thin-film is applied.
  - Duplicate all found `BSDF` elements and recompute them using the thin-film `ior` and `thickness` parameters. This will ensure correct results for elements that use the `BSDF` outputs before having the thin-film layer applied. However, this doubles their computational cost, which can be critical in certain scenarios (i.e. complex materials, real-time rendering, etc.).

[...]

PR #1413, opened 2023-07-17
Simplify thin-film layering into thin-film BSDF parameters

niklasharrysson:
This change list simplifies the way thin-film iridescence is specified in the PBR library, as discussed in issue #1384.


jstone-lucasfilm (commented):
@niklasharrysson This looks like a great improvement, but perhaps we should leave our data libraries (e.g. `libraries/bxdf`) unchanged for now, so that we can continue validating the upgrade process during development on 1.39?  Once the 1.39 codebase is complete, we can then upgrade our data libraries all at once, removing the extra work of applying the upgrade process to these files.

jstone-lucasfilm (approved):

PR #1416, opened 2023-07-18
Fixes for 3-way comparison in tests_to_html.py

krohmerNV:
Tried to do a comparison between MDL 1.6, 1.7, and 1.8 and found two issues:
- the computed file paths are not correct when passing different `--inputdir1`, `--inputdir2`, and `--inputdir3`
- when comparing basically mdl against mdl and mdl, the file names of the diff images collide

### DIFF
--- a/python/MaterialXTest/tests_to_html.bat
+++ b/python/MaterialXTest/tests_to_html.bat
@@ -1,1 +1,2 @@
-python tests_to_html.py -i1 ../../build
+@echo off
+python tests_to_html.py -i1 ../../build %*
--- a/python/MaterialXTest/tests_to_html.py
+++ b/python/MaterialXTest/tests_to_html.py
@@ -16,11 +16,11 @@ def createDiff(image1Path, image2Path, imageDiffPath):
     try:
         if os.path.exists(imageDiffPath):
             os.remove(imageDiffPath)
-        
+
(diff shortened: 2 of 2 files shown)


jstone-lucasfilm:
@krohmerNV I like the idea of this improvement, though in my local tests it seems to break the simpler two-way comparisons between GLSL and OSL:

TwoWayComparison

krohmerNV:
I'll investigate, thanks!

jstone-lucasfilm:
@krohmerNV Just bumping this thread, in case you have a chance to follow up on this proposal!

krohmerNV:
thanks @jstone-lucasfilm, I need to make time for this in the next week

krohmerNV:
@jstone-lucasfilm updated the default parameters of the script to behave like before.
Sorry it took so long.

jstone-lucasfilm (approved):
Thanks for these improvements, @krohmerNV!

jstone-lucasfilm:
Just as external validation of the PR, here are the latest GLSL/OSL render comparisons with these changes in place:

[MaterialXRenderTests_08_24_2024.pdf](https://github.com/user-attachments/files/16737715/MaterialXRenderTests_08_24_2024.pdf)

PR #1417, opened 2023-07-18
Add a ShaderGen option to select the target MDL version

krohmerNV:
This allows to generate MDL code for older applications while adding new features in the future.

It also to helps structuring changes and features depending on the MDL version.

Work in progress because i would like have feedback first.


krohmerNV:
@jstone-lucasfilm thanks! I'm working on a few test on our side to check if the generated code really works with old versions of our SDK. I don't plan to add major changes anymore. So, if you find time to review, I would very much appreciate it already.

jstone-lucasfilm:
The overall structure of this proposal looks good to me, @krohmerNV, and I'm CC'ing @niklasharrysson for his thoughts and expertise.

source/MaterialXGenMdl/Nodes/SourceCodeNodeMdl.cpp
```diff
@@ -36,6 +73,19 @@ void SourceCodeNodeMdl::initialize(const InterfaceElement& element, GenContext&
...
             string functionName = _functionSource.substr(0, pos);
+
+            const ShaderGenerator& shadergen = context.getShaderGenerator();
+            const MdlShaderGenerator& shadergenMdl = static_cast<const MdlShaderGenerator&>(shadergen);
+            StringVec code = replaceSourceCodeMarkers(getName(), functionName,
+                [&shadergenMdl, &context](const string& marker)
+                {
+                    if (marker == MARKER_MDL_VERSION_SUFFIX)
+                    {
+                        return shadergenMdl.getMdlVersionFilenameSuffix(context);
```
niklasharrysson, line 84: It looks like the result of getMdlVersionFilenameSuffix is constant. Maybe the call could be done outside here and the result captured to the lambda, instead of calling the function every time?

source/MaterialXGenMdl/Nodes/SourceCodeNodeMdl.cpp
```diff
@@ -36,6 +73,19 @@ void SourceCodeNodeMdl::initialize(const InterfaceElement& element, GenContext&
...
+                    }
+                    return string("");
```
niklasharrysson, line 86: Nitpick: There is a predefined EMPTY_STRING that we usually use instead of creating these on the fly.

niklasharrysson:
Sorry for the delay, I noticed this ping today.

This looks good to me as well @krohmerNV , and I only had a few minor comments above.

krohmerNV:
@niklasharrysson thanks for the review.
i missed it as well, sorry.
will try to update today. maybe it's good for 1.38.9

jstone-lucasfilm:
Thanks @krohmerNV, and let's definitely aim to include this change in 1.38.9 if you have the bandwidth to update it.

krohmerNV:
My local tests look fine. Ready from my side.

jstone-lucasfilm (approved):
This looks great to me, thanks @krohmerNV!

PR #1418, opened 2023-07-18
Add fileTextureVerticalFlip option for MDL generation

krohmerNV:
When investigating the differences in the MaterialXTest I implemented the fileTextureVerticalFlip in the MDL shader gen.
In theory this option is not needed because the MaterialX and MDL define the image coordinate both in the lower left.
But it might be useful to align with existing but non standard conform renderers. It could for instance be used in the MaterialXTest to align with the flipped GLSL and OSL results. 

WIP because I guess we want to address/discuss this in the context of the MaterialXTest fixes
and it will probably conflict with #1417


krohmerNV:
@jstone-lucasfilm, I believe the discussion was closed. We want to flip in the test renderer to align with GLSL and OSL. I added comments to explain why this flip is done and how this is to be handled in integrations.

jstone-lucasfilm:
This looks very promising, thanks @krohmerNV, and I'll plan to take a closer look soon!

jstone-lucasfilm (approved):
This looks great to me, thanks @krohmerNV!

PR #1419, opened 2023-07-18
Added option to install resource folder

### LINKED ISSUE #1167: Installation and requirement of 'resources' directory
Hello! When installing MaterialX, f.i. by [building USD with MaterialX](https://github.com/pablode/USD/releases), the 'resources' directory is installed, too.

It's pretty large (130MB+) and it's required by the [MaterialXConfig.cmake.in](https://github.com/AcademySoftwareFoundation/MaterialX/blob/e2bb3898272868115042327d00c0065baf3477d6/cmake/modules/MaterialXConfig.cmake.in#L25) file when trying to link against MaterialX.

Since this folder is not needed by the runtime, unlike the libraries folder, would it be possible to make the installation of this folder optional?

For instance, it could be installed if either `MATERIALX_BUILD_VIEWER` or `MATERIALX_BUILD_TESTS` is set to `ON`.

benjaminbeilharz:
Fixes #1167 

Added an option in the CMake file which toggles the installation of resources if the renderer is built and the user wants to install the resources.


jstone-lucasfilm (approved):
This looks good to me, thanks @benjaminbeilharz!

PR #1420, opened 2023-07-19
Improvements to prefiltered environment maps

ApoorvaJ:
As discussed with @jstone-lucasfilm, MaterialXView's pre-integrated IBL path (i.e. when FIS is turned off) is broken. This PR attempts to fix this issue.

image

## Highlights of this PR

1. In mx_environment_prefilter.glsl, we've moved from the more accurate Fresnel term to Schlick's approximation. We do this because the split-sum approximation for IBL relies on it:

image
_Source: [Moving Frostbite to PBR (PDF)](https://seblagarde.files.wordpress.com/2015/07/course_notes_moving_frostbite_to_pbr_v32.pdf)_

2. Before this PR, MaterialXView uses the default mip-mapping filter on the environment radiance map. This also needs to be changed for correct IBL. In this PR, I pre-convolve the environment radiance map when it is first requested, to generate the correct LD term. I do this for various roughness values and store the results in different mip levels. I hold on to the non-preconvolved map, which is needed when FIS is on, which results in numerical integration. The bulk of the pre-convolution shader code is in mx_pre_convolve_environment.glsl.

## Request for review
I'd like to specifically request review of the following:
1. mx_pre_convolve_environment.glsl - There might  be opportunities here to use some bits of existing code, instead of writing stuff from scratch. E.g. `get_smith_joint_ggx_part_lambda_v()` might be able to be formulated using existing code.
2. I've tested with MaterialXView + Metal on Mac, MaterialXView + OpenGL on Mac, and MaterialXView + OpenGL on Windows. I haven't tested other potential backends and MaterialXGraphEditor.
3. Image.cpp - I think the getMaxMipCount() function was wrong, and it caused a crash in my code. I believe I've fixed it, but worth a double-check.
4. MetalTextureHandler.mm - I needed mipmaps on a renderable texture. I don't know if this change has any deeper repercussions.
5. RenderPipelineGL.cpp, RenderPipelineMetal.mm - I had some trouble with your API abstraction layer, so I might be doing it wrong, or using the API directly.


jstone-lucasfilm (commented):
This looks like an excellent proposal, @ApoorvaJ, and my main recommendation would be to leverage existing GLSL helper functions where possible, so that we don't end up with multiple implementations to optimize and maintain going forward.

As one example, I see this new implementation of height-correlated GGX/Smith masking shadowing:
https://github.com/AcademySoftwareFoundation/MaterialX/pull/1420/files#diff-eb403937b2a487a7fdf5a66ef7476114dd764cebd1cf209cebc77b5f59f1f4f8R88

And if it's possible to leverage (or extend) our existing implementation, that would be better:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L99

Similarly, I see this new implementation of GGX NDF sampling:
https://github.com/AcademySoftwareFoundation/MaterialX/pull/1420/files#diff-eb403937b2a487a7fdf5a66ef7476114dd764cebd1cf209cebc77b5f59f1f4f8R54

And ideally we'd want to leverage (or extend) the existing GGX VNDF sampling that is used elsewhere in the MaterialX GLSL codebase:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L65

In the latter example, we might see improved performance and convergence as well, if the existing VNDF sampling function can be used, but let me know if I'm missing any key differences in the new use case that would require a second helper method.

ApoorvaJ:
Hey @jstone-lucasfilm,

Sorry for the delayed action on this. I've done some more work on this:

- I've removed the GGX/Smith masking shadowing function that I implemented, relying instead on existing code
- I've rebased my PR on top of mainline changes that happened in the meanwhile
- I've fixed a crash I found on Metal, that was happening due to the render target flag being set on textures which didn't need it. I've fixed it properly now.

The one thing I did not manage to do was to leverage (or extend) the [existing GGX VNDF sampling code](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L65). This is primarily due to a big difference between the two approaches, and my lack of understanding of the logic that is in the existing code base. Could I get some technical help from someone on the MaterialX on this? Or are you happy with the code as it currently stands?

jstone-lucasfilm (approved):
Thanks for this great work, @ApoorvaJ, and I believe this is now ready to merge.  My own recent changes have been focused on aligning the GGX importance sampling with our existing approach in MaterialX, using the VNDF distribution and filtered samples.

PR #1422, opened 2023-07-22
Clarification of input node channel output

dbsmythe:
As discussed on Slack, add a line clarifying that an &lt;image> node will output the first N channels from the image file if the image file has more channels than the &lt;image> node.
Also updated the link target for the &lt;surface_unlit> node to conform to naming conventions used in other nodes with underscores in their name.


jstone-lucasfilm (approved):
Looks good to me, thanks @dbsmythe!

PR #1423, opened 2023-07-24
Fixes for OSL implementation of blackbody node

niklasharrysson:
Fix a few syntactical mistakes causing compile errors in `blackbody` implementation.


jstone-lucasfilm (approved):
Thanks for these fixes, @niklasharrysson!

PR #1430, opened 2023-07-26
Handle precedence of generic and target-specific implementations

### LINKED ISSUE #1409: Issues overriding OSL shader nodes implementations.
Hi,
I wanted to override the OSL implementation of the `UsdPrimvarReader` nodes. I'm using a custom target which inherits from `genosl`, and a custom shader generator class which inherits from `OslShaderGenerator`.
In its constructor, I'm registering implementations for the nodes I want to override, like so:
```
    registerImplementation({
            "IMP_UsdPrimvarReader_integer_" + getTarget(),
            "IMP_UsdPrimvarReader_boolean_" + getTarget(),,
            "IMP_UsdPrimvarReader_string_" + getTarget(),,
            "IMP_UsdPrimvarReader_float_" + getTarget(),,
            "IMP_UsdPrimvarReader_vector2_" + getTarget(),,
            "IMP_UsdPrimvarReader_vector3_" + getTarget(),,
            "IMP_UsdPrimvarReader_vector4_" + getTarget(),
        }, [] (void) -> mx::ShaderNodeImplPtr {
            return std::make_shared<CustomPrimvarReaderNode>();
        }
    );
```

But I never enter the "creator" callback.

To debug this, I overloaded the `getImplementation` method of my shader generator, and I checked how the `nodedef.getImplementation(getTarget());` call was behaving for my custom nodes.

Turns out the issue is here: https://github.com/AcademySoftwareFoundation/MaterialX/blob/2c50d91cfa3ac61546b4118bd4e216169373382b/source/MaterialXCore/Definition.cpp#L75-L85

In this loop, `candidateTargets` is equal to `[ "MyTarget", "genosl" ]` as expected, and `interfaces` contains 2 implementations: the first one from OSL, and the second one is the one I registered for my target, *in this order* (this is important)
And the problem is that the `targetStringsMatch` will return true if either of its parameter is an empty string. Which is the case for `interface->getTarget()`... And thus the function returns the first interface, e.g. the one from OSL, instead of mine.

I'm not sure what's the real problem here: should the target be set to `genosl` for all OSL implementations (it feels like it should?) or is it that `targetStringsMatch` returns true if either of its argument is empty? In any case, it feels a bit weird that this `getImplementation` stops on the first things that matches, instead of trying to first look up for a perfect match (e.g. both targets are the exact same) and then fallback to trying to match by inheritance / whatever.

I'm going to try to remove the OSL implementations before registering my own, but I thought this behavior was a bit buggy so I opened this issue to get more info (it might also be something I'm doing wrong :p).

Edit: Hmm, I thought it would be quite easy to remove the default implementations of those nodes, but it turns out it's a bit of a nightmare, so if anyone has a suggestion to workaround this, I'm all ears/eyes :p

kwokcb:
Fixes #1409

## Scenario

It is possible to have a implementation in a derived target which overrides a non-target implementation. The
logic did not try to find implementations with non-empty targets before defaulting to returning an implementation
without a target.

## Fix

When dealing with an explicit target on definition search. First test for implementations which have a non-empty target string.
If not found then fallback to non-target implementations. This should fix returning non-target implementations before trying to find implementations with targets.

# Test
```python
import MaterialX as mx

# Create a document.
doc = mx.createDocument()

# Create a base and derived targetdef
base_targetDef = doc.addTargetDef("base")
derived_targetDef = doc.addTargetDef("derived")
derived_targetDef.setInheritString('base')

# Create an implementation for the base targetdef
base_impl = doc.addImplementation("IM_function_base")
base_impl.setTarget("base")
base_impl.setNodeDefString("ND_function")

# Create an implementation for the derived targetdef
derived_impl = doc.addImplementation("IM_functon_derived")
derived_impl.setTarget("derived")
derived_impl.setNodeDefString("ND_function")

# Create a nodedef
nodeDef = doc.addNodeDef("ND_function", "float3", "function")
```
which creates this documet:
```xml
<materialx version="1.38">
  <targetdef name="base">
  <targetdef name="derived" inherit="base">
  <implementation name="IM_function_base" target="base" nodedef="ND_function">
  <implementation name="IM_functon_derived" target="derived" nodedef="ND_function">
  <nodedef name="ND_function" node="function">
    <output name="out" type="float3">
</materialx>
```
The first test returns "derived" and "base", the second before the change returns "" and "". The derived result is incorrect.
After the change, the derived result is "derived" as expected.

```python
# Test with targets specified for base and derived
impl = nodeDef.getImplementation('derived')
print('Got derived target:', impl.getTarget())
impl = nodeDef.getImplementation('base')
print('Got base target: ', impl.getTarget())

# Test with base without a target
base_impl.setTarget('')
impl = nodeDef.getImplementation('derived')
print('Want derived, but got target: "%s"' % impl.getTarget())
impl = nodeDef.getImplementation()
print('Got no target: "%s"' % impl.getTarget())
```

@dcourtois. Can you see if this will fix your issue? Thanks.

@niklasharrysson, @jstone-lucasfilm : This is the least intrusive change I could think of which does not change
the logic of `targetStringsMatch` as that method is used in other places.


dcourtois:
@kwokcb Hi, sorry for the delay in my answer! I finally got to test this, and yes, it fixes my issue and I can finally get rid of my uggly workaround when I use this pull request :)

kwokcb:
Hi @dcourtois,
No problem -- everyone is busy :).  
Good to hear this fixes things. I'll await review on this.

jstone-lucasfilm (approved):

PR #1431, opened 2023-07-27
Fix graph editor defect for name setting 

ZelboK:
The graph editor has a problem where if you highlight the name in the imgui textbox and press the delete key, it will cause a cycle exception. This isn't caught anywhere, thus leading to crashes. 

As such, just a simple method that checks if the name is empty(and if it is, fixes it) is enough to avoid this issue.


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4200,3 +4200,13 @@ void Graph::writeText(std::string fileName, mx::FilePath filePath)
     writeOptions.elementPredicate = getElementPredicate();
     mx::writeToXmlFile(_graphDoc, filePath, &writeOptions);
 }
+
+template <typename Element>
+void Graph::setName(const Element& node, const std::string &name) {
```
jstone-lucasfilm, line 4205: I'd recommend making this first argument an `ElementPtr`, making it more consistent with the broader codebase, and removing the need for a templated function.  Usually we pass `ElementPtr` objects by value rather than by reference in MaterialX, since these objects are relatively lightweight, and we've found compilers to be very capable in minimizing the overhead of passing shared pointers between functions.

source/MaterialXGraphEditor/Graph.h
```diff
@@ -34,7 +34,7 @@ struct Link
 
 class Graph
 {
-  public:
+public:
```
jstone-lucasfilm, line 37: I see a few accidental edits to whitespace here and elsewhere, and it would be good to revert these before we merge the change.

jstone-lucasfilm (commented):
Thanks for this great proposal, @ZelboK, and I've written up a couple of suggestions for changes below.

source/MaterialXGraphEditor/Graph.cpp
ZelboK: Hey so I actually have this as a templated function because it gets used by two different scenarios. 

There exists a different `setName` method inside of `UiNode` also that gets used. Since those aren't `mx::ElementPtr`s, i decided to use templates here and have all calls just go through this method. 

It would seem that both are relevant to the bug at hand here. I can modify `UiNode`s setName as well but I decided against it to avoid having to include `Element.h` (where `createValidChildName` comes from). 

What do you think?

kwokcb: Hi @ZelboK,
I think the UI node setName does not need to be covered here as it should come from the Element itself, so sanitizing that should be enough. 

The other note is that I though in the issue that the easiest thing to do would be to change the actual Element::setName() method ?

source/MaterialXGraphEditor/Graph.h
kwokcb: I think clang format file can be run this, can't it?

source/MaterialXGraphEditor/Graph.cpp
ZelboK: I was under the impression that rather than changing the core API, sanitizing the input was the way to go

jstone-lucasfilm: @ZelboK @kwokcb I'm open to either approach, though I do think that a fix to `Element::setName` will eventually be needed, so that other applications (both open and proprietary) are protected against similar issues cropping up.

If you'd like to tackle the fix directly in `setName`, that would be fine with me, or you can proceed along the current lines with a specific fix for the graph editor.

jstone-lucasfilm: @ZelboK Taking the broader view, I think that the fix to `Element::setName` would be the simplest, most robust change here, so I'd recommend taking that approach if that works for you.

I believe the gist of the change in `Element::setName` would be to check for an empty input name, and to call `createValidChildName` from the parent of the given element, in a similar manner to this logic:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/c34a1066bb194d36174454b3b763a6d68bc8480d/source/MaterialXCore/Element.cpp#L229

If no parent element is found in `Element::setName`, I believe it's fine to allow the empty name string to be used without modification, as it's common for the root element of a document to be unnamed.

jstone-lucasfilm:
I wanted to follow up on this proposal, @ZelboK, to see if you have the bandwidth to bring it over the finish line.  If so, then there's some discussion above on how to move forward with a forward-looking solution that we can merge.  Let us know!

jstone-lucasfilm:
I wanted to follow up on this proposal, @ZelboK, to see if you have the bandwidth to bring it over the finish line. If so, then there's some discussion above on how to move forward with a broader solution that we can merge.  Let us know!

jstone-lucasfilm:
Thanks for this original proposal, @ZelboK, and I've incorporated these ideas into a broader fix for multiple issues that could trigger crashes when renaming nodes in the graph editor:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1501

PR #1437, opened 2023-08-03
Omit Python data installation in non-Python builds

pmolodo:
Since the addition of [Include standard data libraries in Python package #1237](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1237), we now create a python folder, with a copy of the stdlibs, even if we're not building python.

This is at best redundant, and in my particular case, caused a problem in a downstream build script.

Also - I noticed that the contents of `<INSTALL>/python/MaterialX/libraries` mirrors the contents of `<INSTALL>/libraries`, EXCEPT that `<INSTALL>/python/MaterialX/libraries/mdl` is missing. Is this intentional? Should we be copying the `mdl` folder as well?


jstone-lucasfilm:
@pmolodo This looks like a good improvement to me, though I should note that there's a larger refactoring of Python builds that has been proposed in https://github.com/AcademySoftwareFoundation/MaterialX/pull/1317, with the goal of supporting automated generation of Python packages for PyPI distribution:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1317

We should make sure that these two changes are compatible, but I'd support the idea of merging them both if there are no conflicts.

jstone-lucasfilm:
@pmolodo This looks like a good improvement to me, though I should note that there's a larger refactoring of Python builds that has been proposed in https://github.com/AcademySoftwareFoundation/MaterialX/pull/1317, with the goal of supporting automated generation of Python packages for PyPI distribution.

We should make sure that these two changes are compatible, but I'd support the idea of merging them both if there are no conflicts.

jstone-lucasfilm (approved):
This looks good to me, thanks @pmolodo!

Issue #1439, opened 2023-08-04
Python bindings not working on Windows with Python 3.8+


JGamache-autodesk:
There was an important change in how paths are searched on the Windows platform, and it requires adding the following code (or something similar):
```
# Python 3.8+ on Windows: DLL search paths for dependent
# shared libraries
# Refs.:
# - https://github.com/python/cpython/issues/80266
# - https://docs.python.org/3.8/library/os.html#os.add_dll_directory
import os
import sys
if sys.platform == "win32" and sys.version_info >= (3, 8):
    # On a standard installation, this file is in %INSTALLDIR%\python\MaterialX
    # We need to add %INSTALLDIR%\bin to the DLL path.
    mxdir = os.path.dirname(__file__)
    pydir = os.path.split(mxdir)[0]
    installdir = os.path.split(pydir)[0]
    bindir = os.path.join(installdir, "bin")
    if os.path.exists(bindir):
        os.add_dll_directory(bindir)
```
to the top of the [__init__.py](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/python/MaterialX/__init__.py) file.

PR #1440, opened 2023-08-04
Adjust Windows DLL search path for Python 3.8+

JGamache-autodesk:
Fixes https://github.com/AcademySoftwareFoundation/MaterialX/issues/1439

Required by https://github.com/python/cpython/issues/80266


python/MaterialX/__init__.py
```diff
@@ -1,3 +1,20 @@
...
+import sys
+if sys.platform == "win32" and sys.version_info >= (3, 8):
+    # On a standard installation, this file is in %INSTALLDIR%\python\MaterialX
+    # We need to add %INSTALLDIR%\bin to the DLL path.
+    mxdir = os.path.dirname(__file__)
+    pydir = os.path.split(mxdir)[0]
+    installdir = os.path.split(pydir)[0]
+    bindir = os.path.join(installdir, "bin")
+    if os.path.exists(bindir):
+        os.add_dll_directory(bindir)
```
JeanChristopheMorinPerso, line 16: Once we start shipping MaterialX on PyPI, this could be considered "dangerous" and could cause troubles for MaterialX users.

One thing that could be done is to fence like this:

```python
if sys.platform == "win32" and sys.version_info >= (3, 8):
    try:
        importlib.metadata.version('MaterialX')
    except importlib.metadata.PackageNotFoundError:
        mxdir = os.path.dirname(__file__)
        pydir = os.path.split(mxdir)[0]
        installdir = os.path.split(pydir)[0]
        bindir = os.path.join(installdir, "bin")
        if os.path.exists(bindir):
            os.add_dll_directory(bindir)
```
This would ensure that we don't add "random" stuff to the DLL search directories when installing MaterialX with pip.

python/MaterialX/__init__.py
```diff
@@ -1,3 +1,24 @@
+# Python 3.8+ on Windows: DLL search paths for dependent
+# shared libraries
+# Refs.:
+# - https://github.com/python/cpython/issues/80266
+# - https://docs.python.org/3.8/library/os.html#os.add_dll_directory
+import os
+import sys
+if sys.platform == "win32" and sys.version_info >= (3, 8):
```
JeanChristopheMorinPerso, line 8: Should we handle Python less than 3.8 by appending to `PATH` or `sys.path`?

JGamache-autodesk: Probably not. It was never done before and this code only became necessary in 3.8 when accessing PATH for library lookups was disabled because it was quite a large security risk.

python/MaterialX/__init__.py
```diff
@@ -1,3 +1,24 @@
...
+    import importlib.metadata
+    try:
+        importlib.metadata.version('MaterialX')
+    except importlib.metadata.PackageNotFoundError:
+        # On a non-pip installation, this file is in %INSTALLDIR%\python\MaterialX
+        # We need to add %INSTALLDIR%\bin to the DLL path.
+        mxdir = os.path.dirname(__file__)
+        pydir = os.path.split(mxdir)[0]
+        installdir = os.path.split(pydir)[0]
```
jstone-lucasfilm, line 17: Assuming `os.path.split` performs no split operation when the input path ends in a trailing slash, does this second invocation of `os.path.split` actually have an effect on the result?

jstone-lucasfilm (commented):
Apologies for the long delay on this, @JGamache-autodesk, and I wanted to follow up to see whether this pull request should move forward.

From our latest vantage point in December 2024, is this still the recommended workaround for Python versions >= 3.8 on Windows?

python/MaterialX/__init__.py
JGamache-autodesk: On a non-pip install, the file is `"%INSTALLDIR%\python\MaterialX\_init_.py"`.
os.path.dirname() never ends with a trailing slash, so it will return `"%INSTALLDIR%\python\MaterialX"`.
First os.path.split() will return `("%INSTALLDIR%\python", "MaterialX")`.
Second os.path.split() will return `("%INSTALLDIR%", "python")`.
From there we can find the bin dir and add it to the DLL load search path.

jstone-lucasfilm (approved):
This looks good to me, thanks @JGamache-autodesk!

Issue #1445, opened 2023-08-15
Graph Editor: toplevel input `emission_color` is connected but not rendering correctly?


marwie:
Hello,

I've started to learn about MaterialX and am currently reading through the documentation and spec.     

I'm currrently trying to understand if/why my toplevel `<input>` color3 element connected to a `standard_surface/emission_color` input renders as solid white in the MaterialX Graph Editor.

```mtlx
<?xml version="1.0"?>
<materialx version="1.38">
  <input name="myemission" type="color3" value="1, 0, 0" />
  <standard_surface name="standard_surface_surfaceshader" type="surfaceshader">
    <input name="emission_color" type="color3" interfacename="myemission" />
    <input name="emission" type="float" value="1.0" />
  </standard_surface>
  <surfacematerial name="surfacematerial" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="standard_surface_surfaceshader" />
  </surfacematerial>
</materialx>
```

![image](https://github.com/AcademySoftwareFoundation/MaterialX/assets/5083203/a3667db3-5c47-48ac-b519-0104ecbf9af3)

Using a constant works as expected

```mtlx
<?xml version="1.0"?>
<materialx version="1.38">
  <constant name="my_emission" type="color3">
    <input name="value" type="color3" value="1, 0, 0" />
  </constant>
  <standard_surface name="standard_surface_surfaceshader" type="surfaceshader">
    <input name="emission_color" type="color3" nodename="my_emission" />
    <input name="emission" type="float" value="1.0" />
  </standard_surface>
  <surfacematerial name="surfacematerial" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="standard_surface_surfaceshader" />
  </surfacematerial>
</materialx>
```

![image](https://github.com/AcademySoftwareFoundation/MaterialX/assets/5083203/61b8b5df-b81c-4b9a-84c3-0833c74144ee)

kwokcb:
Hi @marwie ,

This is a known issue for code generation (top level input driver), but good to know this is a simple reoccurring workflow so can check on prioritization. Inputs evaluate properly when inside a node graph as a workaround.

There is no “group selected nodes and make into a graph” yet, so need to start with a graph. 

Look forward to any additional questions / feedback on this.

Thanks.

kwokcb:
[...]

Issue #1385, opened 2023-06-22
Paths to the interface inputs are not returned from code generation


kwokcb:
Currently code generation will return a path to a node input even if it's connected to an interface input.
This is being handled after the fact by patch code for instance in the Material Viewer.

It is thus possible to get / set the incorrect value and may miss colorspace or unit information.

Suggestion is to move patch this in code generation.

Adding @niklasharrysson and @jstone-lucasfilm for thoughts.

Issue #1383, opened 2023-06-22
MaterialXViewer: input values aren't initialized properly


amsXYZ:
Hi there!

I've noticed that, whenever I create a `.mtlx` file that uses `input` elements to expose material properties, these inputs don't get initialised properly with their 'value' attribute in **MaterialXViewer**. They do instead get initialised with the **default** value of the attribute they're connected to.

This is creating some issues in our image-testing setup, since we do read these 'input' values from the '.mtlx' file, and it's giving us completely different rendering results to what **MaterialXViewer** renders out-of-the-box.

Here's an example of the issue:

```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">
  <surface name="surface" type="surfaceshader" xpos="14.688406" ypos="-0.629310">
    <input name="bsdf" type="BSDF" nodename="dielectric_bsdf" />
  </surface>
  <surfacematerial name="surfacematerial" type="material" xpos="16.043478" ypos="-0.698276">
    <input name="surfaceshader" type="surfaceshader" nodename="surface" />
  </surfacematerial>
  <input name="input_color3" type="color3" value="1, 0, 0" xpos="10.442029" ypos="-2.318965" />
  <input name="input_vector2" type="vector2" value="0.05, 0.05" xpos="10.442029" ypos="-0.103448" />
  <dielectric_bsdf name="dielectric_bsdf" type="BSDF" xpos="12.659420" ypos="-1.905172">
    <input name="tint" type="color3" interfacename="input_color3" />
    <input name="ior" type="float" interfacename="input_float" />
    <input name="roughness" type="vector2" interfacename="input_vector2" />
    <input name="scatter_mode" type="string" value="T" />
  </dielectric_bsdf>
  <input name="input_float" type="float" value="1.5" xpos="10.442029" ypos="-1.215517" />
</materialx>
```

We expect the above material to render a red-transmissive `dielectric_bsdf` response (see how the `input_color3` is set to a value of `1, 0, 0`), but instead get this result when rendering it in **MaterialXViewer**:

<img width="1392" alt="image" src="https://github.com/AcademySoftwareFoundation/MaterialX/assets/6959189/b41cfe05-1cf4-41c4-8533-d59f5d133de3">

If we manually set the color to red, we get the expected result rendered:

[...]

PR #1453, opened 2023-08-16
Add support for top level inputs

kwokcb:
## Issue

Fix so that top level inputs are parsed properly and values and path reflection data is correct (for uniform updates).
Fixes: 
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/1385,
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/1445
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/1383

### Defaults

Fix top level input discovery to include `GraphElement`s and not just `NodeGraph`s (a `Document` where top level inputs reside is a `GraphElement`).

Once discovered the proper value and path can be pulled from the interface. The path is important since global uniform update relies on path matching.

### Tests
Added in a simple top level test: `stdlib/nodegraph_inputs/top_level_input.mtlx`

Result in MaterialXView:
image
image

Result in MaterialX Graph Editor:
image
image

### Test files From Issues 
- 1445
image

- 1383
image


source/MaterialXCore/Interface.cpp
```diff
@@ -305,7 +305,7 @@ InputPtr Input::getInterfaceInput() const
 {
     if (hasInterfaceName())
     {
-        ConstNodeGraphPtr graph = getAncestorOfType<NodeGraph>();
+        ConstGraphElementPtr graph = getAncestorOfType<GraphElement>();
```
kwokcb, line 308: This is just an incorrect check as we want GraphElements and not just NodeGraphs.

source/MaterialXGenShader/ShaderNode.cpp
```diff
@@ -381,7 +381,17 @@ void ShaderNode::initialize(const Node& node, const NodeDef& nodeDef, GenContext
         ShaderInput* input = getInput(nodeValue->getName());
         if (input)
         {
-            input->setPath(nodeValue->getNamePath());
+            string path = nodeValue->getNamePath();
```
kwokcb, line 384: Fix shader reflection so that the interface input path will be returned instead of the downstream input. This removes any need for editors to try and find the upstream interface, but also makes it so that direct uniform will find the correct uniform in the uniform block.

kwokcb: I can't leave a comment but the value evaluation at line 349 get's fixe by fix to `getInterfaceInput()` so that the correct value is returned.

jstone-lucasfilm (approved):
This looks great to me, thanks @kwokcb!

PR #1465, opened 2023-08-18
Remove MslRenderer camera initialization

nicolassavva-autodesk:
Allow `ShaderRenderer` camera to be modified by the client in subclass renderers when using Metal.

The PR removes the Metal Renderer specific camera initialization in `MslRenderer` (and now handled by `ShaderRenderer`).

This addresses the same issue in a similar way to what was previously done for the `GlslRenderer`:

**"Improvements to ShaderRenderer"**(https://github.com/AcademySoftwareFoundation/MaterialX/commit/60b105361c65d83ce42bd5d7524d0879b6562bc8)

### DIFF
--- a/source/MaterialXRender/ShaderRenderer.cpp
+++ b/source/MaterialXRender/ShaderRenderer.cpp
@@ -25,11 +25,11 @@ const float DEFAULT_FAR_PLANE = 100.0f;
 // ShaderRenderer methods
 //
 
-ShaderRenderer::ShaderRenderer(unsigned int width, unsigned int height, Image::BaseType baseType, ConventionAPI conventionAPI) :
+ShaderRenderer::ShaderRenderer(unsigned int width, unsigned int height, Image::BaseType baseType, MatrixConvention matrixConvention) :
     _width(width),
     _height(height),
     _baseType(baseType),
-    _conventionAPI(conventionAPI)
+    _matrixConvention(matrixConvention)
 {
     // Initialize a default camera.
     float fH = std::tan(DEFAULT_FIELD_OF_VIEW / 360.0f * PI) * DEFAULT_NEAR_PLANE;
@@ -37,11 +37,11 @@ ShaderRenderer::ShaderRenderer(unsigned int width, unsigned int height, Image::B
     _camera = Camera::create();
     _camera->setViewMatrix(Camera::createViewMatrix(DEFAULT_EYE_POSITION, DEFAULT_TARGET_POSITION, DEFAULT_UP_VECTOR));
 
-    if (_conventionAPI == ShaderRenderer::ConventionAPI::METAL)
+    if (_matrixConvention == ShaderRenderer::MatrixConvention::Metal)
     {
         _camera->setProjectionMatrix(Camera::createPerspectiveMatrixZP(-fW, fW, -fH, fH, DEFAULT_NEAR_PLANE, DEFAULT_FAR_PLANE));
     }
-    else // ConventionAPI::OPENGL (default)
+    else // MatrixConvention::OpenGL (default)
     {
         _camera->setProjectionMatrix(Camera::createPerspectiveMatrix(-fW, fW, -fH, fH, DEFAULT_NEAR_PLANE, DEFAULT_FAR_PLANE));
     }
--- a/source/MaterialXRender/ShaderRenderer.cpp
+++ b/source/MaterialXRender/ShaderRenderer.cpp
@@ -35,7 +35,12 @@ ShaderRenderer::ShaderRenderer(unsigned int width, unsigned int height, Image::B
     float fW = fH * 1.0f;
     _camera = Camera::create();
     _camera->setViewMatrix(Camera::createViewMatrix(DEFAULT_EYE_POSITION, DEFAULT_TARGET_POSITION, DEFAULT_UP_VECTOR));
-    _camera->setProjectionMatrix(Camera::createPerspectiveMatrix(-fW, fW, -fH, fH, DEFAULT_NEAR_PLANE, DEFAULT_FAR_PLANE));
+
+    #if defined (__APPLE__)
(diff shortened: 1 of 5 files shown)


nicolassavva-autodesk:
@ashwinbhat @Morteeza @jstone-lucasfilm: Please have a look when you get a chance

jstone-lucasfilm:
This change looks good to me, thanks @nicolassavva-autodesk, and I'd be interested in thoughts from @Morteeza and @ashwinbhat.

Morteeza:
Without setting view and world information, tests generate wrong results. I believe we need to find another way to set these:

Copper_msl
Copper_glsl

Morteeza:
Also I believe [this line](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1465/commits/de4b3f271685f65f592e3cf5beb00c4ef8bd61e2#diff-d96611f087985c4558eab0d22fce8947f26bcc14aef1a8b432ab5a802cec0001R41) needs to be removed too

source/MaterialXRender/ShaderRenderer.cpp
```diff
@@ -35,7 +35,12 @@ ShaderRenderer::ShaderRenderer(unsigned int width, unsigned int height, Image::B
     float fW = fH * 1.0f;
     _camera = Camera::create();
     _camera->setViewMatrix(Camera::createViewMatrix(DEFAULT_EYE_POSITION, DEFAULT_TARGET_POSITION, DEFAULT_UP_VECTOR));
-    _camera->setProjectionMatrix(Camera::createPerspectiveMatrix(-fW, fW, -fH, fH, DEFAULT_NEAR_PLANE, DEFAULT_FAR_PLANE));
+
+    #if defined (__APPLE__)
```
nicolassavva-autodesk, line 39: We might also want to make this Metal specific so that it can still work with OpenGL based MacOS
@jstone-lucasfilm: what would you recommend for the cleanest way for supporting both here?
Perhaps we can move the MaterialXView specific customization to a reusable place:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXView/CMakeLists.txt#L8-L27

ashwinbhat (commented):
I would recommend removing the default camera initialization ShaderRenderer so that the client can create and supply a camera that is suitable based on the underlying API implementation. 
We already have setCamera API to allow this workflow.

source/MaterialXRender/ShaderRenderer.cpp
jstone-lucasfilm: I'm open to that option, @ashwinbhat, though I'm imagining that there may still be value in providing a default camera for simple use cases.  As long as we continue to provide the option of overriding the default camera, it seems reasonable to make this default as suitable as possible for each platform.

jstone-lucasfilm: @nicolassavva-autodesk I like the suggestion of moving the user's OpenGL/Metal selection into a pre-processor definition that is set in CMakeLists.txt and accessible in MaterialXRender.

This could then be used to select between the two perspective matrix options without needing to check platform flags such as `__APPLE__` in cross-platform classes like ShaderRenderer.

What are your thoughts, @Morteeza and @ashwinbhat?

jstone-lucasfilm: @nicolassavva-autodesk What if we were to pass a perspective matrix convention boolean into the constructor for `ShaderRenderer`, allowing the caller to state which convention should be used for the default perspective matrix?  This would allow the platform-independent ShaderRenderer class to remain agnostic of this platform detail, without requiring us to remove the default camera from ShaderRender entirely.

nicolassavva-autodesk: I introduced an enum to select between OpenGL and Metal (in lieu of a boolean)
This should suffice for now and we should eventually consider propagating the changes into a future refactor of the `Camera` class and the MaterialXViewer

Morteeza:
Thanks @nicolassavva-autodesk  for changes and the rendered test.

> I would recommend removing the default camera initialization ShaderRenderer so that the client can create and supply a camera that is suitable based on the underlying API implementation. We already have setCamera API to allow this workflow. @jstone-lucasfilm @Morteeza any thoughts on this suggestion?

Change look good as is. But as Ashwin suggested maybe it will be more explicit to remove camera GL/Metal camera initialization, and move it to test code implementations; something like this: 
[patch3.diff.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/12421424/patch3.diff.zip)

nicolassavva-autodesk:
Yes, agreed though perhaps that can wait for a future code refactor

jstone-lucasfilm (commented):
This looks like a good approach to me, thanks @nicolassavva-autodesk.

With respect to naming conventions, what would you think about the following enumerated type for the new input argument?

```
enum MatrixConvention
{
    MatrixOpenGL = 0,
    MatrixMetal = 1
};
```

This would help to clarify _which_ convention we're specifying with this argument, and the new enumeration could potentially be moved into a lower-level library (e.g. a math library) if needed further in the future.

nicolassavva-autodesk:
Yes, renaming the enums to something more explicit sounds good to me

nicolassavva-autodesk:
I take it we are still keeping it as an `enum class` and perhaps do not need to repeat "Matrix" in the name so we get:

```
enum class MatrixConvention
{
    OpenGL = 0,
    Metal  = 1
};

`MatrixConvention::Metal`
```

jstone-lucasfilm:
That's completely fine, and although we don't always use enum classes in MaterialX, I think it's a good direction for the future.

nicolassavva-autodesk:
I just renamed the enum as discussed above and rebased the PR branch

jstone-lucasfilm (approved):
This looks good to me, thanks @nicolassavva-autodesk!

Issue #1470, opened 2023-08-20
Graph Editor: selected texture is assigned to wrong input node


marwie:
Hello, when selecting a texture for tghe normal map input it wrongly assigns the selection to the `basecolor` input instead

https://github.com/AcademySoftwareFoundation/MaterialX/assets/5083203/0e74ce4c-e57d-4bf4-a2c1-52cad1593c75

Example material:

```xml
<?xml version="1.0"?>
<materialx name="" version="1.38" >
	<nodegraph name="needle_shadergraph" >
		<!-- Color -->
		<input name="_Color" type="color4" value="1.0, 1.0, 1.0, 0.6745098" />
		<dot name="dot__Color" type="color4" >
			<input name="in" type="color4" interfacename="_Color" />
		</dot>
		<!-- MainTex -->
		<input name="_MainTex" type="filename" value="" />
		<dot name="dot__MainTex" type="filename" >
			<input name="in" type="filename" interfacename="_MainTex" />
		</dot>
		<texcoord name="uv" type="vector3" >
			<input name="index" type="integer" value="0" />
		</texcoord>
		<!-- Tiling -->
		<input name="_Tiling" type="vector4" value="1.0, 1.0, 0.0, 0.0" />
		<dot name="dot__Tiling" type="vector4" >
			<input name="in" type="vector4" interfacename="_Tiling" />
		</dot>
		<multiply name="mult" type="vector4" >
			<input name="in1" type="vector4" nodename="vector3_to_vector4" />
			<input name="in2" type="vector4" nodename="dot__Tiling" />
		</multiply>
		<image name="tex" type="color4" >
			<input name="file" type="filename" nodename="dot__MainTex" />
			<input name="texcoord" type="vector2" nodename="swizzle_0_mult_1" />
			<input name="uaddressmode" type="string" value="mirror" />
			<input name="vaddressmode" type="string" value="mirror" />
			<input name="filtertype" type="string" value="closest" />
		</image>
		<multiply name="mult_1" type="color4" >
			<input name="in1" type="color4" nodename="dot__Color" />
			<input name="in2" type="color4" nodename="tex" />
		</multiply>
		<!-- Normal -->
		<input name="_Normal" type="filename" value="" />
		<dot name="dot__Normal" type="filename" >
			<input name="in" type="filename" interfacename="_Normal" />
		</dot>
		<texcoord name="uv_2" type="vector3" >
			<input name="index" type="integer" value="0" />
[...]

PR #1475, opened 2023-08-21
Fix selected texture being assigned to wrong slot

marwie:
Fix for selecting file being assigned to wrong input. See https://github.com/AcademySoftwareFoundation/MaterialX/issues/1470

This saves the input ptr the file dialogue was opened for and assigns the selected file when ready to the same input.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -1018,6 +1018,7 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
             ImGui::PushItemWidth(-100);
             if (ImGui::Button("Browse"))
             {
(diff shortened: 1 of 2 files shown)


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1018,6 +1018,7 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
             ImGui::PushItemWidth(-100);
             if (ImGui::Button("Browse"))
             {
+                _fileDialogInput = input;
```
kwokcb, line 1021: I think this should be cleared after the file dialog is used in case there are any side-effects with leaving an old input reference around.

Also could we name this something like `_fileDialogImageInput` to make it a bit clearer it's associated with the image dialog. 

Thanks.

PR #1479, opened 2023-08-21
Graph Editor: add log when opening a new file

marwie:
Add small log for when a new material is opened making it easier to see where validation errors in the console belong to (when opening multiple files during a session)

See https://github.com/AcademySoftwareFoundation/MaterialX/issues/1398#issuecomment-1683784647

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -175,6 +175,7 @@ mx::DocumentPtr Graph::loadDocument(mx::FilePath filename)
     {
         if (!filename.isEmpty())
         {
+            std::cout << ">>> Reading file " << filename.asString() << std::endl;
             mx::readFromXmlFile(doc, filename, _searchPath, &readOptions);
             doc->importLibrary(_stdLib);
             std::string message;


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -175,6 +175,7 @@ mx::DocumentPtr Graph::loadDocument(mx::FilePath filename)
     {
         if (!filename.isEmpty())
         {
+            std::cout << ">>> Reading file " << filename.asString() << std::endl;
```
kwokcb, line 178: Thanks for adding this. I agree with your issue that logging should be in the UI, but better than nothing I guess.

marwie: Ideally validation could be visualized in the UI yes but at least with a little log like this we know for sure which file did cause the errors :)

jstone-lucasfilm (commented):
@marwie I like the idea behind this proposal, where we have a way of tracking which files are associated with which errors.  My main concern is that we don't currently use the standard output stream for non-error conditions, which gives text in the console window the meaning of important error conditions that the user should look at closely.  By mixing both error and non-error messages in a single stream, my sense is that we would reduce the value of the messages that are currently written out.  To address this, I believe we'll need to improve our logging system, and that's likely a larger change than you're intending to address here.

marwie:
Hello @jstone-lucasfilm thanks for your input. 

Would perhaps a mid-term solution for this be to include e.g. the filename and perhaps a timestamp in error messages (at the end or the beginning of a message)? That way we don't introduce new types of log messages but still maybe improve the current state by providing a bit more context.  

A potential downside to this could be that it negatively affects readability of the actual message and it would require changes in all places that perform logging (I'm not sure how many there are right now)

jstone-lucasfilm:
Let's hold off on this PR for now, as I believe we should build this feature using a logging API, and we'll need some discussion around how this API should be added to the MaterialX libraries.

PR #1481, opened 2023-08-21
Fixes MaterialXTests' Metal Shading Language rendering tests execution.

Morteeza:
Previously it was failing to find options.mtlx, and not rendering results.


Morteeza:
Unfortunately it seems a few of the tests are broken in MSL. I will try to look into it.

Screenshot 2023-08-21 at 5 02 39 PM

jstone-lucasfilm (approved):
Good catch, @Morteeza, and thanks for the fix!

PR #1482, opened 2023-08-21
Add drag & drop loading to Web Viewer

hybridherbst:
This PR implements drag-drop file loading for
- .mtlx files
- .mtlx files + textures or folders with textures
- folders that contain at least one single .mtlx and textures
- .zip files containing the above

Currently, only the first found .mtlx file from any of the above will be loaded; future improvements could include
- showing a list of dropped .mtlx files to choose from
- hot reload when a .mtlx file on disk or a referenced texture changes again
  - this would turn the web viewer into a pretty nice MaterialX hand editor without having any kind of MaterialX software installed.

### DIFF
--- a/javascript/MaterialXView/source/dropHandling.js
+++ b/javascript/MaterialXView/source/dropHandling.js
@@ -1,0 +1,197 @@
+import * as THREE from 'three';
+
+const debugFileHandling = true;
+let loadingCallback;
+
+export function setLoadingCallback(cb) {
+    loadingCallback = cb;
+}
+
+export function dropHandler(ev) {
+    if (debugFileHandling) console.log('File(s) dropped', ev.dataTransfer.items, ev.dataTransfer.files);
+
+    // Prevent default behavior (Prevent file from being opened)
+    ev.preventDefault();
+
+    if (ev.dataTransfer.items)
+    {
+        const allEntries = [];
+
+        let haveGetAsEntry = false;
+        if (ev.dataTransfer.items.length > 0)
+        haveGetAsEntry = ("getAsEntry" in ev.dataTransfer.items[0]) || ("webkitGetAsEntry" in ev.dataTransfer.items[0]);
+
+        if (haveGetAsEntry) {
+        for (var i = 0; i < ev.dataTransfer.items.length; i++)
+        {
+            let item = ev.dataTransfer.items[i];
+            let entry = ("getAsEntry" in item) ? item.getAsEntry() : item.webkitGetAsEntry();
+            allEntries.push(entry);
+        }
+        handleFilesystemEntries(allEntries);
+        return;
+        }
+
+        for (var i = 0; i < ev.dataTransfer.items.length; i++)
+        {
+        let item = ev.dataTransfer.items[i];
+        
+        // API when there's no "getAsEntry" support
+        console.log(item.kind, item, entry);
+        if (item.kind === 'file')
+        {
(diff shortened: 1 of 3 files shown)


kwokcb:
Hi @hybridherbst ,
This has been requested for some time :) Thanks for putting this together.
I'm curious how are dependent images handled? i.e. what should the URI be to specify the location of the image?

hybridherbst:
Hi @kwokcb, there's no changes needed to make paths work, files and references that work locally should also work here.
The important part is `THREE.Cache.enabled = true;` which allows pre-seeding the cache with the dropped binary files (textures) and text files (mtlx files). So basically everything is loaded into the cache with relative paths to the dropped thing (e.g. a folder or, if multiple files were selected, `/`) and things just work :)

Complexity of drag-drop as implemented here loading just comes from the fact that there are multiple different web APIs and they work slightly different. There may be more efficient ways of doing this (e.g. implementing a custom manager that is able to resolve paths on-the-fly instead of loading all at once) but I think this is a good start.

Some examples:
- download a zip from https://matlib.gpuopen.com/main/materials/all?material=2ee33fd8-0c97-4e8a-beb4-8585c9dec309
- drop the zip on the local page OR
- unpack the zip and drop the unpacked folder OR
- step into the folder, select the .mtlx file and the texture folder, and drop them
- just drop the .mtlx file (currently no UI for this, but will log errors that texture references can't be found)

hybridherbst:
If you're curious, the interesting parts are
- enabling the cache
  https://github.com/AcademySoftwareFoundation/MaterialX/pull/1482/files#diff-00600a31dae106b0f6841a6652381949af88200794a842716e87e6a898bd0209R132

- seeding the cache with images
  https://github.com/AcademySoftwareFoundation/MaterialX/pull/1482/files#diff-bd7a4a21e5443b8ab15c91080430ea312e861919b98a096ae86285905bb8b206R219
  and the mtlx file
  https://github.com/AcademySoftwareFoundation/MaterialX/pull/1482/files#diff-bd7a4a21e5443b8ab15c91080430ea312e861919b98a096ae86285905bb8b206R226
- this ensures that all future requests to these URLs (e.g. the textures referenced inside the mtlx file or the mtlx file itself) are hitting the cache instead of being `fetch`ed, which won't work since they're not really URLs. (the alternative would be turning all the bytes into blob URLs, but that feels less clean).
- unzipping if there's a zip, zipped files are treated the same way as dropped files
  https://github.com/AcademySoftwareFoundation/MaterialX/pull/1482/files#diff-bd7a4a21e5443b8ab15c91080430ea312e861919b98a096ae86285905bb8b206R167
- iterating the dropped data (wherever it came from) and loading it
  https://github.com/AcademySoftwareFoundation/MaterialX/pull/1482/files#diff-bd7a4a21e5443b8ab15c91080430ea312e861919b98a096ae86285905bb8b206R204

kwokcb:
Thanks for the clear explanation. I'll take a closer look at the code for review.

* For posterity If it's not in the code, could you add these annotations. 
* For users, could you update the docs [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/Viewer.md) to denote the new functionality to outline what is supported (basically what you mentioned in the ASWF Slack). It may not be obvious for instance that folks can drop zip file from the GPUOpen library.

Thanks.

kwokcb (commented):
Looks great. Just some minor comments, mostly about comments and docs.

PR #1485, opened 2023-08-23
Improve robustness of Metal render tests

Morteeza:
Follow-up to changes in https://github.com/AcademySoftwareFoundation/MaterialX/commit/a9e06a4fd5ca0863e112cf31087c34f9199e3dd5


Morteeza:
Without this change, MaterialXTests needs to be ran from a specific path; otherwise, some texture loads will fail.

jstone-lucasfilm (approved):

PR #1488, opened 2023-08-24
Add OPENIMAGEIO_UTIL_LIBRARY to OpenImageIO

mikaelsundell:


lgritz:
Aside: Is there a reason to have a FindOpenImageIO, can you not rely on the exported configs that OIIO has had for a few years now?

mikaelsundell:
Good question, this is a quick fix to make sure the ParamValue+List symbols are resolved during build. Should I give it a try and change to the OpenImageIO*.cmake configs?

lgritz:
I think it depends on what is the oldest version of OIIO you want to support.

jstone-lucasfilm:
@lgritz @mikaelsundell Perhaps we can take this in two stages, first addressing the missing functionality in our existing FindOpenImageIO, and then proposing an update that would allow us to remove it entirely?  Before we restrict the range of OpenImageIO versions that we support, I'd ideally like to check with the MaterialX TSC and make sure we're not stranding any existing users.

mikaelsundell:
@lgritz @jstone-lucasfilm Agree, I can prepare another PR using the OIIO configs for consideration in the meantime. Tried it yesterday within my build system and it's not much work.

jstone-lucasfilm (approved):
This looks good to me, thanks @mikaelsundell!

PR #1489, opened 2023-08-24
Add up-axis option to triplanarprojection

crydalch:
Improves usability by letting artists pick an up axis, and therefore which axes are aligned side-by-side. This is useful for textures with a horizontal alignment, such as bricks.

The default is "Z", which gives the same visual appearance as the current default.

image


jstone-lucasfilm (approved):
This looks like a great improvement to me, thanks @crydalch.  I would have no objections to merging this change for 1.38.8, but I'd like to CC @dbsmythe for his thoughts as well.

dbsmythe:
Agreed, this is a good improvement, nice for artistic uses as well as to help deal with matching results between "Z-up" vs "Y-up" systems.

PR #1492, opened 2023-08-25
Changed OpenImageIO to use Config

mikaelsundell:
Updated read_image to non-deprecated version in 2.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Added
 
 - Added an initial NPR (non-photorealistic rendering) data library to MaterialX, supporting the [View Direction](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1631), [Facing Ratio](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1671), and [Gooch Shading](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1674) nodes.
-- Added new nodes to the standard data library, including [Reflect](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1690), [Refract](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1698), [Safe Power](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1689) and [Create Matrix](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1553).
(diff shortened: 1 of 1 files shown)


mikaelsundell:
I don't expect this one to be merged, more for discussion around switching over to Config based OpenImageIO. Left the target_include_directories commented out, should be removed. Kept the uppercase convention to not break the build. Maybe it's worth moving the MATERIALX_BUILD_OIIO and OPENIMAGEIO_FOUND up in the build system to check both at the same time to enable/ disable OpenImageIO if not found. Right now we can request OIIO, still it may not be found? Happens in MaterialXRender and put into parent scope and affects other libraries/ dependencies in the project.

jstone-lucasfilm:
@mikaelsundell I like the idea of including this change in the MaterialX 1.39 release that is scheduled for this summer, though we should discuss the idea with the MaterialX TSC before proceeding.

source/MaterialXRender/CMakeLists.txt
```diff
@@ -30,35 +26,44 @@ if(MATERIALX_BUILD_SHARED_LIBS AND MSVC)
...
     set(OPENIMAGEIO_ROOT_DIR ${MATERIALX_OIIO_DIR})
-    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/External/OpenImageIO")
-    find_package(OpenImageIO REQUIRED)
-    if(OPENIMAGEIO_FOUND)
-        include_directories(${OPENIMAGEIO_INCLUDE_DIR})
-        target_link_libraries(${MATERIALX_MODULE_NAME} ${OPENIMAGEIO_LIBRARIES})
+    find_package(OpenImageIO CONFIG REQUIRED)
+    if(OpenImageIO_FOUND)
+        message("Found OpenImageIO")
+        include_directories(${OpenImageIO_INCLUDE_DIR})
```
lgritz, line 48: If you're using targets (OpenImageIO::OpenImageIO in your target_link_libraries), that already adds the right include things, you shouldn't need a separate include_directories.

source/MaterialXRender/CMakeLists.txt
```diff
@@ -30,35 +26,44 @@ if(MATERIALX_BUILD_SHARED_LIBS AND MSVC)
...
+        target_link_libraries(${MATERIALX_MODULE_NAME} OpenImageIO::OpenImageIO OpenImageIO::OpenImageIO_Util)
```
lgritz, line 49: It's customary here to put either PUBLIC or PRIVATE, depending on whether the OpenImageIO types or functions are exposed through your APIs (I expect not, so should probably be PRIVATE).

mikaelsundell:
Thanks @jstone-lucasfilm and you're right @lgritz :-) this PR was initially a draft will make it final and test against current master and push.

PR #1499, opened 2023-08-27
Improvements to file I/O options in graph editor

kwokcb:
### Issues:

Address some file load/save usability concerns:

- Shortcut to refresh: https://github.com/AcademySoftwareFoundation/MaterialX/issues/1398#issuecomment-1680098103
- Quick save: https://github.com/AcademySoftwareFoundation/MaterialX/issues/1398#issuecomment-1680360553

- CTRL_S as file save (instead of save as)
- CTRL_SHIFT_S as file save-as
- CTRL_R as reload (refresh)

### Logic Changes:

- Change CTRL_S to be save if there is a file loaded. Will not save the initial default node configuration, but will if a specific file was loaded from the command line.
- Add CTRL_SHIFT_S which will bring up the File->Save dialog.
- Add CTRL_R to reload the current document if there is any.
- Add some misc. logging information for these.

( @marwie, Please take a look if this matches the proposed behaviour.
@lfl-eholthouser, Please review if this behaviour is matches what you'd like to see.
Thanks ).

### UI Changes:

- This is how the File dialog looks:

image

### DIFF
--- a/source/MaterialXGraphEditor/Main.cpp
+++ b/source/MaterialXGraphEditor/Main.cpp
@@ -98,6 +98,7 @@ int main(int argc, char* const argv[])
     }
 
     std::string materialFilename = "resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx";
+    bool isDefaultFilename = true;
     std::string meshFilename = "resources/Geometry/shaderball.glb";
     mx::FileSearchPath searchPath = mx::getDefaultDataSearchPath();
     mx::FilePathVec libraryFolders;
@@ -112,7 +113,11 @@ int main(int argc, char* const argv[])
 
         if (token == "--material")
         {
-            materialFilename = nextToken;
+            if (nextToken != materialFilename)
+            {
+                materialFilename = nextToken;
+                isDefaultFilename = true;
+            }
         }
         else if (token == "--mesh")
         {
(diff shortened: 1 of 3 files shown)


source/MaterialXGraphEditor/Main.cpp
```diff
@@ -98,6 +98,7 @@ int main(int argc, char* const argv[])
     }
 
     std::string materialFilename = "resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx";
+    bool isDefaultFilename = true;
```
kwokcb, line 101: The initially loaded content should not be considered an explicitly "user" loaded file. So CTRL-S won't accidently overwrite it.

source/MaterialXGraphEditor/Main.cpp
```diff
@@ -112,7 +113,11 @@ int main(int argc, char* const argv[])
 
         if (token == "--material")
         {
-            materialFilename = nextToken;
+            if (nextToken != materialFilename)
+            {
+                materialFilename = nextToken;
```
kwokcb, line 118: However if the user specifies the file on the command line, then this should cache the filename and allow CTRL-S and CTRL-R to work after load.

kwokcb:
Good catch. The same UI update as load should be applied for reload. I'll fix this.

jstone-lucasfilm (approved):

PR #1502, opened 2023-08-29
Rename texcoord input in 2D shapes

jstone-lucasfilm:
This changelist renames the main 2D input from `sample` to `texcoord` in the 2D shape nodes, aligning them with broader conventions for 2D procedurals in MaterialX.

### DIFF
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -1097,12 +1097,12 @@
 
   <!--
     Node: <line>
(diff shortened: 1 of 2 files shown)


ashwinbhat (approved):
Looks good to me. Consistent with the current convention.

zicher3d:
Looking good. Only one question on like 1692.
Should "sample_double" (and references to it) also be changed to ""texcoord_double"?
It's an internal node, but for readability maybe should be changed.

jstone-lucasfilm:
Good point, @zicher3d, though since this is not part of the artist-facing interface for the node, it's a change that can safely be made further in the future.

PR #1510, opened 2023-09-01
Add method to update uniforms in ShaderRenderer

ashwinbhat:
ShaderRenderer::updateUniform allows user to set update a Uniform value.


ashwinbhat:
@bernardkwok @jstone-lucasfilm This is small change to allow updating of uniforms in ShaderRenderer.

source/MaterialXRender/ShaderRenderer.h
```diff
@@ -110,6 +110,9 @@ class MX_RENDER_API ShaderRenderer
     /// Validate inputs for the program.
     virtual void validateInputs() { }
 
+    /// Update the program with value of the uniform.
+    virtual void updateUniform(ShaderPort* uniform) = 0;
```
jstone-lucasfilm, line 114: I'd recommend giving this virtual method a default definition, so that we don't change this into an abstract base class (which has implications for the complexity of Python bindings).

source/MaterialXRenderMsl/MslRenderer.mm
```diff
@@ -124,6 +124,17 @@
...
 
+void MslRenderer::updateUniform(ShaderPort* uniform)
+{
+
+    if (!uniform)
+    {
+        return;
+    }
+
+    _glProgram->bindUniform(uniform->getVariable(), uniform->getValue());
```
jstone-lucasfilm, line 135: Does the `_glProgram` member variable exist on this platform?

jstone-lucasfilm (commented):
This looks like a promising addition, @ashwinbhat, and I had just two notes on the details.

source/MaterialXRender/ShaderRenderer.h
kwokcb: If we're doing this. Can we throw the "unimplemented" exception here? This will cover all derived classes instead of just `OSLRenderer`. Then you can get rid of the override there.

kwokcb: I would suggest for generality to pass over a `ValuePtr`. Then the update can be done w/o adding in the `ShaderPort` dependence.

source/MaterialXRenderMsl/MslRenderer.mm
kwokcb: Looks like a typo. Guessing it should be `_program`.

kwokcb:
Just a note that this is also done via the Material::updateUniform(). Don't see a way to consolidate this now, but would be worthwhile to consider a way to bridge these somehow?

source/PyMaterialX/PyMaterialXRenderOsl/PyOslRenderer.cpp
```diff
@@ -19,6 +19,7 @@ void bindPyOslRenderer(py::module& mod)
         .def("createProgram", static_cast<void (mx::OslRenderer::*)(const mx::ShaderPtr)>(&mx::OslRenderer::createProgram))
         .def("createProgram", static_cast<void (mx::OslRenderer::*)(const mx::OslRenderer::StageMap&)>(&mx::OslRenderer::createProgram))
         .def("validateInputs", &mx::OslRenderer::validateInputs)
+        .def("updateUniform", &mx::OslRenderer::updateUniform)
```
jstone-lucasfilm, line 22: Perhaps this OSL-specific Python binding is not needed, as `OslRenderer` will inherit the default behavior from `ShaderRenderer`?

jstone-lucasfilm (commented):
Following up on the earlier question from @kwokcb, what is the relationship of this new method to the existing `ShaderMaterial::modifyUniform`?  Should one now be implemented in terms of the other?

Here is the GLSL implementation of `ShaderMaterial::modifyUniform` as an example of this existing pattern:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/6d6c5aa72670c38de95726190a51ccf9bd935485/source/MaterialXRenderGlsl/GlslMaterial.cpp#L345

ashwinbhat:
> Following up on the earlier question from @kwokcb, what is the relationship of this new method to the existing `ShaderMaterial::modifyUniform`? Should one now be implemented in terms of the other?
> 
> Here is the GLSL implementation of `ShaderMaterial::modifyUniform` as an example of this existing pattern:
> 
> https://github.com/AcademySoftwareFoundation/MaterialX/blob/6d6c5aa72670c38de95726190a51ccf9bd935485/source/MaterialXRenderGlsl/GlslMaterial.cpp#L345

There is a bit of a duplicate between the ShaderRenderer and ShaderMaterial. Both ShaderRenderer and ShaderMaterial should eventually be combined as part of a larger refactoring effort. ShaderRenderer has customizable renderer but ShaderMaterial does not, so it would be helpful to merge these capabilities.

ashwinbhat:
> > Following up on the earlier question from @kwokcb, what is the relationship of this new method to the existing `ShaderMaterial::modifyUniform`? Should one now be implemented in terms of the other?
> > Here is the GLSL implementation of `ShaderMaterial::modifyUniform` as an example of this existing pattern:
> > https://github.com/AcademySoftwareFoundation/MaterialX/blob/6d6c5aa72670c38de95726190a51ccf9bd935485/source/MaterialXRenderGlsl/GlslMaterial.cpp#L345
> 
> There is a bit of a duplicate between the ShaderRenderer and ShaderMaterial. Both ShaderRenderer and ShaderMaterial should eventually be combined as part of a larger refactoring effort. ShaderRenderer has customizable renderer but ShaderMaterial does not. ShaderMaterial offers more flexibility on material editing/updates but ShaderRenderer does not. It would be great to merge these capabilities.

kwokcb (approved):
Looks good to me.

jstone-lucasfilm (approved):

PR #1511, opened 2023-09-02
Consolidate CI workflows

jstone-lucasfilm:
- Merge Python Wheel jobs into the Main workflow, reducing overlap and simplifying maintenance.
- Merge Build and Test jobs for Python Wheels, allowing an increase in build parallelism.
- Omit Python Wheel jobs in forks of MaterialX.


.github/workflows/main.yml
```diff
@@ -326,3 +327,66 @@ jobs:
         name: MaterialX_JavaScript
         path: javascript/build/installed/JavaScript/MaterialX        
         if-no-files-found: ignore   
+
+    - name: Build Python SDist
```
JeanChristopheMorinPerso, line 331: I would put this in a separate job or merge it into the `wheel` job to make sure that we don't create an sdist with changes. We want to make sure the sdist is as clean as possible.

Additionally, I think we should build the wheels from the sdist, to confirm that the sdist contains everything necessary to build wheels. This is important because users can do `pip install MaterialX --no-binary=:all:` which will force a compilation of MaterialX from the sdist instead of downloading the pre-compiled binaries. Which means that we should bring back https://github.com/AcademySoftwareFoundation/MaterialX/pull/1511/files#diff-092659a9bd0068791326fb1d08f005532e5aeb983b999a7bd32c51deee0a71d5L67-L70 and https://github.com/AcademySoftwareFoundation/MaterialX/pull/1511/files#diff-092659a9bd0068791326fb1d08f005532e5aeb983b999a7bd32c51deee0a71d5L58-L62.

jstone-lucasfilm: That's a good point, @JeanChristopheMorinPerso, and it's a fine balance I'm aiming to strike with this consolidation, as inter-job dependencies and build-time downloading of artifacts both appear to be sources of build-time stalls in GitHub Actions.

In my local fork of MaterialX, this consolidation greatly reduces the time spent in each build of MaterialX, even when running both the `build` and `wheels` jobs.

On the other hand, I don't want to remove important safety checks that the original system of inter-job dependencies provided, so perhaps additional design thinking is needed here.

jstone-lucasfilm: I like your suggestion of moving the SDist step into the `wheels` job, since the time spent in this step is so small (roughly 17 seconds), so performing the step once for each wheel wouldn't introduce much overhead.

We could then upload just a single copy of the SDist from this job (i.e. Linux with Python 3.11), so that no redundant uploading or storage is introduced.

I'll take a second pass at this proposal with those goals, and see if that approach can work well.

JeanChristopheMorinPerso: Well, now that I think about it, we shouldn't generate one wheel per python version and OS. We might be better splitting it off into a seperate job.

jstone-lucasfilm: I've moved the two SDist steps as high as possible in the `build` job, so that they cannot be affected by later build and test steps, and I believe this should address the first concern you bring up above.

For the second concern, I'm torn about the idea of reintroducing inter-job dependencies, as this seems to be a source of stalls and increased volatility in build times (in my own experience so far with the `python` workflow).

In order to get a better sense of the cost/benefit tradeoff of this choice, I'd be interested in your experience with situations where the Build Python SDist step might generate incorrect results, despite the successful building and validation of Python Wheels on each platform.  Is this something that you've encountered before?

JeanChristopheMorinPerso: We had a broken sdist when we released OTIO 0.15 and had to release a hotfix (0.15.1) and I see broken sdists on a weekly basis in my day job (since my job is to re-package stuff, of which a lot if from PyPI).

I'm guessing the issue you are seeing is that the jobs sometime wait for agents to be available? OCIO and OpenEXR released new versions earlier this week and MaterialX is pushing lots of builds too. I think we can run 100 jobs max organization wide (though I'm not 100%, I don't remember the exact number). @tykeal would know for sure.

jstone-lucasfilm: Thanks for those examples, @JeanChristopheMorinPerso, and I can see the benefit in a separate SDist step to guard against those failure cases.  It should only add one inter-job dependency, and it seems like a reasonable tradeoff with respect to performance versus robustness.  Let me give that a go, and we can see how the build performance plays out in practice.

tykeal: We have a maximum of 500 concurrent jobs (50 if MacOS) with our current setup on GitHub

PR #1512, opened 2023-09-03
Raise minimum version of CMake to 3.5

jstone-lucasfilm:
This changelist raises the minimum version of CMake to 3.5, as earlier versions have been deprecated by the CMake developers.


lgritz:
May I ask why you chose 3.5 in particular? Are you committed to always supporting as far back as the CMake authors are still not disavowing?

You can see in the [VFX Industry Build Matrix](https://docs.google.com/spreadsheets/d/1EwRlz5ZYObEOdBfIk8iTX5thlpTyEAfp3bxOgAfFOiU/) spreadsheet that a bunch of other software in our ecosystem use much more recent versions: OpenEXR requires 3.12. OIIO and OSL have required 3.12 for years and are in the process of moving to 3.15 as their new minimum. OpenColorIO and Alembic require 3.13. USD requires 3.14. OpenVDB and OpenTimelineIO require 3.18.

I think what I'm saying is, so much of the ecosystem requires much newer versions, I don't think you need to be shy about jumping much farther ahead. Anybody building MX from source very likely has to build some of those other packages, so they'll need to have a newer CMake available.

CMake 3.13's addition of `-B` and `-S` alone is worth the price of admission -- it's nice for new users to just be given a simple cmake command line and never again have to explain about the `mkdir build ; cd build ; cmake ..` dance.

jstone-lucasfilm:
I suspect you're right, @lgritz, and I may be reflexively maintaining too broad of a range of CMake versions.  The codebase that I was comparing against is [PyBind11](https://github.com/pybind/pybind11/blob/master/CMakeLists.txt), which still maintains support for CMake 3.5, but you're right that OSL, OIIO, and USD are much meaningful reference points here.

I'll plan to bring this up with the MaterialX TSC, to get a better sense of whether these ancient versions of CMake are worthwhile to support at all, and if they're aligned then we can choose a more modern minimum.

JeanChristopheMorinPerso:
I would probably refer to something like https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html instead of Pybind11 for this, or like Larry said, the other ASWF projects.

As for OTIO, we went with 3.18 because that's the first version that had a decently working FindPython module.

jstone-lucasfilm:
@JeanChristopheMorinPerso It seems worthwhile for us to bring this up at an upcoming MaterialX TSC meeting, to get a sense of whether any studios still rely upon ancient versions of CMake.  Even as recently as two years ago, I've seen pipelines within production studios that defaulted to Python 2.7 and CMake 3.6, but I'm hopeful that the past couple of years have swept these legacy uses aside.

PR #1517, opened 2023-09-06
Improve stream handling in Viewer and GraphEditor

### LINKED ISSUE #1494: GraphEditor / Viewer : <texcoord> index never loads uv set `index` value
Not sure why but if you have a `<texcoord>` node in your document and load it in, the `index` value is always set to 0.
Tried this with both the GraphEditor and Viewer

Just loading in the file and dumping the string for it seems fine so it's not a file I/O issue.

Not sure why this would be, but seems very strange. Tried other nodes with integer inputs and they seem fine.

This is a sample file with <texcoord> + bunch of other nodes with integer inputs.
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <surfacematerial name="MAT_Material" type="material" xpos="17.391304" ypos="0.000000">
    <input name="surfaceshader" type="surfaceshader" nodename="surface_unlit" />
  </surfacematerial>
  <texcoord name="texcoord" type="vector2" xpos="9.362319" ypos="-1.896552">
    <input name="index" type="integer" value="10" />
  </texcoord>
  <look name="look1">
    <materialassign name="materialassign1" material="MAT_Material" geom="Cube" />
  </look>
  <image name="image_color3" type="color3" xpos="11.391304" ypos="-0.646552">
    <input name="texcoord" type="vector2" nodename="texcoord" />
    <input name="file" type="filename" value="D:\Work\materialx\glTF_MaterialX\build\bin\resources\glTF_import\MultiUVTest\glTF\uv1.png" />
    <input name="frameoffset" type="integer" value="1" />
  </image>
  <surface_unlit name="surface_unlit" type="surfaceshader" xpos="13.449275" ypos="-0.508621">
    <input name="emission_color" type="color3" nodename="image_color3" />
  </surface_unlit>
  <constant name="constant_integer" type="integer" xpos="11.217391" ypos="-2.586207">
    <input name="value" type="integer" value="1" />
  </constant>
  <fractal3d name="fractal3d_float" type="float" xpos="13.536232" ypos="-1.896552">
    <input name="octaves" type="integer" value="0" />
  </fractal3d>
  <randomfloat name="randomfloat_float" type="float" xpos="13.666667" ypos="-3.560345">
    <input name="seed" type="integer" value="2" />
  </randomfloat>
  <tangent name="tangent_vector3" type="vector3" xpos="11.217391" ypos="-4.232759">
    <input name="index" type="integer" value="1" />
  </tangent>
  <geomcolor name="geomcolor_float" type="float" xpos="14.115942" ypos="-5.362069">
    <input name="index" type="integer" value="1" />
  </geomcolor>
  <geompropvalue name="geompropvalue_integer" type="integer" xpos="16.384058" ypos="-4.646552">
    <input name="default" type="integer" value="2" />
    <input name="geomprop" type="string" value="" />
  </geompropvalue>
  <place2d name="place2d_vector2" type="vector2" xpos="16.550724" ypos="-2.396552">
    <input name="operationorder" type="integer" value="10" />
  </place2d>
  <ifgreater name="ifgreater_floatI" type="float" xpos="10.173913" ypos="-6.060345">
    <input name="value1" type="integer" value="2" />
    <input name="value2" type="integer" value="3" />
  </ifgreater>
  <convert name="convert_integer_float" type="float" xpos="12.840580" ypos="-6.862069">
    <input name="in" type="integer" value="10" />
  </convert>
  <texcoord name="texcoord_vector3" type="vector3" xpos="8.782609" ypos="-3.551724">
    <input name="index" type="integer" value="100" />
  </texcoord>
</materialx>
```

### LINKED ISSUE #1495: CGLTF loader does not load in multiple color or texcoord sets
Unsure if this was changed or not but the sample GLTF does not load in multiple CPV or UV sets
from a glTF file.

There is code to handle this but it blocked by an early exit to halt multiple set loading.

kwokcb:
## Issue

Fixes #1495 
Fixed #1494

* The glTF loader does not load in multiple texcoords so can't use them if they exist. It does for colors.
* Both the Viewer and GraphEditor force all `texcoord` node lookups to be 0 so no other set can be used.

## Fixes

* Update `Mesh` class to create "fallback" streams if not found. This is similar action to what the Web Viewer does using three.js. Will handle if normal, tangent, bitangent or color or texcoord sets do not exist. Missing texcoord sets map to texcoord set 0. Missing color sets map to position.
* Change `Glsl` and `Msl` renderers to create fallbacks if needed.

### glTF Texcoord Test
* MultiUVTest from glTF Sample Repository. Uses existing texcoord set 1 from glTF file.

image
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <gltf_pbr name="SHD_Material" type="surfaceshader" xpos="13.768116" ypos="-0.672414">
    <input name="base_color" type="color3" nodename="image_basecolor" />
    <input name="metallic" type="float" value="1" />
    <input name="roughness" type="float" value="1" />
    <input name="occlusion" type="float" value="1" />
    <input name="emissive" type="color3" nodename="image_emission" />
  </gltf_pbr>
  <surfacematerial name="MAT_Material" type="material" xpos="17.391304" ypos="0.000000">
    <input name="surfaceshader" type="surfaceshader" nodename="SHD_Material" />
  </surfacematerial>
  <gltf_image name="image_basecolor" type="color3" xpos="10.144928" ypos="-0.896552">
    <input name="file" type="filename" value="uv0.png" colorspace="srgb_texture" />
  </gltf_image>
  <gltf_image name="image_emission" type="color3" xpos="10.144928" ypos="0.422414">
    <input name="file" type="filename" value="uv1.png" colorspace="srgb_texture" />
    <input name="texcoord" type="vector2" nodename="texcoord" />
  </gltf_image>
  <texcoord name="texcoord" type="vector2" xpos="6.463768" ypos="-0.517241">
    <input name="index" type="integer" value="1" />
  </texcoord>
</materialx>
```

### Missing Stream Test
image

```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <texcoord name="texcoord" type="vector2" xpos="8.666667" ypos="0.870690">
    <input name="index" type="integer" value="10" />
  </texcoord>
  <geomcolor name="geomcolor_color3" type="color3" xpos="8.434783" ypos="-0.810345">
    <input name="index" type="integer" value="100" />
  </geomcolor>
  <checkerboard name="checkerboard_color3" type="color3" xpos="10.637681" ypos="1.146552">
    <input name="texcoord" type="vector2" nodename="texcoord" />
    <input name="uvtiling" type="vector2" value="20, 20" />
    <input name="color1" type="color3" value="0.374083, 0.368595, 0.368595" />
  </checkerboard>
  <surface_unlit name="surface_unlit" type="surfaceshader" xpos="13.797101" ypos="-2.258621">
    <input name="emission_color" type="color3" nodename="geomcolor_color3" />
  </surface_unlit>
  <surface_unlit name="surface_unlit2" type="surfaceshader" xpos="14.000000" ypos="0.931035">
    <input name="emission_color" type="color3" nodename="checkerboard_color3" />
  </surface_unlit>
  <tangent name="tangent_vector3" type="vector3" xpos="8.884058" ypos="3.206897">
    <input name="index" type="integer" value="34" />
  </tangent>
  <bitangent name="bitangent_vector3" type="vector3" xpos="8.565217" ypos="5.801724">
    <input name="index" type="integer" value="200" />
  </bitangent>
  <convert name="convert_vector3_surfaceshader" type="surfaceshader" xpos="12.326087" ypos="3.758621">
    <input name="in" type="vector3" nodename="tangent_vector3" />
  </convert>
  <convert name="convert_vector3_surfaceshader2" type="surfaceshader" xpos="12.028986" ypos="6.232759">
    <input name="in" type="vector3" nodename="bitangent_vector3" />
  </convert>
  <surfacematerial name="surfacematerial" type="material" xpos="16.463768" ypos="-1.620690">
    <input name="surfaceshader" type="surfaceshader" nodename="surface_unlit" />
  </surfacematerial>
  <surfacematerial name="surfacematerial2" type="material" xpos="16.434782" ypos="0.439655">
    <input name="surfaceshader" type="surfaceshader" nodename="surface_unlit2" />
  </surfacematerial>
  <surfacematerial name="surfacematerial3" type="material" xpos="16.695652" ypos="3.629310">
    <input name="surfaceshader" type="surfaceshader" nodename="convert_vector3_surfaceshader" />
  </surfacematerial>
  <surfacematerial name="surfacematerial4" type="material" xpos="16.666666" ypos="5.732759">
    <input name="surfaceshader" type="surfaceshader" nodename="convert_vector3_surfaceshader2" />
  </surfacematerial>
</materialx>
```

## Note
* The render test suite already creates fallbacks for any missing streams before-hand (vs on demand here).


jstone-lucasfilm:
This looks really promising, thanks @kwokcb, and I should have a chance to take a closer look soon.

source/MaterialXRenderGlsl/GlslProgram.cpp
```diff
@@ -1054,10 +1055,11 @@ int GlslProgram::mapTypeToOpenGLType(const TypeDesc* type)
...
 const GlslProgram::InputMap& GlslProgram::updateAttributesList()
 {
-    if (_attributeList.size() > 0)
-    {
-        return _attributeList;
-    }
+    //if (_attributeList.size() > 0)
+    //{
+    //    return _attributeList;
+    //}
```
jstone-lucasfilm, line 1061: Is this commented code now safe to remove from GlslProgram::updateAttributesList?

kwokcb: Yes. I forgot to remove the commented out code.
As the attribute list can be dynamically updated we cannot perform an early exit just by testing if the attribute list exists.

source/MaterialXRender/Mesh.cpp
```diff
@@ -41,8 +42,17 @@ Mesh::Mesh(const string& name) :
...
 MeshStreamPtr Mesh::generateNormals(MeshStreamPtr positionStream)
 {
+    const string normalStreamName = "i_" + MeshStream::NORMAL_ATTRIBUTE;
+
+    // Return if already exists
+    MeshStreamPtr normalStream = getStream(normalStreamName, false);
+    if (normalStream)
+    {
+        return normalStream;
+    }
```
jstone-lucasfilm, line 52: If these methods were named `acquireNormals` and `acquireTangents` and used globally as vertex stream accessors, then I could see the value in adding these checks for existing data, but we currently use `generateNormals` and `generateTangents` with the goal of unconditionally generating new vertex streams.  Given that context, I'd recommend removing these checks from the core methods, and making sure that all of our callers have the same expectations from them.

source/MaterialXTest/MaterialXRender/RenderUtil.cpp
```diff
@@ -422,7 +422,7 @@ void ShaderRenderTester::addAdditionalTestStreams(mx::MeshPtr mesh)
 
     const std::string TEXCOORD_STREAM1_NAME("i_" + mx::MeshStream::TEXCOORD_ATTRIBUTE + "_1");
     mx::MeshFloatBuffer* texCoordData2 = nullptr;
-    if (!mesh->getStream(TEXCOORD_STREAM1_NAME))
+    if (!mesh->getStream(TEXCOORD_STREAM1_NAME, false))
```
jstone-lucasfilm, line 425: I like your decision to make `false` the default for allowing fallbacks in this method, as this is definitely the behavior users will expect.  Given that decision, though, I'd recommend that we omit the explicit argument when using the default behavior, as this seems more clear and compact.

jstone-lucasfilm (commented):
This looks nearly ready to merge, thanks @kwokcb, and I had just a few recommendations for adjustments.

source/MaterialXRender/Mesh.cpp
kwokcb: Actually all callers currently check for existence before calling these methods so this just places the logic inside the methods vs all callers having to add in the check.

source/MaterialXTest/MaterialXRender/RenderUtil.cpp
kwokcb: I'll fix this.

source/MaterialXRender/Mesh.cpp
jstone-lucasfilm: All true points, though I wouldn't recommend establishing a pattern where these checks occur *both* inside and outside these calls.

Since the names of these methods only suggest that data will be *generated*, and all clients currently check outside of the calls, perhaps the best approach is to omit the new inside checks?

jstone-lucasfilm: All true points, but I wouldn't recommend establishing a pattern where checks are performed *both* inside and outside these calls, as this reduces the clarity of where checks are truly needed.

Since the names of these methods only suggest that data will be *generated*, not cached and returned on future calls, and all clients currently perform checks for existing data, perhaps the best approach is to omit these new internal checks?

source/MaterialXRender/Mesh.cpp
```diff
@@ -42,8 +43,17 @@ Mesh::Mesh(const string& name) :
...
 MeshStreamPtr Mesh::generateNormals(MeshStreamPtr positionStream)
 {
+    const string normalStreamName = "i_" + MeshStream::NORMAL_ATTRIBUTE;
+
+    // Return if already exists
+    MeshStreamPtr normalStream = getStream(normalStreamName);
+    if (normalStream)
+    {
+        return normalStream;
+    }
```
jstone-lucasfilm, line 53: Following up on the discussion above, wouldn't we want to omit these existence checks from the `generateNormals`, `generateTextureCoordinates`, and related helper methods?  Ideally we don't want this check to be performed twice, and the use of "generate" in the names of these methods suggests that it's the caller that should be responsible for any caching logic.

kwokcb: I've made the change to perform existence test outside now.

source/MaterialXRender/Mesh.cpp
```diff
@@ -312,4 +323,116 @@ void MeshStream::transform(const Matrix44& matrix)
...
+MeshStreamPtr Mesh::getStream(const string& name, bool allowFallback)
+{
+    for (const auto& stream : _streams)
+    {
+        if (stream->getName() == name)
+        {
+            return stream;
+        }
+    }
+    if (allowFallback)
```
jstone-lucasfilm, line 335: I'm not sure about this design choice, where the formerly const method `Mesh::getStream` will silently create new streams when data is missing.

As an alternative design, what if we were to leave `Mesh::getStream` as a const method, and move the stream construction logic into a new `Mesh::createStream` method.  In the two locations where we need fallbacks for missing streams, we would simply check for a null result from `Mesh::getStream` and then call `Mesh::createStream` explicitly.

source/MaterialXRender/Mesh.h
```diff
@@ -275,39 +275,25 @@ class MX_RENDER_API Mesh
...
-            }
-        }
-        return MeshStreamPtr();
-    }
+    MeshStreamPtr getStream(const string& type, unsigned int index, bool allowFallback = false);
 
     /// Add a mesh stream
     void addStream(MeshStreamPtr stream)
     {
+        for (MeshStreamPtr s : _streams)
```
jstone-lucasfilm, line 292: This seems like a surprising behavior for `Mesh::addStream`, where it will silently skip the requested addition when a stream by the same name already exists.  In this situation, the caller has no way to know that the mesh stream that they have constructed is not actually referenced or needed by the mesh.

Would it instead be better to thrown an exception in this situation, so that the logic bug leading to this situation can be addressed?

jstone-lucasfilm (commented):
This is looking very promising, thanks @kwokcb, and I had two notes about design choices in the review comments.

kwokcb:
Don't really have time to continue pursuing this. Wil close for now.

PR #1520, opened 2023-09-07
Fix input/output count in Element::clearContent

rasmusbonnedal:
When clearContent() is called on an InterfaceElement its _inputCount and _outputCount were not updated even though all inputs and outputs were removed. This led to problems including potential crash in Document::validate().


jstone-lucasfilm (approved):
Thanks for the fix, @rasmusbonnedal, and this looks good to me!

PR #1522, opened 2023-09-08
Refine dot node elision and remove unused code

JGamache-autodesk:


source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1027,48 +1027,16 @@ void ShaderGraph::optimize(GenContext& context)
...
+            // Filename dot nodes must be elided so they do not create extra samplers.
             ShaderInput* in = node->getInput("in");
-            if (in->getChannels().empty())
+            if (in->getChannels().empty() && in->getType() == Type::FILENAME)
             {
                 bypass(context, node, 0);
                 ++numEdits;
             }
         }
-        else if (node->hasClassification(ShaderNode::Classification::IFELSE))
```
JGamache-autodesk, line 1038: This used to target a node called "compare" which was removed in 1.37.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1027,48 +1027,16 @@ void ShaderGraph::optimize(GenContext& context)
...
-                const float intestValue = value ? value->asA<float>() : 0.0f;
-                const int branch = (intestValue <= cutoff->getValue()->asA<float>() ? 2 : 3);
-
-                // Bypass the conditional using the taken branch
-                bypass(context, node, branch);
-
-                ++numEdits;
-            }
-        }
-        else if (node->hasClassification(ShaderNode::Classification::SWITCH))
```
JGamache-autodesk, line 1056: This worked in 1.37 where the "which" input was a "parameter" instead of an "input". Since this input is not marked as "uniform", it should not be elided.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1027,48 +1027,16 @@ void ShaderGraph::optimize(GenContext& context)
         }
         else if (node->hasClassification(ShaderNode::Classification::DOT))
         {
-            // Dot nodes without modifiers can be elided by moving their connection downstream.
+            // Filename dot nodes must be elided so they do not create extra samplers.
             ShaderInput* in = node->getInput("in");
-            if (in->getChannels().empty())
+            if (in->getChannels().empty() && in->getType() == Type::FILENAME)
```
JGamache-autodesk, line 1032: Still solves the issue with extra samplers, but preserves the distinction between "dot" nodes and "constant" nodes.

jstone-lucasfilm:
This looks very promising, thanks @JGamache-autodesk.  Would it be reasonable to consider this proposal for after today's release of MaterialX 1.38.8, so that we have sufficient time to think through all of the potential edge cases?

JGamache-autodesk:
Hi @jstone-lucasfilm. I would consider the automatic elision of `dot` nodes regardless of type a showstopper. I can provide a PR with only a [one line fix](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1522/files#diff-9b4db05a8a32064452d6bd30f0aa5671e525148c2dd1aebc7ffc1ba72764a0d0R1032) to restrict to filenames if that can help acceptance.

jstone-lucasfilm:
Don't worry about simplifying the changelist, @JGamache-autodesk, and let's start testing the proposal as it currently stands.  Thoughts from @kwokcb would be appreciated as well!

JGamache-autodesk:
Dot nodes were elided in #1152 to prevent creating extra samplers in rasterized shadergen.
This preserves the change written by @kwokcb while keeping the distinction between a topological `constant` node and a non-topological `dot` node.
In scenarios where you want to feed a single color to multiple shader nodes in a complex material, using a `constant` node will force recompilation every time the color is tweaked, we want users to be able to use a `dot` node to feed the graph, which should allow tweaking the color without having to recompile the shader. Also allows re-using the shader code for material instances differing on this input color.

resources/Materials/Examples/StandardSurface/standard_surface_wood_tiled.mtlx
```diff
@@ -1,8 +1,12 @@
 <?xml version="1.0"?>
 <materialx version="1.38" colorspace="lin_rec709">
   <nodegraph name="NG_wood1" fileprefix="../../../Images/">
+    <dot name="color_filename" type="filename">
+      <input name="in" type="filename" value="wood_color.jpg" />
+      <input name="note" type="string" value="Checking for spurious samplers" />
+    </dot>
     <tiledimage name="image_color" type="color3">
-      <input name="file" type="filename" value="wood_color.jpg" colorspace="srgb_texture" />
+      <input name="file" type="filename" nodename="color_filename" colorspace="srgb_texture" />
```
JGamache-autodesk, line 9: Sufficient to trigger the issue fixed in #1152. This will not generate valid GLSL code unless dot elision is active for filenames.

jstone-lucasfilm (approved):
Thanks for the fix, @JGamache-autodesk!  Once all tests have passed, I'll move forward and include this in the release.

kwokcb:
Just a comment -- after the fact now since this is merged.

Since it is possible to not perform an upgrade on a document, Thus, I'm unsure what it means to remove support for older nodes. if someone loads in a document and tries to code-gen. Could be things will not work at all, but just wanted to add this note.

source/MaterialXGenShader/ShaderNode.cpp
```diff
@@ -292,14 +290,6 @@ ShaderNodePtr ShaderNode::create(const ShaderGraph* parent, const string& name,
     {
         newNode->_classification = Classification::TEXTURE | Classification::DOT;
     }
-    else if (nodeDef.getNodeString() == COMPARE)
-    {
-        newNode->_classification = Classification::TEXTURE | Classification::CONDITIONAL | Classification::IFELSE;
-    }
-    else if (nodeDef.getNodeString() == SWITCH)
```
kwokcb, line 299: I'm not sure way it means to remove the texture classification here. I think nodes like try to perform filtered sample upstream will no longer work. But it could be that it does not work anyways.

source/MaterialXGenShader/ShaderGraph.cpp
kwokcb: I think this should be restored as it checks to see if indeed a uniform unconnected value. I have log an new issue and can discuss there, but now new shaders will always add more code in.  (https://github.com/AcademySoftwareFoundation/MaterialX/issues/1526)

Issue #1208, opened 2023-01-31
ninja error while enabled building Viewer and Node Graph Editor 


igor-elovikov:
With default settings if I set to build both I'm experiencing this error:
`ninja: error: build.ninja:3662: multiple rules generate lib/glfw3.lib [-w dupbuild=err]` I guess there is some conflict with imgui and NanoGUI backends

I've managed to built both apps using `-DMATERIALX_BUILD_SHARED_LIBS:BOOL=ON` though this one isn't default.

* Cmake 3.24.2 Windows 11

dgovil:
Hitting the same here. I'd switched to the Make backend which worked, but yeah your suggestion works around the issue for Ninja

jstone-lucasfilm:
Good catch, @igor-elovikov, and this similar to an issue that we encountered in CMake/Make builds, since NanoGUI includes its own local copy of the `glfw` library.

To resolve the CMake/Make issue, we made small modifications to the copy of `glfw` used in the Graph Editor, so that its project name wouldn't overlap the one in NanoGUI.

Could it be that a similar renaming of library outputs might resolve the issue in CMake/Ninja builds?

dgovil:
I think either renaming the target, or sharing the target should work. ie only configuring the GLFW target for the graph editor if the one for NanoGUI isn't setup already (and vice versa, though I think you can guarantee the GraphEditor comes after the Viewer)

igor-elovikov:
I've made a small PR that I'm using to build graph and viewer at the same time
essentially just renamed lib name for graph editor

jstone-lucasfilm:
Thanks to @igor-elovikov for the fix!

PR #1533, opened 2023-09-13
Fix Ninja builds of both Viewer and Editor

igor-elovikov:
Related to: https://github.com/AcademySoftwareFoundation/MaterialX/issues/1208


source/MaterialXGraphEditor/External/Glfw/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ if (BUILD_SHARED_LIBS AND UNIX)
     # On Unix-like systems, shared libraries can use the soname system.
     set(GLFW_LIB_NAME glfw)
 else()
-    set(GLFW_LIB_NAME glfw3)
+    set(GLFW_LIB_NAME glfw3_gred)
```
jstone-lucasfilm, line 44: This looks like a good catch, thanks @igor-elovikov.  To align this new name with similar ones that distinguish the version of GLFW in the MaterialX Graph Editor, could we name this library `glfw3_minimal`?

PR #1536, opened 2023-09-18
GPU instance test

kwokcb:
This is a test PR to try out GPU instances for CI.
Just tests current software render and tries new gpu render instance.


jstone-lucasfilm:
Thanks for getting this testing started, @kwokcb.

It sounds like there's a conversation between the Linux Foundation and the GitHub Actions team about the status of GPU runners on ASWF projects, and I'll close out this pull request until we have new details from that discussion.

PR #1538, opened 2023-09-21
Align default precision across shading languages

jstone-lucasfilm:
This changelist removes explicit overrides of floating-point precision in shader generation, making the default precision 6 decimal places across shading languages, and allowing the client to adjust the precision of generated shaders via Value::setFloatPrecision.

This addresses a slight visual regression in generated OSL (with a maximum visual error on the order of 1e-3), caused by a hard-coding of floating-point precision to 3 decimal places in OslSyntax.

### DIFF
--- a/source/MaterialXGenMdl/MdlShaderGenerator.cpp
+++ b/source/MaterialXGenMdl/MdlShaderGenerator.cpp
@@ -198,6 +198,9 @@ ShaderPtr MdlShaderGenerator::generate(const string& name, ElementPtr element, G
 
     ShaderPtr shader = createShader(name, element, context);
 
+    // Request fixed floating-point notation for consistency across targets.
+    ScopedFloatFormatting fmt(Value::FloatFormatFixed);
+
     ShaderGraph& graph = shader->getGraph();
     ShaderStage& stage = shader->getStage(Stage::PIXEL);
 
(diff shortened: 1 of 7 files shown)


jstone-lucasfilm:
Attaching the latest render comparisons between GLSL and OSL:

[MaterialXRenderTests_09_20_2023.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/12691220/MaterialXRenderTests_09_20_2023.pdf)

niklasharrysson (approved):
This looks great to me!

rasmusbonnedal:
I tested the branch and I get the expected 6 decimal fixed precision in OSL, GLSL and MDL. Looks good!

source/MaterialXGenMdl/MdlShaderGenerator.cpp
```diff
@@ -198,6 +198,9 @@ ShaderPtr MdlShaderGenerator::generate(const string& name, ElementPtr element, G
 
     ShaderPtr shader = createShader(name, element, context);
 
+    // Request fixed floating-point notation for consistency across targets.
```
kwokcb, line 201: I'm wondering if this should be moved to a new method to set the precision that all generators call into ? Then you don't need to modify this in N different places for N generators if this changes.

jstone-lucasfilm: That's a good thought, though I wonder if there's really any simpler expression of this setup than the current one-liner.  In order to wrap an RAII class, you'd need a second RAII class instantiating the first one, which might just end up being additional indirection without improving clarity or universality across languages.

In the future, one option to consider would be a base ShaderGenerator::generate method that handles common steps, though for now I think it's clearer to have this explicit step in each language, so that developers know that this precision formatting is being invoked here, and can test different precisions locally for custom generators.

PR #1539, opened 2023-09-22
Update overlay node logic to match reference

rasmusbonnedal:
This commit changes the overlay node to match the overlay operation in the Adobe PDF spec.

```
Overlay(cb, cs) = { 2 * cs * cb                 [if cb <= 0.5]
                    1 - 2 * (1 - cs) * (1 - cb) [if cb > 0.5]
```
(cb = B, cs = F in MaterialX specification notation)

It also changes the implementation to graph based instead of code snippets.


rasmusbonnedal:
Before the fix:
image

rasmusbonnedal:
After the fix:
image

rasmusbonnedal:
[before.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/12699860/before.pdf)
[after.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/12699861/after.pdf)

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -3343,8 +3343,8 @@
   <!--
     Node: <overlay>
     Take two 1-4 channel inputs and apply the same operator to all channels:
-      2FB if F<0.5;
-      1-(1-F)(1-B) if F>=0.5
+      2FB if B<0.5;
```
kwokcb, line 3346: A note for @jstone-lucasfilm, I would like to suggest to start migrating this documentation to the `doc` attribute for `nodedef`s. ( All of these comments are not easily discoverable currently, and assume folks won't go to the spec to find out what they do. )

kwokcb (approved):
Looks good to me. Thanks for spending the time to do this.
I left a small suggestion to add docs for the node since it's being edited but this is a minor thing.

libraries/stdlib/stdlib_defs.mtlx
jstone-lucasfilm: That's a good note, thanks @kwokcb, and I believe the SideFX team and other early MaterialX clients have brought that idea up as well  The key balance we'll need to strike is between brevity and clarity, since our `doc` strings will are likely to appear in smaller spaces within application UIs.  One option would be to maintain the more complete descriptions in our specification, and to use simpler one-sentence descriptions in `doc` strings.

jstone-lucasfilm:
Great work, @rasmusbonnedal, and I really like the decision to implement this node as a language-independent graph.  I don't see any issues that concern me in this change, though I'd like to run this through local tests before merging.

libraries/stdlib/stdlib_defs.mtlx
jstone-lucasfilm: That's a good note, thanks @kwokcb, and I believe the SideFX team and other early MaterialX clients have brought that idea up as well The key balance we'll need to strike is between brevity and clarity, since our doc strings are likely to appear in smaller spaces within application UIs. One option would be to maintain the more complete descriptions in our specification, and to use simpler one-sentence descriptions in doc strings.

rasmusbonnedal: I suggest `doc="Combination of multiply and screen blend mode."` for the overlay node. What do you think @kwokcb @jstone-lucasfilm?

kwokcb: I'm wondering if it's better to say that it performs the same logic as the Adobe overlay operator instead (perhaps with a reference) ?
In Slack it was noted the match to this spec logic: https://printtechnologies.org/wp-content/uploads/2020/03/pdf-reference-1.6-addendum-blend-modes.pdf.

jstone-lucasfilm: I'd recommend keeping our `doc` strings completely self-contained, without any links to external references, as these links are not likely to behave as we'd expect in DCC tools.  The suggestion from @rasmusbonnedal above sounds good to me as a starting point.

kwokcb: Not having links is good but it still seems useful to add that this matches the Adobe `overlay` operator logic so maybe add that in ?  I'm just not sure what "Combination of multiply and screen blend mode" -- even "mode" has no meaning without context.

Minianikum: Artists might not be familiar with what different blend-modes mean, as you say @kwokcb.
I think that the second part of the description of overlay in Wikipedia is actually pretty good way of putting it:
"Where the base layer is light, the top layer becomes lighter; where the base layer is dark, the top becomes darker; where the base layer is mid grey, the top is unaffected"

It might need to become shorter though...

jstone-lucasfilm (commented):
This looks good to me, thanks @rasmusbonnedal!

jstone-lucasfilm (approved):

Issue #1483, opened 2023-08-23
Clarify where to build MaterialX from for first time developers


kwokcb:
The documentation can be clarified a bit [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/MainPage.md#quick-start-for-developers) to show where to build from.

The .`gitignore` explicitly ignores the build subfolder and this is what CI does. This means users don't need to add other folder filters by default.

Suggest to add this information in as well as a reference to the CI workflow yaml file for reference.

Thanks @marwie for mentioning this out as a "first timer" :).

MustafaJafar:
Hello, 
Thank you for the `dev days` event.
I've made PR #2388 to resolve this issue.

jstone-lucasfilm:
@MustafaJafar has made great initial progress on this task, and it just needs additional feedback and review from the MaterialX team, so I'll remove the Good First Issue tag from this task for now.

PR #1540, opened 2023-09-25
Developer documentation update

kwokcb:
Update #1483 

- Some minor update to address where to build from.
- Also some other suggested updates to docs on the developer and main page, including linking in graph editor and javascript bindings docs, and mentioning `PyPi` packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ MaterialX is an open standard for representing rich material and look-developmen
 ### Quick Start for Developers
 
 - Download the latest version of the [CMake](https://cmake.org/) build system.
-- Point CMake to the root of the MaterialX library and generate C++ projects for your platform and compiler.
+- Create a new sub-folder called `build` under the root of the MaterialX library.
+- From the `build` folder, point CMake to the root of the MaterialX library and generate C++ projects for your platform and compiler.
 - Select the `MATERIALX_BUILD_PYTHON` option to build Python bindings.
 - Select the `MATERIALX_BUILD_VIEWER` option to build the MaterialX viewer.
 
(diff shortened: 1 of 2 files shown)


kwokcb:
@jstone-lucasfilm , I think the `build` folder blurb will help to clarify as some folks have run into this and think it will help with dev-days. The other changes are sugs to round out some minor pieces of docs.

README.md
```diff
@@ -14,7 +14,8 @@ MaterialX is an open standard for representing rich material and look-developmen
 ### Quick Start for Developers
 
 - Download the latest version of the [CMake](https://cmake.org/) build system.
-- Point CMake to the root of the MaterialX library and generate C++ projects for your platform and compiler.
+- Create a new sub-folder called `build` under the root of the MaterialX library.
```
jstone-lucasfilm, line 17: I like the idea of clarifying our start-up process, but by adding a manual step where the user creates a `build` subfolder, we're effectively steering the user towards command-line usage of CMake and out-of-source builds, since this step is not required from the CMake GUI or for in-source builds.

Maybe we should leave our Quick Start as it stands, but provide details for a few different CMake scenarios in our full Developer Guide?

I'm definitely open to thoughts on this, and I'm now so used to CMake (both on the command-line and from the GUI) that I can no longer be objective about which instructions are valuable.

documents/DeveloperGuide/MainPage.md
```diff
@@ -46,7 +48,10 @@ Additional options for the generation of MaterialX Python include the following:
 
 #### Building The MaterialX Viewer
 
-Select the `MATERIALX_BUILD_VIEWER` option to build the MaterialX Viewer.  Installation will copy the **MaterialXView** executable to a `bin/` directory within the selected install folder.
+Select the `MATERIALX_BUILD_VIEWER` option to build the MaterialX viewer.  Installation will copy the **MaterialXView** executable to a `bin/` directory within the selected install folder.
+
+#### Building The MaterialX Graph Editor
```
jstone-lucasfilm, line 53: I'd add a newline here for consistency with our other sections.

jstone-lucasfilm (commented):
I really like this proposal from a high level, though I have a few notes about the details.

README.md
kwokcb: Sounds good to not clutter this up and leave it in the Developer Guide.

I think for the Developer Guide, I propose something like this instead:

"It is recommended to create a separate build folder, with the default folder name of `build` to match the default Github setup.`, then the blurb about running the build from that folder.

BTW, I found that on some CMake configs that it errors out without using `-S ..` vs `..` for the source directory.

PR #1544, opened 2023-09-28
Add XML tags to MaterialX blocks in specification

kwokcb:
### Issue

Improve readability of Markdown docs MaterialX code snippets

### Changes

- Add xml formatting tag for spec markdown docs.
- Existing docs like shader gen already use the appropriate formatting tags.
- Tested this with converters such as pandoc, and VScode markdown converter and should make no difference as they
will just end up as preformatted text `<pre></pre>`.

### Example
Snippet from one portion of main spec doc. Right side is preview of how it would look formatted.

image


kwokcb:
@dbsmythe, @jstone-lucasfilm : Let me know what you think, but it does make it easier to read, especially for longer snippets. The `xml` tag was already being used for shadergen docs, so I just added them to the spec docs. Thanks.

Issue #1463, opened 2023-08-18
Structs used in final generated OSL shaders, makes it hard for DCC UI's to cope


ZapAndersson:
The code generation for OSL, when used to generate a final shader, it uses a struct parameter:

```
...
struct textureresource { string filename; string colorspace; };
...
shader image_color3
[[
    string mtlx_category = "image",
    string mtlx_name = "image_color3"
]]
(
    textureresource  file = {"", ""}
    [[
        string widget = "filename"
    ]],
    string layer = "",
    color default1 = color(0, 0, 0),
    vector2 texcoord = {u,v},
    string uaddressmode = "periodic",
    string vaddressmode = "periodic",
    string filtertype = "linear",
    string framerange = "",
    int frameoffset = 0
    [[
        string widget = "number"
    ]],
    string frameendaction = "constant",
    output color out = color(0.0)
)
{
    color image_color3_out = color(0.0);
    mx_image_color3(file, layer, default1, texcoord, uaddressmode, vaddressmode, filtertype, framerange, frameoffset, frameendaction, image_color3_out);
    out = image_color3_out;
}
```

This causes issues, not all DCC's can handle struct inputs (I've been told Katana can not), but even for those that do, the metadata gets applied to the whole parameter, not individual struct items. So this means that both file.filename and file.colorspace ands up being treated as "filename" widgets and get a file picker, which is suboptimal for sure.

Talking to @niklasharrysson he said it would be fairly easy to remove the struct from the generated code, so it just shows up as regular parameters.

I propose to do just that, and instead have a convention for parameters, so it would end up as

string file [[ string widget="filename" ]],
string file_ColorSpace [[ string widget="colorspace" ]],

...or something similar instead.

/Z

niklasharrysson:
@ZapAndersson Thanks for creating the issue.

I have made a fix to this in my fork that you can try. The branch is here: https://github.com/niklasharrysson/MaterialX/tree/osl_filename_params

Let me know if it's not working as expected.

ZapAndersson:
Thanks, we will take a look.

In the meantime we made a python hack workaround, and in making it, I noticed the color space never gets put in the OSL texture(). I know this is not implemented yet on the OSL side (just talked to @lgritz about that the other day) but it should be safe to put it in as an optional parameter already.

/Z

ZapAndersson:
@niklasharrysson I tested your branch, and while this solves the issue with the parameters, the color space ends up never being used, basically we need

[...]

PR #1546, opened 2023-10-02
Use explicit colorspace parameters in OSL texture calls

niklasharrysson:
This change list adds the parameter "colorspace" to the texture call in OSL, for sampling color3/color4 textures. Previously this was supported under the covers, by token replacement that a custom OSL generator had to setup, and only done for the Arnold OSL generator.

As discussed in issue #1463 there was a request for having this parameter used by default. Now color transformations can be done internally in OSL, if/when using an OSL version that supports this. The color transform node insertions, that are enabled by default, should then be disabled by setting `GenOptions.emitColorTransforms = false`.


jstone-lucasfilm (approved):

PR #1547, opened 2023-10-02
Improve OSL generation of filename inputs

niklasharrysson:
This change list modifies the generation of public filename parameters from being a `textureresource` struct, to being two separate string parameters for filename and colorspace.

This is to solve the issue discussed in #1463.


source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -493,111 +512,75 @@ void OslShaderGenerator::emitLibraryIncludes(ShaderStage& stage, GenContext& con
...
         {
-            value = _syntax->getDefaultValue(input->getType());
-        }
-        emitString(" = " + value, stage);
-
-        //
-        // Add shader input metadata.
-        //
-
-        auto widgetMetadataIt = UI_WIDGET_METADATA.find(input->getType());
```
niklasharrysson, line 564: Metadata generation was moved out to a separate method, see below.

source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -493,111 +512,75 @@ void OslShaderGenerator::emitLibraryIncludes(ShaderStage& stage, GenContext& con
...
+            // Add the file string input
+            emitLineBegin(stage);
+            emitString("string " + input->getVariable() + " = \"" + valueStr + "\"", stage);
+            emitMetadata(input, stage);
+            emitString(",", stage);
+            emitLineEnd(stage, false);
+
+            // Add the colorspace string input
+            emitLineBegin(stage);
+            emitString("string " + input->getVariable() + "_colorspace = \"" + input->getColorSpace() + "\"", stage);
```
kwokcb, line 556: Hi @niklasharrysson ,
- Won't this always try to add a colorspace input regardless of whether it was defined or not (via `getColorSpace()` 
- It seems also ignore the generation option as to whether or not to emit colorspace ?

- BTW I forget, but how does this get propagated into a "colorspace" input on a OSL `texture()` ? Does the "widget" declaration indicate to add it ?
Thanks.

niklasharrysson: Hi @bernardkwok,

Yes, the parameter for colorspace is always added, but this is the same behavior as before. The change is only to split the parameter from a `textureresource` struct input (that also always hade the colorspace string) into two separate string inputs. The main reason for this is that UI can easier be built from this when we use string inputs instead of a custom struct input. And we can include metadata for these inputs that hints that a widget for colorspace is expected to be used.

So the change is to go from this parameterization:
```
    textureresource  file = {"", ""}
    [[
        string widget = "filename"
    ]],

```
into this new parameterization:
```
    string file = ""
    [[
        string widget = "filename"
    ]],
    string file_colorspace = ""
    [[
        string widget = "colorspace"
    ]],

```
If there is no colorspace set as default value an empty string is set. But the user can of course change this later in the UI built for this shader.

The generation option `GenOption.emitColorTransforms` is used to control if color transform nodes should be inserted or not during generation. I don't think that should affect the shader interface. For example, if an OSL version is used that supports color transformations internally in the `texture` call, then we still want the same shader interface but we want to disable the color transform node insertions to avoid getting double transformations.

The propagation of these parameter values are done same as before, using a `textureresource` struct. A corresponding struct is created from the two string in the main function of the shader and then sent to the image sampler functions. A fix to explicitly set the colorspace string in the `texture` call has been done in PR #1546.

kwokcb: Hi @niklasharrysson ,
Thanks for the explanation. It all makes sense now (been a while since I looked at this :)).
I didn't know you had already changed the `texture()` signature, and was mostly thinking of the case of clearing the `colorspace` string if you turn on the `emitColorTransforms` option.

jstone-lucasfilm (approved):

PR #1559, opened 2023-10-11
Fix edge case in HW shader generation for texcoords

### LINKED ISSUE #1464: Graph Editor: `Implicit cast from "vec3" to "vec2"` - using texcoord both as vector3 and vector2
This seems to be related to `texcoord` 0 being used for the normalmap as well as the basecolor - I'm still investigating this and will try to reproduce it with a fresh material but for the sake of documentation I'll open an issue here now 😅 

image

```xml
<?xml version="1.0"?>
<materialx name="" version="1.38" >
	<nodegraph name="needle_shadergraph" >
		<!-- MainTex -->
		<input name="_MainTex" type="filename" value="Tile_Albedo.jpg" />
		<dot name="dot__MainTex" type="filename" >
			<input name="in" type="filename" interfacename="_MainTex" />
		</dot>
		<texcoord name="uv" type="vector3" >
		</texcoord>
		<convert name="convert_to_vector2" type="vector2" >
			<input name="in" type="vector3" nodename="uv" />
		</convert>
		<image name="tex" type="color4" >
			<input name="file" type="filename" nodename="dot__MainTex" />
			<input name="texcoord" type="vector2" nodename="convert_to_vector2" />
		</image>
		<convert name="convert_to_color3" type="color3" >
			<input name="in" type="color4" nodename="tex" />
		</convert>
		<output name="out" type="color3" nodename="convert_to_color3" />
		<!-- Normal -->
		<input name="_Normal" type="filename" value="Tile_Normal.jpg" />
		<dot name="dot__Normal" type="filename" >
			<input name="in" type="filename" interfacename="_Normal" />
		</dot>
		<texcoord name="uv_2" type="vector2" >
		</texcoord>
		<image name="tex_1" type="color4" >
			<input name="file" type="filename" nodename="dot__Normal" />
			<input name="texcoord" type="vector2" nodename="uv_2" />
		</image>
		<swizzle name="swizzle_0_tex_1" type="vector3" >
			<input name="in" type="color4" nodename="tex_1" />
			<input name="channels" type="string" value="rgb" />
		</swizzle>
		<output name="out_1" type="vector3" nodename="normalmap" />
		<normalmap name="normalmap" type="vector3" >
			<input name="in" type="vector3" nodename="swizzle_0_tex_1" />
		</normalmap>
		<constant name="specular_roughness" type="float" >
			<input name="value" type="float" value="0.0" />
		</constant>
		<output name="out_2" type="float" nodename="invert" />
		<invert name="invert" type="float" >
			<input name="in" type="float" nodename="specular_roughness" />
		</invert>
	</nodegraph>
	<standard_surface name="needle_standard_surface" type="surfaceshader" >
		<!-- BaseColor -->
		<input name="base_color" type="color3" nodegraph="needle_shadergraph" output="out" />
		<!-- NormalTS -->
		<input name="normal" type="vector3" nodegraph="needle_shadergraph" output="out_1" />
		<!-- Smoothness -->
		<input name="specular_roughness" type="float" nodegraph="needle_shadergraph" output="out_2" />
	</standard_surface>
	<surfacematerial name="Default" type="material" >
		<input name="surfaceshader" type="surfaceshader" nodename="needle_standard_surface" />
	</surfacematerial>
</materialx>

```

<details>
<summary>Textures</summary>

Tile_Albedo

Tile_Normal

</details>

madmann91:
This merges both `TexCoordNodeMsl` and `TexCoordNodeGlsl` into a single `HwTexCoordNode`. The parts of the node that are configurable (e.g. the input index) can be changed via re-implementation of the virtual member function `getIndexInput()`.

The new mechanism works regardless of syntax, and uses the smallest texcoord width, depending on what is used in the shader graph.

Fixes #1464


jstone-lucasfilm:
This looks really promising, thanks @madmann91!  If the EasyCLA system isn't working as expected, feel free to submit a support request ticket as described above, and the Linux Foundation is usually very quick in resolving these issues.

source/MaterialXGenShader/ShaderStage.cpp
```diff
@@ -70,6 +70,12 @@ ShaderPort* VariableBlock::add(const TypeDesc* type, const string& name, ValuePt
     auto it = _variableMap.find(name);
     if (it != _variableMap.end())
     {
+        // Automatically try to widen the type of the shader port if the requested type differs from
```
kwokcb, line 73: I assume this also affect / "fix" color if you ask for the same color stream with different number of channels ?

source/MaterialXGenShader/Nodes/HwTexCoordNode.cpp
```diff
@@ -54,9 +54,25 @@ void TexCoordNodeGlsl::emitFunctionCall(const ShaderNode& node, GenContext& cont
         ShaderPort* texcoord = vertexData[variable];
         shadergen.emitLineBegin(stage);
         shadergen.emitOutput(node.getOutput(), true, false, context, stage);
-        shadergen.emitString(" = " + prefix + texcoord->getVariable(), stage);
+
+        // Extract the requested number of components from the texture coordinates (which may be a
```
kwokcb, line 58: This will extract for texcoords, but as noted before if you can widen other streams like color then extraction is needed for those stream types as well. This only seems to be for color now, but might want to take into consideration other streams may have different channel counts in the future. (e.g. if  position-4 or tangent-4 is added)

madmann91: From what I understand, color nodes seem to already perform component extraction: see https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/Nodes/GeomColorNodeGlsl.cpp#L52. Is this correct?

Looking at the code though, I believe `GeomColorNode{Glsl,Msl}` should be merged into a `HwGeomColorNode`, as they are also almost identical, but perhaps this should be better done in a different PR as this does not seem to be linked with this issue.

madmann91: If you are worried that new nodes can now widen port types by mistake, it should be possible to add a boolean parameter to `addStageInput` such that the widening behavior is turned off by default (and an exception is thrown if the types do not match in that case). Would that help in that regard?

madmann91: I've now implemented the checks, as per the comment above, so there should be no risk of a node accidentally widening a port type by mistake. This also adds a check that was not there before, and should protect against requesting a type with a different type than the existing port type. This new behavior also documented in the header file.

I also took the liberty of applying the suggestions from the other PR (#1560), which were to use static strings for port names.

kwokcb: Hi @madmann91 ,

- Thanks a lot for looking into this. 
- I'll take another look at this, but my main thought was do you always widen and trust that each type with perform an extraction from the widened type, or do you want an explicit "firewall" and as more types require widen the "firewall". 

- BTW, I have another query in  that there are derived code generators (such as the one in HDStorm), and wondering if those generators require any changes. Best if they do not but if so, we may want to make a note of that as a heads up.

source/MaterialXGenShader/ShaderStage.cpp
```diff
@@ -65,11 +65,26 @@ ShaderPort* VariableBlock::find(const ShaderPortPredicate& predicate)
...
+        if (shouldWiden)
+        {
+            // Automatically try to widen the type of the shader port if the requested type differs from
+            // the existing port's type.
+            if (it->second->getType()->getSize() < type->getSize())
+            {
+                it->second->setType(type);
+            }
+        }
+        else if (type != it->second->getType())
```
kwokcb, line 82: Thanks for adding this exception, then it's obvious the condition.

source/MaterialXGenShader/ShaderStage.h
```diff
@@ -113,7 +113,14 @@ class MX_GENSHADER_API VariableBlock
...
     /// Add a new shader port to this block.
-    ShaderPort* add(const TypeDesc* type, const string& name, ValuePtr value = nullptr);
+    /// @param type The desired shader port type
+    /// @param name The shader port name
+    /// @param value The value to attach to the shader port
+    /// @param shouldWiden When false, an exception is thrown if the type of the existing port with
+    ///   the same name does not match the requested type. When true, the types can mismatch, and the
+    ///   type of any existing port is widened to match the requested type when necessary.
+    /// @return A new shader port, or a pre-existing shader port with the same name.
+    ShaderPort* add(const TypeDesc* type, const string& name, ValuePtr value = nullptr, bool shouldWiden = false);
```
kwokcb, line 123: Can you do a quick check on the Python wrapper for this class (in `PyMaterialXGenShader`). I didn't see any issue with the change in signature since I don't think it's exposed but best to double-check :). Thanks.

source/MaterialXGenShader/Nodes/HwTexCoordNode.cpp
madmann91: I believe that it's up to you to decide what is the expected behavior. From looking at the code, it seems that certain nodes are "polymorphic", in the sense that they can be created with different types (e.g. texcoord, ...), but others are not. Since this behavior can be specific to the backend or node type, I believe the current solution of explicitly requiring `shouldWiden = true` for polymorphic nodes is safe, as this makes sure the behavior remains the same as before, unless widening is requested: I thus believe that no code change is necessary anywhere else (at least if the current behavior is expected to stay the same). In any case, the new exception check should catch the case when nodes are incorrectly wired.

source/MaterialXGenShader/ShaderStage.h
madmann91: I didn't see any mention of `add` in `PyMaterialX/PyMaterialXGenShader/PyShaderStage.cpp`. Could it be that this is not exposed?

resources/Materials/TestSuite/stdlib/texture/texcoord.mtlx
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
```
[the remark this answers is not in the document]
madmann91: Actually, color nodes do not suffer from the same problem because they are always typed as `COLOR4` in the generated GLSL/MSL code, regardless of their actual width (so a graph which only has a `geomcolor_color3` will still emit an input variable of type `color4` in the variable block).

I think it would be best to first merge this PR and create another one if we want to support codegen with a different type depending on the width of the color nodes in the graph.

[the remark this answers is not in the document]
madmann91: See e.g. `MaterialXGenGlsl/Nodes/GeomColorNodeMsl.cpp`.

jstone-lucasfilm (approved):
This looks good to me, thanks @madmann91!

PR #1560, opened 2023-10-12
Fix edge case in HW shader generation for transform nodes

### LINKED ISSUE #1425: Default transform<X> nodes can crash shader generation (GLSL)
## Issue

There are no default "from" and "to" space values so they are left empty on creation.
If you try and generate code from them, then there are a few missing empty value checks missing and code gen can crash.

## Setup

This is a simple example:
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <transformnormal name="transformnormal_vector3" type="vector3">
    <input name="in" type="vector3" value="0.0, 0.0, 1.0" />
    <input name="fromspace" type="string" value="" uniform="true" />
    <input name="tospace" type="string" value="" uniform="true" />
  </transformnormal>
  <convert name="shader_transformnormal_vector3_out" type="surfaceshader">
    <input name="in" type="vector3" nodename="transformnormal_vector3" />
  </convert>
  <surfacematerial name="material_transformnormal_vector3_out" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="shader_transformnormal_vector3_out" />
  </surfacematerial>
</materialx>
```

## Initial Triage
Loading this in will crash the view and editor where only `GLSL` was tried out. The fix appears to be in `TransformVectorNodeGLSL.cpp` but didn't check other code generators, and it seems all of the "transform" nodes will need to be patched, so wondering if there is a higher level place ? 

## Proposed Fix
Add in firewall checks to see if there is a value and default to empty if none:
- https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/Nodes/TransformVectorNodeGlsl.cpp#L20-L23
- https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenGlsl/Nodes/TransformVectorNodeGlsl.cpp#L45-L49
- ... and other required places ...

## Secondary Proposal
Propose these nodes should have some reasonable default and be enumerations if it makes sense. Otherwise it's mostly
a guess as to what strings are required and documentation validation will fail if the string is incorrect.

madmann91:
This merges the `TransformXXXNodeGlsl` and `TransformXXXNodeMsl` classes into a `HwTransformNode` which has 3 subclasses,
`HwTransformVectorNode`, `HwTransformPointNode`, and `HwTransformNormalNode`, all of which can be configured by overriding the virtual methods of `HwTransformNode`.

Fixes #1425


source/MaterialXGenShader/Nodes/HwTransformNode.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+const ShaderInput* HwTransformNode::getFromSpaceInput(const ShaderNode& node) const
+{
+    return node.getInput("fromspace");
+}
+
+const ShaderInput* HwTransformNode::getToSpaceInput(const ShaderNode& node) const
+{
+    return node.getInput("tospace");
```
kwokcb, line 80: Might be worthwhile to keep "tospace" and "fromspace" as statics / consts to avoid reallocation of the same string.
Also if the name changes this won't break.

source/MaterialXGenShader/Nodes/HwTransformNode.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <MaterialXGenShader/HwShaderGenerator.h>
+#include <MaterialXGenShader/Shader.h>
+
+MATERIALX_NAMESPACE_BEGIN
+
+string HwTransformNode::MODEL = "model";
+string HwTransformNode::OBJECT = "object";
+string HwTransformNode::WORLD = "world";
+
+static inline string getSpaceName(const ShaderInput* spaceInput)
```
kwokcb, line 16: This can probably be moved to a new method `getValueString(`) on `ShaderInput`.

source/MaterialXGenShader/ShaderNode.cpp
```diff
@@ -32,6 +32,11 @@ string ShaderPort::getFullName() const
     return (_node->getName() + "_" + _name);
 }
 
+string ShaderPort::getValueString() const
```
kwokcb, line 35: Sorry to keep hassling you on niggly things :), 
- For completeness, this can be exposed by adding an entry in the Python wrapper in `PyShader.cpp` in the `PyMaterialXGenShader` module.
- Also I think you can return a const reference to a string in this method.

madmann91: I'll add a python wrapper, but I don't think this can return a const reference, given that `Value::getValueString()` returns a `string` by value.

kwokcb (approved):

jstone-lucasfilm (approved):
Great work, @madmann91, and this is a very elegant solution to the posted issue!

PR #1561, opened 2023-10-12
Command hooks for Graph Editor

kwokcb:
## Issue

Update #1563

## Proposed Change

Add new arguments to specify additional menu items, and commands to run.

- A new `--outputCommand` option allows for a command to run on the current document on disk, or a string version of the document.
- A new `--editCommand` option allows a command to be run on the current file and then reloads the file. 
- For both options, the string argument has a `,` which separates the menu item string name from command string.

- `%F` means to save out the current doc.
- `%f` means to use the current doc name (no write)
- `%s` means to get the document as a string
- `%S` means to strip out carriage returns, linefeed chars and get document as string.
- `%r` means to get the current renderable element.

## Testing
- Below is an example which opens Visual Studio Code on the current file, or allows a custom Python script to run.
  ```
  MaterialXGraphEditor--outputCommand "Open In Visual Studio,code %F" --editCommand "Test Python Command,python 
  testEdit.py %F"
  ```

  https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/ce5301b5-a371-453c-b84c-1260e6d5dbab

- Register to generate shader code for currently selected renderable node (`%r`). OSL selected in this case.
```
 MaterialXGraphEditor --outputCommand "Generate Code,python generateShader.py --target osl %F --shaderPath %r"
 ```


jstone-lucasfilm:
Following the discussion in https://github.com/AcademySoftwareFoundation/MaterialX/issues/1563, let's hold off on this PR for now, and we can revisit this as we learn more about what types of customization the community would find valuable.

PR #1562, opened 2023-10-12
Restructure Add Node List

### LINKED ISSUE #1541: Graph Editor: Add Node popup restructure
The add node popup can be confusing for users. To combat this the order of the categories should be changed to better match  specification documents, e.g. Source Nodes (Texture, procedural, geometric, etc.) first, then Operator Nodes (math, adjustment, compositing, conditional, channel, etc.), then organizational and nodegraph-like nodes, then Shader BSDF Nodes, Shader EDF Nodes, etc. followed by Material nodes, etc. Also instead of having multiple nodes of each type for add etc have a second sub group of just add and then have all of the types of add within that.

beersandrew:
Fixed #1541
- added a list of nodes to the best of my ability to be the order they should be displayed


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3566,97 +3566,133 @@ void Graph::addNodePopup(bool cursor)
         // Input string length
         // Filter extra nodes - includes inputs, outputs, groups, and node graphs
         const std::string NODEGRAPH_ENTRY = "Node Graph";
-        for (std::unordered_map<std::string, std::vector<std::vector<std::string>>>::iterator it = _extraNodes.begin(); it != _extraNodes.end(); ++it)
+
+        std::vector<std::string> ordered_keys = {
```
beersandrew, line 3570: 1. is this a reasonable way to approach this? is this annoying to update as new groups are added? how often are groups added?

2. I realize this probably isn't the best place for this list to live, is there a recommended place to keep a list like this?

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3566,97 +3566,133 @@ void Graph::addNodePopup(bool cursor)
...
-        // Filter nodedefs and add to menu if matches filter
-        for (std::unordered_map<std::string, std::vector<mx::NodeDefPtr>>::iterator it = _nodesToAdd.begin(); it != _nodesToAdd.end(); ++it)
-        {
-            // Filter out list of nodes
-            if (subs.size() > 0)
+            else
             {
-                ImGui::SetNextWindowSizeConstraints(ImVec2(250.0f, 300.0f), ImVec2(-1.0f, 500.0f));
-                for (size_t i = 0; i < it->second.size(); i++)
+                auto it = _extraNodes.find(key);
```
beersandrew, line 3646: I didn't like reusing `auto it` here, looking for some advice on a better structure, changing the name is also possible but felt kind of sloppy/lazy to me

source/MaterialXGraphEditor/Graph.cpp
jstone-lucasfilm: @beersandrew This choice seems fine to me, and further in the future we can move this string vector into a global constant that's accessible through the MaterialX API.  Often new MaterialX features are developed in this way, where initially they're local data or functions in an example application, and they're moved into the core API as a second step.

source/MaterialXGraphEditor/Graph.cpp
jstone-lucasfilm: Since the original `it` variable is known to be invalid in this code path, perhaps you can just reuse the variable itself, and omit the `auto` keyword from this line?

beersandrew: I wanted to do that originally, but the types weren't the same so it didn't allow it, I've refactored the code to consolidate _nodesToAdd and _extraNodes so I think it's better now

source/MaterialXGraphEditor/Graph.cpp
beersandrew: Awesome, thanks for the note, I did move it up to a higher function per @bernardkwok's suggestion on Slack, still might not be the end location but I think it's better than where I had it in the first place

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1224,23 +1221,59 @@ void Graph::setUiNodeInfo(UiNodePtr node, const std::string& type, const std::st
 void Graph::createNodeUIList(mx::DocumentPtr doc)
 {
     _nodesToAdd.clear();
-    const std::string EXTRA_GROUP_NAME = "extra";
-    for (mx::NodeDefPtr nodeDef : doc->getNodeDefs())
+    const std::string EXTRA_NODES = "EXTRA_NODES";
+    std::vector<std::string> ordered_groups = {
```
kwokcb, line 1225: Hi @beersandrew ,
I think it's still more robust to dynamically build this list, since a scan for all of the nodedefs is done.
This allows nodegroups to be added / modified in the standard or 3rd party libraries overtime without having to rebuild this code.

For example, the Houdini and Maya nodes used their own `nodegroup` identifiers so it would work if you import the definitions from those libraries.

beersandrew: I agree, however I was trying to follow the description of the issue (https://github.com/AcademySoftwareFoundation/MaterialX/issues/1541) 

>  To combat this the order of the categories should be changed to better match specification documents, e.g. Source Nodes (Texture, procedural, geometric, etc.) first, then Operator Nodes (math, adjustment, compositing, conditional, channel, etc.), then organizational and nodegraph-like nodes, then Shader BSDF Nodes, Shader EDF Nodes, etc. followed by Material nodes, etc.

@lfl-eholthouser does alphabetically make sense here to you too? Does it make sense to order by some specified list of groups and include additional groups last? 

Happy to do anything here, just trying to balance the requests

jstone-lucasfilm: @kwokcb Although alphabetical order is an alternative worth considering, I like the "specification order" that was originally proposed for this task.

@lfl-eholthouser Does that sound right to you as well?

lfl-eholthouser: I agree, I think aligning with the specification would be best, and @beersandrew I think additional groups last makes sense.

lfl-eholthouser (commented):
I think it looks great to me, thanks @beersandrew !

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1224,20 +1221,57 @@ void Graph::setUiNodeInfo(UiNodePtr node, const std::string& type, const std::st
...
+    auto nodeDefs = doc->getNodeDefs();
+    std::unordered_map<std::string, std::vector<mx::NodeDefPtr>> groupToNodeDef;
+
+    for (const auto& nodeDef : nodeDefs)
     {
-        // NodeDef is the key for the map
         std::string group = nodeDef->getNodeGroup();
-        if (group.empty())
+
+        if (groupToNodeDef.find(group) == groupToNodeDef.end())
```
kwokcb, line 1256: Want to check here, if there is no node group (empty string), what will occur ? Is it better to create some string name for this group ? (While this was cleaned up in the std library, there is no requirement to have a nodegroup and I've seen in DCCs where it's not set).

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3566,96 +3600,66 @@ void Graph::addNodePopup(bool cursor)
...
             {
                 ImGui::SetNextWindowSizeConstraints(ImVec2(100, 10), ImVec2(-1, 300));
-                if (ImGui::BeginMenu(it->first.c_str()))
+                if (ImGui::BeginMenu(node[3].c_str()))
                 {
                     ImGui::SetWindowFontScale(_fontScale);
-                    for (size_t j = 0; j < it->second.size(); j++)
+                    std::string name = node[0];
+                    std::string prefix = "ND_";
+                    if (name.compare(0, prefix.size(), prefix) == 0 && name.compare(prefix.size(), std::string::npos, node[2]) == 0)
```
kwokcb, line 3641: Sorry. I'm not clear what this does ? Just checking as not all definitions start with "ND_".

source/MaterialXGraphEditor/Graph.h
```diff
@@ -246,8 +246,7 @@ class Graph
     std::vector<std::string> _currGraphName;
 
     // for adding new nodes
-    std::unordered_map<std::string, std::vector<mx::NodeDefPtr>> _nodesToAdd;
-    std::unordered_map<std::string, std::vector<std::vector<std::string>>> _extraNodes;
+    std::vector<std::vector<std::string>> _nodesToAdd;
```
kwokcb, line 249: A niggly thing. This could be a vec of `StringVec` or could add a new new type `StringVecVec` :). 
?

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3566,96 +3600,66 @@ void Graph::addNodePopup(bool cursor)
...
-                        std::string name = it->second[i]->getName();
-                        std::string val = getUserNodeDefName(name);
-                        if (ImGui::MenuItem(val.c_str()) || (ImGui::IsItemFocused() && ImGui::IsKeyPressedMap(ImGuiKey_Enter)))
+                        if (ImGui::BeginMenu(node[2].c_str()))
                         {
-                            addNode(it->second[i]->getNodeString(), val, it->second[i]->getType());
-                            _addNewNode = true;
+                            if (ImGui::MenuItem(getUserNodeDefName(name).c_str()) || (ImGui::IsItemFocused() && ImGui::IsKeyPressedMap(ImGuiKey_Enter)))
+                            {
+                                addNode(node[2], getUserNodeDefName(name), node[1]);
```
kwokcb, line 3655: The magic numbers to access array elements in the code is a bit hard to read. Maybe some named constants would help readability e.g. what does it mean to access position 2 in the array

source/MaterialXGraphEditor/Graph.cpp
beersandrew: This is an attempt to codify the 'rule' for nesting menus from this slack comment (https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1697132570854279?thread_ts=1697130472.487569&cid=C0230LWBE2X). In looking at the point_light node, the actual node name is `ND_point_light` but the category (or getNodeString()) returns `point_light`. To me the rule then is:

Wherever name == category (without the ND_ prefix) we just show that node in the 2nd layer, otherwise we add the category as a menu item, and put the name as the node in the 3rd layer.

source/MaterialXGraphEditor/Graph.cpp
beersandrew: Would it be reasonable to create a new class here? coming from C# it pained me to write this code with all this weird array indexing and sort of faking an object. It seems like creating a `MenuItem` class with properties of name, type, group, and category would clean up both the vec<vec<string>> into a vec<MenuItem> and this weird indexing. 

not sure if there's an obvious place for that to live or if there are any gotchas / reasons to not do this

source/MaterialXGraphEditor/Graph.cpp
beersandrew: Based on my understanding, these would not get added to the menu at all. To me that seems like a bug. @lfl-eholthouser would it make sense to put all nodes that are not in the hard coded list of groups at the bottom of the menu?

source/MaterialXGraphEditor/Graph.cpp
jstone-lucasfilm: That sounds like the right approach to me, @beersandrew, adding a `MenuItem` class that's local to the Graph Editor (e.g. in Graph.cpp or Graph.h), and leveraging this instead of a standard vector.

source/MaterialXGraphEditor/Graph.cpp
lfl-eholthouser: Yes, I would be interested in knowing when nodedefs would not have a NodeGroup. But I think that would definitely make sense maybe in something like a Misc group? Just so as not to greatly increase the size of the main menu

beersandrew: I've added `no_group` as an option for those that do not have a group

source/MaterialXGraphEditor/Graph.h
beersandrew: I've added a class `MenuItem` to clean this up

source/MaterialXGraphEditor/Graph.cpp
beersandrew: Added the `MenuItem` class to clean this up

source/MaterialXGraphEditor/Graph.cpp
kwokcb: Thanks @beersandrew. I leave it up to others on the name of "no_group".  Maybe "Ungrouped" instead as it shows up in the UI. Also you might want to take care in case someone calls their group "no_group" :).

source/MaterialXGraphEditor/Graph.h
```diff
@@ -14,6 +14,30 @@
 
 #include <stack>
 
+class MenuItem
```
kwokcb, line 17: Thanks for adding this in. Much easier to read the code.

kwokcb (approved):
Looks good to me. Thanks.

(Beyond the scope of this PR, but a note for @jstone-lucasfilm. As discussed in a earlier TSC meeting, it would be good to put define "nodegroup" names in the library so it's formalized. Then you can get rid of the hard-coded list here :) ).

jstone-lucasfilm (approved):
Great work, thanks @beersandrew, and thanks to @kwokcb and @lfl-eholthouser for their review!

This improvement points out future work that would be valuable for the MaterialX project, e.g. clarifying the rules that we use to categorize standard nodes.

Visualized accurately in the new interface, some of our current categorizations seem questionable (e.g. the `procedural` category for `constant` nodes), and I think this increased visibility for node categories should help us greatly in refining these choices going forward.

beersandrew:
This was a blast and exactly what I was hoping to get out of Dev Days, thanks all for the help! :pray:

PR #1565, opened 2023-10-12
Add floating popup to pins when hovered

### LINKED ISSUE #1543: Graph Editor: Add floating popup to pins when hovered
To improve usability and remove uncertainty regarding UiNode pins without needed to always look at the node property, a floating window or static info pane indicating the Type of the input/output, and either its value (if set, or the default value if not) or "Connected to out" if it's connected when hovering over node input or node ouput would be extremely useful.

ethomare-ilm:
Added a tooltip/popup showing the name, type, value and connection of a pin!
Screenshot 2023-10-12 at 6 06 31 PM
Screenshot 2023-10-12 at 6 06 26 PM
Screenshot 2023-10-12 at 6 06 36 PM

Fixes #1543


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3707,6 +3707,39 @@ void Graph::searchNodePopup(bool cursor)
...
+void Graph::addPinPopup() {
+    // Add a floating popup to pin when hovered
+    if (isPinHovered()) {
+        ed::Suspend();
+        UiPinPtr pin = getPin(ed::GetHoveredPin());
+        std::string connected = "";
+        std::string value = "";
+        if (pin->_connected) {
+            connected = "\nConnected to";
+            if (pin->_name == "out") {
```
kwokcb, line 3725: I think this may miss some outputs as not all output pins are called "out". e.g. UsdUvTexture, the outputs are called "r", "g" , "b" , "a" , "rgb" and "rgba". 

Also the else case, you can't assume that the input pin is connected to an output pin called "out".

ethomare-ilm: Okay, I added some connections back and removed that if statement so now it works for both input and output pins

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3707,6 +3710,40 @@ void Graph::searchNodePopup(bool cursor)
...
+        ed::Suspend();
+        UiPinPtr pin = getPin(ed::GetHoveredPin());
+        std::string connected = "";
+        std::string value = "";
+        if (pin->_connected)
+        {
+            connected = "\nConnected to";
+            for (UiPinPtr connectedPins : pin->getConnections())
+            {
+                connected = connected + " " + connectedPins->_name + ",";
```
jstone-lucasfilm, line 3734: @ethomare-ilm I'd recommend modifying this logic so that the comma is only added _between_ entries in the vector, and not after the last entry.

jstone-lucasfilm (commented):
This looks good to me, thanks @ethomare-ilm, and I had just one suggestion for an improvement to string formatting.

jstone-lucasfilm (approved):
This looks great to me, thanks @ethomare-ilm!

PR #1566, opened 2023-10-12
Update Catch2 to version 2.13.10

ashwinbhat:
- Update Catch2 version to v2.13.10
- Catch2 v2.13.0 supports BENCHMARK tests.
- A new BENCHMARK test "GenShader: ShaderGen Performance" to benchmark document load, validation and shadergen times


jstone-lucasfilm:
This looks very promising, thanks @ashwinbhat!

Looking at the build times, though, I see that our Dynamic Analysis tests take quite a while to complete the new benchmark step.

Would it be possible to exclude the benchmark step from Dynamic Analysis builds, e.g. by setting a preprocessor flag when MATERIALX_DYNAMIC_ANALYSIS is enabled in a build, and checking for this before running this specific test?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/CMakeLists.txt#L56C8-L56C34

jstone-lucasfilm:
Thanks for that update, @ashwinbhat, though I'm not sure this is having the desired effect!

Looking at the builds in GitHub Actions, most are taking 25-30 minutes, but the two most recent builds from PR 1566 are taking more than 40 minutes, with the Dynamic Analysis build taking an unusually long time to complete (31 minutes, as compared to 8 minutes for other builds).

One recommendation on the approach to controlling which tests run in Dynamic Analysis builds:

What if we were to set a preprocessor flag named `MATERIALX_DYNAMIC_ANALYSIS` when the user has selected the option of the same name through CMake, and then simply wrap benchmark tests with a check for this preprocessor flag?  That would make it more straightforward to change this behavior in the future, extending it to new tests or removing it from other tests, without needing to synchronize test names in CMake.

Here's an example of this pattern for setting preprocessor flags matching user options in the root CMake file:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/CMakeLists.txt#L168

ashwinbhat:
Hi @jstone-lucasfilm I think I've updated it to follow you suggestion where PerformanceTest: ShaderGen and Validation is run only with MATERIALX_DYNAMIC_ANALYSIS = ON. I see that the last DYNAMIC_ANALYSIS build took 51m 19s. Should we turn it off completely if it's going to hurt PR time and developer productivity

jstone-lucasfilm:
@ashwinbhat Oh, I may have miscommunicated my thinking here, and it's only when MATERIALX_DYNAMIC_ANALYSIS is **off** that we want to run these new benchmarking tests.

In a dynamic analysis build, the compiler inserts additional error-checking instructions into the generated code, allowing us to enforce that conditions such as memory boundary overwrites never occur in our unit tests.

But because these error-checking instructions dramatically slow down runtime performance, we want to omit our new benchmark tests when this mode has been enabled in the compiler, and we can check for this condition using the MATERIALX_DYNAMIC_ANALYSIS preprocessor flag that you've added.

source/MaterialXTest/CMakeLists.txt
```diff
@@ -121,6 +121,9 @@ set_target_properties(
     VERSION "${MATERIALX_LIBRARY_VERSION}"
     SOVERSION "${MATERIALX_MAJOR_VERSION}")
 
+# Enable Catch benchmarking
+target_compile_definitions(MaterialXTest PRIVATE -DCATCH_CONFIG_ENABLE_BENCHMARKING)
```
jstone-lucasfilm, line 125: Shouldn't this step be contingent upon `MATERIALX_BUILD_BENCHMARK_TESTS` being ON, so that we don't increase compile times for standard builds?

CMakeLists.txt
```diff
@@ -177,6 +179,12 @@ endif()
 if(MATERIALX_TEST_RENDER)
     add_definitions(-DMATERIALX_TEST_RENDER)
 endif()
+if (MATERIALX_DYNAMIC_ANALYSIS)
+    add_definitions(-DMATERIALX_DYNAMIC_ANALYSIS)
```
jstone-lucasfilm, line 183: Perhaps we should revert the addition of this preprocessor flag, since it's no longer used in the code?

jstone-lucasfilm (commented):
This looks very close to the mark, and I had just a couple of suggestions for improvements.

jstone-lucasfilm (approved):

PR #1569, opened 2023-10-13
Update link connections by dragging between ports

erikaharrison-adsk:
Currently, one must explicitly delete an existing link before adding a new link for a given node. This change enables dragging to update a link connection, and automatically deletes the existing link on the input pin if one exists. 

Note: An output pin on a node may have many outgoing links, but an input pin may only have one incoming link. This behaviour is unchanged.

Example video: https://github.com/AcademySoftwareFoundation/MaterialX/assets/85182970/7a825225-09f7-4936-abb9-f294978ebaaf

Also:
* Renames a few variables for consistency on start/end and input/output pins
* Ensures you can't add a link from an input to an input, or from an output to an output
* Can also drag from an output to an input and have it correctly delete existing link

Original issue/sub-issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/1398#issuecomment-1680198084


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2483,12 +2483,22 @@ void Graph::setDefaults(mx::InputPtr input)
     }
 }
 
-void Graph::addLink(ed::PinId inputPinId, ed::PinId outputPinId)
+void Graph::addLink(ed::PinId startPinId, ed::PinId endPinId)
```
erikaharrison-adsk, line 2486: I found input vs. output, start vs. end a bit confusing. Have tried to make this more accurate by checking the input pin is actually an input pin.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2505,187 +2515,206 @@ void Graph::addLink(ed::PinId inputPinId, ed::PinId outputPinId)
         return;
     }
 
-    if (inputPin->_connected == false)
+    // Perform kind check
+    bool kindsMatch = (outputPin->_kind == inputPin->_kind);
```
erikaharrison-adsk, line 2519: Added check that the output and input pins are not the same kinds (eg. not both Output and not both Input).

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2505,187 +2515,206 @@ void Graph::addLink(ed::PinId inputPinId, ed::PinId outputPinId)
...
+    if (kindsMatch)
     {
-        int upNode = getNodeId(inputPinId);
```
erikaharrison-adsk, line 2510: Aside: This diff looked much better on my computer. Sorry its come across this way.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2505,187 +2515,206 @@ void Graph::addLink(ed::PinId inputPinId, ed::PinId outputPinId)
...
+            showLabel("Invalid Connection: Node does not have an implementation", ImColor(50, 50, 50, 255));
             return;
         }
-        else if (uiUpNode->getNode())
+    }
+
+    if (ed::AcceptNewItem())
+    {
+        // If the accepting node already has a link, remove it
+        if (inputPin->_connected)
```
erikaharrison-adsk, line 2569: This is the crux of the new work - if we do have a connection for our input pin, we remove the existing link associated with it.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2505,187 +2515,206 @@ void Graph::addLink(ed::PinId inputPinId, ed::PinId outputPinId)
         return;
     }
 
-    if (inputPin->_connected == false)
```
erikaharrison-adsk, line 2508: We no longer skip if the input pin has a connection. Note: This resulted in the rest of the work being tabbed back (and likely caused the resulting ugly diff - it looked better on my computer, sorry).

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4045,12 +4074,12 @@ void Graph::drawGraph(ImVec2 mousePos)
         // Add new link
         if (ed::BeginCreate())
         {
-            ed::PinId inputPinId, outputPinId, filterPinId;
```
erikaharrison-adsk, line 4048: Also tweaked the terminology here. imgui's node editor simply returns a start/end -- could be output to input or input to output. Changed the terminology here to reflect that.

lfl-eholthouser (approved):
Thanks @erikaharrison-adsk ! This looks great, looking forward to trying it out.

source/MaterialXGraphEditor/Graph.cpp
lfl-eholthouser: Thanks for making that change! It's a lot clearer now.

jstone-lucasfilm (approved):
Great work on this improvement, thanks @erikaharrison-adsk!

Issue #1471, opened 2023-08-21
Graph Editor is creating imgui.ini file


marwie:
Steps to reproduce:

1) Open the editor
2) Save a material at an empty folder on your system
3) The directory now contains one file (your material)
4) Close the editor
5) Notice that the folder now contains a file named `imgui.ini`

Windows 10

Also note: the imgui.ini file contains information about position and size of the window but it doesnt seem to be used at all

meshula:
Dear Imgui itself reads the ini file. When new windows and panels are created it uses that information to restore the state and panel layout of the most recent previous user session. Deleting the file restores the window layout to the default arrangement indicated in the viewer's source code.

marwie:
The file seems to never be modified nor does changing the values in it affect the graph view when i restart / open the graph editor again

meshula:
Yes, that is definitely the default behavior of Dear ImGui. There is a discussion about it here, https://github.com/ocornut/imgui/issues/4294 including various solutions. Perhaps the thing to do is to get some consensus on the behavior. For example, it might make sense to always store the ini file in a platform specific AppData directory.

This might be a good topic to bring up in the Academy Software Foundation Slack, in the #materialx channel. Perhaps the collective wisdom there might help come to a consensus about a good behavior.

jstone-lucasfilm:
My initial thought is that we should _disable_ the automatic writing/reading of INI files in the MaterialX Graph Editor, since it's not capturing useful information about the user's session state.

As an example, here's the output INI file in a recent session, where it's only capturing the state of the default debug window, and not the state of active user windows:

```
[Window][Debug##Default]
Pos=60,60
Size=400,400
Collapsed=0
```

From the discussion thread that @meshula links above, here are instructions on how we can potentially disable this in the MaterialX Graph Editor:

https://github.com/ocornut/imgui/issues/4294#issuecomment-874720489

dgovil:
A related PR is https://github.com/AcademySoftwareFoundation/MaterialX/pull/1257

I'd set up the pathing to use default config locations on posix systems. I'd left windows alone for lack of a system to test the conventions. It would be good if someone could make windows have parity.

kwokcb:
To disable, it looks like you can just null the filenames as [mentioned here ](https://github.com/ocornut/imgui/issues/5169), but you need a reference, which is what's done in Main.cpp. So we'd need these lines:
```
io.IniFilename = NULL;
io.LogFilename = NULL;
```
[...]

PR #1570, opened 2023-10-13
Stop Graph Editor from creating imgui.ini file

iukpo-lucasfilm:
Added code to stop creation of imgui.ini after saving material after saving then closing Graph Editor (#1471)


jstone-lucasfilm (commented):
This looks good to me, thanks @iukpo-lucasfilm!

Although we may want to revisit the notion of configuration INI files in future iterations, I would recommend that we remove the `getConfigPath` helper function from Main.cpp as part of your change, since it was added specifically for the graph editor INI use case.

The function will always be available in the GitHub history of the MaterialX project, and we can bring it back if configuration INI files become valuable in the future.

iukpo-lucasfilm:
@jstone-lucasfilm Thank you! I will remove `getConfigPath`, as requested.

jstone-lucasfilm (approved):
This looks ready to merge, thanks @iukpo-lucasfilm!

Issue #1556, opened 2023-10-08
Create MTLX file from textures in a directory


Cinifreak:
I made an example for that feature a while ago. @jstone-lucasfilm and @bernardkwok found it interesting. So, I can contribute it as joining Dev-Days.

The idea is to have a python script to create a MTLX NodeGraph from textures in directory with educated guess according to the best shading model match to that textures.

By default, we will have a relative paths for textures with optional flags to make it absolute.
Also, there will be another optional flag to force it to use a specific shading model configuration. if not, the script will make it's educated guess for the best match.

jstone-lucasfilm:
Thanks for this original proposal, @Cinifreak, and this is now addressed in #1746.

PR #1572, opened 2023-10-14
Create material from texture

Cinifreak:
This PR is an the starting phase of that issue `https://github.com/AcademySoftwareFoundation/MaterialX/issues/1556` we created during the Dev-Days.

me and @michaelerdaa collaborate for that PR to create kind of a smart python script to create `mtlx` files from scanning a desired directory that has textures relaying on a `json` file that has a configuration for file extensions and texture types.

This is just the initial run, It will be updated during this weekend multiple times to fulfil the `TODO` list. So, any comments or ideas to be added to this PR would be appreciated.


kwokcb:
Hi @Cinifreak ,
This looks great so far!

A couple suggestions for consideration:

- It's probably more robust to use `<child> = <parent>.createValueChildName(<child name>)` to get back a unique child name. It probably won't occur but safer than just using the `<name>` when creating new nodes. You should call `<child>.getName()` to get the unique name back. Could be useful to avoid issues if you allow for a set of image folders to be loaded into a single document, or allow adding to an existing document.

- As part of the config, it would be nice to consider the inputs for the "unlit" shader available in MaterialX as part of the default regex checks. e.g. testing for 'emission' would have different mappings for the "unlit" shader (which has no "base color").

- it would be nice to be able to define what texture/image type to use. e.g. I'd like to use `tileimage` vs image, and if using a to `UsdUVTexture` as the input image you don't require a `normalmap`.

---
And some other misc items:

- I'm not sure if you want to use the MaterialX FilePath/FileSearchPath utilities to do file name / path manipulation.

- I'm wondering if it's useful to add the UDIM parsing code back to MaterialX as a utility. Think it would be useful and could reside with the existing UDIM helpers. Another application would be allow to set a filename on an image node when a user selects a file to use as the input. The UDIM pattern can be detected as the `<UDIM>` token added in. (DId this for Maya way back :) )

- I like the usage of JSON and as it's a script easy to get support for and if written in Typescript would be web friendly. I like that it can be extended as well. Do you foresee have a config file per DCC / library convention? For instance. I'd like to try this out with Substance - so would that be a separate config file ?

cary-ilm:
Looks like there might be a problem with the CLA, could you kindly hold off on merging this for a bit? It should be easy to straighten out, I'll follow up offline and report back shortly.

Issue #1558, opened 2023-10-10
MaterialX should allow specifying a material without a new materialX document


gareth-morgan-autodesk:
Currently the MaterialX API requires a unique MaterialX Document in order to specify a material.  This means at the very least an XML document must be parsed for every material in the scene, and most likely the generation of the MaterialX graph in memory, as well as, in many cases, the code generation.  For real life non-trivial scenes with 1000s of material this sets a upper bound on the performance if the materials in a scene are to be represented as MaterialX.

Real life scenes (such as the kind typically encountered in content creation tools and game engines) will frequently have 1000s of unique materials, but they will not have 1000s of unique material networks, typically they will have a small number (<10) actual unique material networks (e.g. different combinations of textures, such as diffuse texture, normal map, roughness map, etc, or different procedural networks), with the remaining 1000s of materials only varying by a small set of simple properties (e.g. base color RGB color vector, roughness float, normal map filename, etc) that are associated with the individual material while sharing the underlying material network.  

What is more, in ,most real life scenarios this information is part of the scene description, there is large set of materials, each associated with one of a small number of material "definitions" (meaning the shader or graph defining the material),  there is no need to infer this information at runtime.  The MaterialX API should have a away of specifying a collection of materials this way, with a single definition (represented by a MaterialX document) and multiple materials each represented by a flat list of material properties.  Each property should be a simple name-value pair, with the string name being a MaterialX path, and the value being a "plain old data" MaterialX value (float, int, vector, matrix, or file path)

jstone-lucasfilm:
@gareth-morgan-autodesk Just to clarify, are you looking for a different approach than the one found in the MaterialX Chess Set, where a set of related materials are defined within a single MaterialX document?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/Examples/StandardSurface/standard_surface_chess_set.mtlx

This approach should also extend to materials that bind uniform values directly to the inputs of a shading model, and there's no requirement that materials reference node graphs in MaterialX:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/Examples/StandardSurface/standard_surface_glass_tinted.mtlx

gareth-morgan-autodesk:
[...]

PR #1573, opened 2023-10-17
DRAFT: Add the concept of an Override to the MaterialX C++ API

gareth-morgan-autodesk:
As discussed in this [Jira](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1558), the current MaterialX API has no means of specifying a MaterialX material without a specifying a new MaterialX network graph (which must be parsed, code generated and compiled separately).  There is no way to specify that a material is set a of data-only changes and does not require a new material network, which is the common case in most real world scenes (where you have a handful of actually distinct material networks, and a large number of materials that only vary by a few vector or scalar data items.).  This an important omission from the API and which as well as losing important information when a scene's materials are represented with the MaterialX API, has huge performance implications due to the extra parsing, code generation and compilation entailed (see [this comment](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1558#issuecomment-1763249574) for a detailed breakdown of the performance implications of this)

This PR introduces the concept of an `Override` to the MaterialXCore API.  This is a small addition to the API, the `Override` class is constructed with an existing MaterialX document and flat list of named properties (each being the string path to an input in the original document):
```
			override = mx::Override::create(doc, { "SimpleMaterial/base_color" });
```

The properties within the `Override` can then be changed individually by passing a valid MaterialX value type to the `setValue` method:
``` 
		override->setValue("SimpleMaterial/base_color", mx::Color3(r, g, b));
```

In this way the `Override` class adds the ability to specify a material (that is still a well defined MaterialX material, with a standardized appearance just as if it was specified with a MaterialX document directly) that is just a set of "plain old data" overrides from a pre-existing network, without forcing the renderer to parse and code generate a new material network unnecessarily.  The renderer does not have to treat an override material differently, it can just create a new material network for each one, and the results will be identical, but huge performance gains can be achieved by taking advantage of the extra information provided by the Override API.

To demonstrate this, and the massive performance improvement this allows, the PR also contains a proof-of-concept implementation of using `Override` in the `MaterialXRenderGlsl` library.  This adds a `setOverride` method to `ShaderMaterial` class allowing multiple materials to share the same document, element and node, but vary with different overrides:
```
	for (int i = 0; i < numMaterials; i++) {
		        mx::MaterialPtr pMtl = mx::GlslMaterial::create();
			pMtl->setDocument(doc);
			pMtl->setElement(typedElem);
			pMtl->setMaterialNode(materialNode);
			pMtl->setOverride(overrides[i]);
     }
```
The POC implementation adds a new bind method which `bindUniformOverrides` will bind the override properties to the current render state.:
```
				pMtl->bindViewInformation(viewCamera);
				pMtl->bindLighting(lightHandler, imageHandler, shadowState);
				pMtl->bindShader();
				pMtl->bindImages(imageHandler, { "" });
				pMtl->bindUniformOverrides();
   
```

These two Gists show the difference between having to create a [new material network for each material](https://gist.github.com/gareth-morgan-autodesk/704364c359a7a0c7ee9280feebd0b24e), and [using material overrides](https://gist.github.com/gareth-morgan-autodesk/74362c042afbc76fe49ca26b8c4e4408).

### DIFF
--- a/source/MaterialXCore/Override.h
+++ b/source/MaterialXCore/Override.h
@@ -30,6 +30,7 @@ class MX_CORE_API Override
         std::shared_ptr<Document> doc,
         const vector<string>& properties,
         const vector<ValuePtr>& values = {});
+    Override(const Override& other);
     ~Override() { }
 
     /// Create an override for this document with a list of properties.
@@ -45,6 +46,12 @@ class MX_CORE_API Override
         return std::make_shared<Override>(doc, properties, values);
     }
 
+    /// Create override from an existing override object.
+    /// @param other An existing Override to copy.
+    static OverridePtr copy(OverridePtr other)
+    {
+        return std::make_shared<Override>(*other);
+    }
 
     /// Set the value of the named override property.
     /// @param name The name of the property to set.
--- a/source/MaterialXCore/Override.h
+++ b/source/MaterialXCore/Override.h
@@ -1,0 +1,68 @@
+//
+// Copyright Contributors to the MaterialX Project
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#ifndef MATERIALX_OVERRIDE_H
+#define MATERIALX_OVERRIDE_H
+
+/// @file
+/// Geometric element subclasses
+
+#include <MaterialXCore/Export.h>
+#include <MaterialXCore/Value.h>
+
+MATERIALX_NAMESPACE_BEGIN
+
+class Document;
+class Element;
+class Input;
+class Value;
+
+using OverridePtr = shared_ptr<class Override>;
+
+/// @class Override
+/// .
+class MX_CORE_API Override
+{
+  public:
+    Override(
+        std::shared_ptr<Document> doc,
+        const vector<string>& properties,
+        const vector<ValuePtr>& values = {});
+    ~Override() { }
+
+    static OverridePtr create(
+        std::shared_ptr<Document> doc,
+        const vector<string>& properties,
+        const vector<ValuePtr>& values = {})
+    {
+        return std::make_shared<Override>(doc, properties, values);
+    }
+
+    template<typename T>
+    void setValue(const string& name, const T& val) {
+        int idx = getIndex(name);
+        _values[idx] = Value::createValue(val);
+    }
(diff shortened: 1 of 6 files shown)


source/MaterialXCore/Override.h
```diff
@@ -0,0 +1,110 @@
...
+
+    /// Get the name of the Nth property being overriden.
+    /// @return The full path name of the input associated with the override property.
+    string getPropertyName(int n) { return _properties[n]; }
+
+    /// Get the number of properties in this override.
+    /// @return The property count.
+    int getPropertyCount() { return int(_properties.size()); }
+
+    // TODO: Add the ability to serialize overrides (to/from XML, JSON, etc.)
```
gareth-morgan-autodesk, line 98: As mentioned we should have serialize functions as with regular materialX documents, and the ability to produce a document from an override (for implementations that do not want to implement a different data path for override materials)

source/MaterialXRender/ShaderMaterial.h
```diff
@@ -34,6 +35,60 @@ class MX_RENDER_API ShadowState
     float ambientOcclusionGain = 0.0f;
 };
 
+/// @struct ShaderMaterialDefinition
```
gareth-morgan-autodesk, line 38: In this POC we add the concept of a ShaderMaterial definition and state, allowing multiple ShaderMaterial objects to share the same state (contain hwShader and glProgram)

source/MaterialXRenderGlsl/GlslMaterial.cpp
```diff
@@ -222,25 +234,42 @@ ImagePtr GlslMaterial::bindImage(const FilePath& filePath, const std::string& un
...
     return nullptr;
 }
 
+void GlslMaterial::bindUniformOverrides()
+{
+    if (!_override)
+        return;
+
+    // TODO: This is proof-of-concept code, in production this should probably use uniform blocks, or at least precompute the uniform locations.
+    for (int i = 0; i < _override->getPropertyCount(); i++) {
```
gareth-morgan-autodesk, line 250: In this POC we just bind each property as a uniform, as mentioned in comment we should use uniform blocks here.

PR #1576, opened 2023-10-18
Remove unused uniforms in HW generation for heighttonormal

nicolassavva-autodesk:
This PR addresses the undesirable shaderGen for Metal and GLSL targets when using the HeightToNormalNode.
The parent ConvolutionNode class emits a couple of filter kernel arrays, for box and gaussian weights, that may cause shader compiler issues when they remain unused.

An example diff of a dumped out Metal shader source using the `HeightToNormalNode` before (righthand side) and after (lefthand side):

image

### DIFF
--- a/source/MaterialXGenGlsl/Nodes/HeightToNormalNodeGlsl.cpp
+++ b/source/MaterialXGenGlsl/Nodes/HeightToNormalNodeGlsl.cpp
@@ -33,9 +33,9 @@ ShaderNodeImplPtr HeightToNormalNodeGlsl::create()
     return std::make_shared<HeightToNormalNodeGlsl>();
 }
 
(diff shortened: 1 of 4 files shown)


nicolassavva-autodesk:
@Morteeza @ashwinbhat @jstone-lucasfilm: This resolves `c_box_filter_weights` and `c_gaussian_filter_weights` we encountered when compiling shaders for some targets.
The proposed patch specifically addresses the previously observed
`error: field may not be qualified with an address space` under MacOS-Metal

jstone-lucasfilm:
Thanks for the quick fix, @nicolassavva-autodesk, and this approach looks good to me.

jstone-lucasfilm (approved):
Looks good to me, thanks @nicolassavva-autodesk!

Issue #1581, opened 2023-10-22
Crash in `MaterialXCore::Element::setChildIndex()` due to off-by-one error in index check


StefanHabel:
I believe I found an off-by-one crash bug in [`MaterialXCore::Element::setChildIndex()`](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Element.cpp#L162-L178).

Minimal repro via Python bindings:
```bash
$ python3 -c 'import MaterialX as mx; d = mx.createDocument(); b = d.addBackdrop(); d.setChildIndex(b.getName(), 1)'
Segmentation fault
```
Expected behavior:
```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
LookupError: Invalid child index
```
Fix:
```diff
-    if (index < 0 || index > (int) _childOrder.size())
+    if (index < 0 || index >= (int) _childOrder.size())
     {
         throw Exception("Invalid child index");
     }
```

Found while working on https://github.com/AcademySoftwareFoundation/MaterialX/issues/342

jstone-lucasfilm:
This looks like a good catch, thanks @StefanHabel, and feel free to post your fix if you have the bandwidth!

jstone-lucasfilm:
Thanks for this report, @StefanHabel, as well as for the fix in #1582!

StefanHabel:
Thanks very much @jstone-lucasfilm ! 🙌

PR #1582, opened 2023-10-22
Fixed an off-by-one index check in `Element::setChildIndex()`.

StefanHabel:
Extended the existing test case:
valid values for `index` in the test case are `0` and `1` only.

Minimal repro via Python bindings:
```bash
$ python3 -c 'import MaterialX as mx; d = mx.createDocument(); b = d.addBackdrop(); d.setChildIndex(b.getName(), 1)'
```
Without this patch:
```bash
Segmentation fault
```
With this patch:
```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
LookupError: Invalid child index
```

Ran the tests locally via `make && make test`:
```bash
100% tests passed, 0 tests failed out of 73
```

Update #1581

### DIFF
--- a/source/MaterialXCore/Element.cpp
+++ b/source/MaterialXCore/Element.cpp
@@ -168,7 +168,7 @@ void Element::setChildIndex(const string& name, int index)
         return;
     }
 
-    if (index < 0 || index > (int) _childOrder.size())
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm (approved):
Good catch, @StefanHabel, and thanks for the fix!

PR #1584, opened 2023-10-26
Fix MDL implementation of mx_hsvtorgb

### LINKED ISSUE #1583: MDL implementation of mx_hsvtorgb is incorrect
I've been checking out the Blender Hydra/MaterialX support, and believe to have stumbled upon a bug in the MDL implementation of the HSV to RGB color space transformation logic.

The following images show shading networks parametrized with different Hue values in (1) Blender's Cycles, (2)  Storm (via Hydra) using the MaterialX GLSL backend, (3) Gatling (via Hydra) using the MaterialX MDL backend and (4) Gatling, with the correct logic.
Cycles Storm GLSL
Gatling MDL (old) Gatling MDL (new)

Discrepancies between the MaterialX GLSL and MDL backends can only be seen for high hue values (0.9, 1.0) and stem from a differing line of code in their `mx_hsvtorgb` functions. For GLSL, the hue is first floored and then brought into the range 0-6:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/926ac2756c36f5541d44f6c3f8814bfebf228f3b/libraries/stdlib/genglsl/lib/mx_hsv.glsl#L47-L49
whereas for MDL, the value is first brought into the range 0-6 and is then floored:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/926ac2756c36f5541d44f6c3f8814bfebf228f3b/source/MaterialXGenMdl/mdl/materialx/hsv.mdl#L29-L31

I can't tell whether this is an implementation mistake or an error in the referenced literature (Reinhard et al.), as the book is not openly available. However, the GLSL logic also matches OpenColorIO:
https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/8add374dade5a7e3ee802269b71e1c0b75f8b8f0/src/OpenColorIO/ops/fixedfunction/FixedFunctionOpGPU.cpp#L413

cc @krohmerNV

pablode:
This fixes #1583.

### DIFF
--- a/source/MaterialXGenMdl/mdl/materialx/hsv.mdl
+++ b/source/MaterialXGenMdl/mdl/materialx/hsv.mdl
@@ -25,16 +25,16 @@ import ::limits::*;
 export float3 mx_hsvtorgb(float3 hsv) {
     // from "Color Imaging, Fundamentals and Applications", Reinhard et al., p. 442
 
-    // A hue of 1.0 is questionably valid, and needs to be interpreted as 0.0f
-    float h_prime = (hsv.x < 1.0f) ? hsv.x * 6.0f : 0.0f; // H * 360.0/60.0
-    float h_floor = math::floor(h_prime);
-    float f = h_prime - h_floor;
+    float h = 6.0 * (hsv.x - math::floor(hsv.x));
+    int hi = int(h); // truncate
+    float f = h - float(hi);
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm (approved):
This change looks good to me, thanks @pablode!

PR #1586, opened 2023-10-26
Add uiScale option to the Graph Editor

virokannas:
The branch has an additional command line variable for MaterialXGraphEditor, --uiScale, that will override the automatic calculation and application of DPI scaling.

Using this flag on MacOS with a retina display, will result in a normal-sized UI.


kwokcb:
Hi @simpassi ,
I think it would be better if the dpi detection was corrected instead to return the appropriate value for retina displays.
Would you have time to look at this? Thanks.

simpassi:
Sure, I can check what all would be involved, but the underlying mechanism causing this is that GLFW needs a special hint to enable full-resolution backing, and at that point DPI scaling will need to be handled per monitor. Also, the situation where a window is moved from one monitor to another needs to be caught and handled.

Right now as the hint isn't enabled, GLFW will just use a logical, not physical, resolution for everything and as a side effect the UI is correct size on any monitor you move it to, as the OS is handling the viewport scaling.

I'm sure there'll still be situations where disabling scaling is desired. Another option is to have a variable to set the scaling factor manually - some UI frameworks like QT make this possible through environment variables.

kwokcb:
If this get's too complicated (even just for the single monitor case) than I think you suggestion of allowing a scale factor override is a good alternative and more flexible than an on/off switch.

simpassi:
I changed it to take an optional --scaleFactor float argument. On a retina Mac, 1.0 will yield an acceptable result, but of course now it can be used to simply scale the UI up or down too.

simpassi:
@kwokcb forgot to tag you in the above message; let me know if there's anything you'd like changed in the merge req.

source/MaterialXGraphEditor/Main.cpp
```diff
@@ -100,6 +102,10 @@ int main(int argc, char* const argv[])
         {
             parseToken(nextToken, "string", captureFilename);
         }
+        else if (token == "--scaleFactor")
+        {
+            parseToken(nextToken, "float", scaleFactor);
```
kwokcb, line 107: The only firewall test I suggest is to make sure negative numbers aren't allowed.

kwokcb (approved):
Looks good to me. Sorry for the delay in review as I missed the notification.

kwokcb (approved):

source/MaterialXGraphEditor/Main.cpp
```diff
@@ -30,6 +30,7 @@ const std::string options =
     "    --path [FILEPATH]              Specify an additional data search path location (e.g. '/projects/MaterialX').  This absolute path will be queried when locating data libraries, XInclude references, and referenced images.\n"
     "    --library [FILEPATH]           Specify an additional data library folder (e.g. 'vendorlib', 'studiolib').  This relative path will be appended to each location in the data search path when loading data libraries.\n"
     "    --captureFilename [FILENAME]   Specify the filename to which the first rendered frame should be written\n"
+    "    --scaleFactor [FACTOR]         Manually specify a HiDPI scaling factor\n"
```
jstone-lucasfilm, line 33: The substance of this change looks good to me, @virokannas, and my one recommendation would be to rename this command-line option from `scaleFactor` to `dpiScale` for improved user clarity.

virokannas: I can do that @jstone-lucasfilm if you want - but wouldn't having the word "dpi" in the option name hint towards having to use a DPI value or a DPI-dependent value rather than a scaling factor? The actual variable being driven by this is a factor that is unaware of the screen DPI.

I'm sure at least some who would use this are also familiar with QT, which has the same setting labeled as QT_SCALE_FACTOR. Linux window managers call it a scaling factor, with varying ranges, Windows uses percentages, etc..

But if dpiScale fits better within the MaterialX ecosystem, I'll change it no problem.

jstone-lucasfilm: @virokannas That's a good point, and I'm open to better ideas for terminology here.  In the ImGui documentation, they provide the following definition of ImGuiStyle::ScaleAllSizes, which is the method that is influenced by this command-line option.

>// To scale your entire UI (e.g. if you want your app to use High DPI or generally be DPI aware) you may use this helper function. Scaling the fonts is done separately and is up to you.

Mainly, I'd like to clarify that this option affects UI/DPI scaling, and is not related to the scale of the viewport render.

virokannas: @jstone-lucasfilm Ah, of course.

How about "uiScale" or "uiScaleFactor"?

jstone-lucasfilm: Sure, let's go with `uiScale` to keep things simple.

virokannas:
@jstone-lucasfilm now it's changed to uiScale

jstone-lucasfilm (approved):
This looks good to me, thanks @virokannas!

Issue #1424, opened 2023-07-24
Graph Editor: Keyboard entry affects all panels causing unexpected behavior


kwokcb:
Currently there are hotkeys are being interpreted regardless of which panel is in focus. This causes unexpected especially in an field in a panel for which the action is unexpected. 

Examples:
- Pressing `u` key for entering text goes up a level unexpectedly
- Pressing `tab` key to go to next field brings up the 'add node` dialog.
- Pressing the `delete` key can end up deleting a node instead of a character in a text field.

This has also been noted by @zicher3d where these examples are paraphrased from.

Proposed Changes:
1. Only allow hotkeys meant to work in the node graph panel to apply there. The property editor is one example where node graph hotkeys are affecting text input.
2. Conversely allow property editor keys to not be overridden by hotkeys for the node graph panel. e.g. tabbing.
3. Fix the logic for exclusion from file-dialogs to exclude all dialogs as that those get affected as well.
4. Exclude the rendering panel as well to allow any current or future hotkeys to work there.

**On purpose the adverse affect of mouse interaction is excluded from here as there is already a separate issue.**

HudsonHN:
I'll take a look at this one too

jstone-lucasfilm:
Thanks for this original report, @kwokcb, and thanks to @HudsonHN for the fix in #1596.

PR #1596, opened 2023-11-07
Fix keyboard input focus in Graph Editor

HudsonHN:
Changed IsWindowFocused flags for inputs mentioned in #1424 to prevent all panels from responding.


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4135,16 +4136,19 @@ void Graph::drawGraph(ImVec2 mousePos)
             ed::NavigateToContent(0.0f);
         }
```
eytienne, line 4137: from my local experiments, I believe this small "centering" block could move above so you can get rid of the `shouldProcessInput` variable and wrap everything under the same `if`

eytienne (changes requested):

source/MaterialXGraphEditor/Graph.cpp
HudsonHN: Good catch! Changes pushed.

jstone-lucasfilm:
@HudsonHN It would be great if we can include this improvement in MaterialX 1.39, which is scheduled for release this summer.  Would you mind retargeting your pull request to the `dev_1.39` branch, so that we can consider your change in that context?

jstone-lucasfilm:
@HudsonHN We've now merged development work on MaterialX 1.39 back to the main branch of MaterialX, in preparation for wrapping up development on the 1.39.0 release. When you have a chance, could you retarget this pull request back to the main branch as well?

HudsonHN:
> @HudsonHN We've now merged development work on MaterialX 1.39 back to the main branch of MaterialX, in preparation for wrapping up development on the 1.39.0 release. When you have a chance, could you retarget this pull request back to the main branch as well?

Fumbled a bit but it's done 😄

jstone-lucasfilm (approved):
This looks good to me, thanks @HudsonHN!

PR #1597, opened 2023-11-07
Improve "Add Node" workflows in the graph editor

GregoryD2017:
## Context Aware "Add Node" Workflow
Previously, the tab menu was the only way to add new nodes to the graph editor. Now users can select any pin in the graph, and drag a link off of that pin. If this link does not connect to another pin, on mouse release, the user will be prompted to add a new node. The list of nodes that users can add will be filtered to only include nodes that can be connected to the pin from which this link originates.

This context-aware node filtering in the "Add Node" popup can be toggled on/off with the "Smart Filter" check box.

### Demo 🥳 

context-aware-add
_Notice how as you toggle the `Smart Filter` on and off, the list of allowable nodes changes as well. When `Smart Filter` is off, every  defined node is allowed to be added._

### One future improvement:
Right now, when dragging a link from a pin, nodes that have valid connections are shown in the "Add Node" popup. We could make this popup even more granular -- show both the node name and the specific pin of that node that would be connected by this link. Certain nodes have multiple input or output pins of the same type. By having separate entries for each of the specific pins as well, users can select which pin they want to be connected in the newly added node, and when that node is added, we can automatically draw that link for them.


GregoryD2017:
@jstone-lucasfilm Ach so sorry about that -- I wasn't getting email notifications for github since I haven't signed into this email account since college 😅 . I'll get the license and build errors fixed ASAP. Apologies!

GregoryD2017:
License + build errors fixed @jstone-lucasfilm 😃

jstone-lucasfilm:
@GregoryD2017 In order to keep this proposal moving forward, make sure to resolve the merge conflicts mentioned above!  This change still seems promising, but we'll need to keep it up-to-date with the main branch in order to provide meaningful review.

jstone-lucasfilm:
@GregoryD2017 If you're interested in continuing work on this proposal, we'll be starting Dev Days 2024 on September 25th!

jstone-lucasfilm:
I'll close out this pull request for now, @GregoryD2017, but feel free to pick up work on this in a future Dev Days event!

PR #1601, opened 2023-11-17
Fixed null check in Graph::propertyEditor

### LINKED ISSUE #1600: Crash in `Graph::propertyEditor` due to absence of null check
I reproduced with the following steps:

1. Open some simple mtlx containing a nodegraph.
2. Click on the nodegraph to select it then doubleclick to open it.
3. Click on a surfaceshader node.
4. Click on the name property editor and press `Backspace` to drop some chars. Immediate seg fault!

Using gdb, I find out the null pointer problem causing crash in some shared_ptr code. I have the incoming PR.

Found while experimenting with #1596.

eytienne:
Fixes #1600

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3726,7 +3726,6 @@ void Graph::addNodePopup(bool cursor)
                         }
                     }
 
-
                     ImGui::EndMenu();
                 }
             }
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3893,7 +3893,6 @@ void Graph::handleRenderViewInputs()
     {
         _renderer->setScrollEvent(scrollAmt);
     }
-
 }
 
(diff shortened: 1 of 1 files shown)


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3349,7 +3349,7 @@ void Graph::propertyEditor()
                         }
                         else
                         {
-                            nodes->getOutput()->setConnectedNode(_currUiNode->getNode());
+                            uiNode->getOutput()->setConnectedNode(_currUiNode->getNode());
```
eytienne, line 3352: Code smell: looking at the enclosing if sequence, `_currUiNode->getNode()` will always be null.

jstone-lucasfilm:
This looks very promising, thanks @eytienne!  I'm CC'ing @lfl-eholthouser for her thoughts on the proposed fix.

lfl-eholthouser (approved):

jstone-lucasfilm (approved):
Thanks for the fix, @eytienne!

PR #1603, opened 2023-11-22
Link to the OpenGL::GL target instead of OPENGL_LIBRARIES

pmolodo:
This prevents absolute paths from creeping into the output MaterialXConfig.cmake, which makes it not portable.

Also, in general in modern CMake, targets are preferred over baked paths.  Additionally, at least on Linux,
${OPENGL_LIBRARIES} includes libGLU, which is not actually a required dependency


pmolodo:
Fixed an issue with requiring OpenGL on iOS builds.  Refactored the way COMMON_LIBRARIES was used to make the iOS library changes clearer / grouped with other target_link_library statements

jstone-lucasfilm:
This looks like a good proposal, @pmolodo, and the one caveat is that we'll need to raise our minimum required version of CMake, which is currently set to 3.5:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/CMakeLists.txt#L8

We've been interested in raising this minimum version for quite a while, and one option would be to align with OpenImageIO and raise the minimum version of CMake to 3.15:

https://github.com/AcademySoftwareFoundation/OpenImageIO/blob/master/CMakeLists.txt#L5

Let's bring this up as a topic in the MaterialX channel of the ASWF Slack, so that we can gather consensus on what minimum version we should align with.

jstone-lucasfilm:
Following up on a discussion on the MaterialX channel of the ASWF Slack, we've raised the minimum CMake version to 3.16:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1607

pmolodo:
Great, thanks!

PR #1606, opened 2023-11-27
Link to X11 targets instead of CMake variables

pmolodo:
# NOTE: please review / merge #1603 before this, as this PR depends on it
*(If it is determined that #1603 is unacceptable for some reason, in a way that does NOT apply to this PR, I can rebase / refactor this MR to not rely on #1603)*

-----------------------------------

ie, ${X11_LIBRARIES} and ${X11_Xt_LIB}

This prevents absolute paths from creeping into the output
MaterialXConfig.cmake, which makes it not portable.

Also, in general in modern CMake, targets are preferred over baked
paths.


jstone-lucasfilm (approved):
This change looks good as well, thanks @pmolodo!

PR #1612, opened 2023-12-05
Apply Clang formatting

jstone-lucasfilm:
This changelist applies Clang formatting to source files in the MaterialX codebase, and includes no functional changes.

### DIFF
--- a/source/MaterialXCore/Definition.cpp
+++ b/source/MaterialXCore/Definition.cpp
@@ -84,7 +84,7 @@ InterfaceElementPtr NodeDef::getImplementation(const string& target) const
             }
         }
     }
-    
+
     // Then search for a generic match.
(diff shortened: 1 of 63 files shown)


lgritz:
Do you know about `.git-blame-ignore-revs`?

jstone-lucasfilm:
@lgritz I'd heard about this feature, but I hadn't yet deeply considered its usage on MaterialX.  Perhaps it would make our changes more straightforward to track, by removing pure coding style changes from consideration?

For tracking down the history of individual lines in a codebase, I've found the native GitHub feature to be really good, and it allows you to trace the history backwards on a per-line basis:

https://github.com/AcademySoftwareFoundation/MaterialX/blame/main/source/MaterialXCore/Definition.cpp

PR #1617, opened 2023-12-18
Fix for custom node groups in graph editor menu

ld-kerley:
Currently if you have node definitions added to groups that aren't in the fixed list that MaterialXGraphEditor understands, then they won't be added to the menu.


jstone-lucasfilm:
Thanks for this fix, @ld-kerley, and I'm CC'ing @lfl-eholthouser in case she has thoughts on the approach.

lfl-eholthouser:
This look great! Thanks for adding it @ld-kerley

jstone-lucasfilm (approved):

PR #1623, opened 2023-12-22
Add support for UI attributes in web viewer

kwokcb:
## Changes

- ui min, max, step support for property editor inputs
- label ui advanced folders
- enable complete vs reduced to get exposure of consistent inputs as with other utils.
- fix nodedef input lookup to handle non-root nodes
- use x,y,z,w for vector input names vs full name and ".0".

## Results:
- Slider update and advanced tagging
image

- Proper set of inputs for upstream nodes
image

- Vector UI naming:
image

- No sliders for enumerations (Leave "TODO" to add drop-drowns for now)
image


jstone-lucasfilm (approved):
Great work, thanks @kwokcb!

PR #1631, opened 2024-01-03
Initial nprlib with viewdirection node

jstone-lucasfilm:
This changelist introduces an NPR (non-photorealistic rendering) data library to MaterialX, initially consisting of a single `viewdirection` node.

The NPR data library is designed for the portable expression of artistic, non-physically-based materials, with common examples being cartoon shading, architectural diagrams, and edge highlighting.  In future versions of MaterialX, we expect this data library to be extended with additional primitives and artist-facing graphs as they are requested by the graphics community and approved by the MaterialX TSC.

For shader generation, MaterialX initially supports the NPR data library in GLSL, ESSL, MSL, and OSL, with MDL being omitted for now, based on intentional restrictions in its design philosophy.


jstone-lucasfilm:
This changelist includes a handful of very simple NPR examples, including Edge Brighten, Environment Map, and Starfield, in order to test implementations and evaluate parity across languages.

Artists are encouraged to create their own examples, and we'd be happy to include them in the future!

GraphEditor_EdgeBrighten
GraphEditor_EnvironmentMap
GraphEditor_Starfield

ashwinbhat:
Very excited to see this work towards nprlib. Thank you!!

source/MaterialXView/Viewer.cpp
```diff
@@ -2395,7 +2343,7 @@ mx::MaterialPtr Viewer::getEnvironmentMaterial()
 {
     if (!_envMaterial)
     {
-        mx::FilePath envFilename = _searchPath.find(mx::FilePath("resources/Lights/envmap_shader.mtlx"));
+        mx::FilePath envFilename = _searchPath.find(mx::FilePath("resources/Materials/TestSuite/nprlib/environment_map.mtlx"));
```
ashwinbhat, line 2346: IIRC MaterialXViewer usually does not refer to resources in TestSuite. Is envmap being relocated to npr due to dependency on viewdirection node?

jstone-lucasfilm: That's a good point, and we could indeed keep this resource in the "Lights" folder if that's more consistent with existing usage.  Originally, I moved it to the TestSuite so that it would be validated in MaterialXTest, but we have other `viewdirection` examples in that folder that can serve in this role, and the environment map shader doesn't strictly need to be there.

PR #1632, opened 2024-01-03
Add enumeration support to web viewer

kwokcb:
## Update
- Scan for `enum` and `enumvalues` if attributes exist on input. If no `enumvalues` exist then map to "default" of 0..&lt;enum list size&gt;.
- Create a drop-down for each enumerated input.

## Results:

Example from a `tiledimage` unit test file:

| | |
|-|-|
|  |  |


jstone-lucasfilm (approved):

PR #1635, opened 2024-01-04
Web viewer formatting improvements and fixes

kwokcb:
## Changes

- Add colouring to items and folders which can be set vis CSS. This makes it easier to tell what areas are under what folders.
- Close "ui advanced" groups and all but first material on load. Both make it so that the drop-down is not really long. Especially for the chess set or any file with more than a few materials which requires a fair bit of scrolling to find the materials (and collapse them).

## Fixes
- Fix string disable setting (was using old API)
- Fix parenting of enum widgets to be under current folder instead of top level.

## Results:

| Chess set with 1st material exanded and subsequent materials collapsed |  Standard surface to show coloring. Closing `uiadvanced` allows for the the inputs to fit. | Enumerated properly parented (Alpha Mode) |
| :--: | :--: | :--: |
|  |  |  |

For inclusion, adding in other shading models. (Only OpenPBR also has `uiadvanced`)

| OpenPBR | gltf_pbr | UsdPreviewSurface |
| :--: | :-: | :-: |
| image | image | image |

### DIFF
--- a/javascript/MaterialXView/index.ejs
+++ b/javascript/MaterialXView/index.ejs
@@ -10,6 +10,15 @@
         margin: 0;
         font-family: Arial
       }
+
+      /* Property editor item color */
+      .peditoritem {
+        background-color: #334444;
+      }
+      /* Property editor folder color */
+      .peditorfolder {
+        background-color: #333333;
+      }
     </style>
   </head>
   <body style="margin: 0px; overflow: hidden;">
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -1128,7 +1128,7 @@ export class Material
                             if (uniformToUpdate && value != null) {
                                 item = currentFolder.add(material.uniforms[name], 'value');
                                 item.name(path);
-                                item.readonly(true);
+                                item.disable(true);
                                 item.domElement.classList.add('peditoritem');                                
                             }
                             break;
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -860,15 +860,7 @@ export class Material
                                     currentFolder = folderList[newFolderName];
                                     if (!currentFolder) 
                                     {
-                                        if (nodeDefInput.hasAttribute('uiadvanced'))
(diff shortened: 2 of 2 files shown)


javascript/MaterialXView/index.ejs
```diff
@@ -10,6 +10,15 @@
         margin: 0;
         font-family: Arial
       }
+
+      /* Property editor item color */
+      .peditoritem {
```
kwokcb, line 15: This can easily be changed to modify the preferred colors.

javascript/MaterialXView/source/viewer.js
```diff
@@ -1017,8 +1035,9 @@ export class Material
                                         } 
                                     }                                    
                                     const defaultOption = enumList[value]; // Set the default selected option
-                                    const dropdownController = gui.add(enumeration, defaultOption, enumeration).name(path);
+                                    const dropdownController = currentFolder.add(enumeration, defaultOption, enumeration).name(path);
```
kwokcb, line 1038: Was adding to the wrong parent. Should add to the current folder.

javascript/MaterialXView/source/viewer.js
```diff
@@ -1106,7 +1128,8 @@ export class Material
                             if (uniformToUpdate && value != null) {
                                 item = currentFolder.add(material.uniforms[name], 'value');
                                 item.name(path);
-                                item.readonly(true);
+                                item.disable(true);
+                                item.domElement.classList.add('peditoritem');
```
kwokcb, line 1132: API readonly does not exist anymore.

jstone-lucasfilm:
Thanks for putting this proposal together, @kwokcb!

I really like the added colors and fixes to API calls, but I'm not as sure about the marking of folders as "Advanced", which has the odd visual impact of making them look like section headers (e.g. making them more visually prominent rather than hiding them).  Since the "Advanced" folder idea likely needs additional visual refinement, perhaps it should be considered a separate proposal?

kwokcb:
Sure thing. I'll revert the advanced formatting changes.

jstone-lucasfilm (approved):

PR #1636, opened 2024-01-04
WebViewer: Add canvas snapshot option

kwokcb:
## Changes

Add snapshot code to trigger on 's' key.

## Example

- UI View
image

- Saved Image
 screenshot


jstone-lucasfilm:
This is a great idea for a new feature, though perhaps we should use the same key binding as the MaterialX Viewer, with the "F" key representing "Frame Capture"?

kwokcb:
Yes. we should try to be consistent. I just remembered the incorrect key. Will change to `f`.

jstone-lucasfilm (approved):

PR #1637, opened 2024-01-05
Update TSC stakeholders

jstone-lucasfilm:
This changelist assigns Andréa Machizaud and Chris Kulla as the representatives from Adobe and Epic Games, as confirmed by votes at a recent MaterialX TSC meeting.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -48,7 +48,7 @@ The current Voting Members of the MaterialX TSC are:
 - Doug Smythe - Industrial Light & Magic
 - Niklas Harrysson - Lumiere Software
 - Orn Gunnarsson - Autodesk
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -68,8 +68,8 @@ The current Stakeholders of the MaterialX TSC are:
 - Rafal Jaroszkiewicz - SideFX
 - Lee Kerley - Sony Pictures Imageworks
```
lgritz, line 69: As an aside, Lee is no longer at SPI, so his affiliation should be corrected.

And we should appoint a new stakeholder, I guess.

jstone-lucasfilm: Good point, and let's chat offline about appointing a new representative from SPI.

Issue #1633, opened 2024-01-04
Web Viewer: When loading file that contains both surfacematerial and a nodegraph with single output, only one material is displayed


hybridherbst:
Relevant Slack discussion: https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1704323996349749

The attached file contains two materials, one implicit (single-output nodegraph) and one explicit (named surfacematerial).
According to the discussion, this is valid.
[two-materials.mtlx.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/13824462/two-materials.mtlx.zip)

I think it would be good if the viewer
- showed all materials in a file (it seems to already do this for the chess_set, just not here)
- allows choosing which material to display on the default shaderball (so in this case I could choose between the two in the file)

kwokcb:
Some "rambling" thoughts on this :).  Whatever is done it would be good to have consistent workflows between all sample viewers / editors if plausible.

The simple logic currently on load is 
1. to use at material assignments if present `<materialassign>`. Thus the chess set and, the "look" examples and the boombox have shaders assigned to the appropriate glTF geometry node.
2. otherwise take the first shader and assign it to all geometry.

The geometry is whatever geometry is currently loaded though the default is a shaderball.

The current state for desktop is:
 - A. Viewer: does 1,2 but allows for a current geometry to be set and a material selected to use against that geometry. A bit clumsy with any real scene viewer or interactive assignment. Assignments are not saved though maybe they should be?
- B. The Node Editor will take the first renderable and use it against the current geometry on load. Material assignments are not considered. Node selection determines the closest downstream "renderable" and uses that. Currently a renderable is either a material or surfaceshader node. This can be extended as "renderables" also include outputs which are scalar (float, vec2/3/4, color3/4).

Thus:
- I. There isn't really a "show all materials" option and not sure what that implies.
- II. There isn't really a consistent way to choose what renderable to display.  

I think the choice of how to add this in depends on the intent of the viewer / editor. For:
- I. This could be a preview "swatch" per renderable, or a separate geometry per renderable (instance the shadeball). This could be be consistently done for all editors/viewers. 
- II. This could be a assignment mechanism (someone like a geometry selection + renderable selection then assignment workflow). Or this could be done by "solo"ing the renderable (choose a renderable to show on all geometry while "solo"ing and then the current assignments when not "solo"ing. This would work for the node editor as well.

hybridherbst:
[...]

PR #1644, opened 2024-01-09
Renderable logic improvements to web viewer

kwokcb:
# Changes

- Fixes: #1633
- Adds in proper parsing of renderable elements vs always just picking the first surface shader found.
  - The test suite files which have (multiple) nodegraph outputs and top level outputs will now load properly.
- Adds in UI to mark folders as renderable (using a shaderball icon) 
- Adds in "soloing" capability to allow picking a renderable and have it show up on all geometry. When nothing is solo'ed the default material assignment is used.
- Includes fixes for:
  - Dag path assignment matching .
  - Missing exposure of NodeGraph::getDownStreamPorts() in JS.
  - Addressing the big performance hit when binding materials to geometry in ThreeJS. The code by default is some quite slow code for reflection / debugging purposes which is now turned off. Chess set load is seconds vs minutes. This affects the 'solo'ing workflow significantly as each switch is a geometry re-bind. If the material is not already cached then slow code will be hit which can cause seconds to pass when selecting a new material -- which appears like a "hang" from a user perspective.

# Results

## Videos
- Shows basic interaction of picking different renderables to solo and editing inputs on solo'ed item.

https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/908536a3-4a3b-4b3d-99b3-4cad1fb2507b

- Shows improved bind time performance for 2 cases: 
  - Load and bind all programs on document load (entire chess set)
  - Load all programs but incrementally binding on renderable switch
Once cached the program bind time overhead is reduced.

https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/e17324b4-4b26-4aa2-bc14-93eaead3e8ae

## Snapshots
- Includes case where are already assignments. Solo'ing is non-destructive and `materialassign` assigments found on load will be restored on unsolo'ing. (wood example)

| On load | Solo'ed |
| :--: | :--: | 
| image | image |
| image | image  image |
|  image |  image  image |


javascript/MaterialXView/source/viewer.js
```diff
@@ -769,31 +926,108 @@ export class Material
...
+
+        // Create and cache associated UI
+        var matUI = gui.addFolder(elemPath);        
+        matassign.setMaterialUI(matUI);
+
+        let matTitle = matUI.domElement.getElementsByClassName('title')[0];
+        // Add a icon to the title to allow for assigning the material to geometry
+        // Clicking on the icon will "solo" the material to the geometry.
+        // Clicking on the title will open/close the material folder.
+        matTitle.innerHTML = "<img id='" + elemPath + "' src='public/shader_ball.svg' width='16' height='16' style='vertical-align:middle; margin-right: 5px;'>" + elem.getNamePath();
```
kwokcb, line 1016: This mechanism of inserting additional DOM elements + event handler allows for insertion of additional interaction. We can do this anywhere we'd like to add in additional action "buttons".

javascript/MaterialXView/source/viewer.js
```diff
@@ -769,31 +926,108 @@ export class Material
...
+                img.src = 'public/shader_ball.svg';
+                //matTitle.classList.remove('peditor_material_unassigned');
+            }
+        }
+    }
+
+    static updateSoloMaterial(viewer, elemPath, materials, event)
+    {
+        // Prevent the event from being passed to parent folder
+        event.stopPropagation();
```
kwokcb, line 957: This is the only "trick" to capture the "icon" events to not being passed to the parent action which is to collapse the folder.

javascript/MaterialXView/source/viewer.js
```diff
@@ -623,15 +667,110 @@ export class Material
...
+            {
+                let nodeGraph = nodeGraphs[i];
+                if (nodeGraph)
+                {
+                    if (nodeGraph.hasAttribute('nodedef') || nodeGraph.hasSourceUri())
+                    {
+                        continue;
+                    } 
+                    // Skip any nodegraph that is connected to something downstream
+                    if (nodeGraph.getDownstreamPorts().length > 0)
```
kwokcb, line 712: This needs the missing API exposure. Prevents re-including nodegraphs which are already connected to downstream shaders.

jstone-lucasfilm (approved):

hybridherbst:
Awesome! Thank you so much!

Issue #1537, opened 2023-09-20
Potential IBL regression


ApoorvaJ:
The behavior of --envRad has significantly changed between the last two versions. I'm not sure if this is a regression.

This only happens when an irradiance image file is not found. In both versions, I get the warning: `Image file not found: /path/to/MaterialX/irradiance/foo.hdr`. But in the newer version, the result is much darker.

Is this a bug or is it on purpose?

Command: `bin/Debug/MaterialXView --material "Emerald_Peaks_Wallpaper.mtlx" --envRad "foo.hdr"`

On v1.38.7:
<img width="1392" alt="image" src="https://github.com/AcademySoftwareFoundation/MaterialX/assets/2312221/54af9340-83b7-4146-9bba-e70e367325c2">

On v1.38.8:
<img width="1392" alt="image" src="https://github.com/AcademySoftwareFoundation/MaterialX/assets/2312221/7ed89e5c-bfe6-4c9a-a5f1-5cbae8f3081a">

The material can be found [here](https://matlib.gpuopen.com/main/materials/all?material=8c9de375-59a8-49a8-a553-639880e0b3f9).

jstone-lucasfilm:
Good catch, @ApoorvaJ, and I'm testing a proposed fix for this issue, which we can integrate into MaterialX 1.38.9.

jstone-lucasfilm:
Fixed in #1647.

PR #1647, opened 2024-01-11
Fix irradiance generation in MaterialXView

jstone-lucasfilm:
This changelist fixes a regression to environment irradiance generation in MaterialXView, caused by a change to image caching logic for GLSL rendering in MaterialX 1.38.8.  This fix restores the ability to render environment maps that are missing their pre-generated irradiance data, with irradiance being generated on the fly via spherical harmonics.


jstone-lucasfilm:
Fixes #1537

PR #1650, opened 2024-01-15
Fix for type comparison for glsl programs

### LINKED ISSUE #1608: Regression in generation of symbols for ShaderGen Type Globals
## Issue

The logic to generate globals appears to have "broken"  sometime between Nov 15 and 29,2203 -- at least for Python libraries but perhaps for shared libraries in general.

This shows up for example in this code for GLSL uniforms.
 The code perform pointers comparisons which can now fail when running outside the test suite.

```c++
int GlslProgram::mapTypeToOpenGLType(const TypeDesc* type)
{
    if (type == Type::INTEGER)
        return GL_INT;
    else if (type == Type::BOOLEAN) <-- can fail
        return GL_BOOL;
    else if (type == Type::FLOAT) <-- can fail
        return GL_FLOAT;
    else if (type->isFloat2())
        return GL_FLOAT_VEC2;
    else if (type->isFloat3())
        return GL_FLOAT_VEC3;
    else if (type->isFloat4())
        return GL_FLOAT_VEC4;
    else if (type == Type::MATRIX33) <-- can fail
        return GL_FLOAT_MAT3;
    else if (type == Type::MATRIX44) <-- can fail
        return GL_FLOAT_MAT4;
    else if (type == Type::FILENAME)
    {
        // A "filename" is not indicative of type, so just return a 2d sampler.
        return GL_SAMPLER_2D;
    }

    return GlslProgram::Input::INVALID_OPENGL_TYPE;
}
```

You get these errors when trying to render from Python:
```shell
Errors: "GLSL uniform parsing error
Vertex shader uniform block type mismatch [PrivateUniforms]. Name: "u_worldMatrix". Type: "matrix44". Semantic: "". Value: "<none>". Unit: "<none>". Colorspace: "<none>". GLType: -1
Vertex shader uniform block type mismatch [PrivateUniforms]. Name: "u_viewProjectionMatrix". Type: "matrix44". Semantic: "". Value: "<none>". Unit: "<none>". Colorspace: "<none>". GLType: -1"
```

The **root** issue appears that globals defined for Type's get duplicated per Python module. This can be checked by adding debugging statements to Type constructor here: https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/TypeDesc.cpp#L42

This does not appear to be cause by this module dependency change: https://github.com/AcademySoftwareFoundation/MaterialX/pull/1595 
so it is still unknown why this regression has occurred.

The simplest test is to add debugging prints to the global declaration initializations and having a Python file which does the following:
```python
import MaterialX.PyMaterialXGenShader as mx_gen_shader
import MaterialX.PyMaterialXGenGlsl as mx_gen_glsl
```
Globals appear to be initialized twice -- once for each module.

```

kwokcb:
## Issue

Fixes #1608 where type comparison will fail and hence uniforms are not discoverable for rendering.

## Fix

The pointer comparison can fail so use name type comparison instead.

### DIFF
--- a/source/MaterialXRenderGlsl/GlslProgram.cpp
+++ b/source/MaterialXRenderGlsl/GlslProgram.cpp
@@ -1040,9 +1040,9 @@ int GlslProgram::mapTypeToOpenGLType(const TypeDesc* type)
 {
     if (type == Type::INTEGER)
         return GL_INT;
-    else if (type == Type::BOOLEAN)
+    else if (type->getName() == Type::BOOLEAN->getName())
         return GL_BOOL;
-    else if (type == Type::FLOAT)
+    else if (type->getName() == Type::FLOAT->getName())
         return GL_FLOAT;
     else if (type->isFloat2())
         return GL_FLOAT_VEC2;
@@ -1050,10 +1050,12 @@ int GlslProgram::mapTypeToOpenGLType(const TypeDesc* type)
         return GL_FLOAT_VEC3;
     else if (type->isFloat4())
         return GL_FLOAT_VEC4;
-    else if (type == Type::MATRIX33)
+    else if (type->getName() == Type::MATRIX33->getName())
         return GL_FLOAT_MAT3;
-    else if (type == Type::MATRIX44)
+    else if (type->getName() == Type::MATRIX44->getName())
+    {
         return GL_FLOAT_MAT4;
+    }
     else if (type == Type::FILENAME)
     {
         // A "filename" is not indicative of type, so just return a 2d sampler.
--- a/source/MaterialXRenderGlsl/GlslProgram.cpp
+++ b/source/MaterialXRenderGlsl/GlslProgram.cpp
@@ -1053,9 +1053,7 @@ int GlslProgram::mapTypeToOpenGLType(const TypeDesc* type)
     else if (type->getName() == Type::MATRIX33->getName())
         return GL_FLOAT_MAT3;
     else if (type->getName() == Type::MATRIX44->getName())
-    {
         return GL_FLOAT_MAT4;
-    }
     else if (type == Type::FILENAME)
     {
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
@kwokcb Do you have a sense of why the original pointer comparisons are no longer robust?  Since we use this pattern throughout the shader generation system, and not just in this one function, I'd recommend that we address the source of the pointer comparison failures, rather than working around it in just one location.

kwokcb:
@jstone-lucasfilm. I'm not sure what may be the cause of this. It may perhaps never be safe to perform pointer comparisons and just worked since all code is called within the same module -- I found this output by using GLSLProgram outside the module via Python.

jstone-lucasfilm:
@kwokcb That sounds like a good theory of the case, and if these pointer comparisons are indeed unsafe across modules, then we should refactor this PR as a proposal to replace the pointer comparisons globally.

CC'ing @niklasharrysson for his thoughts!

niklasharrysson:
@kwokcb @jstone-lucasfilm It is very strange if these pointer compares have become unsafe. The `TypeDesc` class has been there since the early days of shader generation many many years ago .And these type compare are being done all over the code base, in many different modules.

It's also hard to see why it would fail here. Since when you get to the points where this `mapTypeToOpenGLType` is being called a shader has already been generated successfully, and hundreds of these type compares must have been performed already.

Bernard, could you get some more information out from a debugger? If you look at the TypeDesc pointers and data pointed to, and compare to what is sent in, how does that look? 

We can take this one offline and discuss more.

jstone-lucasfilm:
I'll close out this specific pull request, since the fix will likely be broader than the changes described here, and we'll need additional research into the cause.

We can continue the discussion in https://github.com/AcademySoftwareFoundation/MaterialX/issues/1608 open, which is a good summary of the issue that needs to be addressed.

PR #1653, opened 2024-01-17
Improvements to noise implementations

jstone-lucasfilm:
- Leverage node graphs to share the conversion aspects of noise implementations across languages.
- Simplify noise unit tests to cover only unique implementations.


kwokcb:
@jstone-lucasfilm , can we do a check that this does not increase the uniform interface expose for GLSL like with OSL. As discussed in a side thread with @niklasharrysson, my worry here is if all uniforms in the graph are exposed instead of just the `noise` functions inputs. 

I don't think this should not occur but want to double check.

jstone-lucasfilm:
@kwokcb Good question, and I don't see a change in the visibility of shader uniforms with this proposal.

PR #1659, opened 2024-01-18
1.39 update - rename atan2 inputs

ld-kerley:
Renaming the input ports to align correctly with the spec
* `in1` -> `iny`
* `in2` -> `inx`

Add upgrade code to migrate port names found in documents.


jstone-lucasfilm (approved):
This looks great to me, thanks @ld-kerley, and I'm CC'ing @dbsmythe and @niklasharrysson for their thoughts.

jstone-lucasfilm:
Bumping this thread, in case @dbsmythe or @niklasharrysson have thoughts; otherwise, I think we should plan to begin merging these proposals for 1.39.

dbsmythe:
I have no objections!

PR #1660, opened 2024-01-19
Upgrade switch node to ten inputs in 1.39

ld-kerley:
Also added an upgrade function, which may need some closer attention.   The upgrade function attempts to retain the behavior from 1.38 where any value of `which` that is above the accepted range that would previously been served the default (`in1`) value.  A `which` value of 6 would have previously return `in1`, but now it would return one of the newly added inputs.  To retain the previous renderer image, the document upgrader sets the value of `which` to the default of 0.

I'm very much open to input if this doesn't feel right.


jstone-lucasfilm (approved):
This looks good to me, @ld-kerley, and I'm CC'ing @dbsmythe and @niklasharrysson for their thoughts on this alignment of our codebase with the specification.

niklasharrysson (approved):
This looks good to me too. Thanks @ld-kerley

PR #1661, opened 2024-01-19
1.39 update - add 'round' node definition.

ld-kerley:
Mirror the existing `ceil` node definition.

No upgrade function necessary, as the node definition never existed.


jstone-lucasfilm:
This proposal looks great to me, thanks @ld-kerley!

Since this change is purely additive, and requires no upgrade path for 1.38 documents, it falls in the categories of changes that we can potentially include in the upcoming 1.38.9 release.

Would it be beneficial to propose this as a pull request to the **main** branch instead of the 1.39 branch, so that it can be included in our very next release?

ld-kerley:
looks like my rebase went sideways - I'm gonna just close this and open a new PR against main.

Issue #1022, opened 2022-07-06
Discussion about breaking changes and versioning


ix-dcourtois:
Hi,

I'd like to open a discussion about a little friction when updating MaterialX library: breaking changes :)
This is a bit touchy because I know that from a library's maintainers' point of view, it's very tempting to evolve the API to keep it as clean as possible, at the risk of breaking things for user code. But on the other hand, every time I update MaterialX in our build system, I have to spend a fair bit of time to fix things. Sometimes it was quite simple, some other times a lot trickier. And most of the time at places where I didn't expect it :)

Anyway, MaterialX being quite new, I understand that the API is not stable enough yet, and trying to make it evolve while trying to keep it backward compatible could quickly become a maintenance nightmare for you. But I feel there must be a sort of middle ground that would allow us to update without too many headaches, and you to make changes without dritying things too much, thus this discussion.

For breaking changes in the C++ API, a solution could be to flag things as deprecated while keeping them, and add the new things in parallel. And once in a while, cleanup the deprecated APIs. Ideally, the cleanup (e.g. breaking change) would happen on minor version update (e.g. from `1.38.5` to `1.39.0` for instance) This way, patch version update could still introduce new functionality, but it would guarantee that it wouldn't break user code: we would be able to update a lot easier, get new functionality and fixes without problem, and we would have time to convert the deprecated APIs that we're using before updating to a new minor version.

The second kind of breaking changes are ones related to data: for instance between the 1.38.3 and the 1.38.5, the directory structure of `libraries/genosl` changed, and the osl headers `mx_funcs.h` moved around. This broke compilation of generated OSL code because the include paths were no longer set correctly, so we had to patch our compilation options to account for that. Those kind of changes are a bit more difficult to handle I think because there is no clean deprecation mechanism for those... Maybe at least defer those changes to a minor version update? That way patch version update would guarantee that we could update without problem?

Sorry for the length :)
Any thoughts, ideas?

Regards,
Damien.

kwokcb:
A patch identifier was considered before so maybe something like this could work (?):

1. If there is any non-backwards compatible change or a data representation change then the patch level is bumped. It may or may not be possible to backwards compatible all the time. Deprecated API can be marked as such. 

So next patch would be `1.39.5.1`. 

2. Patch level is not part of the version. 
Issues I've run into the past is that there was no indication in the data file that within 2 build
releases or within 2 minor releases anything had changed. A suggestion from before was to
[...]

PR #1664, opened 2024-01-19
Add versioning rules to Developer Guide

jstone-lucasfilm:
This changelist adds two new sections to the Developer Guide, describing the categories of changes to the MaterialX API and data libraries that are allowed in version upgrades.

### DIFF
--- a/documents/DeveloperGuide/MainPage.md
+++ b/documents/DeveloperGuide/MainPage.md
@@ -58,9 +58,24 @@ Building the `install` target of your project will install the MaterialX C++ and
 
 ### MaterialX Versioning
 
-The MaterialX codebase uses a modified semantic versioning system where the *major* and *minor* versions match that of the corresponding MaterialX [specification](https://www.materialx.org/Specification.html), and the *build* version represents engineering advances within that specification version.  MaterialX documents are similarly marked with the specification version they were authored in, and they are valid to load into any MaterialX codebase with an equal or higher specification version.
+The MaterialX codebase uses a modified semantic versioning system where the *major* and *minor* versions match that of the corresponding MaterialX [specification](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md), and the *build* version represents engineering advances within that specification version.  MaterialX documents are similarly marked with the specification version they were authored in, and they are valid to load into any MaterialX codebase with an equal or higher specification version.
 
-Upgrading of MaterialX documents from earlier versions is handled at import time by the Document::upgradeVersion method, which applies the syntax and node interface upgrades that have occurred in previous specification revisions.  This allows the syntax conventions of MaterialX and the names and interfaces of nodes to evolve over time, without invalidating documents from earlier versions.
+Upgrading of MaterialX documents from earlier versions is handled at import time by the `Document::upgradeVersion` method, which applies the syntax and node interface upgrades that have occurred in previous specification revisions.  This allows the syntax conventions of MaterialX and the names and interfaces of nodes to evolve over time, without invalidating documents from earlier versions.
+
+#### MaterialX API Changes
+
+The following rules describe the categories of changes to the [MaterialX API](https://materialx.org/docs/api/classes.html) that are allowed in version upgrades:
+
+- In *build* version upgrades, only non-breaking changes to the MaterialX API are allowed.  For any API call that is modified in a build version upgrade, backwards compatibility should be maintained using deprecated C++ and Python wrappers for the original API call.
+- In *minor* and *major* version upgrades, breaking changes to the MaterialX API are allowed, though their benefit should be carefully weighed against their cost.  Any breaking changes to API calls should be highlighted in the release notes for the new version.
+
+#### MaterialX Data Library Changes
+
+The following rules describe the categories of changes to the [MaterialX Data Libraries](https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/libraries) that are allowed in version upgrades:
+
+- In *build* version upgrades, only additive changes to the MaterialX data libraries are allowed, e.g. adding new nodes or node inputs with backwards-compatible default values.  Because additive changes are backwards-compatible by nature, they do not require corresponding version upgrade logic in the codebase.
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This proposal follows up on a discussion at our last MaterialX TSC meeting, and additionally incorporates ideas from the discussion in https://github.com/AcademySoftwareFoundation/MaterialX/issues/1022.

documents/DeveloperGuide/MainPage.md
```diff
@@ -58,9 +58,24 @@ Building the `install` target of your project will install the MaterialX C++ and
...
+Upgrading of MaterialX documents from earlier versions is handled at import time by the `Document::upgradeVersion` method, which applies the syntax and node interface upgrades that have occurred in previous specification revisions.  This allows the syntax conventions of MaterialX and the names and interfaces of nodes to evolve over time, without invalidating documents from earlier versions.
+
+#### MaterialX API Changes
+
+The following rules describe the categories of changes to the [MaterialX API](https://materialx.org/docs/api/classes.html) that are allowed in version upgrades:
+
+- In *build* version upgrades, only non-breaking changes to the MaterialX API are allowed.  For any API call that is modified in a build version upgrade, backwards compatibility should be maintained using deprecated C++ and Python wrappers for the original API call.
```
kwokcb, line 69: - I think it useful to have something that says what is a major, minor and build number in a release. At least for build. Something like "the syntax of a version is `<major>.<minor>.<build>` ?
- Probably want to add in Javascript wrappers as well.

documents/DeveloperGuide/MainPage.md
```diff
@@ -58,9 +58,24 @@ Building the `install` target of your project will install the MaterialX C++ and
...
+- In *minor* and *major* version upgrades, breaking changes to the MaterialX API are allowed, though their benefit should be carefully weighed against their cost.  Any breaking changes to API calls should be highlighted in the release notes for the new version.
+
+#### MaterialX Data Library Changes
+
+The following rules describe the categories of changes to the [MaterialX Data Libraries](https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/libraries) that are allowed in version upgrades:
+
+- In *build* version upgrades, only additive changes to the MaterialX data libraries are allowed, e.g. adding new nodes or node inputs with backwards-compatible default values.  Because additive changes are backwards-compatible by nature, they do not require corresponding version upgrade logic in the codebase.
```
kwokcb, line 76: Do we note that since these are "non-breaking" that the `build` version is not required as part of the document version which has only `<major>.<minor>`.

kwokcb: Q from before. Do we want to say that new versions of nodes are allowed as these can be forward-compatible.

documents/DeveloperGuide/MainPage.md
```diff
@@ -58,9 +58,24 @@ Building the `install` target of your project will install the MaterialX C++ and
...
+- In *minor* version upgrades, changes to the names and interfaces of MaterialX nodes are allowed, with the requirement that version upgrade logic be used to maintain the validity and visual interpretation of documents from earlier versions.
```
kwokcb, line 77: We might want to be careful as changes to the `nodedef` `name` I don't think should ever be allowed. Changes to the nodedef `node` can work.

documents/DeveloperGuide/MainPage.md
kwokcb: Do we want to mentioned that a nodegraph implementation for a nodedef can be modified as long as the interface does not?

documents/DeveloperGuide/MainPage.md
```diff
@@ -58,9 +58,24 @@ Building the `install` target of your project will install the MaterialX C++ and
...
+- In *major* version upgrades, changes to the syntax rules of MaterialX documents are allowed, with the requirement that version upgrade logic be used to maintain the validity and visual interpretation of documents from earlier versions.  These changes usually require synchronized updates to both the MaterialX API and data libraries.
```
kwokcb, line 78: Does this mean we can't change the nodegraph / implementations for nodes to make them "visuallly" better in minor / branch upgrades? e.g. if we change std surf or openpbr ? It just kind of sounds like that IMO.

documents/DeveloperGuide/MainPage.md
jstone-lucasfilm: For defining the major, minor, and build versions, my sense is that we cover this in the [section above](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/MainPage.md#materialx-versioning), and I'd like to keep this new section as simple as possible.

That's a good note about JavaScript, though we don't yet hold this API to the same level of backwards-compatibility as C++ and Python, since it's not yet integrated into major industry projects.  In the future, I think you're definitely right that we'll want to hold the JavaScript API to the same level of rigorous compatibility, but we're not quite there yet.

documents/DeveloperGuide/MainPage.md
jstone-lucasfilm: As long as both content documents and data libraries are marked with the MaterialX versions in which they were authored, I think we can safely update nodedef names in addition to node names, using the same Document::upgradeVersion logic.  There are subtleties to consider for the future, e.g. USD containers without document versioning, but the rules in this section are focused on the case of MaterialX documents that are fully versioned.

documents/DeveloperGuide/MainPage.md
jstone-lucasfilm: This is a good point, and I would probably refine this language to say that "fixes that align node implementations with the specification" are allowed in build updates.  In these cases, we're correcting defects in a node implementation, not redefining the node in the standard, and these fixes should be allowed.

documents/DeveloperGuide/MainPage.md
jstone-lucasfilm: Although these are accurate statements, I think they might be in the category of "too much information" for this simple summary of our versioning rules.  Hopefully these edge cases, along with many others that we don't have room to state, will become clear through usage in version updates over time.

documents/DeveloperGuide/MainPage.md
jstone-lucasfilm: Thanks for this note, and I've adjusted the language to clarify this.

PR #1665, opened 2024-01-20
Improve robustness of TypeDesc pointer comparisons

kwokcb:
## Issue

Update #1608 

- The globals presets defined for TypeDesc are pointers which may not be shared between shared modules. This occurs for instance  in Python where the pointers are declared locally for each module.
- Any pointer comparison between the same TypeDesc preset  can thus result in a failure status.

## Workaround

The current non API changing patch is to change any existing direct pointer references to to reference compares using new `TypeDesc::operater==` and `TypeDesc::operator!=` method.

This has been tested with RTS:
[rts.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/13998018/rts.pdf)
[rts2.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/13998017/rts2.pdf)
- 
and separate rendering tests using a renderer written in Python:
- (https://github.com/kwokcb/MaterialX_Learn/blob/main/pymaterialx/mtlxutils/mxrenderer.py)

This workaround has been discussed with @niklasharrysson and @jstone-lucasfilm offline.

## Caveat

This is not a long term fix but is a safe patch. A recommendation is to add  a release note about this so that integrations know about this issue.

### DIFF
--- a/source/MaterialXGenShader/TypeDesc.h
+++ b/source/MaterialXGenShader/TypeDesc.h
@@ -57,6 +57,12 @@ class MX_GENSHADER_API TypeDesc
     static const TypeDesc* registerType(const string& name, unsigned char basetype, unsigned char semantic = SEMANTIC_NONE,
                                         size_t size = 1, bool editable = true, const ChannelMap& channelMapping = ChannelMap());
 
+    /// Equality operator overload
+    bool operator==(const TypeDesc& other) const;
+
+    /// Inequality operator overload
+    bool operator!=(const TypeDesc& other) const;
+
     /// Get a type descriptor for given name.
     /// Returns an empty shared pointer if no type with the given name is found.
     static const TypeDesc* get(const string& name);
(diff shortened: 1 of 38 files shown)


source/MaterialXGenShader/TypeDesc.h
```diff
@@ -57,6 +57,12 @@ class MX_GENSHADER_API TypeDesc
     static const TypeDesc* registerType(const string& name, unsigned char basetype, unsigned char semantic = SEMANTIC_NONE,
                                         size_t size = 1, bool editable = true, const ChannelMap& channelMapping = ChannelMap());
 
+    /// Equality operator overload
+    bool operator==(const TypeDesc& other) const;
```
jstone-lucasfilm, line 61: For consistency with other comparison operators in MaterialX, I would recommend naming this input `rhs`:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Types.h#L73

jstone-lucasfilm (commented):
This looks like a valuable fix, thanks @kwokcb, and I had just one recommendation on variable names.

source/MaterialXGenShader/TypeDesc.h
kwokcb: Fixed.

jstone-lucasfilm (approved):

PR #1667, opened 2024-01-21
Fix orphaned links when deleting node in graph editor

### LINKED ISSUE #1577: Graph Editor: Double input on nodes following edit
After editing the input to a nodegraph I end up with a node having two inputs,  interfacename and nodename
image

Steps to repro:
 - Open the file bug_double_connection_before.mtlx
 - Open the nodegraph
 - You will see a multiply node whose input is an intefacename (input node)
image
 - Add a constant_color3 node
 - Delete the input_color3 node
 - Connect the output of the constant_color3 node to the multiply
image
 - Save the MTLX file
The multiply node now has 2 inputs as in the top image

Loading the saved MTLX now gives an error:
image

MTLX files:
[materialX_double_connection.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/13035330/materialX_double_connection.zip)

### LINKED ISSUE #1666: Graph Editor: Deleting a node with multiple output does not update down-nodes properly
When deleting a node with multiple outputs connected, only the first linked input pin appears to be properly updated.
See before and after deletion:

image

Steps to reproduce:
- Open the Graph Editor and create a new material (File->New)
- Add a node with multiple output pins (Node A)
- Add a node with multiple input pins of the same type (Node B) (also works with multiple nodes)
- Connect all of the outputs of Node A to Node B
- Delete Node A
- Notice that only the pin that was connected to Node A's first output is displayed as disconnected

Minimal reproduction material just before deletion: [minimalrepro_bug_multiple_output.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/14002110/minimalrepro_bug_multiple_output.zip)

This issue was discovered while investigating #1577. It appears to be related: a node deletion does not completely clean the orphaned links.

LeoBelda:
This PR introduces fixes related to the removal of orphaned links when deleting a node in the Graph Editor:
- remove the attribute `INTERFACE_NAME_ATTRIBUTE` of input pins that were connected to the deleted node
(Fixes #1577) 
- iterate over all of the output pins instead of only handling the first one.
(Fixes #1666)

Note that the part of the code which handles the attribute removal was copied from `Graph::deleteLinkInfo` after noticing that deleting a link did not triggered the issue #1577.

This would probably be too much of a change for the present PR, but in order to remove duplicate code, `Graph::deleteLinkInfo` could probably be re-used when deleting a node, as demonstrated here: https://github.com/AcademySoftwareFoundation/MaterialX/commit/bf8dd81281eb9bf25a0f4b4b94510985bf9feac6

However, it would be less efficient since `Graph::deleteLinkInfo` internally iterates to find the right pin, and would have to be reworked to allow for a caller to directly provide the pin.


jstone-lucasfilm:
Thanks for writing up this fix, @LeoBelda, and I'm CC'ing @lfl-eholthouser for her thoughts on the approach.

jstone-lucasfilm (approved):
Thanks for this fix, @LeoBelda!

lfl-eholthouser (approved):

PR #1674, opened 2024-01-24
Add Gooch Shading node to NPR library

ashwinbhat:
Initial implementation of a  `gooch` node to the NPR data library that represents [Gooch Shading](https://en.wikipedia.org/wiki/Gooch_shading) used for technical illustration. This is built on prior work by @bernardkwok 
image


libraries/nprlib/nprlib_defs.mtlx
```diff
@@ -39,4 +39,18 @@
     <output name="out" type="float" default="0.0" />
   </nodedef>
 
+  <!--
+    Node: <gooch>
```
jstone-lucasfilm, line 43: The name of this node would benefit from more detail, and one option that comes to mind is `gooch_npr`.  If we went with this naming convention, then future NPR nodes that incorporate a lighting computation would similarly have the `npr` suffix, creating a parallel with `bsdf` in the physically based shading library.

libraries/nprlib/nprlib_defs.mtlx
```diff
@@ -39,4 +39,18 @@
...
+    Compute Gooch shading.
+  -->
+  <nodedef name="ND_gooch" node="gooch" doc="Compute Gooch shading">
```
jstone-lucasfilm, line 46: We should assign this new node to the `npr` node group, so that application interfaces display it in this folder.

libraries/nprlib/nprlib_defs.mtlx
```diff
@@ -39,4 +39,18 @@
...
+    <input name="Warm" type="color3" value="0.6, 0, 0" uiname="Warm Color" doc="Warm diffuse color"/>
+    <input name="Cool" type="color3" value="0, 0, 0.6" uiname="Cool Color" doc="Cool diffuse color"/>
+    <input name="Shininess" type="float" value="8" uimin="1" uimax="64" doc="Specular shininess"/>
+    <input name="SpecularColor" type="color3" value="0.6, 0.6, 0.6" uiname="Specular Color" doc="Specular color"/>
+    <input name="LightDirection" type="vector3" value="1, 0.5, 0.5" uiname="Light Direction" doc="Light vector"/>
```
jstone-lucasfilm, line 51: For consistency with other standard node interfaces, we should use our lower-case and underscore convention for inputs, so the name of this input would be `light_direction`.

jstone-lucasfilm (commented):
This looks like a great start, thanks @kwokcb and @ashwinbhat, and I had a few initial notes on interface conventions and categorization.

libraries/nprlib/nprlib_defs.mtlx
jstone-lucasfilm: Another option for this node name would be `gooch_shade`, which suggests the presence of a lighting computation without implying a physically based light integration of the type found in the PBR library.

The name `gooch_shade` would also echo the common term for [Gooch Shading](https://en.wikipedia.org/wiki/Gooch_shading) in the industry, which might help to clarify its purpose.

ashwinbhat: gooch_shade sounds reasonable. I'll update it.

PR #1675, opened 2024-01-25
Added missing classification on VolumeShader nodes.

dgovil:
ShaderNodes.cpp had missing classification information for Volume Shaders. This PR is a simple addition of that classification.


jstone-lucasfilm (commented):
This looks like a good fix to me, @dgovil, and I'm Cc'ing @niklasharrysson for his thoughts.

niklasharrysson:
This looks good to me too. Thanks for the fix @dgovil

PR #1676, opened 2024-01-25
Add invert node to specification

ld-kerley:
The node exists in the standard library code, but is missing from the specification.


jstone-lucasfilm (approved):

jstone-lucasfilm:
As with your proposal to implement the round node, though, perhaps this ought to be a pull request to the *main* branch rather than the dev_1.39 branch, so that it becomes available in our very next release of MaterialX.

Does that seem reasonable to you as well?

ld-kerley:
That makes sense, I will move both of those PRs over to 'main'

ld-kerley:
This one had zero conflicts :)

PR #1677, opened 2024-01-26
Rename ND_normalmap to ND_normalmap_float in 1.39

ld-kerley:
Renaming `ND_normalmap` -> `ND_normalmap_float`

Added upgrade mechanism to handle the node definition rename.

### DIFF
--- a/libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
+++ b/libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
@@ -42,9 +42,8 @@
   </implementation>
 
   <!-- <normalmap> -->
-  <implementation name="IM_normalmap_float_genglsl" nodedef="ND_normalmap" file="mx_normalmap.glsl" function="mx_normalmap_float" target="genglsl" />
+  <implementation name="IM_normalmap_float_genglsl" nodedef="ND_normalmap_float" file="mx_normalmap.glsl" function="mx_normalmap_float" target="genglsl" />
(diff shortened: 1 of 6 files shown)


jstone-lucasfilm (commented):
This looks great to me, @ld-kerley, and I'm CC'ing @ashwinbhat and @JGamache-autodesk to make sure we're not missing any implications for MaterialX 1.38 content embedded in USD stages.

JGamache-autodesk:
USD stages will systematically hardcode the nodedef string into the `info:id` attribute of the UsdShade nodes. This means all USD persisted data will need an upgrade path to 1.39, and also potentially require storing the MaterialX version somewhere in the document to know if an upgrade is required. 

We need to start the discussion with the AOUSD WG because that might become an adoption blocker for MaterialX 1.39.

jstone-lucasfilm:
Thanks for that context, @JGamache-autodesk, and that's the perspective I was hoping you could provide here.  I wonder if the USD project could begin versioning its own documents when it upgrades to MaterialX 1.39, which would allow MaterialX 1.38 content embedded in legacy USD stages to be robustly upgraded to MaterialX 1.39.  The logic that @ld-kerley has proposed here would then work universally for MaterialX content, regardless of the file format in which it has been encoded.

jstone-lucasfilm:
We'll need to address the USD versioning question before the OpenUSD project can upgrade to MaterialX 1.39, as even the simplest possible update to a node interface (e.g. https://github.com/AcademySoftwareFoundation/MaterialX/pull/1659) will require that the MaterialX version of the source document is known.

My sense is that we should move forward with this specific change from @ld-kerley, keeping the need for USD versioning in our minds as an important topic for upcoming USD/MaterialX working group meetings.

jstone-lucasfilm (approved):

PR #1680, opened 2024-01-29
Always build GLFW as a static library

### LINKED ISSUE #1245: libGLFW is not installed as part of the MaterialXGraphEditor target
The MaterialXGraphEditor target has a dependency on libGLFW. However this does not get installed when building the graph editor, causing it to be non functional unless the libraries get copied over. 

I see that it sets `GLFW_INSTALL` to `OFF` , which is oddly what the NanoGUI build does for the Viewer too.
I do however see that NanoGUI says this in its cmake

```
# Two targets have now been defined: `glfw_objects`, which will be merged into
  # NanoGUI at the end, and `glfw`.  The `glfw` target is the library itself
  # (e.g., libglfw.so), but can be skipped as we do not need to link against it
  # (because we merge `glfw_objects` into NanoGUI).  Skipping is required for
  # XCode, but preferable for all build systems (reduces build artifacts).
  ```

Perhaps something similar needs to be done for the GraphEditor so it doesn't have a hard link against libGLFW? or we need to enable the install.

Enabling the install however does cause this error to pop up during the build despite that file existing.

`source/MaterialXGraphEditor/External/Glfw/CMake/cmake_uninstall.cmake.in does not exist.`

However I do see the file exists on the nanogui side: `MaterialXView/NanoGUI/ext/glfw/CMake/cmake_uninstall.cmake.in`
So I just wanted to see if there was any insight on why the file was removed, and why the install is disabled? also if there were any ways we could move forward and either inline the symbols or install the lib?

### PR DESCRIPTION
**Pull request by ld-kerley**
Currently the embedded glfw build for MaterialXGraphEditor inherits the value of `BUILD_SHARED_LIBS` from `MATERIALX_BUILD_SHARED_LIBS`, but we're not installing `libglfw`, per #1245 the intention was to statically link.

closes #1245

### DIFF
--- a/source/MaterialXGraphEditor/CMakeLists.txt
+++ b/source/MaterialXGraphEditor/CMakeLists.txt
@@ -94,7 +94,7 @@ target_link_libraries(
     MaterialXGraphEditor
     PRIVATE
     ${MATERIALX_LIBRARIES}
- 	glfw)
+ 	glfw_minimal)
 
 install(TARGETS MaterialXGraphEditor
     EXPORT MaterialX
--- a/source/MaterialXGraphEditor/CMakeLists.txt
+++ b/source/MaterialXGraphEditor/CMakeLists.txt
@@ -60,6 +60,12 @@ set(GLFW_BUILD_TESTS OFF)
 set(GLFW_BUILD_DOCS OFF)
 set(GLFW_INSTALL OFF)
 
+# we explicitly force glfw to build static libraries, as we don't install them.
+# we don't need to restore the state after we call add_subdirectory here because
+# we're not adding any new targets after this, and the state is unmodified at a higher
+# cmake level.
+set(BUILD_SHARED_LIBS OFF)
+
 add_subdirectory(External/Glfw)
 
 if(MSVC)
--- a/source/MaterialXGraphEditor/CMakeLists.txt
+++ b/source/MaterialXGraphEditor/CMakeLists.txt
@@ -94,7 +94,7 @@ target_link_libraries(
     MaterialXGraphEditor
     PRIVATE
     ${MATERIALX_LIBRARIES}
- 	glfw_minimal)
+ 	glfw)
 
 install(TARGETS MaterialXGraphEditor
     EXPORT MaterialX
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: source/MaterialXGraphEditor/CMakeLists.txt [hunk 1]
```diff
@@ -88,7 +94,7 @@ target_link_libraries(
     MaterialXGraphEditor
     PRIVATE
     ${MATERIALX_LIBRARIES}
- 	glfw_minimal)
+ 	glfw)
```
jstone-lucasfilm, line 97: This specific line change may need additional clarification, as we're still using the "glfw_minimal" name for the unique copy of GLFW in the MaterialX Graph Editor:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGraphEditor/External/Glfw/CMakeLists.txt#L44

Is this unique name no longer needed, or are there perhaps situations where we should continue referencing the "glfw_minimal" name?

jstone-lucasfilm (commented):
This change to shared libraries looks good to me, though I had a question about the change to the target_link_libraries call below.

source/MaterialXGraphEditor/CMakeLists.txt
jstone-lucasfilm: I wonder if changing the name to `glfw` is causing the build process to find a different installation of GLFW entirely, not the one that's provided in the MaterialX Graph Editor?

The motivation for the unique name is that CMake doesn't allow two projects with the same name to exist within a build, and NanoGUI already contains a copy of GLFW using its default name:

https://github.com/mitsuba-renderer/nanogui/tree/f5020e2f3e5114d517642e67afbb21cb88cf04c0/ext

ld-kerley: your hunch was correct - I had a homebrew install of glfw that was confusing things - once I removed that - the `glfw_minimal` target worked. sorry about that.

jstone-lucasfilm: Thanks for verifying, and otherwise this change looks good to me!

jstone-lucasfilm (approved):

PR #1689, opened 2024-02-02
Add safepower node to data libraries

ld-kerley:
Add node definition and nodegraph for `safepower` which is already in the v1.39 specification, but currently missing in the source code.

As this is an additive change we can safely add this to `main` for inclusion in any v.138 release.


jstone-lucasfilm (approved):

Issue #1692, opened 2024-02-03
Suggestion: Publish MaterialX Core Javascript package


kwokcb:
## Issue

It is currently possible to access the shader generation package but not just the core package
e.g

```html
<script src="https://academysoftwarefoundation.github.io/MaterialX/JsMaterialXGenShader.js" type="text/javascript"></script>
```
will load the Shader gen package.

## Proposal

- Currently this package is published as part of the Web Viewer build. It only copies over the `JsMaterialXGenShader.js`
package. It should not be hard to also copy over the smaller `JsMaterialXCore` package.
- It would be useful to update the Javascript documentation to let folks know that they do not have to build these packages
themselves. It only mentions how to load a local package.
- Currently it's pretty slow to load over a network and the size of the wasm files I believe are one reason why the web viewer has the appearance of being "slow" to start. (@jstone-lucasfilm -- this is my hunch). Once loaded if it's cached then it's okay.
  - Note that the packages have **not** not been optimized and probably should be. e.g. see
https://emscripten.org/docs/compiling/Deploying-Pages.html?highlight=wasm).

* Leaving a ping with @ashwinbhat as we have discussed how accessible MaterialX Javascript packages are.

PR #1693, opened 2024-02-04
Support hosting of JsMaterialXCore library

kwokcb:
## Issue

Update #1692 

- Expose JsMaterialXCore as part of published github pages.
- This allows users to load in the core package only for Web usage without pulling in the much larger shader gen package
(JsMaterialXGenShader) -- which is often undesired.
- These are the current package / data sizes:
  image

## Changes

- Add JsMaterialXCore WASM and JS as part of distribution.

## Example Usage

Basic loading of just core (which includes format).

```html
     <!-- Grab core package -->
    <script src="https://academysoftwarefoundation.github.io/MaterialX/JsMaterialXCore.js" type="text/javascript"></script> 
    <script type="text/javascript">
  
        let mx = null;

        function doStuff(mx) {
            doc = mx.createDocument();
            console.log(mx.prettyPrint(doc));
        }

        function loadMaterialX() {
            return new Promise((resolve, reject) => {
                MaterialX().then((mtlx) => 
                {
                    mx = mtlx;
                    console.log("MaterialX loaded.");
                    resolve();
                }).catch((error) => {
                    reject(error);
                });
            });
        }

        loadMaterialX().then(() => {
            console.log("Performing additional logic with mx:", mx);
            doStuff(mx);
        }).catch((error) => {
            console.error("Error loading MaterialX:", error);
        });
  </script>
```

### DIFF
--- a/javascript/MaterialXView/webpack.config.js
+++ b/javascript/MaterialXView/webpack.config.js
@@ -71,6 +71,8 @@ module.exports = {
                 { from: stdSurfaceMaterials, to: stdSurfaceMaterialsBaseURL },
                 { from: usdSurfaceMaterials, to: usdSurfaceMaterialsBaseURL },
                 { from: gltfSurfaceMaterials, to: gltfSurfaceMaterialsBaseURL },
+                { from: "../build/bin/JsMaterialXCore.wasm" },
+                { from: "../build/bin/JsMaterialXCore.js" },
                 { from: "../build/bin/JsMaterialXGenShader.wasm" },
                 { from: "../build/bin/JsMaterialXGenShader.js" },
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
@kwokcb I could use some additional details on how this change would impact existing users of the MaterialX Web Viewer.  Would this increase the amount of data that they need to download at startup?  If so, then should we instead be exploring options for publishing these two libraries independently of the viewer itself?

kwokcb:
There isn't any impact since this package is not part of the web viewer. This would just reside in a location which is accessible from the web. Those that want to access this can do so independently of the viewer.

jstone-lucasfilm (approved):

PR #1697, opened 2024-02-05
Improve input filtering in Web Viewer

kwokcb:
## Issue

There is no current "edit" exposure control for code generation. We thus end up with too many
non-interface inputs and non-inputs being exposed when using "COMPLETE" option.

## Work-around

Add some additional filtering to remove interior inputs, non-inputs.
Also check for repeated inputs.

## Results

| Test | Resulting UI |
| :--: | :--: |
| Procedural Brick Example |  image |
| Marble Example |  image
 | Top level input |  image |
 | Wood (graph has no interfaces exposed)  |  image | 
 
@jstone-lucasfilm , @niklasharrysson . Until we get some new filtering mechanism I'm adding in post filtering to exposed uniforms.

### DIFF
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -1082,6 +1082,12 @@ export class Material
                     {
                         continue;
                     }
+                    
+                    // Skip non-input types and anything > 2 levels deep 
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm (approved):
This looks like a useful workaround for now, thanks @kwokcb.

PR #1698, opened 2024-02-06
Add refract node

ld-kerley:
Adding `refract` node to partner with the `reflect` node, with a similar named interface, and implementation as a nodegraph taken from the OSL spec 
```
vector refract (vector I, vector N, float eta)
{
         float IdotN = dot (I, N);
         float k = 1 - eta*eta * (1 - IdotN*IdotN);
         return (k < 0) ? vector(0,0,0) : (eta*I - N * (eta*IdotN + sqrt(k)));
}
```
which also lines up with the [Kronos implementation](https://registry.khronos.org/OpenGL-Refpages/gl4/html/refract.xhtml) here too.

### DIFF
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -2767,7 +2767,7 @@
     The normal vector is assumed to be normalized.
   -->
   <nodedef name="ND_refract_vector3" node="refract" nodegroup="math" doc="Compute the refraction direction of an input vector.">
-    <input name="in" type="vector3" value="1.0,0.0,0.0" doc="Input vector to be reflected."  />
+    <input name="in" type="vector3" value="1.0,0.0,0.0" doc="Input vector to be refracted."  />
     <input name="normal" type="vector3" defaultgeomprop="Nworld" doc="Normal vector." />
     <input name="eta" type="float" value="1.0" doc="Ratio of the indices of refraction."/>
     <output name="out" type="vector3" />
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -2761,6 +2761,18 @@
     <output name="out" type="vector3" />
   </nodedef>
 
+  <!--
+    Node: <refract>
+    Compute the refraction direction of an input vector against a normal vector.
+    The normal vector is assumed to be normalized.
+  -->
+  <nodedef name="ND_refract_vector3" node="refract" nodegroup="math" doc="Compute the refraction direction of an input vector.">
+    <input name="in" type="vector3" value="1.0,0.0,0.0" doc="Input vector to be reflected."  />
+    <input name="normal" type="vector3" defaultgeomprop="Nworld" doc="Normal vector." />
+    <input name="eta" type="float" value="1.0" doc="Ratio of the indices of refraction."/>
+    <output name="out" type="vector3" />
+  </nodedef>
+
   <!-- ======================================================================== -->
   <!-- Adjustment nodes                                                         -->
   <!-- ======================================================================== -->
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -2768,7 +2768,7 @@
   <nodedef name="ND_refract_vector3" node="refract" nodegroup="math" doc="Compute the refraction vector">
     <input name="in" type="vector3" value="1.0, 0.0, 0.0" doc="Incident vector"  />
     <input name="normal" type="vector3" defaultgeomprop="Nworld" doc="Surface normal" />
-    <input name="eta" type="float" value="1.0" doc="Index of refraction" />
+    <input name="ior" type="float" value="1.0" doc="Index of refraction" />
     <output name="out" type="vector3" />
   </nodedef>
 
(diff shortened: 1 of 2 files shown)


libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -2761,6 +2761,18 @@
...
 
+  <!--
+    Node: <refract>
+    Compute the refraction direction of an input vector against a normal vector.
+    The normal vector is assumed to be normalized.
+  -->
+  <nodedef name="ND_refract_vector3" node="refract" nodegroup="math" doc="Compute the refraction direction of an input vector.">
+    <input name="in" type="vector3" value="1.0,0.0,0.0" doc="Input vector to be reflected."  />
+    <input name="normal" type="vector3" defaultgeomprop="Nworld" doc="Normal vector." />
+    <input name="eta" type="float" value="1.0" doc="Ratio of the indices of refraction."/>
```
kwokcb, line 2772: Is there an input range that applies here? Assume you don't want 0.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -2761,6 +2761,18 @@
     <output name="out" type="vector3" />
   </nodedef>
 
+  <!--
+    Node: <refract>
+    Compute the refraction direction of an input vector against a normal vector.
+    The normal vector is assumed to be normalized.
+  -->
+  <nodedef name="ND_refract_vector3" node="refract" nodegroup="math" doc="Compute the refraction direction of an input vector.">
+    <input name="in" type="vector3" value="1.0,0.0,0.0" doc="Input vector to be reflected."  />
```
kwokcb, line 2770: typo: "refracted" vs "reflected" :)

libraries/stdlib/stdlib_defs.mtlx
ld-kerley: Agreed, we want the value to be bigger than 0, but do we have a way to express that in the current metadata?  I only really see `uimin` and `uisoftmin`.  What would we consider a safe epsilon above zero? 

We're a bit inconsistent on  ui hints for IOR in other nodes, `dielectric_bsdf` and other nodes in pbrlib don't provide any.  `standard_surface` and `UsdPreviewSurface` say `uimin="0.0" uisoftmin="1.0"  uisoftmax="3.0"`, `gltf_pbr` says ` uimin="1" uisoftmax="3" `

I'm open to guidance on what to set here.

libraries/stdlib/stdlib_defs.mtlx
ld-kerley: pesky copy/paste bug - will fix- thanks

libraries/stdlib/stdlib_defs.mtlx
jstone-lucasfilm: I wouldn't worry about setting these ranges in the initial implementation, but going forward I would recommend the `uisoftmin="1.0"  uisoftmax="3.0"` of UsdPreviewSurface, since there are artistic use cases for values outside this range.

jstone-lucasfilm:
@ld-kerley @kwokcb One remaining question in my mind is whether this index of refraction should be named `ior` rather than `eta`, following the pattern in MaterialX nodes where the real component of IOR is named `ior` and the imaginary component is named `extinction`.  I believe this choice has historically been motivated by the goal of making node interfaces as artist-focused as possible, where `eta` and `kappa` are less artistically descriptive.

Here are two examples of the current pattern for IOR in MaterialX nodes:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/pbrlib_defs.mtlx#L136
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/pbrlib_defs.mtlx#L79

ld-kerley:
I think consistency is important - so +1 vote for `ior` here if that the prevailing direction for MaterialX.

kwokcb:
Correct me if I have the wrong interpretation but isn't `eta` the  ratio of the `source ior` / `destination ior` so maybe `relative_ior`  or `ior_ratio` ?

ld-kerley:
I alway refer to [The Bible](https://pbr-book.org/4ed/Reflection_Models/Specular_Reflection_and_Transmission#TheIndexofRefraction) whenever I'm confused. 

"The speed reduction is commonly summarized using the index of refraction (IOR). For example, a material with an IOR of 2 propagates light at half the speed of light. For common materials, the value is in the range 1.0–2.5 and furthermore varies with the wavelength of light. We will use the Greek letter , pronounced “eta,” to denote this quantity."

jstone-lucasfilm:
@kwokcb All of the IOR inputs in MaterialX nodes can be thought of as being relative to the IOR of the interior medium (which is 1.0 by default), but we don't make such a distinction in our naming conventions.

jstone-lucasfilm:
@kwokcb All of the IOR inputs in MaterialX nodes can be thought of as being relative to the IOR of the exterior medium (which is 1.0 by default), but we don't make such a distinction in our naming conventions.  Following that convention, I can see the name `ior` as being the most consistent here, though I'm open to other ideas.

kwokcb:
Sounds good to go with `ior`. And thanks for the `eta` reference @ld-kerley.

jstone-lucasfilm (approved):
This looks excellent, thanks @ld-kerley!

PR #1699, opened 2024-02-07
MDL implementation for viewdirection

krohmerNV:
Draft of an MDL implemention for mx_viewdirection.
This assumes that the viewdirection is defined by direction from camera position to shading point.
In the context of a path tracer, the camera position is used as well and NOT the ray origin of the current path segment.


jstone-lucasfilm (approved):

PR #1707, opened 2024-02-09
Remove arrayappend node definition

ld-kerley:
The `arrayappend` node has been removed from the 1.39 specification, so we're removing it from the codebase.

This node has never been implemented in any shader generator or application (we think), so it should be safe to remove without any upgrade functionality.


jstone-lucasfilm (approved):

PR #1709, opened 2024-02-12
Update specification/stdlib_defs for new nodes

dbsmythe:
Refactor procedural nodes into procedural, noise and shape nodegroups.
Add descriptions of new shape nodes to specification.
Add two colorspaces requested by Pixar USD team.


dbsmythe:
While I was adding these documentation changes, I had a few questions that should be discussed and resolved:

- randomfloat/randomcolor: why do these have an "in" input?  They also have a "seed", so what does "in" do?
- What does the "center" input for <line> do?
- Should the default value for "center" in <circle>, <cloverleaf> and <hexagon> be (0.5, 0.5) and not the currrent (0,0)?
- I would assume that <grid> produces white gridlines on a black background, is that correct?
- Description of the "staggered" input for <crosshatch>?

kwokcb:
FYI: As mentioned in the TSC sync : 

The `nodegroup` meta-data tag reliance can be seen [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/ShaderNode.cpp#L302 )for instance in code generation, and would hence "break" if `procedural2d` and `procedural3d` were changed. I'm mention this in the PR as well.

There a check for geometry stream identification as well which is also "brittle".

jstone-lucasfilm:
@dbsmythe Now that we've shifted our focus to the `dev_1.39` branch for development of MaterialX 1.39, would you mind retargeting this pull request to that branch, and we can prepare it to be merged there?

jstone-lucasfilm:
@dbsmythe Just bumping this thread, as it would be great to integrate these improvements into the `dev_1.39` branch, making them part of the 1.39 release this summer.

PR #1710, opened 2024-02-13
Align vector spaces for npr nodes

kwokcb:
## Issue

- Gooch was using `<normal>` in "object" space and `<viewdirection>` in "world" space.
This causes incorrect rendering when the geometric transform is not identity.
- The toon example was also doing this.

## Fix

- Correct to consistently use "world" space.


jstone-lucasfilm (approved):
This looks like a good catch, thanks @kwokcb.  Going forward, we should support viewdirection in all spaces in GLSL/MSL, but this fix should correctly align spaces in the meantime.

PR #1711, opened 2024-02-13
Redesign of TypeDesc class

niklasharrysson:
This change list implements a redesign of the `TypeDesc` class used for data type descriptions in MaterialX.

The main purpose of this change is to solve issue #1608. But the new design also gives performance improvements to the use of this class.

Since the `TypeDesc` class is used frequently many files are touched by this change and there are some API breakage. Function wrappers with `[[deprecated]]` are added to minimize API breakage, but it's not possible for everything that changed. For example, the `TypeDesc` instances are now stored and accessed by value rather then by pointer, requiring a change to the access syntax.

Here are the benefits to this new design:

- The `TypeDesc` class is now a POD type, and just 64-bits, making it suitable to be stored and passed by value.
- It is now 100% safe to use in multiple modules / shared libraries / from python. No pointers are used anywhere, only copies of the instances.
- All comparisons and hash operations are performed using a precomputed hash on the type name, so it's equally efficient as the old pointer compares.
- Performance in general is improved as the data locality is better, passing and accessing by value instead of by pointer.
- The global type instances are now created as compile time constants using `constexp`, so a compiler can optimize their use in various ways.
- Their size is minimized to 64-bits, so they use the same storage as with the old pointer method.
- We can use compare operators without pointer dereferencing. So `TypeA == TypeB` is used instead of the unpleasant `*TypeA == *TypeB` that we had to introduce as a quick workaround for issue 1608.

The cons are:

- The change is API breaking (I hope this amount of breakage is ok for v1.39?).
- The change requires C++17. We need support for `std::string_view` in `constexpr` in order to create the types as compile time constants (the move to C++17 was discussed at the last TSC meeting)
- `TypeDesc::getName()` has a worse performance then before. The name string is now stored in the global type registry rather than stored on the instances. So `getName()` requires a map lookup, but it's rarely used and I think the other improvements make up for this by far.


meshula:
I think the TSC discussed and approved moving to C++17 already? If I'm remembering that correctly, I think it would make sense to first land the C++17 requirement in a trivial separate PR and let that percolate through and build this PR on top of that.

niklasharrysson:
@meshula Yes, that was my impression as well that the TSC approved a move to C++17. And later discussions on Slack has confirmed this.

Your suggestion sounds good, I will pull out the C++ version upgrade to a separate PR.

Thanks

jstone-lucasfilm (approved):
This change looks good to me, thanks @niklasharrysson!

PR #1712, opened 2024-02-14
Move CPP standard to C++17

niklasharrysson:
Trivial change to move CPP standard to C++17 for v1.39. As agreed on by the MaterialX TSC.


jstone-lucasfilm (approved):

PR #1715, opened 2024-02-16
CMake configure based OpenImageIO

mikaelsundell:
@jstone-lucasfilm and @lgritz:
Accidentally deleted the previous PR after sync fork, sorry. Here is a fresh one with all changes. Kept the "plain" signature for target_link_libraries, need to change the previous target_link_libraries (33) to use keywords for targets MaterialXGenShader + ${CMAKE_DL_LIBS}, ok?

Deleted FindOpenImageIO.cmake + cleaned up a few of the non used variables, also fixed the deprecated call to read_image to include subimage and miplevel (2.0).


jstone-lucasfilm:
@mikaelsundell Since we're aiming to include this proposal in MaterialX 1.39, would you mind making this a pull request for the `dev_1.39` branch where active development on that version of MaterialX is ongoing?

mikaelsundell:
Yes! Will have it fixed in a day or two from now.

mikaelsundell:
Now pushed as #1734

PR #1717, opened 2024-02-17
Simplify render test suite

jstone-lucasfilm:
- Remove bake and wedge tests from the render suite, as they have side effects that obscure accurate comparisons across languages.  Moving forward, we should likely emphasize testing this functionality in Python, following our current approach using baketextures.py and translateshader.py in GitHub Actions.
- Simplify material and shader node names to improve clarity in render comparisons.


jstone-lucasfilm:
Attaching updated render comparisons between GLSL and OSL, with reference render settings active:

[MaterialXRenderTests_02_17_2024.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/14320466/MaterialXRenderTests_02_17_2024.pdf)

PR #1721, opened 2024-02-23
Improvements to color transform handling

jstone-lucasfilm:
- Add support for a wider set of use cases for color space bindings, including surface shader nodes with filename-type inputs.
- Add support for scoped color space bindings in shader translation.
- Add a test suite example that covers newly-supported use cases.

Supporting changes include:

- Merge the protected method ShaderGraph::createSurfaceShader into ShaderGraph::create, allowing surface shader nodes to be handled in the same fashion as other nodes.
- Update the interface of ShaderGraph::createNode, allowing greater flexibility in accessing parent nodes during shader generation.  We've omitted a deprecated wrapper for this API update, as it would not be straightforward to write one in a memory-safe fashion, but fortunately this method only appears to be used within the ShaderGraph class.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -93,7 +93,7 @@ void ShaderGraph::createConnectedNodes(const ElementPtr& downstreamElement,
     ShaderNode* newNode = getNode(newNodeName);
     if (!newNode)
     {
-        newNode = createNode(*upstreamNode, context);
+        newNode = createNode(upstreamNode, context);
     }
 
     //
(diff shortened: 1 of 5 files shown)


jstone-lucasfilm:
Attaching updated render comparisons between GLSL and OSL, with reference render settings active:

[Uploading MaterialXRenderTests_02_23_2024.pdf…]()

jstone-lucasfilm:
Attaching updated render comparisons between GLSL and OSL, with reference render settings active:

[MaterialXRenderTests_02_23_2024.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/14390924/MaterialXRenderTests_02_23_2024.pdf)

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -843,35 +670,47 @@ ShaderNode* ShaderGraph::createNode(const Node& node, GenContext& context)
...
+                populateColorTransformMap(colorManagementSystem, shaderOutput, sourceColorSpace, targetColorSpace, false);
                 populateUnitTransformMap(unitSystem, shaderOutput, input, targetDistanceUnit, false);
             }
-        }
-        else
-        {
-            ShaderInput* shaderInput = newNode->getInput(input->getName());
-            populateColorTransformMap(colorManagementSystem, shaderInput, input, targetColorSpace, true);
-            populateUnitTransformMap(unitSystem, shaderInput, input, targetDistanceUnit, true);
+            else
```
kwokcb, line 708: I think this logic is inconsistent? i.e. interface inputs are checked for filename but not for non-filenames ?

jstone-lucasfilm: This is a great question, and the reason that this change is only required for filename inputs has to do with the unusual way that we process color transforms for filename inputs, applying them to the **output** of the associated node.

Imagine the common case of an image node, where the colorspace is assigned to its file input, but the associated transform must be applied to the output of the image node, not to the literal filename-type input.

For this reason, our usual handling of interface names doesn't automatically apply to color space bindings for filename inputs, and we must manually handle this case in the logic of our color transform system.

PR #1725, opened 2024-02-28
Add MATERIALX_BUILD_MONOLITHIC option

ld-kerley:
Working towards building Frameworks for macOS and IOS, this PR adds the option to build MaterialX as a single monolithic library. 

CMake aliases are exported for the standard library targets aliased to the monolithic library to avoid changes being necessary downstream.


meshula:
I'm looking forward to this change;  it will make my life easier on iOS in particular, but I think the mx cmake script has already reached a threshold of complexity that IMO necessitates a refactoring. This new change consistently does essentially the same thing per library, which runs a maintenance risk, moving forward, of someone fixing a bug or making an improvement in one place but not all places. I would recommend that rather than repeating the pattern of checking for the monolithic build and setting setting project properties in every sub library, it would good to instead replace all instances of `add_library` with a new routine in the root CMakeLists.txt, `mx_add_library` (with appropriate parameters), so that the new pattern is applied identically by virtual of funneling through a single function.

ld-kerley:
Great idea @meshula - and turned out pretty clean I think - I stole some inspiration from USD. I think there is some further clean up that can be done on the MaterialXRenderXXX modules, but I'd like to merge this first and do that in a follow up PR.  The extracting the common logic out to the `mx_add_library()` function made the commonalities easier to see.

jstone-lucasfilm:
This looks like an excellent proposal to me, thanks @ld-kerley!

Rather than adding a new `Public.cmake` module, what would you think about adding the new `mx_add_library` function directly to the root level CMakeLists.txt file, just after the existing definition of `assign_source_group`?

Although I'm not against the idea of adding and maintaining an external CMake helper module, my thinking is that keeping the functions local makes them more visible to developers, and that we can take the step of making them external if needed further in the future.

meshula:
I do find this refactoring so much easier to follow, thank you for that :)

ld-kerley:
No strong opinion either way - I was just mirroring what USD did to be honest, but if you'd prefer to avoid the external cmake helper, I'm fine with moving it over.

jstone-lucasfilm (approved):
This looks like a significant improvement to me, thanks @ld-kerley!

PR #1727, opened 2024-03-04
Add color82 to generalized_schlick_bsdf

jstone-lucasfilm:
This changelist adds a color82 input to the generalized_schlick_bsdf node in MaterialX, with a full implementation in hardware shading languages (GLSL/MSL/ESSL) and simple fallback behavior in closure-based shading languages (OSL/MDL).

The real-time implementation is based on Naty Hoffman's generalization of Adobe Fresnel (https://renderwonk.com/publications/wp-generalization-adobe/gen-adobe.pdf), which is visually consistent with previous versions of generalized_schlick_bsdf when color82 is left at its default setting.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -21,6 +21,7 @@ struct FresnelData
 
     // Generalized Schlick Fresnel
     vec3 F0;
+    vec3 F82;
     vec3 F90;
     float exponent;
 
@@ -36,6 +37,7 @@ FresnelData(int   _model        = 0,
             vec3  _ior          = vec3(0.0f),
             vec3  _extinction   = vec3(0.0f),
             vec3  _F0           = vec3(0.0f),
+            vec3  _F82          = vec3(0.0f),
             vec3  _F90          = vec3(0.0f),
             float _exponent     = 0.0f,
             float _tf_thickness = 0.0f,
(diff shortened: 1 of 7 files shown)


jstone-lucasfilm:
Generalized Schlick BSDF with white color82:
GeneralizedSchlickBSDF_Color82_White

Generalized Schlick BSDF with gold color82:
GeneralizedSchlickBSDF_Color82_Gold

Generalized Schlick BSDF with blue color82:
GeneralizedSchlickBSDF_Color82_Blue

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
```diff
@@ -247,6 +249,17 @@ float mx_fresnel_dielectric(float cosTheta, float ior)
     return 0.5 * x * x * (1.0 + y * y);
 }
 
+// https://renderwonk.com/publications/wp-generalization-adobe/gen-adobe.pdf
+vec3 mx_fresnel_hoffman_schlick(float cosTheta, vec3 F0, vec3 F82, vec3 F90, float exponent)
+{
+    const float COS_THETA_MAX = 1.0 / 7.0;
+    const float COS_THETA_FACTOR = 1.0 / (COS_THETA_MAX * pow(1.0 - COS_THETA_MAX, 6.0));
```
niklasharrysson, line 256: For consistency you could use the added mx_pow6 function here as well.

niklasharrysson:
This looks good to me @jstone-lucasfilm, and I only had a nitpick comment above.

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
jstone-lucasfilm: That's a good note, @niklasharrysson, and I had wanted to use the approach you describe, but some GLSL compilers require that a `const` value be defined entirely in terms of constant expressions.

So I actually see a compile-time error in my GLSL environment if I use the non-const `mx_pow6` as a sub-expression of COS_THETA_FACTOR.

For simplicity, I went back to using the built-in `pow` function, but there may be a more clever way of notating `mx_pow6` so that it can also be used in this way.

niklasharrysson: Ah, that's fine, this was just a tiny thing I noticed.

PR #1729, opened 2024-03-07
Improvements to document validation

ld-kerley:
This PR adds two new pieces to the correctness checking within MaterialX, these were discovered while doing work on MaterialX library files.  I found them helpful, so I thought it might help to upstream the changes for others to benefit from too.

When loading libraries, there is now an optional boolean `errorOnDuplicates`, that will raise an error if a library contains an element is already present, previously this case would silently continue.  I did wrestle with if this should just be called `strict`, incase there are other cases we want to add in the future.  Open to input there.

Secondly I have added validation to `TypedElement` to validate if the type used has actually been registered in the document as a type.  This very helpfully tracked down a couple of typos in work-in-progress code. 

Finally I added the ability for the `mxvalidate.py` script to directly accept a path to alternate library locations.


source/MaterialXCore/Element.cpp
```diff
@@ -476,6 +476,31 @@ TypeDefPtr TypedElement::getTypeDef() const
...
+    bool res = true;
+
+    if (hasType())
+    {
+        // check that the type specified is declared as a type in the document
+        const auto& typeStr = getType();
+        bool isValidType = getDocument()->getTypeDef(typeStr) != nullptr;
+
+        if (!isValidType) {
+            // there is one special type "multioutput" that can be used on node
```
kwokcb, line 490: As I recall there is at least one one which has no type. I think it's the `background` or `dot` or some "application" node(s). It's special cased elsewhere but (sorry) I don't recalled where.

source/MaterialXCore/Document.cpp
```diff
@@ -266,7 +266,7 @@ NodeDefPtr Document::addNodeDefFromGraph(const NodeGraphPtr nodeGraph, const str
     return nodeDef;
 }
 
-void Document::importLibrary(const ConstDocumentPtr& library)
+void Document::importLibrary(const ConstDocumentPtr& library, bool errorOnDuplicates)
```
kwokcb, line 269: This error condition used to reside within the read options at some point in the past. It might be better to put it there instead ? AFAIK it still resides in the Autodesk fork so adding a ping for @ashwinbhat for any feedback.

kwokcb (commented):
Hi @ld-kerley 
A couple of background points:
1. This error condition did exist before and it sounds reasonable to add it back in but will defer to Jonathan.

2. There is a case to be made to allow the same libraries to be loaded in more than once and ignore duplicates. I think this came / comes up more when folks import libraries into 2 or more main documents and then try and merge them together. Also xincludes can do this -- though there is no "firewall" check to avoid duplicate xincludes.

3. There are some nodes without types. I left a note.

4. Type checking is useful to have but will fail validation if the std lib file is not loaded in which may be the case sometimes. I'm not sure what is the "best" way to handle this as types are not declared based on the stdlib file itself (data-driven). It's in C++ and also happens to be in the file. The same issue can occur for the unit typedef.

ld-kerley:
Thanks for the feedback @kwokcb.  This was certainly not posted as final work, and I wanted to get input on what I had thrown together internally to get me unstuck.

source/MaterialXCore/Document.cpp
ld-kerley: What I was actually trying to catch here was the same node definition being defined in two different library files, so I believe this wouldn't be possible to catch with a read options flag.  As I understood things the read options only apply to when a single file is loaded.

Adding the check here, allows clashes from different files to be validates.  I understand that there are certainly use-cases for not wanting this to be an error, hence the optional guard.

source/MaterialXCore/Element.cpp
ld-kerley: Thanks for flagging this, obviously if we decide to move forward with this, then I'll make sure we're passing all the unit tests, and hopefully that flags any existing corner cases as you mention.

ld-kerley:
After a brief chat with @jstone-lucasfilm I added a `ValidationOptions` class to pass around configuration for the validation stage, this way we can add additional validations, without changing existing behavior.

jstone-lucasfilm:
Just on this specific build error, it looks like we'd need to add the new argument for `Document::importLibrary` to its JavaScript binding, so that existing calls continue to work as before:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/JsMaterialX/JsMaterialXCore/JsDocument.cpp#L26

ld-kerley:
I took a stab at resolving the JS bindings - but I don't have a local build setup to build them, and I also don't have a lot of experience with them either - perhaps someone with more JS chops could give me some pointers (or patch this PR)

jstone-lucasfilm:
Thanks for taking a look, @ld-kerley, and either @kwokcb or I can help to resolve that last JavaScript issue.

jstone-lucasfilm:

jstone-lucasfilm:
@ld-kerley I'll close out this pull request for now, since it's framed in terms of a branch that we're planning to delete, but feel free to propose a similar pull request for the `main` branch in the future!

PR #1730, opened 2024-03-07
Integrate MDL enhancements to public repository

msuzuki-nvidia:
Merge internal enhancements into public repository.

This PR includes:

- Enable `multiscatter_tint`
- Fix bump was not working in `mx_generalized_schlick_bsdf`


source/MaterialXGenMdl/mdl/materialx/cm.mdl
```diff
@@ -0,0 +1,245 @@
+//
+// Copyright Contributors to the MaterialX Project
+// SPDX-License-Identifier: Apache-2.0
+//
+// Default color management transforms for MDL
```
jstone-lucasfilm, line 5: Although I see some good proposed changes in this pull request, I think the color management MDL file may be from an _earlier_ version of MaterialX, and that it was removed when we implemented the color management library as language-independent graphs:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1352

source/MaterialXGenMdl/mdl/materialx/stdlib_1_6.mdl
```diff
@@ -2872,6 +2872,66 @@ export core::color4 mx_screen_color4(
 	return core::color4(rgb,a);
 }
 
+float mx_overlay(float mxp_fg, float mxp_bg)
```
jstone-lucasfilm, line 2875: Similarly, I think this MDL function for the overlay node may be from an _earlier_ version of MaterialX, and it was removed when Rasmus Bonnedal implemented the overlay node as a language-independent graph:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1539

source/MaterialXGenMdl/mdl/materialx/cm.mdl
msuzuki-nvidia: Thank you for the review and giving the history. Removed this file.

source/MaterialXGenMdl/mdl/materialx/stdlib_1_6.mdl
msuzuki-nvidia: Thanks. Removed mx_overlay functions.

source/MaterialXGenMdl/mdl/materialx/stdlib_1_6.mdl
```diff
@@ -4426,6 +4426,71 @@ export float4 mx_noise2d_float4(
     return float4(v.x, v.y, v.z, w) * mxp_amplitude + mxp_pivot;
 }
 
+export color mx_noise2d_color3(
```
jstone-lucasfilm, line 4429: It looks like these MDL-specific definitions for type variations of noise nodes were also replaced by language-independent graphs in a pull request for MaterialX 1.38.8, so we don't need to restore them to the codebase!

https://github.com/AcademySoftwareFoundation/MaterialX/pull/1653

msuzuki-nvidia: 😓
Thanks, looks like we have not been synced our repo for a while. The extra noise functions are removed.

jstone-lucasfilm (approved):
This change looks good to me, thanks @msuzuki-nvidia!

PR #1731, opened 2024-03-08
Build API docs using a custom command with dependencies

ld-kerley:
I was bumping up against the documentation not building without using a custom target, and then stumbled across this super old issue (#9), and while reading it realized I think there's a way to have the best of both world.

Transitioning to using `add_custom_command` instead of `add_custom_target` avoids the "always build" issue, and adding `DEPENDS` arguments allows for the correct dependency tracking for the docs building, so it will only get triggered if necessary.  Also for what it's worth it didn't seem that the docs take that long to build anyway.

Anyway, just putting this up here for discussion, I don't know how many people are building the docs locally, but having it part of the regular cmake build without the need for an additional target seemed like a nice improvement.

I also changed the `find_package(Doxygen)` to add the `REQUIRED` keyword that way the cmake configure will fail if `doxygen` isn't present.  I think this is ok as the entire `documents/CMakeLists.txt` is guarded behind the `MATERIALX_BUILD_DOCS` option.


jstone-lucasfilm (approved):

PR #1735, opened 2024-03-11
Use CMake configuration of OpenImageIO in 1.39

mikaelsundell:
PR for CMake based configuration of OpenImageIO for dev_1.39


source/MaterialXRender/CMakeLists.txt
```diff
@@ -38,16 +38,11 @@ target_link_libraries(
...
-        include_directories(${OPENIMAGEIO_INCLUDE_DIR})
-        target_link_libraries(${MATERIALX_MODULE_NAME} ${OPENIMAGEIO_LIBRARIES})
-        # Also needed by MaterialXTest:
-        set(OPENIMAGEIO_FOUND "${OPENIMAGEIO_FOUND}" PARENT_SCOPE)
-        set(OPENIMAGEIO_INCLUDE_DIR "${OPENIMAGEIO_INCLUDE_DIR}" PARENT_SCOPE)
-        set(OPENIMAGEIO_LIBRARIES "${OPENIMAGEIO_LIBRARIES}" PARENT_SCOPE)
-        set(OPENIMAGEIO_LIBRARY_DIRS "${OPENIMAGEIO_LIBRARY_DIRS}" PARENT_SCOPE)
-        set(OPENIMAGEIO_ROOT_DIR "${OPENIMAGEIO_ROOT_DIR}" PARENT_SCOPE)
+    find_package(OpenImageIO CONFIG REQUIRED)
+    if(OpenImageIO_FOUND)
```
ld-kerley, line 42: Given you're using `REQUIRED` in the `find_package()` line above, I think this `if()` should never fail, as if the `find_package()` call fails to find `OpenImageIO` then it will raise a CMake configure error.

source/MaterialXRender/CMakeLists.txt
```diff
@@ -38,16 +38,11 @@ target_link_libraries(
...
+        target_link_libraries(${MATERIALX_MODULE_NAME} OpenImageIO::OpenImageIO OpenImageIO::OpenImageIO_Util)
+        # Also needed by MaterialXView:
+        set(OPENIMAGEIO_ROOT_DIR "${OpenImageIO_INCLUDE_DIR}/../" PARENT_SCOPE)
```
ld-kerley, line 45: Perhaps not completely related to this PR - but this line got me intrigued as to what `MaterialXView` was using `OPENIMAGEIO_ROOT_DIR` for, and it appears to be copying the binaries from OIIO in to the local cmake build folder.  I'm not exactly sure why this is, but if it's not necessary to do this anymore, we could clean up this line too.

If we need to keep it, then I might vote to move the `find_package()` call to a higher level CMakeLists.txt file, instead of using `PARENT_SCOPE` here, but thats more of a style choice than anything else.

ld-kerley:
Big fan of more modern CMake! thanks for this.

source/MaterialXRender/CMakeLists.txt
mikaelsundell: I agree, now fixed.

source/MaterialXRender/CMakeLists.txt
mikaelsundell: I agree, should I give it a try for review?

ld-kerley: Per @jstone-lucasfilm comment in slack - I would say yes. It's an easy enough thing to revert if it causes people issues.

jstone-lucasfilm: Agreed that this would be a great improvement, though I think it's fine to consider that as an orthogonal change to the current PR.

jstone-lucasfilm (approved):
This looks good to me, thanks @mikaelsundell, and let's consider the simplification of OIIO handling in the MaterialX Viewer to be a future improvement.

PR #1741, opened 2024-03-16
Use framework linkage for OpenGL on macOS

dgovil:
This commit fixes the output of the `MaterialXTargets.cmake` to have proper system linkage for OpenGL in the cmake config.
Without this change, downstream projects on a Mac like USD cannot build against MaterialX properly.

There was an issue introduced in b50dea8ef6c6eda52f7ecabafeb45272a48cfe11 where the output started being like so

```
set_target_properties(MaterialXRenderGlsl PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "GLAD_GLAPI_EXPORT"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "MaterialXRenderHw;MaterialXGenGlsl;OpenGL::GL;-framework Foundation;-framework Cocoa;-framework Metal"
)
```

Prior to that commit it would end up being a hardcoded path to the SDK which is also not great, because it means any SDK update would make the file unusable.

`/Library/Developer/CommandLineTools/SDKs/MacOSX13.5.sdk/System/Library/Frameworks/OpenGL.framework;`

Switching to `-framework` like the rest of the system links fixes that and makes it portable regardless of what SDK paths are on a system


jstone-lucasfilm:
This looks like a good improvement, thanks @dgovil!

Since we're currently focusing on MaterialX 1.39 for release this summer, would you mind retargeting this pull request to the `dev_1.39` branch, and we can review it there?

dgovil:
Changed the target to dev_1.39

jstone-lucasfilm:
@dgovil It should be fine to create a new 1.38.10 release as well, if there are important fixes that need to be included in the 1.38.x series, but so far we've been focused on the 1.39 release this summer.  Let us know if you believe this change is important enough to warrant an additional 1.38.10 release, and we can bring that up in the next MaterialX TSC.

dgovil:
I think it might be worth having a 1.38.10 to back port some fixes like this. My worry is this one makes it harder for macOS devs to use 1.38.9, and I'm not sure they'd be ready to move to 1.39

I can bring it up in the next TSC meeting.

jstone-lucasfilm (approved):
This looks good to me, thanks @dgovil!

PR #1744, opened 2024-03-20
Remove 'baked' suffix from TextureBaker names

ashwinbhat:
The texturebaker appended a "_baked" token suffix to the generated material names that changed material name. Some pipelines get impacted by this rename, since the original material name is lost. Removing the suffix to retain original material name.


jstone-lucasfilm (approved):

PR #1750, opened 2024-03-25
Move light uniforms into referencing GLSL files

jstone-lucasfilm:
This change moves the new environment light intensity uniforms into their referencing GLSL files, addressing shader compilation errors when GenOptions::hwDirectionalAlbedoMethod is set to DIRECTIONAL_ALBEDO_TABLE or DIRECTIONAL_ALBEDO_MONTE_CARLO by the client.

Additionally, the change moves the environment light intensity option from the UI to the command line of the viewer, since isolated adjustments of the indirect lighting term are an unintuitive rendering control for most users.


ashwinbhat (approved):
Thanks for the update @jstone-lucasfilm I'm fine with moving the control from UI to command line.

PR #1752, opened 2024-03-26
Use framework linkage for OpenGL on macOS

dgovil:
Backporting https://github.com/AcademySoftwareFoundation/MaterialX/pull/1741 to Main for 1.38.10


jstone-lucasfilm (approved):
Thanks for the fix, @dgovil!

PR #1754, opened 2024-03-26
Improve shader source duplication tests

jstone-lucasfilm:
This changelist improves the robustness of shader source duplication tests in MaterialXGenShader, and adds a new example material that highlights edge cases in the original logic.


JGamache-autodesk:
Yes, we just hit that one in MaterialX 1.38.10. Create a MaterialX shader that has both smoothstep_float and smoothstep_color3 nodes and this will hit the issue that was fixed in the ShaderSourceNode.

I need to backport this in MayaUSD without touching the ShaderStage class, which can probably be done by using `shadergen.emitLibraryInclude()` if I am not mistaken.

PR #1756, opened 2024-03-29
GLSL improvements to Oren-Nayar diffuse

jstone-lucasfilm:
- Remove an extra factor of PI in the GLSL computation of Oren-Nayar diffuse, improving its alignment with corresponding implementations in OSL and MDL.
- Include the required directional albedo factor when computing GLSL environment lighting for Oren-Nayar diffuse.
- Add an analytic approximation for the directional albedo of Oren-Nayar diffuse, using a rational quadratic fit to Monte Carlo data.
- Generalize the interfaces of mx_oren_nayar_diffuse and mx_burley_diffuse, enabling their use in tangent-space computations.


jstone-lucasfilm:
Here are the resulting render comparisons between GLSL and OSL for the code in this changelist:

[MaterialXRenderTests_03_29_2024.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/14808097/MaterialXRenderTests_03_29_2024.pdf)

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
```diff
@@ -2,28 +2,87 @@
...
 {
-    float LdotV = clamp(dot(L, V), M_FLOAT_EPS, 1.0);
-    float NdotV = clamp(dot(N, V), M_FLOAT_EPS, 1.0);
     float s = LdotV - NdotL * NdotV;
     float stinv = (s > 0.0f) ? s / max(NdotL, NdotV) : 0.0;
 
-    float sigma2 = mx_square(roughness * M_PI);
+    float sigma2 = mx_square(roughness);
     float A = 1.0 - 0.5 * (sigma2 / (sigma2 + 0.33));
     float B = 0.45 * sigma2 / (sigma2 + 0.09);
```
portsmouth, line 13: Note that this implementation is not actually the Fujii model as described in his blog post.

The correct form for that would be:
```glsl
    float stinv = (s > 0.0f) ? s / max(NdotL, NdotV) : s;
    float A = 1.0 / (1.0 + (0.5 - 2.0/(3.0*PI)) * sigma);
    float B = sigma * A;
    return A + B * stinv;
```
Otherwise this is just "qualitative Oren-Nayar", which has dark circle artifacts. (Both models suffer from fairly extreme darkening also).

And I think the `sigma` should just equal `roughness`, there's no need for the squaring, at least there's nothing in the blog post or other literature that suggest that should be done.

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
```diff
@@ -2,28 +2,87 @@
...
 
     return A + B * stinv;
 }
 
+float mx_oren_nayar_diffuse_dir_albedo_monte_carlo(float NdotV, float roughness)
+{
+    NdotV = clamp(NdotV, M_FLOAT_EPS, 1.0);
+    vec3 V = vec3(sqrt(1.0f - mx_square(NdotV)), 0, NdotV);
+
```
portsmouth, line 21: I've derived an analytical formula for this albedo, which may be preferable to this (64 iteration!) loop.

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
jstone-lucasfilm: For now, we're aligning the GLSL implementation of Oren-Nayar with the OSL implementation, which is defined here, and let me know if you notice any divergence between the two:

https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/7fe43cec6e947e584cecfda426d83a77df5af05a/src/testrender/shading.cpp#L479

To your second point, we're treating sigma and roughness as synonyms here, and we're only squaring the sigma value to generate sigma2 (the square of sigma).

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
jstone-lucasfilm: I'd definitely be interested in the analytical formula for Oren-Nayar directional albedo, though we'll need to measure its performance carefully against the analytic approximation we're currently using.

Note that this Monte-Carlo computation is only used to generate data tables, which are then used as source data for the analytic fit in MATLAB.

We greatly increase this sample count when generating data for the analytic fit, but I've left the sample count at 64 to enable real-time comparisons between the analytic fit and the Monte Carlo results.

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
jstone-lucasfilm: I've updated the comment to clarify that OSL is the source for the GLSL implementation, and have removed the reference to Fujii.

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
jstone-lucasfilm: We may want to start with our rational quadratic approximation for now, since it can exactly match the math in the GLSL/OSL implementation of Oren-Nayar, and we'd need to optimize the analytic Fujii version before it would be real-time friendly.

But it's really valuable to know that such an analytic solution is possible, and perhaps we can switch to this approach in a future iteration of Oren-Nayar across all languages.

portsmouth: Using Mathematica I found a similarly fast approximate expression for the albedo of the Fujii model (FON), expressing the G function as a polynomial in the direction cosine:
```glsl

float FON_albedo_fast_approx(float mu, float sigma)
{
    #define constant1_FON (0.5 - 2.0/(3.0*PI))
    #define Gcoeffs_FON mat2(0.287021, -0.17486, -0.306961, 0.193945)
    #define GoverPI_FON dot((Gcoeffs_FON * vec2(1,mu)) * vec2(1,mu*mu), vec2(1,1))
    return (1.0 + GoverPI_FON) / (1.0 + constant1_FON*sigma);
}
```

This has fractional error of at most about 6e-4 or 0.06% over all angles. (In comparison, the MC fit here of QON has RMSE 0.0028, which I assume corresponds to roughly max fraction error 3e-3 or 0.3%). 

image

I checked the assembly in https://shader-playground.timjones.io/ and this Fujii fit also has fewer instructions than the MC fit (2 adds, 3 muls, 1 dot, 1 div versus 3 adds, 5 muls, 1 div), so should be faster.

https://shader-playground.timjones.io/746a35c803f9af419ba7b846b67490a5

However, I wasn't able to find as good a fit for the qualitative Oren-Nayar model (QON), as in your implementation here. The error is about 3% for the same instruction count as my FON approx. (Seems the functional form of the Fujii model albedo is more easily fitted to polynomials, perhaps to be expected since it was designed to avoid discontinuities).

This further adds to my feeling that in future models (e.g. OpenPBR) we should switch to the Fujii form.

niklasharrysson (approved):
This change looks great @jstone-lucasfilm. 

(Only review of the changes to make GLSL match OSL here. The ideas to improve on this ON model sounds great too, but I suppose that's a larger discussion and will apply to all target languages).

PR #1757, opened 2024-04-01
Enable color and unit transforms on root nodes

jstone-lucasfilm:
This changelist adds support for color and unit transforms on the root nodes that are passed into shader generation, where previously they were only supported on non-root nodes in the material graph.

In MaterialX 1.38.9, a special-case pathway for root surface shader nodes was merged into the main logic, causing color transforms in some USD/MaterialX contexts to be omitted, and this changelist restores and generalizes the original approach to handle all root nodes.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -569,6 +569,7 @@ ShaderGraphPtr ShaderGraph::create(const ShaderGraph* parent, const string& name
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:
New GLSL/OSL renders using the code in this changelist:

[MaterialXRenderTests_04_01_2024.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/14829313/MaterialXRenderTests_04_01_2024.pdf)

PR #1762, opened 2024-04-02
Add missing vertex data prefix to Vulkan ShaderGen

klucknav:
@clach found that the VkShaderGenerator::getVertexDataPrefix() is returning the empty string when it seems like it should be returning vertexData.GetInstance() + "." similar to the glsl and metal shader gens do.

### DIFF
--- a/source/MaterialXGenGlsl/VkShaderGenerator.cpp
+++ b/source/MaterialXGenGlsl/VkShaderGenerator.cpp
@@ -69,9 +69,9 @@ void VkShaderGenerator::emitInputs(GenContext& context, ShaderStage& stage) cons
     }
 }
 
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm (commented):
Thanks for this fix, @klucknav!  I'm CC'ing @ashwinbhat for additional thoughts from the Autodesk perspective, but this looks fine to me.

ashwinbhat:
> Thanks for this fix, @klucknav! I'm CC'ing @ashwinbhat for additional thoughts from the Autodesk perspective, but this looks fine to me.

Looks good to me. Thanks for fix @klucknav @clach

jstone-lucasfilm (approved):

PR #1769, opened 2024-04-09
Add API change document (proposal)

kwokcb:
## Proposal

- Add a short document explaining how to make API breaking changes.
- Adds in C++, Python and Javascript examples for when:
  - The new method has the same name as the old one (signature change only)
  - The new method has a different name than the old one (replace old method)


jstone-lucasfilm:
@kwokcb Let's hold off on this contribution for now, since we're still working through the details of how deprecated methods should be contributed to MaterialX, and this proposal isn't yet quite aligned with our existing documentation on API changes:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/MainPage.md#materialx-api-changes

PR #1770, opened 2024-04-10
Improve Robustness of Input Connection Setting

kwokcb:
## Issue Addressed

This change adds detection for the case where both a connection and a value are specified, adds and update connection APIs to enforce not creating invalid documents.

## Changes

- Adds a new `Input::setConnectedInterfaceName()` interface which clears the `value` attributes to avoid accidently creating invalid documents.
- Updates other existing connection APIs to also clear the `value` attribute appropriately.
- Updates validation on an input to mark when both a value and a connection are created.

## Tests
- Update invalid documents:
  - Fix existing error in one of the render test suite files (`color_management.mtlx`).
  - `mxvalidate.py` and code generation tests caught this error. Ran `mxvalidate` on everything in `libraries` and `resources`.
- Add a new `node` unit test to check interface setting and validity checking.
- Change the node graph editor to use the new interface and remove previous manual `value` and `interfacename` clearing calls.


kwokcb:
@JGamache-autodesk, @zicher3d. I believe the change should address both the lack of validation and make the interface more robust in general and in the node editor.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1869,7 +1869,6 @@ void Graph::copyInputs()
                         }
 
                         copyNode->inputPins[count]->setConnected(true);
-                        copyNode->inputPins[count]->_input->removeAttribute(mx::ValueElement::VALUE_ATTRIBUTE);
```
kwokcb, line 1872: Just removing the value can actually leave a state of no value and no connection.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1834,7 +1834,7 @@ void Graph::copyInputs()
                             else if (upNode->getInput())
                             {
 
-                                copyNode->inputPins[count]->_input->setInterfaceName(upNode->getName());
+                                copyNode->inputPins[count]->_input->setInterfaceInput(upNode->getName());
```
kwokcb, line 1837: Every call is now "save" as it will remove the `value` attribute if a non-empty name is set, otherwise it will remove the `interfacename`. A `interfacename` with an empty string could also be set which could create an invalid doc.

source/MaterialXTest/MaterialXCore/Node.cpp
```diff
@@ -32,6 +32,57 @@ bool isTopologicalOrder(const std::vector<mx::ElementPtr>& elems)
...
+    std::string validationErrors;
+
+    mx::FileSearchPath searchPath = mx::getDefaultDataSearchPath();
+    mx::DocumentPtr doc = mx::createDocument();
+    mx::loadLibraries({ "libraries" }, searchPath, doc);
+
+    // Test inside nodegraph
+    mx::GraphElementPtr nodegraph = doc->addNodeGraph("graph1");
+
+    std::vector<mx::GraphElementPtr> graphs = { doc, nodegraph };
```
kwokcb, line 46: Test within a nodegraph and at top level (document).

resources/Materials/TestSuite/stdlib/color_management/color_management.mtlx
```diff
@@ -54,14 +54,14 @@
       <input name="file" type="filename" value="resources/Images/grid.png" colorspace="adobergb" />
     </image>
     <standard_surface name="image_adobergb_standard_surface7" type="surfaceshader">
-      <input name="base_color" type="color3" value="1, 1, 1" nodename="image_adobergb" />
+      <input name="base_color" type="color3" nodename="image_adobergb" />
```
kwokcb, line 57: This is now caught.

source/MaterialXCore/Interface.cpp
```diff
@@ -301,6 +304,23 @@ NodePtr Input::getConnectedNode() const
     return PortElement::getConnectedNode();
 }
 
+void Input::setInterfaceInput(const string& interfaceName)
+{
+    if (interfaceName != EMPTY_STRING)
```
kwokcb, line 309: Does proper validity checks for existence of interface and clears properly by removing the `interfacename` (disallows setting to empty string).

source/MaterialXCore/Interface.cpp
```diff
@@ -339,6 +359,7 @@ bool Input::validate(string* message) const
     {
         bool hasValueBinding = hasValue();
         bool hasConnection = hasNodeName() || hasNodeGraphString() || hasOutputString() || hasInterfaceName();
+        validateRequire(!(hasValueBinding && hasConnection), res, message, "Node input binds to both a value and a connection");
```
kwokcb, line 362: Additional check for value + connection.

JGamache-autodesk, line 362: What about an input that has both `interfacename`  and `nodename`? Should really be "only one of". You can do the combinatorial, or use an integer:
```C++
int numBindings = 0;
if (hasValue()) numBindings ++;
if (hasNodeName()) numBindings ++;
if (hasNodeGraphStrings()) numBindings ++;
if (hasInterfaceName()) numBindings ++;
if (hasOutputString() && !(hasNodeName() || hasNodeGraphString()))  numBindings ++;
validateRequire(numBindings, res, message, "Node input binds no value or connection");
validateRequire(numBindings <= 1, res, message, "Node input has too many bindings");
```

JGamache-autodesk (approved):
Very good. Thanks!

source/MaterialXCore/Interface.cpp
kwokcb: Thanks for the logic! I've replaced it with your code and added in additional attribute removes.

JGamache-autodesk (approved):

python/Scripts/mxvalidate.py
```diff
@@ -13,63 +13,78 @@ def main():
     parser.add_argument("--resolve", dest="resolve", action="store_true", help="Resolve inheritance and string substitutions.")
     parser.add_argument("--verbose", dest="verbose", action="store_true", help="Print summary of elements found in the document.")
     parser.add_argument("--stdlib", dest="stdlib", action="store_true", help="Import standard MaterialX libraries into the document.")
-    parser.add_argument(dest="inputFilename", help="Filename of the input document.")
```
jstone-lucasfilm, line 16: Although it's worthwhile to consider extending our `mxvalidate.py` script from individual documents to folders, this seems distinct from the current proposal to make input connections more robust, and I'd recommend omitting it here.

kwokcb: Sounds good. I can open a different PR.

source/MaterialXCore/Interface.cpp
```diff
@@ -301,6 +306,26 @@ NodePtr Input::getConnectedNode() const
     return PortElement::getConnectedNode();
 }
 
+void Input::setInterfaceInput(const string& interfaceName)
+{
+    if (interfaceName != EMPTY_STRING)
```
jstone-lucasfilm, line 311: Minor: I'd recommend expressing this as `!interfaceName.empty()` for consistency and efficiency.

source/MaterialXCore/Interface.h
```diff
@@ -247,6 +247,12 @@ class MX_CORE_API Input : public PortElement
     /// Return the node, if any, to which this input is connected.
     NodePtr getConnectedNode() const override;
 
+    /// Set the input interface. If the interface name specified is not empty,
+    /// the the interface is set if a corresponding input exists in the graph or document
+    /// containing the input's node. If the interface name is empty then the 
+    /// interface reference removed.
+    void setInterfaceInput(const string& interfaceName);
```
jstone-lucasfilm, line 254: The proposed name and documentation for this method seem very unclear to me, and it's my sense that it would confuse a new developer.

Maybe what you really want is to refactor the existing `setInterfaceName` method, allowing it to both set the interface name string and clear all previous bindings?  I see that this method is currently on `ValueElement`, so that it's available for `Token` elements as well, but this current structure doesn't need to remain as it stands, and you could potentially move it to any appropriate level of the class hierarchy to achieve the goal of an improved `setInterfaceName` method that works as expected for all classes.

jstone-lucasfilm (commented):
This looks really promising, thanks @kwokcb, and I had a few thoughts on how this could be clarified through refactoring.

JGamache-autodesk (commented):
Unsure about the naming @kwokcb. I would have expected a function taking a string parameter to end with `String` as in `setInterfaceString()`. This would make it consistent with `setNodeGraphString()` and `setOutputString()`.

I would expect a function named `setConnectedInterface()` to instead take an `InputPtr`, just like `setConnectedNode()`.

Can you change either the name or the parameter type? From what I see in the remainder of the edits, passing the InputPtr looks feasible.

jstone-lucasfilm:
I see your point, @JGamache-autodesk, and I wonder if `setConnectedInterfaceName` might be a better fit here.  If a similar method taking an `InputPtr` were later added, then it could be named `setConnectedInterfaceInput`.

I'm definitely open to additional thoughts on naming conventions, and my sense is that we're writing the beginnings of a connection-focused API for MaterialX, where syntactic details are handled automatically under the hood.

kwokcb:
Yeah the API naming for whatever reason is pretty inconsistent.  

The "String" post-fix attached which always confused me I think only 1 of the set of methods use this.

I'd actually like to keep it a "logical" name instead of an "implementation argument" name. i.e. The logical action is to set a connection akin to how USD does not specify the argument as part of [it's API](https://openusd.org/dev/api/class_usd_shade_connectable_a_p_i.html)

Thus, if we're going with a forward-looking name, my suggestion would `connectToInterfaceInput()` as it explicitly is an action to connect to a input interface. It can be a string, or an `InputPtr` -- though I'd prefer to add in a real "dag path" class and use that instead -- like how `FilePath` was added way back to replace strings for file paths.

(BTW, I was "lazy" and just added a `setInterfaceInput` to match the existing `getInterfaceInput` method).

jstone-lucasfilm:
@kwokcb Just to clarify the rules for the usage of `string` in MaterialXCore accessors, it's used when the bare name of an accessor would overlap the name of a MaterialX class, e.g. `setOutput`.

In these cases, we always add the word `string` to clarify the type of the argument (e.g. `setOutputString`, keeping the door open to future accessors that take the full class as their argument.

Given that precedent, I'd agree with @JGamache-autodesk that a method named `setConnectedInterfaceInput` would be expected to take an `InputPtr`, so it may be more appropriate to assign this method a name along the lines of `setConnectedInterfaceName`.

Looking towards a v2.0 release further in the future, we should indeed consider refactoring our API more fully, taking into account all of the learnings from our v1.x approach.  But for now I'd recommend that we follow our own precedent for method names, and not make piecemeal changes that suggest new naming conventions without a full v2.0 redesign.

kwokcb:
I didn't know about the `String` post-fix convention.  `setConnectedInterfaceName()` it is then.

JGamache-autodesk (approved):

jstone-lucasfilm:
@kwokcb We've now merged development work on MaterialX 1.39 back to the `main` branch of MaterialX, in preparation for the final weeks of development on the 1.39.0 release.  When you have a chance, could you retarget this pull request back to the `main` branch as well?

PR #1771, opened 2024-04-10
JavaScript support for additional shader generators

kwokcb:
## Proposal

Make it possible to build or deploy a non-ESSL shader generation for the Javascript module: MaterialxGenShader.

## Changes

Allow building a deployment of other "core" code generators based on build flags.
Remove forced disabling of generators in root Cmake file.


kwokcb:
Hi @jstone-lucasfilm .
Do you think this would be possible to get in for 1.39? This is the only current difference I have in order to build the web utilities on top of. Thanks.

jstone-lucasfilm:
@kwokcb What are you imagining as the use case for this feature, where a JavaScript application would generate OSL or MDL code?

kwokcb:
OSL can be transpiled to hardware shading languages so is worthwhile to consider. I think MDL could be dropped -- I already don't include it in my local builds.

jstone-lucasfilm:
@kwokcb Is there a specific use case that you've run into, where this additional functionality in MaterialX would be beneficial?

Since this proposal adds new source source code, I'm just aiming to understand the potential benefits of the change to developers and users.

jstone-lucasfilm:
@kwokcb Is there a specific use case that you've run into, where this additional functionality in MaterialX would be beneficial?

Since this proposal adds new source code, I'm aiming to understand the potential benefits of the change to developers and users, so that we can weigh this against the additional code complexity.

kwokcb:
The aim was to provide a way to see the output for supported languages and see what different code generation options / optimizations will do for different shader graphs and be able to examine and pull out source + shader reflection data. This could be used for shader uniqueness comparison or transpiling.

I don't think it really adds any code complexity. It just allows builds to have different generator options. Currently you have to create an ESSL code generator explicitly (same as in C++ and Python). BTW, to me, the real complexity is the mixing of all languages code + definitions which would be a second step to only pack / load required dependencies for a given language.

ld-kerley:
I think this idea a lot - would allow the creation of a website like godbolt, but for MaterialX files.

jstone-lucasfilm:
Got it, and I'm sold on the idea of Godbolt for MaterialX files.  Let's move forward with this change, and we can see how it ends up being used in the future.

jstone-lucasfilm (approved):

PR #1773, opened 2024-04-12
Add ability to validate folders of documents to mxvalidate.py

kwokcb:
## Issue

It is cumbersome to try and validate a set of documents using `mxvalidate`.

## Change

- Add ability to scan folders for MaterialX documents for `mxvalidate`.
- It will also be more efficient as standard libraries will only be loaded once for the set of documents to validate.

### DIFF
--- a/python/Scripts/mxvalidate.py
+++ b/python/Scripts/mxvalidate.py
@@ -4,7 +4,7 @@ Verify that the given file is a valid MaterialX document.
 '''
 
 import argparse
-import sys
+import sys, os
 
 import MaterialX as mx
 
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This is an interesting proposal, @kwokcb, and my main hesitation is that it splits our `mxvalidate.py` example script into two separate modes (file and folder), and we haven't taken this approach with any of our other examples.

If I ran into a production situation that required batch validation, my initial instinct would be to use `mxvalidate.py` in a `for` loop, and of course you can move beyond our example scripts to custom Python code if more sophisticated logic is required.

What are your thoughts?

jstone-lucasfilm:
@kwokcb We've now merged development work on MaterialX 1.39 back to the `main` branch of MaterialX, in preparation for the final weeks of development on the 1.39.0 release.  When you have a chance, could you retarget this pull request back to the `main` branch as well?

jstone-lucasfilm:
@kwokcb I'll close out this pull request for now, since it's framed in terms of a branch that we're planning to delete, but feel free to propose similar changes to the `main` branch if you're still interested in this idea!

PR #1777, opened 2024-04-17
Add ND_add_integer and ND_subtract_integer

ld-kerley:
Adding integer variations of `add` and `subtract`.


jstone-lucasfilm (approved):
This looks good to me, @ld-kerley, and I'm CC'ing @dbsmythe in case he has additional thoughts!

PR #1779, opened 2024-04-17
Add matrix type switch nodes

ld-kerley:
* `ND_switch_matrix33`
* `ND_switch_matrix44`
* `ND_switch_matrix33I`
* `ND_switch_matrix44I`


jstone-lucasfilm (commented):
This looks great to me, thanks @ld-kerley, and I'm CC'ing @krohmerNV and @dbsmythe for their thoughts on the MDL implementation and specification revisions.

jstone-lucasfilm (approved):

krohmerNV:
@jstone-lucasfilm this will probably not compile with MDL 1.6 and 1.7 because the stdlib_1_6.mdl needs the definitions too (or it should be there only).. anyhow I'll fix this as soon as I can.

But a general question. Isn't it possible to implement the switch in mtlx directly using a graph? this hard-coded number of options doesn't look nice.

ld-kerley:
@krohmerNV - are the additions you're suggesting are needed to stdlib_1_6.mdl in this PR already - [here](https://github.com/AcademySoftwareFoundation/MaterialX/pull/1779/files#diff-5bc92d18ace315927743ebdf41c9028565fb8e5ab9a1f1713bd39ef07b638943R3768) ?

Or are other changes needed?

krohmerNV:
@ld-kerley, you are right. That's what I meant. Sorry. 
I'll check locally to verify and let you know.

PR #1793, opened 2024-04-28
Remove swizzle node for 1.39

jstone-lucasfilm:
This changelist removes the swizzle node from the MaterialX data libraries for v1.39, adding an upgrade pathway for instances of this node in legacy documents.

Additionally, it addresses a minor inconsistency in the specification for the extract and extractrowvector nodes, which take zero-based indices rather than positive-valued selections.


jstone-lucasfilm:
Attaching updated comparison renders in GLSL and OSL:

[MaterialXRenderTests_04_28_2024.pdf](https://github.com/AcademySoftwareFoundation/MaterialX/files/15143611/MaterialXRenderTests_04_28_2024.pdf)

source/MaterialXCore/Version.cpp
```diff
@@ -1077,6 +1097,107 @@ void Document::upgradeVersion()
                     }
                 }
             }
+            else if (nodeCategory == "swizzle")
+            {
+                InputPtr inInput = node->getInput("in");
+                InputPtr channelsInput = node->getInput("channels");
+                if (inInput &&
```
ld-kerley, line 1104: Do you need to handle the case where `inInput` doesn't exist?  All the node defs appear to return `0`, but it's possible the `swizzle` node was wired in to a port that doesn't have `0` as its default?

jstone-lucasfilm:
@ld-kerley Absolutely, and there's a dedicated file in the MDL shader generation system that can now be removed as well, but I'm taking this change in stages!

ld-kerley (commented):
The upgrade function is really cleanly written :) - I was really worried it would be a lot more complex than that. Nice work!

source/MaterialXCore/Version.cpp
jstone-lucasfilm: That's a good point, and perhaps we should handle this case by converting the `swizzle` node to a `constant` node containing the appropriate values.  I haven't encountered this case in our test assets, but it's theoretically possible.

PR #1796, opened 2024-04-30
Remove swizzle support code

jstone-lucasfilm:
Following PR #1793. which removes the swizzle node from the MaterialX data libraries, this changelist removes unused support code for swizzle nodes in GLSL, MSL, OSL, and MDL shader generation.


jstone-lucasfilm:
I'm CC'ing @krohmerNV for his thoughts on this proposed change, since some of the largest changes are in the MDL support libraries, where a sizeable codebase was previously required to support swizzle nodes.

krohmerNV:
looks like all swizzles are removed, so I would replace the usages of the float2 swizzles left in the MDL code by a temporary variable and a constructor float2(t.x, t.y). 
I guess this also needs a bit of testing on my side. I'm not sure about the field access operators on structures which are similar to swizzles in the MDL code generator.

jstone-lucasfilm:
That's a great suggestion, @krohmerNV, though it might be slightly above my level of MDL knowledge, as some of the instances of `::swizzle::xy` were inlined in parameter lists.  Would it be reasonable for us to start with the current simplification, and we can remove `mx_swizzle_xy` in a future step?

PR #1800, opened 2024-05-05
Update implementations of extract and separate

jstone-lucasfilm:
This changelist updates the implementations of the extract and separate nodes, removing their dependence on channels attributes.

This is a first step towards converting channels attributes to nodes in the upgrade logic for MaterialX 1.39.


jstone-lucasfilm:
I've had a chance to validate these changes in GLSL and OSL, and I'm CC'ing @krohmerNV for his thoughts on the proposed changes in MDL.

PR #1821, opened 2024-05-18
Fix handling of image default values in MDL

msuzuki-nvidia:
Fixed an issue where the `<image>` node did not return the correct value in MDL back-end when the file path was invalid.


jstone-lucasfilm (approved):
Thanks for the fix, @msuzuki-nvidia!

PR #1822, opened 2024-05-19
Initial support for EON diffuse

jstone-lucasfilm:
This changelist adds initial support for the EON diffuse model in MaterialX, enabled through the use of an `energy_compensation` option on the `oren_nayar_diffuse_bsdf` node.

For now, MaterialX shader generation only supports the new option in GLSL, ESSL, and MSL, with other shading languages to be added once this initial implementation has been vetted by the community.

See the OpenPBR whitepaper at openpbr.org for full details on the EON diffuse model, including its formal definition and goals.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
@@ -128,7 +128,6 @@ vec3 mx_oren_nayar_compensated_diffuse(float NdotV, float NdotL, float LdotV, fl
 
 vec3 mx_oren_nayar_compensated_diffuse_dir_albedo(float cosTheta, float roughness, vec3 color)
 {
-    float A = 1.0 / (1.0 + FUJII_CONSTANT_1 * roughness);
     float dirAlbedoFujii = mx_oren_nayar_fujii_diffuse_dir_albedo(cosTheta, roughness);
     float avgAlbedoFujii = mx_oren_nayar_fujii_diffuse_avg_albedo(roughness);
     vec3 colorMultiScatter = mx_square(color) * avgAlbedoFujii /
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
@@ -1,7 +1,7 @@
 #include "mx_microfacet.glsl"
 
-// Based on the implementation of Oren-Nayar diffuse in Open Shading Language.
-// https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/main/src/testrender/shading.cpp
+// Qualitative Oren-Nayar diffuse with improvements from Fujii:
+// https://mimosa-pudica.net/improved-oren-nayar.html
 float mx_oren_nayar_diffuse(float NdotV, float NdotL, float LdotV, float roughness)
 {
     float s = LdotV - NdotL * NdotV;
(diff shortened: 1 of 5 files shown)


jstone-lucasfilm:
CC'ing @portsmouth, @peterkutz, @fpsunflower, and @niklasharrysson for their thoughts and recommendations.

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
```diff
@@ -1,7 +1,7 @@
 #include "mx_microfacet.glsl"
 
-// Based on the implementation of Oren-Nayar diffuse in Open Shading Language.
-// https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/main/src/testrender/shading.cpp
+// Qualitative Oren-Nayar diffuse with improvements from Fujii:
+// https://mimosa-pudica.net/improved-oren-nayar.html
 float mx_oren_nayar_diffuse(float NdotV, float NdotL, float LdotV, float roughness)
 {
     float s = LdotV - NdotL * NdotV;
```
portsmouth, line 7: To match the Fujii model, the code for `s` needs to be:
```glsl
float stinv = (s > 0.0f) ? s / max(NdotL, NdotV) : s;
//float stinv = (s > 0.0f) ? s / max(NdotL, NdotV) : 0.0;   // <-- not this
```

fpsunflower:
Related PR on the OSL side: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1817

I've adopted the same "energy_compensation" int parameter.

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
```diff
@@ -79,6 +79,33 @@ float mx_oren_nayar_diffuse_dir_albedo(float NdotV, float roughness)
     return clamp(dirAlbedo, 0.0, 1.0);
 }
 
+float mx_oren_nayar_diffuse_avg_albedo(float NdotV, float roughness)
+{
+    float sigma2 = mx_square(roughness);
+    float A = 1.0 - 0.5 * (sigma2 / (sigma2 + 0.33));
+    float B = 0.45 * sigma2 / (sigma2 + 0.09);
+    return A + (2.0 / 3.0 - 64.0 / (45.0 * M_PI)) * B;
```
portsmouth, line 87: This is what we called the QON model in the paper, not FON.
If you compensate this, you don't get what we called EON (which is compensated FON). 

For example, it will have the dark circle artifacts, and (subjectively) doesn't look quite as nice as EON.

I'd suggest that the "compensated Oren-Nayar" should actually modify the QON term to FON, as in the paper.

portsmouth:
I see you're implementing this as basically the existing Oren-Nayar in MaterialX (i.e. QON), augmented with the QON energy-compensation from the paper. We didn't recommend doing that. 

I'd suggest instead to have the compensated Oren-Nayar be a separate model, i.e. what we called EON, which is obtained by compensating Fujii's Oren-Nayar model (FON) -- not QON.

Otherwise you will inherit from QON some artifacts, and the compensation is a bit weird since QON violates energy conservation, meaning the compensation term has to be negative (so in the colored case, this might not conserve energy -- since the compensation will be _substracting_ a smaller amount).

jstone-lucasfilm:
@portsmouth That makes sense, and I was originally thinking that only the energy compensation term would differ when the `energy_compensation` option was enabled, with the original A and B terms remaining unchanged, but I see that the EON proposal changes more than just the energy compensation.

I'll take a second pass at this logic, taking the full extent of the EON proposal into account, and see how things look.

libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl
jstone-lucasfilm: @portsmouth I was aiming to match @fpsunflower's implementation of QON in Open Shading Language, which has the `0.0` fallback as well.  Should the change that you propose be made there as well?

jstone-lucasfilm:
@portsmouth @fpsunflower I believe the latest logic should be ready for review, and I've now added support for multi-scatter color correction in the indirect lighting path, which uses the real-time-friendly approach of storing pre-integrated irradiance in a texture.

portsmouth:
Do we want  `base_weight` to be multiplied into the BRDF (as the multiplicative "weight" of the MaterialX BSDF -- as you implemented it),  or instead have the parameter $\rho$ be `base_weight * base_color` (and have no overall weight multiplier)? The latter is actually what we currently say in the OpenPBR spec, i.e. base_weight is a "scalar multiplier to base_color".

It depends what we want `base_weight` to mean. Perhaps your the current interpretation actually makes more sense though than what we say in the spec, since then the `base_weight` functions simply as a linear multiplier of the BRDF (and thus the albedo).

I'd suggest we consider changing the OpenPBR spec itself to match that.

jstone-lucasfilm:
That sounds like the right approach to me as well, @portsmouth, and we can follow up on the OpenPBR PR once it's posted.

PR #1823, opened 2024-05-21
Add missing inversesqrt support to MSL

shill-lucasfilm:
While developing the GLSL implementation of a new sheen model for MaterialX, I noticed that `inversesqrt()` was not supported when running MaterialXView on macOS. This trivial change addresses that.


shill-lucasfilm:
Oops, it appears that I checked in with the wrong account. I'll reopen with a new PR.

PR #1825, opened 2024-05-21
Initial support for Zeltner sheen

shill-lucasfilm:
This is a first version of the GLSL implementation, which is currently inactive but can be enabled in place of the existing sheen model via the `SHEEN_METHOD` `#define` in `mx_sheen_bsdf.glsl`.

In contrast to the reference implementation (https://github.com/tizian/ltc-sheen), this version uses analytic fits instead of a table, to ease integration. I plan to improve the quality of the fits in the future (particularly the directional albedo), and also add a sampling function.


shill-lucasfilm:
The macOS shader validation test is currently failing due to the MSL backend lacking support for `inversesqrt`. I have a PR already open for that (https://github.com/AcademySoftwareFoundation/MaterialX/pull/1824).

jstone-lucasfilm (approved):
This looks great to me, thanks @shill-lucasfilm, and the new code is lightweight enough that I believe we can move forward with a merge, making additional refinements in future pull requests.

PR #1827, opened 2024-05-21
Fixes to swizzle and channels upgrade

### LINKED ISSUE #1826: 1.39 "channels" upgrade issue due to value vs. connection priority logic
## Issue

If a node has an` input` which happens to have both a `value` and a connection (which isn't recommended but can be loaded), then an invalid graph is created as a `swizzle` with no value is created and upstream connected node is left unconnected.

This seems to come up enough as some DCCs / libraries write both value + connections.

Prior to this upgrade the graph loads in properly connected.

### Input data
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <gltf_pbr name="SHD_olives" type="surfaceshader" xpos="13.768116" ypos="-0.379310">
    <input name="metallic" type="float" value="1" nodename="image_occlusion" />
    <input name="roughness" type="float" value="1" nodename="image_orm" channels="y" />
    <input name="emissive" type="color3" value="0, 0, 0" />
  </gltf_pbr>
  <gltf_image name="image_orm" type="vector3">
    <input name="file" type="filename" value="image1.png" />
  </gltf_image>
  <gltf_image name="image_occlusion" type="float">
    <input name="file" type="filename" value="image2.png" />
  </gltf_image>
</materialx>
```

### Output result
```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <constant name="swizzle" type="float">
    <input name="value" type="float" value="" />
  </constant>
  <gltf_pbr name="SHD_olives" type="surfaceshader">
    <input name="metallic" type="float" value="1" nodename="image_occlusion" />
    <input name="roughness" type="float" value="1" nodename="swizzle" />
    <input name="emissive" type="color3" value="0, 0, 0" />
  </gltf_pbr>
  <gltf_image name="image_orm" type="vector3">
    <input name="file" type="filename" value="image1.png" />
  </gltf_image>
  <gltf_image name="image_occlusion" type="float">
    <input name="file" type="filename" value="image2.png" />
  </gltf_image>
</materialx>
```

The validation error output is:
```shell
 Invalid value: <input name="value" type="float" value="">
```

### Proposed Fix

If a value and a connection are both specified then the connection should take precedence when creating the temporary `swizzle`.
The original `value` should be kept to preserve existing behaviour.

kwokcb:
## Change

Update fixes #1826 .

- Add in a connection check when a value is found. Do not add in a constant but instead allow for other connection logic to be used.
- (Aside: Fix one test which should have always failed validation since it has value + connection).

Update fixes ##1828.

- Make float->float extraction simply remove the `channels` attribute as it's either erroneous or a no-op.


source/MaterialXCore/Version.cpp
```diff
@@ -1148,9 +1148,10 @@ void Document::upgradeVersion()
                     string channelString = channelsInput ? channelsInput->getValueString() : EMPTY_STRING;
                     size_t sourceChannelCount = CHANNEL_COUNT_MAP.at(inInput->getType());
                     size_t destChannelCount = CHANNEL_COUNT_MAP.at(node->getType());
-                    if (inInput->hasValue())
+                    const bool isDisconnected = inInput->getNodeName().empty() && inInput->getNodeGraphString().empty() && inInput->getInterfaceName().empty();
```
kwokcb, line 1151: Question to @jstone-lucasfilm : Should the `value` be removed as it will be caught in validation ? I've left it there so it preserves existing behaviour.

jstone-lucasfilm: Since the value has no effect on the rendered result, and the connection will take precedence, I would recommend that we simply remove the value in the upgrade to 1.39.

jstone-lucasfilm: Either approach seems fine here, as long as the connection continues to take precedence in the upgraded 1.39 document.

source/MaterialXCore/Version.cpp
```diff
@@ -1148,9 +1155,10 @@ void Document::upgradeVersion()
...
```
jstone-lucasfilm, line 1158: Here I'd recommend naming the boolean `hasConnection` for consistency, and ideally we should use any available helper methods (e.g. `hasInterfaceName`) to write this logic more clearly.

source/MaterialXCore/Version.cpp
kwokcb: I've removed it rather than preserve an invalid configuration.

source/MaterialXCore/Version.cpp
```diff
@@ -578,6 +578,7 @@ void Document::upgradeVersion()
     {
         // Convert color2 types to vector2
         const StringMap COLOR2_CHANNEL_MAP = { { "r", "x" }, { "a", "y" } };
+        const string CHANNELS_ATTRIBUTE_STRING = "channels";
```
kwokcb, line 581: Put in a const string since it's repeated in a number of places.

jstone-lucasfilm: In this case, it might be clearer to use the string literals, so that readers remain aware that these constants don't actually exist in the MaterialX codebase.  That's the pattern we've been using elsewhere in the version upgrade system, especially in more recently-contributed changes.

jstone-lucasfilm (approved):

PR #1829, opened 2024-05-21
Fix MSL syntax bug

ld-kerley:
After [commit 2007ebf](https://github.com/AcademySoftwareFoundation/MaterialX/commit/2007ebf3ff887e098c8b3240036da3d438c3ec17) was merged to `dev_1.39` branch the unit tests for MSL rendering started failing.

This patch rectifies the syntax error created.


jstone-lucasfilm (approved):
Ah, that's my fault, and thanks for the catch, @ld-kerley!

PR #1830, opened 2024-05-21
Use MTLDataType in MaterialXRenderMsl

ld-kerley:
While poking around inside `MaterialXRenderMSL` I found myself poking around in headers to look up what magical numbers like `58` meant.  Metal has an enum we can use instead to make the code a little easier to read.


jstone-lucasfilm (approved):

PR #1831, opened 2024-05-22
Initial implementation of structs in MaterialX

ld-kerley:
This PR introduces support for custom structs types.

Custom structs are defined using the `<typedef>` element.

```
<typedef name="texcoord_struct">
    <member name="ss" type="float" value="0.5"/>
    <member name="tt" type="float" value="0.5"/>
</typedef>
```

`<member>` elements are used to define the members of the custom struct. The `value` here is the default value used if no initializer is specified.

Note, that the name of the custom struct is suffixed by `_struct`, this is a proposed naming convention to ensure that new type names introduced to the generated shader code don't conflict with other variable names.

Once defined, custom structs can then be used as types for ports in node definitions.

```
<nodedef name="ND_extract_s_texcoord" node="extracts" nodegroup="shader" >
    <input name="in" type="texcoord_struct" value="{0.1;0.1}" />
    <output name="out" type="float" value="0.0" />
</nodedef>
```

The `value` provided here in an initializer list style string.  It is wrapped with braces, separted by a semi-colon, and should contain initializer values for every member of the struct.  If a custom struct contains a member that itself is a custom struct, then these it initializer lists should be nested.

```
<typedef name="texcoordGroup_struct">
    <member name="st_0" type="texcoord_struct" value="{0.1;0.1}"/>
    <member name="st_1" type="texcoord_struct" value="{0.5;0.5}"/>
    <member name="st_2" type="texcoord_struct" value="{0.9;0.9}"/>
</typedef>

<nodedef name="ND_extract_s_texcoordGroup" node="extractsgroup" nodegroup="shader" >
    <input name="in" type="texcoordGroup_struct" value="{{0.1;0.2};{0.3;0.4};{0.5;0.6}}"/>
    <input name="index" type="integer" value="0"/>
    <output name="out" type="float" value="0.0" />
</nodedef>
```


ld-kerley:
Tagging @jstone-lucasfilm @niklasharrysson for discussion - happy to jump on a call if thats easier too.

niklasharrysson:
@ld-kerley Great to see this work! I like this direction a lot, where structs can remain unfolded in generated code.

Some feedback on your questions:

* About value tokenization: I think it would be great to introduce a `CompoundValue` (or `AggregateValue`) class, as a subclass of `Value`, in order to do the value parsing in one central place, instead of having to deal with the valuestring parsing in many places. This would also let us validate struct values during document validation, and not only during code generation.

*  For name clashes I'm surprised this doesn't work out of the box. There is a call to `registerReservedWords` inside `registerTypeSyntax`, that should handle this. So try tracing from there and see why the struct name is not registered as a reserved word.
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/Syntax.cpp#L55

* For loading in struct types from the document, perhaps we could add initialize/deinitialize methods to `ShaderGenerator`, to handle such things in a central place. We already have other similar methods that need to be called to initialize generation, for example the `registerShaderMetadata` call. So all such calls could be collected into a single init method.

* For nested structs, it looks like the approach you have take should be able to support this. So that would be nice. That's one of the pros of not unfolding the structs, as that would have been more messy I think.

niklasharrysson:
Ah I see! Yes, obfuscating the struct type names with a suffix sounds like a good idea then.

ld-kerley:
@niklasharrysson I hit a roadblock when adding a randomly generated suffix to the struct names in the generated shader code.  It means the struct names are not know when the node definition author is writing the shader source, and they need to declare the type of the inputs to their functions in the shader source.

We could opt for a static suffix.  `texcoord` -> `texcoord_typedef`, or something, and then node authors would be able to use the less obvious `texcoord_typedef` type name in their shader code.
 
Or, we could just not suffix at all, and instead find a way to make the error messages more informative? 

The most recent push includes the `AggregateValue` update, and I was considering postponing the centralized initialize/deinitialize refactor for the `ShaderGenerator` to a separate PR, just to keep this one manageable.

niklasharrysson:
@ld-kerley Aha, yes that's a roadblock indeed! :)

A codegen added static suffix feels a bit awkward, since as you note, node authors need to be aware of this and append the suffix in all shader code using this data type. Perhaps it's better then to not add a suffix at all during codegen, and instead introduce a naming convention for custom data types, that authors must follow?

For example, something like:
_A custom type name should have a suffix that declare the semantic of the type. Example: `texcoord_struct`
In addition, if the type is declared outside of stdlib, it should have a namespace prefix to minimize name collisions. Example:  `pxr_texcoord_struct`, `maya_light_struct`, etc._

Would be great to have better error messages if collisions do occur. I'm not sure how though, since it's the target language compiler that finds this.

Sounds good to postpone the centralized initialize/deinitialize to another PR.

ld-kerley:
@niklasharrysson I really like the `_struct` suffix idea.  We can add a step in the document validation to check for it.  Do we want to block a render, ie. fail code gen, if we don't have the suffix?  I can imagine someone could very well get lucky and not clash.  Do we have other examples (I think we do because I've hit them) where a document could be invalid, but still render exactly as intended?

Also you suggest using the semantic of the type, does that mean that this example from the current spec
```<typedef name="spectrum" semantic="color"/>```
should be re-written to 
```<typedef name="spectrum_color" semantic="color"/>``` 
?

Or perhaps this suggestion infers a completely new XMLTag/MaterialX type?  `<structdef>` ? and we would leave `<typedef>` alone as its written in the new spec (with no `<member>` child tag). That way the new concept can more cleanly have its own set of rules?
```
<structdef name="texcoord_struct">
  <member name="ss" type="float" value="0.5"/>
  <member name="tt" type="float" value="0.5"/>
</structdef>
```

Given this feature never worked before, it doesn't seem too difficult, from a backwards compatibility perspective, to pivot this way?  Perhaps @dbsmythe or @jstone-lucasfilm have input here?

**Comment by niklasharrysson** [comment]
@ld-kerley About validation of a naming convention, I'm not sure we need to do that, since even if we do there is no guarantee that this will help to resolve name conflicts. At least in theory, a node author could use the same name for a variable in the code added for a node implementation, even if both a prefix and suffix is used.

So perhaps the best we can do is to provide guidelines for how to name you data types and identifiers. And if there is a conflict anyway, resulting in shader compile errors, it's up to the node author to resolve it.

With this in mind perhaps it would be best to just stick with the `typedef` element, even for structs. It feels more consistent since after all it's a type you're defining.

As for the naming convention to use, the one I gave was just an example, and it would be great to get others feedback on this as well.

**Comment by jstone-lucasfilm** [comment]

**Comment by ld-kerley** [comment]
@jstone-lucasfilm - done - and also pushed my latest.

@niklasharrysson - this includes the `_struct` suffix, to the example only - no validation.  And also now should allow for structs-of-structs - including in the GLSL/MSL rendering path. 

Logged an issue (#1850) for the common init/de-init for ShaderGeneration.

**Review by niklasharrysson — COMMENTED** [review]
This is coming along nicely! I added some suggestions above.

Apart from those there seems to be some build errors caught by CI, which mainly looks like compile warnings that should be fixed.

File: source/MaterialXGenShader/TypeDesc.h [hunk 1]
```diff
@@ -154,7 +165,8 @@ class MX_GENSHADER_API TypeDesc
     uint32_t _id;
     uint8_t _basetype;
     uint8_t _semantic;
-    uint16_t _size;
+    uint8_t _size;
+    int8_t _structIndex;
```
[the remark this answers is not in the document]
ld-kerley: Yeah I was wondering the same thing as I was writing it, if we could migrate to dynamically registering the existing non-struct types.  

I was looking at the code again, and I think maybe we can go with a uint8_t here for `_structIndex` and just ensure we're calling `isStruct()` before relying on the index.  That way we can double the number of structs available.  Also Is the size of TypeDesc critical?  if we needed more in the future could we just increase the size of `_structIndex` ?

ld-kerley:
I think all the tests are passing now and I addressed the @niklasharrysson notes above - moving the the example scenes to the test suite required a little re-working of the testing framework to add another libraries location to the search path.

source/MaterialXGenShader/TypeDesc.h
```diff
@@ -83,6 +85,9 @@ class MX_GENSHADER_API TypeDesc
     /// Return the semantic for the type.
     unsigned char getSemantic() const { return _semantic; }
 
+    /// Return the index for the struct member information in StructTypeDesc, valid is invalid if `isStruct()` returns false.
```
niklasharrysson, line 88: Nitpick: Is there a typo in this comment "valid is invalid"

source/MaterialXGenShader/TypeDesc.h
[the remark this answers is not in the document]
niklasharrysson: Sounds great to use the sign bit as well to get 256 struct types.

I was happy to be able to squeeze the whole class down to 64 bits, so we can pass these around by value for free. But considering everything else going on during shader generation, which all the string operations etc., it's probably not a big deal if this class gets a bit more heavy in the future, if we need more struct types some day.

niklasharrysson (approved):
These changes looks great to me. I only added a few extra comments above.

Thanks @ld-kerley!

ld-kerley:
Addressing some notes from @jstone-lucasfilm - and also removing the unit testing changes - these will be added at a later date once this is merged.

source/MaterialXCore/Value.h
```diff
@@ -73,7 +74,7 @@ class MX_CORE_API Value
     /// Create a new value instance from value and type strings.
     /// @return A shared pointer to a typed value, or an empty shared pointer
     ///    if the conversion to the given data type cannot be performed.
-    static ValuePtr createValueFromStrings(const string& value, const string& type);
+    static ValuePtr createValueFromStrings(const string& value, const string& type, ConstTypeDefPtr typeDefPtr = nullptr);
```
jstone-lucasfilm, line 77: For consistency, I'd recommend `typeDef` as the name of this argument, as we usually omit `ptr` from variable and argument names.

source/MaterialXCore/Value.h
```diff
@@ -193,6 +194,65 @@ template <class T> class MX_CORE_API TypedValue : public Value
...
+        auto result = createAggregateValue(_typeName);
+        for (const auto& val : _data)
+        {
+            result->appendValue(val->copy());
+        }
+        return result;
+    }
+
+    /// Append a member value to the aggregate.
+    void appendValue(ConstValuePtr valuePtr) {
```
jstone-lucasfilm, line 219: For consistency, these brackets should use Allman conventions, where each bracket is placed on its own line.  The same applies to other brackets in this changelist, but I'll let this comment stand in for all of them.

source/MaterialXGenShader/Syntax.cpp
```diff
@@ -190,6 +190,36 @@ bool Syntax::remapEnumeration(const string&, TypeDesc, const string&, std::pair<
...
 }
 
+void Syntax::registerStructTypeDescSyntax()
+{
+    for (const auto& typeName : StructTypeDesc::getStructTypeNames())
+    {
+        const auto& typeDesc = TypeDesc::get(typeName);
+        const auto& structTypeDesc = StructTypeDesc::get(typeDesc.getStructIndex());
+
+        std::string structTypeName = typeName;
```
jstone-lucasfilm, line 200: For consistency, we can remove the `std::` prefix from these string types.

source/MaterialXGenShader/Syntax.cpp
```diff
@@ -246,4 +276,38 @@ string AggregateTypeSyntax::getValue(const Value& value, bool /*uniform*/) const
...
+
+    string separator = "";
+    for (const auto& memberValue : aggValue.getMembers()) {
+        result += separator;
+        separator = ";";
+
+        auto memberTypeName = memberValue->getTypeString();
+        auto memberTypeDesc = TypeDesc::get(memberTypeName);
+
+        // recursively use the syntax to generate the output, so we can supported nested structs.
```
jstone-lucasfilm, line 302: For consistency, we should capitalize the first letter of each sentence in comments.

source/MaterialXGenShader/TypeDesc.cpp
```diff
@@ -28,6 +30,18 @@ TypeDescNameMap& typeNameMap()
     return map;
 }
 
+
+    //////////////////////////////////////////////////////////////////////////
```
jstone-lucasfilm, line 34: For consistency, I would omit these double lines of forward slashes, and the declarations below should be fully left-aligned, as seen in examples like the following:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Util.cpp#L14

source/MaterialXGenShader/TypeDesc.cpp
```diff
@@ -86,4 +144,77 @@ TYPEDESC_REGISTER_TYPE(MATERIAL, "material")
 
 } // namespace Type
 
+
+
+///////////////////////////////////////////////////////////////////
```
jstone-lucasfilm, line 149: As above, I would omit these triple lines of forward slashes, and you might instead add a comment noting that these are "StructTypeDesc methods", as seen in examples like the following:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Element.cpp#L52

source/MaterialXGenShader/TypeDesc.h
```diff
@@ -60,15 +61,15 @@ class MX_GENSHADER_API TypeDesc
     };
 
     /// Empty constructor.
-    constexpr TypeDesc() noexcept :
-        _id(0), _basetype(BASETYPE_NONE), _semantic(SEMANTIC_NONE), _size(0) { }
+    constexpr TypeDesc() noexcept : _id(0), _basetype(BASETYPE_NONE), _semantic(SEMANTIC_NONE), _size(0), _structIndex(0) {}
```
jstone-lucasfilm, line 64: For consistency, I'd recommend placing these initializers on their own line, either using one line for each variable or a single line for all variables.

source/MaterialXGenShader/TypeDesc.h
```diff
@@ -156,14 +169,15 @@ class MX_GENSHADER_API TypeDesc
...
+    uint8_t _size;
+    uint8_t _structIndex;
 };
 
 /// @class TypeDescRegistry
 /// Helper class for type registration.
 class MX_GENSHADER_API TypeDescRegistry
 {
-  public:
+public:
```
jstone-lucasfilm, line 180: This spacing change was probably just an automatic reformat by your editor, but I'd recommend restoring the original spacing.

source/MaterialXGraphEditor/RenderView.cpp
```diff
@@ -630,6 +630,9 @@ void RenderView::initContext(mx::GenContext& context)
     unitSystem->setUnitConverterRegistry(_unitRegistry);
     context.getShaderGenerator().setUnitSystem(unitSystem);
     context.getOptions().targetDistanceUnit = "meter";
+
+    // Register struct type defintions
```
jstone-lucasfilm, line 634: Minor spelling error for "definitions".

jstone-lucasfilm (commented):
This change is looking very close to the mark, and my initial notes are mainly formatting and naming convention requests.

source/MaterialXGenShader/TypeDesc.cpp
```diff
@@ -46,7 +53,53 @@ TypeDesc TypeDesc::get(const string& name)
...
+    typenames.erase(it->second.typeId());
+    types.erase(it);
+}
+
+ValuePtr TypeDesc::createValueFromStrings(const string& value) const
+{
+    ValuePtr newValue = Value::createValueFromStrings(value, getName());
+
+    // Value::createValueFromStrings() will return a value (including potentially a nullptr) if the
+    // type has been registered, otherwise
```
jstone-lucasfilm, line 75: I get the sense that this is meant to be interpreted as one continuous sentence, where the "otherwise" clause is continued below, but it may be clearer just to write this as a single sentence expressing the behavior of both code paths.

source/MaterialXGenShader/TypeDesc.h
```diff
@@ -61,14 +62,21 @@ class MX_GENSHADER_API TypeDesc
...
+        _basetype(BASETYPE_NONE),
+        _semantic(SEMANTIC_NONE),
+        _size(0),
+        _structIndex(0)
+    {
+    }
 
     /// Constructor.
-    constexpr TypeDesc(std::string_view name, uint8_t basetype, uint8_t semantic = SEMANTIC_NONE, uint16_t size = 1) noexcept :
+    constexpr TypeDesc(std::string_view name, uint8_t basetype, uint8_t semantic = SEMANTIC_NONE, uint8_t size = 1, uint8_t structIndex = 0) noexcept :
```
jstone-lucasfilm, line 74: Rather than focusing on efficient packing, perhaps it's better to make both of these values `uint_16_t` type, so that the hardcoded maximum number of struct indices is beyond what we would expect to see in any real-world asset.

I wouldn't expect to see any performance or memory consequences from instances of this class being one machine word larger!

source/MaterialXGenShader/TypeDesc.h
```diff
@@ -85,6 +93,9 @@ class MX_GENSHADER_API TypeDesc
     /// Return the semantic for the type.
     unsigned char getSemantic() const { return _semantic; }
 
+    /// Return the index for the struct member information in StructTypeDesc, the result is invalid if `isStruct()` returns false.
```
jstone-lucasfilm, line 96: Perhaps we should declare this method just below `isStruct`, so that the reference to `isStruct` in the comment is immediately clear to the reader.

jstone-lucasfilm (commented):
This is looking really good, thanks @ld-kerley, and I had just a few additional notes on implementation details.

source/MaterialXGenShader/TypeDesc.cpp
ld-kerley: Yeah, I think that comment was written earlier in the code evolution, I'll try and re-word it to something clearer.

source/MaterialXGenShader/TypeDesc.h
ld-kerley: I'd defer to @niklasharrysson here - If I recall there was a comment/conversation with him where he was happy the size of this struct had not changed.  Perhaps related to the API changes introduced in MatX 1.39 where we're passing the `TypeDesc` object by value?

source/MaterialXGenShader/TypeDesc.h
ld-kerley: I thought it was most consistent to keep all the `getXXX()` functions together and all the `isXXX()` functions together.

If you feel strongly about moving the function so that the referenced function in the comment is close by then I'm ok with that,  just let me know.

source/MaterialXGenShader/TypeDesc.h
niklasharrysson: Yeah, it was nice to fit the class into a single machine word, since these are passed by value. But I think you are right Jonathan, that on modern hardware, and compared to everything else going on in the API, there should be no noticeable difference to make it slightly larger.

So it's probably better to be on the safe side and make sure `size` and `structIndex` will never overflow.

source/MaterialXGenShader/TypeDesc.cpp
```diff
@@ -28,6 +28,13 @@ TypeDescNameMap& typeNameMap()
     return map;
 }
 
+using StructTypeDescStorage = vector<StructTypeDesc>;
+StructTypeDescStorage& structTypeStorage()
+{
+    static StructTypeDescStorage  storage;
```
jstone-lucasfilm, line 34: Ah, I hadn't noticed this singleton pattern until now, and this seems like a design decision that we should discuss.

Would this break in situations where MaterialX is leveraged across multiple threads, with each thread performing shader generation on a different document with different usage of structs?

Traditionally we forbid singletons in MaterialX to avoid this class of multi-threading hazards, and if there's another engineering solution that stores this data in a thread-local object such as the `GenContext` (which was explicitly designed to handle this category of storage), that would be strongly preferable.

ld-kerley: This use intentionally parallels the existing pattern for the TypeDesc system [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/TypeDesc.cpp#L21)

```
TypeDescMap& typeMap()
{
    static TypeDescMap map;
    return map;
}
```

If we want to refactor the type system to use storage in `GenContext` I would propose that we do that as a separate PR.

jstone-lucasfilm: Got it, and it sounds we have some existing threading issues to address in our shader generation, independent of this proposal.  I'll close out this request for now, but I'm CC'ing @niklasharrysson so that he's in the loop.

jstone-lucasfilm (approved):
This looks like a great improvement, thanks @ld-kerley!

jstone-lucasfilm:
@ld-kerley This change looks ready to merge, but it would benefit from a new description, since it's no longer a draft seeking feedback from the team!  Would you like to write up a new description before we merge this?

source/MaterialXGenShader/TypeDesc.cpp
niklasharrysson: @jstone-lucasfilm @ld-kerley Just adding my 2 cents on the threading issue:

The use of a singleton for `typeMap()` does not expose threading issues, since write access to this data is strictly private to this file and only done once during program initialization. This is because the registration of TypeDesc's for native data types is hard coded and only done here:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/TypeDesc.cpp#L118

After this we only do read access.

Whereas for StructTypeDesc's, the registration is dynamic based on the loaded documents, so in this case there are potential threading issues with using the same pattern.

If we at some point want to make registration of TypeDesc's data driven as well, we would need to update both of these singleton use cases though.

PR #1837, opened 2024-05-25
Remove references to thin_film_bsdf in specification

pablode:
The node has been removed but there were still references in the 1.39 specification.


jstone-lucasfilm (approved):
Good catch, thanks @pablode!

PR #1851, opened 2024-05-31
Fix errors in MDL modules introduced in recent work

niklasharrysson:
This change list fixes a bunch of compile errors in the MDL modules.


jstone-lucasfilm (approved):
Thanks for the fixes, @niklasharrysson!

PR #1854, opened 2024-05-31
Remove Convert C++ node

ld-kerley:
replace with direct source code implementation.

remove unneeded getSwizzledVariable()

Issue #1382, opened 2023-06-17
Add VFX Platform 2023 to CI build matrix


jfpanisset:
Although the MaterialX CI currently doesn't use the [aswf-docker](https://github.com/AcademySoftwareFoundation/aswf-docker) build containers, there is now a VFX Platform 2023 container which should contain the pre-requisites to build MaterialX in a VFX Platform 2023 compliant environment:

* [aswf/ci-materialx](https://hub.docker.com/r/aswf/ci-materialx)
* more specifically: [aswf/ci-materialx:2023](https://hub.docker.com/layers/aswf/ci-materialx/2023/images/sha256-5e5bce9bc6b9e931dd426060e9712dcdfe09576bb26cc741f5b7363f8cbafd32?context=explore)

Any feedback as to missing pre-requisites in this build container would be welcome.

jstone-lucasfilm:
This is a good suggestion, @jfpanisset, and we could potentially benefit from this by building with OpenImageIO enabled in an ASWF Docker container, which we don't yet support in our GitHub Actions matrix.

jstone-lucasfilm:
Following up on this thread in 2025, I think the broader idea of adding Docker builds to our GitHub CI is worthwhile, but we can likely wait until there's a use case demanding this.  I'll close this GitHub Issue out for now, and we can return to this idea in the future.

PR #1855, opened 2024-05-31
Remove Combine C++ node

ld-kerley:
Replace with direct source code implementation for `combine2`, `combine3` and `combine4`.


libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
```diff
@@ -704,18 +704,18 @@
   <implementation name="IM_convert_integer_float_genosl" nodedef="ND_convert_integer_float" target="genosl" sourcecode="float({{in}})" />
 
   <!-- <combine2> -->
-  <implementation name="IM_combine2_vector2_genosl" nodedef="ND_combine2_vector2" target="genosl" />
-  <implementation name="IM_combine2_color4CF_genosl" nodedef="ND_combine2_color4CF" target="genosl" />
-  <implementation name="IM_combine2_vector4VF_genosl" nodedef="ND_combine2_vector4VF" target="genosl" />
-  <implementation name="IM_combine2_vector4VV_genosl" nodedef="ND_combine2_vector4VV" target="genosl" />
+  <implementation name="IM_combine2_vector2_genosl" nodedef="ND_combine2_vector2" target="genosl" sourcecode="{ {{in1}},{{in2}} }"/>
```
jstone-lucasfilm, line 707: Do these implementations compile successfully in OSL?  I was expecting to see type-specific constructors such as `vector2` and `color4` in these implementations, though I haven't yet tested this pull request with an OSL build.

ld-kerley: I believe that OSL allows the use of initializer lists, and the full test suite passes - I was assuming that that would exercise the osl compiler to validate?

certainly happy to revert to type'd constructors if thats preferred.  I do see that might be clearer to read.

ld-kerley: On closer inspection, it does appear that the CI is not running `oslc` to validate the OSL source code.  I wonder if this could be changed? And due to a mixup on my side, I wasn't running it locally either.  This uncovered both syntax errors here, as well as in the already merged #1854, so I'll put a PR to fix that asap as well.

jstone-lucasfilm: Thanks for these fixes, and we're very open to ideas on improving the integration of OSL into render testing.  On my side, I currently set the flags described here when running OSL render tests, and there's definitely room to simplify this:

https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/source/MaterialXTest#per-language-render-setup

ld-kerley: Do you think it's possible to get these to run in the CI for the PRs?  I'm not familiar with the MaterialX GitHub CI setup.

jstone-lucasfilm: The main issue is that we only allow source code to be used in GitHub CI if we can build it from scratch, since downloading binaries creates potential security issues.  And building Open Shading Language from scratch takes a bit too long for us to include this in continuous builds.  One could imagine including OSL in a nightly or weekly build of MaterialX, though, if we end up creating such as build type in the future.

jstone-lucasfilm: Another possibility would be to leverage one of the VFX Platform build environments, as described in this issue:
https://github.com/AcademySoftwareFoundation/MaterialX/issues/1382

If the VFX Platform build environment already includes OSL, then we should be able to safely reference this in our render test suite.

jstone-lucasfilm (approved):

PR #1863, opened 2024-06-03
Fix OSL syntax for Convert nodes

ld-kerley:
PR #1854, previously removed the C++ implementations for `convert` and replaced them with direct source code implementations.  The testing done wasn't correctly running `oslc` so a few syntax errors slipped through.

These are corrected in this PR.

Note:
`vector2`, `vector4` and `color4` are implemented as structs in OSL, and so do not have constructors in the typical sense.   The following is not legal OSL syntax.
```
vector4 a = vector4(0,0,0,0);
```
instead you must use a list initializer to initialize the struct.
```
vector4 a = {0,0,0,0};
```

Conversely, currently it is NOT legal to initialize a builtin type, for instance `color` or `vector` using a list initializer. 
The following is not legal OSL syntax.
```
color a = {0,0,0};
```
instead you must use the type constructor.
```
color a = color(0,0,0);
```

Also note, subscript access also is also not legal OSL syntax because these types are implemented as structs. For example to access the green component of a `color4`, you must use `myColor.rgb[1]`.

### DIFF
--- a/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
+++ b/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
@@ -686,20 +686,20 @@
 
   <!-- <convert> -->
   <implementation name="IM_convert_float_color3_genosl" nodedef="ND_convert_float_color3" target="genosl" sourcecode="color({{in}},{{in}},{{in}})" />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm (approved):
Thanks for these fixes, @ld-kerley!

Issue #1716, opened 2024-02-16
UDIMs uv_scale and uv_offset shader uniform parameters not exposed for tiledimage and UsdUVTexture nodes


JGamache-autodesk:
Looking at the current implementation of UDIMs there seems to be two choices:

1- Magic uv_scale and uv_offset ports appearing on top-level shader interface keyed to the `image` node name.
2- Global `UDIM_SET_PROPERTY` declaration affecting every single image in the MaterialX document.

Whenever an `image` node is hidden in a nodegraph, the uv_scale and uv_offset attributes do not propagate to the final shader interface, requiring option 2 to be used.

I would like to suggest expanding option 1, but this requires propagating scale/offset inputs thru parent nodegraph interfaces. Since I will be mentioning changes that will break the shader ABI this will be done only if explicitly requested by the DCC dev by setting a new ShaderGen option.

Let's take an arbitrary NodeGraph implementing the following NodeDef as an example:
```xml
  <nodedef name="ND_rustedimage_color3" node="rustedimage" nodegroup="texture2d">
    <input name="rustfile" type="filename" value="" uniform="true" />
    <input name="diffusefile" type="filename" value="" uniform="true" />
    <!-- Many other inputs... -->
    <output name="out" type="color3" default="0.0, 0.0, 0.0" />
  </nodedef>
```

The NodeGraph implementation of this definition will use two image nodes connected to the `rustfile` and `diffusefile` inputs. Could be other types of NodeGraph-implemented image nodes because the algo must be recursive to completely exit all the NodeGraph layers.

Let's say we are doing shadergen for a material using one `rustedimage` node called `Rusty1`. 

We dig inside and end up at some point calling `HwImageNode::addInputs()` for the image nodes handling the `rustfile`. 
1- The image ShaderNode being processed gets two new inputs, named `file_uv_scale` and `file_uv_offset`
2- We then look at the parent NodeGraph, find out that input `rustfile` is connected to the `file` input of the ShaderNode being processed, so we add `rustfile_uv_scale` connected to `file_uv_scale` and `rustfile_uv_offset` connected to `file_uv_offset`

Similarly we end up with `diffusefile_uv_scale` and `diffusefile_uv_offset` for the other image ShaderNode connected to the `diffusefile` filename input. Now you see why we use the name of the filename port as prefix.

[...]

PR #1864, opened 2024-06-03
Propagate uv scale and offset

JGamache-autodesk:
Exactly the same as https://github.com/AcademySoftwareFoundation/MaterialX/pull/1840 but on the main branch.

Fixes https://github.com/AcademySoftwareFoundation/MaterialX/issues/1716

Propagate internally added ports when found inside a NodeGraph implementation of a NodeDef.

Testing:

Run render GLSL tests with the shader write option enabled.
Open build/bin/resources/Materials/TestSuite/stdlib/color_management/filename_cm_test/filename_CM_Test_ps.glsl
See that all three image nodes (bl, bl1, and tr1) have uv_scale and uv_offset uniform parameters
See that both NG_tm_test() and NG_tm_retest() now have scale and offset parameters and that these parameters are passed to the mx_image_color3() function.
See that in main() we use all scale and offset uniform parameters when calling either the NodeGraph functions or the regular image node.


source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -442,6 +446,35 @@ ShaderGraphPtr ShaderGraph::create(const ShaderGraph* parent, const NodeGraph& n
...
+                // Only add an input socket once. This means that a NodeGraph
+                // with three embedded image nodes will only get a single
+                // pair of uv_scale and uv_offset inputs shared between
+                // all three nodes.
+                // TODO: Have as many uv_scale inputs as there are embedded image
+                //       nodes. Requires a scheme to make their names unique. Bonus
+                //       points if that name can be related to the filename input.
+                // NOTE: This is enough to have UDIMs working with ND_gltf_colorimage
+                //       and ND_UsdUVTexture, so we stop here at this time.
+                inputSocket = addInputSocket(name, nodeInput->getType());
```
JGamache-autodesk, line 469: Original comment from @kwokcb:
> Hi @JGamache-autodesk,
> I don't think this is enough and is inconsistent with how other sockets are "published" using <node name>_ <input > name> . This would also resolve the issue that you should be able to support different UDIM scale/offsets for > different image inputs -- i.e. there is no guarantee there is only one set of UDIM information.

Reply by @JGamache-autodesk:
> We should definitely split the issue, especially since this PR only attempts to fix single image wrappers like those found for glTf, USD, ADSK, Arnold, etc. Image nodes wrapped in simple NodeGraph implementations.
> The second part where we support multi-image nodes with multiple UDIM scales should be indeed discussed and implemented correctly, but maybe should not be a showstopper for first fixing ND_UsdUVTexture and ND_gltf_colorimage.

JGamache-autodesk: @kwokcb please advise: Is it a case where I should rework the comment, or is it a case where a fix should only be submitted if it also covers the edge case mentioned in the comment?

ld-kerley:
Just an FYI - I have some work in progress where I'm trying to deprecate the `HwImage.cpp` node completely in favor of a data driven node.

jstone-lucasfilm:
@ld-kerley @JGamache-autodesk Would you prefer to wait for the upcoming data-driven `image` node, after which this change would be refactored?  Or would you prefer to move forward with this proposal, after which it would be integrated into the upcoming data-driven `image` node?

JGamache-autodesk:
@jstone-lucasfilm I think we should close this one as it is definitely not compatible with the goals of @ld-kerley.

I will see once the data driven image node lands how much work will be required to have UDIMs working. I already have an evil plan that edits in memory the MaterialX library to add `uv_scale` and `uv_offset` ports in case the new data driven image node totally lack any UDIM support of the texture atlas kind. 

There is currently no UDIM support in USD at this time unless the shader is the native UsdPreviewSurface. So we will need to see what USD currently does for its UDIM workflow and what this requires on the MaterialX side.

PR #1866, opened 2024-06-04
Updates to double-sided materials for 1.39

niklasharrysson:
This change list updates the library files so they match the definitions for `surfacematerial` and `surface` nodes in v1.39.


libraries/pbrlib/pbrlib_defs.mtlx
```diff
@@ -221,6 +221,7 @@
     <input name="bsdf" type="BSDF" value="" doc="Distribution function for surface scattering." />
     <input name="edf" type="EDF" value="" doc="Distribution function for surface emission." />
     <input name="opacity" type="float" value="1.0" doc="Surface cutout opacity" />
+    <input name="thin_walled" type="boolean" value="false" uniform="true" doc="Option to make the surface thin-walled." />
     <output name="out" type="surfaceshader" />
   </nodedef>
```
niklasharrysson, line 227: Note that the old deprecated node `thin_surface` still remains below, and we can remove this in another PR. 

I'm not sure if we need to add code for upgrading from `thin_surface` to `surface` with thin_walled enabled? The thin_surface node was never implemented in any backends, except maybe MDL, so I'm not sure if this was ever used by anyone.

jstone-lucasfilm: That's a good question, and my own sense is that the deprecated `thin_surface` node has not yet been used in real-world assets, because it's only been supported on a single target.  So I'd say that it's likely safe to remove this node without an upgrade path in a future change.

jstone-lucasfilm (approved):

PR #1867, opened 2024-06-04
Simplify TypeSyntax API 

ld-kerley:
After the recent refactor of `<convert>` (#1854) and `<combine>` (#1855) the `getValue()` function overload is no longer needed.


jstone-lucasfilm (approved):
Good catch, thanks @ld-kerley!

PR #1868, opened 2024-06-04
Simplify OSL test setup

ld-kerley:
Use the OSL cmake exports to configure the variables needed to enable OSL testing.

Also remove the need for specifying the standard OSL include path, an `oslc` install will already know where these are.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,10 +180,20 @@ if (MATERIALX_BUILD_USE_CCACHE)
     endif()
 endif()
 
+# Attempt to configure OSL testing if it can be found by cmake.
+# This will not override any explicitly provided oslc and testrender
+find_package(OSL)
+if(OSL_FOUND)
+    if(NOT MATERIALX_OSL_BINARY_OSLC)
+        set(MATERIALX_OSL_BINARY_OSLC $<TARGET_FILE:OSL::oslc>)
+    endif()
+    if(NOT MATERIALX_OSL_BINARY_TESTRENDER)
+        # currently OSL does not export a cmake target for testrender, but once that's added this can be simplified.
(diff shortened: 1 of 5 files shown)


ld-kerley:
This is ready to go - pending someone who works on Windows validating that this still allows OSL to be used with windows.

jstone-lucasfilm:
This proposal looks very promising to me, and I should have a chance to test it soon with Windows OSL (which uses a less standardized build approach and folder structure).

jstone-lucasfilm:
@ld-kerley In my local test, I get an error message from CMake before I have a chance to select any build options, so this seems to occur even if the user doesn't choose to build with OSL support:

```
CMake Warning at CMakeLists.txt:185 (find_package):
  By not providing "FindOSL.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "OSL", but
  CMake did not find one.

  Could not find a package configuration file provided by "OSL" with any of
  the following names:

    OSLConfig.cmake
    osl-config.cmake

  Add the installation prefix of "OSL" to CMAKE_PREFIX_PATH or set "OSL_DIR"
  to a directory containing one of the above files.  If "OSL" provides a
  separate development package or SDK, be sure it has been installed.
```

ld-kerley:
@jstone-lucasfilm - was this an error message in CMake, or just a warning and things continued?  I would expect things to have just continued, albeit with an annoyingly loud warning.

I just pushed a PR where I'm quietening the warning, and also guarding the check to only when we need it (testing OSL).

jstone-lucasfilm:
@ld-kerley I believe you're correct that this was just a warning, though CMake makes it look as if the warning is fatal through the use of red text.  Let me try the latest build and see how things are looking now.

jstone-lucasfilm (approved):

PR #1869, opened 2024-06-04
Add ccache support

ld-kerley:
Add optional support for using ccache to speed up builds.

Guarded by `MATERIALX_BUILD_USE_CCACHE` if people need to disable for some reason.


jstone-lucasfilm (approved):

PR #1872, opened 2024-06-06
Fix Monolithic build on Windows

ashwinbhat:
Configure MaterialX with MATERIALX_BUILD_MONOLITHIC=ON on Windows, building this with with VS2022 causes couple issues:
1. error C1128: number of sections exceeded object file format fixed by adding /bigobj flag
2. Linker error from MaterialX_v1_39_0::OiioImageLoader::saveImage() fixed by muting OiioImageLoader using MATERIALX_BUILD_OIIO


CMakeLists.txt
```diff
@@ -236,6 +236,7 @@ endif()
 # Adjust compiler settings
 if(MSVC)
     add_compile_options(/MP)
+    add_compile_options(/bigobj)
```
jstone-lucasfilm, line 239: Is there an advantage to including this setting in non-monolithic Windows builds of MaterialX, or should we restrict this change using MATERIALX_BUILD_MONOLITHIC?

meshula: /bigobj just increases the number of sections that can be linked, and shouldn't in general by measurable in any way otherwise. The only "downside" of setting it on always is that if someone is compiling with vs2019+ they won't be able to give the library to someone linking with vs2005 or earlier. The advantage presumably for always setting it is simpler build logic.

ashwinbhat: To be on the safer side, I think it's better to enable it for MATERIALX_BUILD_MONOLITHIC. I'll update the PR.

jstone-lucasfilm (approved):
Thanks for the fix, @ashwinbhat!

PR #1879, opened 2024-06-10
Improve robustness of input connections

kwokcb:
## Issue Addressed

This change adds detection for the case where both a connection and a value are specified, adds and update connection APIs to enforce not creating invalid documents.

## Changes

- Adds a new `Input::setConnectedInterfaceName()` interface which clears the `value` attributes to avoid accidently creating invalid documents.
- Updates other existing connection APIs to also clear the `value` attribute appropriately.
- Updates validation on an input to mark when both a value and a connection are created.

## Tests
- Update invalid documents:
  - Fix existing error in one of the render test suite files (`color_management.mtlx`).
  - `mxvalidate.py` and code generation tests caught this error. Ran `mxvalidate` on everything in `libraries` and `resources`.
- Add a new `node` unit test to check interface setting and validity checking.
- Change the node graph editor to use the new interface and remove previous manual `value` and `interfacename` clearing calls.


kwokcb:
@jstone-lucasfilm . Found it was painful to try to rebase or update the remote branch. Created this new one in main from https://github.com/AcademySoftwareFoundation/MaterialX/pull/1770.

jstone-lucasfilm (approved):

PR #1880, opened 2024-06-10
Add mode input to sheen_bsdf

jstone-lucasfilm:
This changelist adds a `mode` input to `sheen_bsdf`, allowing the user to select between Conty-Kulla and Zeltner sheen models.

The Zeltner model is initially supported only in GLSL, ESSL, and MSL, integrating recent work from Stephen Hill, with OSL and MDL falling back to the Conty-Kulla model for now.

The graph implementation of `open_pbr_surface` has been updated to take advantage of this new feature, and the `open_pbr_velvet.mtlx` example has been adjusted to take the visual behavior of Zeltner sheen into account.

### DIFF
--- a/libraries/pbrlib/genglsl/mx_sheen_bsdf.glsl
+++ b/libraries/pbrlib/genglsl/mx_sheen_bsdf.glsl
@@ -46,20 +46,18 @@ void mx_sheen_bsdf_indirect(vec3 V, float weight, vec3 color, float roughness, v
     N = mx_forward_facing_normal(N, V);
     float NdotV = clamp(dot(N, V), M_FLOAT_EPS, 1.0);
 
-    vec3 Li = mx_environment_irradiance(N);
-
+    float dirAlbedo;
     if (mode == 0)
     {
-        float dirAlbedo = mx_imageworks_sheen_dir_albedo(NdotV, roughness);
-        bsdf.throughput = vec3(1.0 - dirAlbedo * weight);
-        bsdf.response = Li * color * dirAlbedo * weight;
+        dirAlbedo = mx_imageworks_sheen_dir_albedo(NdotV, roughness);
     }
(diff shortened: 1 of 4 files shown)


libraries/pbrlib/genglsl/mx_sheen_bsdf.glsl
```diff
@@ -1,89 +1,65 @@
...
-    float NdotV = min(dot(N, V), 1.0);
-    roughness = clamp(roughness, 0.01, 1.0); // Clamp to range of original impl.
-
-    float dirAlbedo = mx_zeltner_sheen_dir_albedo(NdotV, roughness);
-    bsdf.throughput = vec3(1.0 - dirAlbedo * weight);
-
-    vec3 Li = mx_environment_irradiance(N);
-    bsdf.response = Li * color * dirAlbedo * weight;
+        float dirAlbedo = mx_zeltner_sheen_dir_albedo(NdotV, roughness);
+        bsdf.throughput = vec3(1.0 - dirAlbedo * weight);
```
shill-lucasfilm, line 62: Since these lines are the same as in the if clause, you could move them to the end (and make `dirAlbedo` non-local).

libraries/pbrlib/genglsl/mx_sheen_bsdf.glsl
```diff
@@ -1,89 +1,65 @@
...
+        bsdf.response = Li * color * dirAlbedo * weight;
```
shill-lucasfilm, line 63: Since these lines are the same as in the if clause, you could move them to the end (and make `dirAlbedo` non-local).

jstone-lucasfilm: Good suggestion, and will do!

PR #1885, opened 2024-06-11
Update LamaConductor Node

mkuo-lucasfilm:
1. Removed the iridescenceThickness and iridescenceIOR inputs, as these are not present in the RenderMan definition of the node.

2. Removed the exteriorIor input, as we expect relative IOR to be computed by the renderer.


jstone-lucasfilm (approved):

PR #1896, opened 2024-06-21
Add logical operator node definitions

ld-kerley:
When experimenting with some of the `<convert>` nodes I found I really wanted a logcical `<not>` to implement `ND_convert_integer_boolean` as a nodegraph using `<ifequal>`.

We actually have these logical operators in the RealityKit MaterialX nodes already (recently shared publicly at WWDC 2024), so I thought it might be useful to propose we add them to the standard library.


jstone-lucasfilm (approved):
This looks like an excellent contribution, @ld-kerley!  I'm CC'ing @dbsmythe to make sure the functionality and documentation meets his standards.

PR #1901, opened 2024-06-24
Adding LamaSurface Node

mkuo-lucasfilm:
Adding initial support for LamaSurface node (pixar renderman definition: https://rmanwiki.pixar.com/display/REN26/LamaSurface)
- LamaSurface has two inputs of BSDF type and one input of float type, and returns a surfacematerial
- LamaSurface node example is commented out, as there's an issue with the shader generation process, which doesn't yet support the notion of a custom node that returns a surfacematerial


jstone-lucasfilm (approved):

PR #1904, opened 2024-06-25
Fix CMake targets relative pathing

ld-kerley:
Introduced with the monolithic build PR #1725 the cmake export targets no longer had relative paths.

An errant additional `target_include_directories` slipped in.


jstone-lucasfilm (approved):

PR #1911, opened 2024-06-30
Update normalmap node to 1.39

pablode:
- I have added the null-check from GLSL and MSL to the OSL and MDL backends for consistency. As an alternative, it could be removed from all implementations.

- Tangent vector renormalization in OSL seems like a concern that should be handled by the implementation providing the geometry streams. I removed it.

- I did not check if the OSL and MDL implementations compile and pass the tests


pablode:
Ready for review.

jstone-lucasfilm:
Thanks for the update, @pablode!

I've been noticing that Linux build failure in Python Wheels for a day or two, and it's safe to ignore that in the context of this pull request.

PR #1917, opened 2024-07-03
Add partial and optional support for OCIO to MaterialX

JGamache-autodesk:
The default color management system in MaterialX has support for a very limited set of color spaces. Some commonly used spaces such as linear Rec.2020 or sRGB-encoded AP1 are not available. MaterialX is often used in applications that already use OCIO for color management and so artists are sometimes confused as to why the color spaces in a menu that is available for textures in another part of the application are not available for textures in MaterialX. In some cases, a color space may even be mathematically the same as one in the MaterialX set, but it is simply named differently. This may lead to the artist picking the wrong color space and wasting time. In situations like this, the application is able to offer a better user experience by offloading the interpretation of color space names to OCIO. This allows artists to use the same color spaces with all textures rather than having a different set for MaterialX.
However, not all OCIO color spaces are appropriate for use in a MaterialX document and this PR does not offer support for full OCIO. Supported color spaces include any of those that are mathematically compatible with one of the default MaterialX spaces.
This PR does not change where the implicit color conversions happen in a MaterialX shader, and it does not provide any additional functionality to creators of MaterialX documents, other than expanding the set of allowed color space names. The MaterialX API already allows a client to override the default mx::ColorManagementSystem, this PR illustrates how that may be done in order to take advantage of a subset of OCIO.
It should be noted that only minimal changes would be needed to adapt this PR for use with either OCIO or the nanoColor library currently under development.
This integrates the OCIO shadergen into the MaterialX shadergen to provide on-the-fly color management for OpenGL, Vulkan, Metal, and OSL.
Targeting 1.39.1. Definitely not 1.39.0.
Tested with GLSL and OSL on Linux.


source/MaterialXGenShader/Nodes/OpenColorIONode.cpp
```diff
@@ -0,0 +1,151 @@
...
+        auto ocioManager = std::dynamic_pointer_cast<OpenColorIOManagementSystem>(context.getShaderGenerator().getColorManagementSystem());
+
+        auto gpuProcessor = ocioManager->getGpuProcessor(getName());
+        OCIO::GpuShaderDescRcPtr shaderDesc = OCIO::GpuShaderDesc::CreateShaderDesc();
+
+        // TODO: Extend to essl and MDL and possibly SLang.
+        if (context.getShaderGenerator().getTarget() == "genglsl") {
+            shaderDesc->setLanguage(OCIO::GPU_LANGUAGE_GLSL_4_0);
+        } else if (context.getShaderGenerator().getTarget() == "genmsl") {
+            shaderDesc->setLanguage(OCIO::GPU_LANGUAGE_MSL_2_0);
```
ld-kerley, line 68: Would be nice if we could keep this version in sync somehow with 

`const string MslShaderGenerator::VERSION = "2.3";`

from [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/30f694f42670c4eb5e7386ce931a6e3690bbe2a3/source/MaterialXGenMsl/MslShaderGenerator.cpp#L46)

source/MaterialXGenShader/CMakeLists.txt
```diff
@@ -9,5 +16,14 @@ mx_add_library(MaterialXGenShader
...
         MaterialXCore
+        ${optionalLibraries}
     EXPORT_DEFINE
         MATERIALX_GENSHADER_EXPORTS)
+
+if(MATERIALX_BUILD_OCIO)
+    if (NOT MATERIALX_BUILD_MONOLITHIC)
+        target_compile_definitions(MaterialXGenShader PUBLIC MATERIALX_BUILD_OCIO)
+    else()
+        target_compile_definitions(${MATERIALX_MONOLITHIC_TARGET} PUBLIC MATERIALX_BUILD_OCIO)
```
ld-kerley, line 27: You should be able to use `${TARGET_NAME}` to collect the correct monolithic/non-monolithic target to use.

It might also be cleaner to append the OCIO library to the target that way too, so then the entire OCIO addition here could just be added inside one guard

```
if(MATERIALX_BUILD_OCIO)
    find_package(OpenColorIO REQUIRED)
    target_link_libraries(${TARGET_NAME} OpenColorIO::OpenColorIO)
    target_compile_definitions(${TARGET_NAME} PUBLIC MATERIALX_BUILD_OCIO)
endif()
```

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.cpp
```diff
@@ -324,8 +328,23 @@ void shaderGenPerformanceTest(mx::GenContext& context)
     context.registerSourceCodeSearchPath(libSearchPath);
 
     // Enable Color Management
-    mx::ColorManagementSystemPtr colorManagementSystem =
+    mx::ColorManagementSystemPtr colorManagementSystem;
+#ifdef MATERIALX_BUILD_OCIO
+    try
+    {
+        auto config = OCIO::Config::CreateFromBuiltinConfig("ocio://studio-config-latest");
```
ld-kerley, line 335: I'm not an OCIO expert - but does this pull some pre-baked config from inside OCIO.  Will people need to provide custom OCIO configs at some point?

ld-kerley (commented):
This looks like really great work - when I get some cycles I'd be happy to kick the tires on the MSL side (might be next week at this point).  I left a few observations below.

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.cpp
doug-walker: @ld-kerley , the PR supports use of arbitrary OCIO configs, but to keep things simple for the purposes of this unit test, Jerry is using one of the configs that's built into the OCIO library itself.

In the real world, apps could create the config using OCIO::Config::CreateFromFile(path) rather than CreateFromBuiltinConfig.

JGamache-autodesk:
Well, OSL was not as trivial as expected, but still manageable. Results are good:
OCIO_results
And we can see the OCIO code being injected in GLSL:
```GLSL
vec4 ocio_color_conversion_8182d4da8df3d64c153c5e4252dbbf(vec4 inPixel)
{
  vec4 outColor = inPixel;
  
  // Add Gamma 'monCurveFwd' processing
  
  {
    vec4 breakPnt = vec4(0.0392857157, 0.0392857157, 0.0392857157, 1.);
    vec4 slope = vec4(0.077380158, 0.077380158, 0.077380158, 1.);
    vec4 scale = vec4(0.947867274, 0.947867274, 0.947867274, 0.999998987);
    vec4 offset = vec4(0.0521326996, 0.0521326996, 0.0521326996, 9.99998974e-07);
    vec4 gamma = vec4(2.4000001, 2.4000001, 2.4000001, 1.00000095);
    vec4 isAboveBreak = vec4(greaterThan( outColor, breakPnt));
    vec4 linSeg = outColor * slope;
    vec4 powSeg = pow( max( vec4(0., 0., 0., 0.), scale * outColor + offset), gamma);
    vec4 res = isAboveBreak * powSeg + ( vec4(1., 1., 1., 1.) - isAboveBreak ) * linSeg;
    outColor.rgb = vec3(res.x, res.y, res.z);
    outColor.a = res.w;
  }

  return outColor;
}
```
and in the OSL code:
```C
color4 ocio_color_conversion_8182d4da8df3d64c153c5e4252dbbf(color4 inPixel)
{
  color4 outColor = inPixel;
  
  // Add Gamma 'monCurveFwd' processing
  
  {
    vector4 breakPnt = vector4(0.0392857157, 0.0392857157, 0.0392857157, 1.);
    vector4 slope = vector4(0.077380158, 0.077380158, 0.077380158, 1.);
    vector4 scale = vector4(0.947867274, 0.947867274, 0.947867274, 0.999998987);
    vector4 offset = vector4(0.0521326996, 0.0521326996, 0.0521326996, 9.99998974e-07);
    vector4 gamma = vector4(2.4000001, 2.4000001, 2.4000001, 1.00000095);
    vector4 isAboveBreak = vector4((outColor.rgb.r > breakPnt.x) ? 1.0 : 0.0, (outColor.rgb.g > breakPnt.y) ? 1.0 : 0.0, (outColor.rgb.b > breakPnt.z) ? 1.0 : 0.0, (outColor.a > breakPnt.w) ? 1.0 : 0.0);
    vector4 linSeg = outColor * slope;
    vector4 powSeg = pow( max( vector4(0., 0., 0., 0.), scale * outColor + offset), gamma);
    vector4 res = isAboveBreak * powSeg + ( vector4(1., 1., 1., 1.) - isAboveBreak ) * linSeg;
    outColor.rgb = vector(res.x, res.y, res.z);
    outColor.a = res.w;
  }

  return outColor;
}
```

doug-walker:
@jstone-lucasfilm, I wanted to ask about this during our meeting today but we ran out of time. Now that Jerry has implemented Bernard's request, is it possible to move forward with this PR?  (Although this work is compatible with NanoColor, it meets a separate need, and thus seems like it could proceed without waiting for the conclusion of that initiative.)

jstone-lucasfilm:
@doug-walker Agreed that this pull request is still valuable, independent of the ultimate design of NanoColor, and I'm currently thinking that we should include it in our next MaterialX release, v1.39.2.  Right now we're in the release candidate phase for v1.39.1, and this change deserves more time than we'd have to include it in that release.

source/MaterialXGenShader/OpenColorIOManagementSystem.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+} // anonymous namespace
+
+//
+// OpenColorIOManagementSystem methods
+//
+
+OpenColorIOManagementSystemPtr OpenColorIOManagementSystem::create(const OCIO::ConstConfigRcPtr& config, const string& target)
+{
+    if (target != "genglsl" && target != "genmsl" && target != "genosl")
```
ld-kerley, line 51: targets can also be inherited from each other, and I believe that maybe OpenUSD's Storm uses this concept.  Instead of just having a fix list of base target names, we may perhaps need to respect inherited targets as well?

source/MaterialXGenShader/OpenColorIOManagementSystem.h
```diff
@@ -0,0 +1,81 @@
...
+
+/// @class OpenColorIOManagementSystem
+/// Class for a default color management system.
+class MX_GENSHADER_API OpenColorIOManagementSystem : public DefaultColorManagementSystem
+{
+  public:
+    virtual ~OpenColorIOManagementSystem() { }
+
+    /// Create a new OpenColorIOManagementSystem
+    static OpenColorIOManagementSystemPtr create(const OCIO::ConstConfigRcPtr& config, const string& target);
```
ld-kerley, line 38: How is a `OCIO::ConstConfigRcPtr` constructed?  

This seems like the only instance where OCIO would be in the public MaterialX API, and it maybe be nice if could avoid that need. 

Perhaps if we exposed wrappers around `OCIO::Config::CreateFromEnv()`  `OCIO::Config::CreateFromFile()` `OCIO::Config::CreateFromBuiltinConfig()` ?

libraries/stdlib/genosl/lib/vector4_extra_ops.osl
```diff
@@ -0,0 +1,45 @@
+// Adds some syntactic sugar allowing mixing vector4 and color4 as
+// arguments of some binary operators used by OCIO transform code.
+
+vector4 __operator__mul__(matrix m, vector4 v)
+{
+  return vector4(v.x * m[0][0] + v.y * m[0][1] + v.z * m[0][2] + v.w * m[0][3],
```
ld-kerley, line 6: I was going to suggest this wrap the existing `transform(m, v)` function that exists in `libraries/stdlib/genosl/lib/vector_4.h`, but then when I was looking it looks like the math isn't actually the same.

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/genosl/include/vector4.h#L414

I remember @jstone-lucasfilm mentioning something about extensive documentation about matrix vector multiplication in MaterialX but I can't find it right now - but I would like to propose that these two function at least agree with each other - or even better be a single shared implementation.

jstone-lucasfilm: Here's the documentation for matrix-vector multiplication in MaterialXCore, and our shading language implementations ought to follow this same pattern:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Types.h#L653

source/MaterialXGenShader/OpenColorIOManagementSystem.h
JGamache-autodesk: Indeed. I made everything opaque to not expose a dependency to OCIO.

source/MaterialXGenShader/CMakeLists.txt
```diff
@@ -11,3 +11,9 @@ mx_add_library(MaterialXGenShader
         MaterialXCore
     EXPORT_DEFINE
         MATERIALX_GENSHADER_EXPORTS)
+
+if(MATERIALX_BUILD_OCIO)
+    find_package(OpenColorIO REQUIRED)
+    target_link_libraries(${TARGET_NAME} PRIVATE OpenColorIO::OpenColorIO)
```
JGamache-autodesk, line 17: Now PRIVATE. Downstream clients need not be aware of OCIO.

resources/Materials/TestSuite/stdlib/color_management/ocio_color_management.mtlx
JGamache-autodesk, line 1: With OCIO enabled:
OCIO_Demo_glsl

Without OCIO:
OCIO_Demo_glsl

source/PyMaterialX/PyMaterialXGenShader/PyColorManagement.cpp
```diff
@@ -58,4 +62,12 @@ void bindPyColorManagement(py::module& mod)
     py::class_<mx::DefaultColorManagementSystem, mx::DefaultColorManagementSystemPtr, mx::ColorManagementSystem>(mod, "DefaultColorManagementSystem")
         .def_static("create", &mx::DefaultColorManagementSystem::create)
         .def("getName", &mx::DefaultColorManagementSystem::getName);
+
+#ifdef MATERIALX_BUILD_OCIO
+    py::class_<mx::OpenColorIOManagementSystem, mx::OpenColorIOManagementSystemPtr, mx::ColorManagementSystem>(mod, "OpenColorIOManagementSystem")
+        .def_static("createFromEnv", &mx::OpenColorIOManagementSystem::createFromEnv)
```
JGamache-autodesk, line 68: Keeping OCIO opaque really pays off here.

source/MaterialXGenShader/Nodes/OpenColorIONode.cpp
hodoulp: > I suspect this enum values indicates a minimum required MSL version to have valid syntax and function calls to support the code emitted by OCIO. If a future version of MSL has breaking syntax changes or improved functionality that simplifies the emitted code, then it might be of interest to add a new implementation generating upgraded GPU_LANGUAGE_MSL_2_3 code. 

Yes

resources/Materials/TestSuite/stdlib/color_management/ocio_color_management.mtlx
```diff
@@ -0,0 +1,48 @@
...
+      <input name="file" type="filename" value="resources/Images/color_palette_srgb.exr" colorspace="sRGB - Texture" />
+    </image>
+    <image name="image2" type="color3">
+      <input name="file" type="filename" value="resources/Images/color_palette_lin_p3d65.exr" colorspace="sRGB Encoded P3-D65 - Texture" />
+    </image>
+    <image name="image3" type="color3">
+      <input name="file" type="filename" value="resources/Images/color_palette_g18_rec709_tx.exr" colorspace="Gamma 1.8 Rec.709 - Texture" />
+    </image>
+    <image name="image4" type="color3">
+      <input name="file" type="filename" value="resources/Images/color_palette_lin_ap0.exr" colorspace="lin_ap0" />
```
meshula, line 29: Question about all the color space names here ~ although the color spaces here are all in the CIF suite, none of them use the CIF names. I assume that it is meant to trigger delegation to OCIO to resolve the color spaces, rather than invoking any of the built in transformations? That approach makes sense to me, but I had to study the diff a fair bit to be sure I had come to the right conclusion. I was thinking it might be good for maintenance purposes to add documentation to this file so that when people encounter it in the future they don't inadvertently "clean it up" by changing the names to canonical names and thus silently bypass actually testing the OCIO functionality?

libraries/stdlib/genosl/lib/vector4_extra_ops.osl
```diff
@@ -3,10 +3,7 @@
 
 vector4 __operator__mul__(matrix m, vector4 v)
 {
-  return vector4(v.x * m[0][0] + v.y * m[0][1] + v.z * m[0][2] + v.w * m[0][3], 
-                 v.x * m[1][0] + v.y * m[1][1] + v.z * m[1][2] + v.w * m[1][3], 
-                 v.x * m[2][0] + v.y * m[2][1] + v.z * m[2][2] + v.w * m[2][3], 
-                 v.x * m[3][0] + v.y * m[3][1] + v.z * m[3][2] + v.w * m[3][3]);
+  return transform(m, v);
```
meshula, line 6: these changes here to the OSL functions ~ they are unrelated to the OCIO integration, right? I think it would make sense to commit them independently so as not to conflate unrelated changes in one commit?

meshula (commented):
these changes lgtm, i just have a couple of minor questions.

libraries/stdlib/genosl/lib/vector4_extra_ops.osl
JGamache-autodesk: This file is introduced by this pull request. Would be nice if it was correct on the first try. Latest changes recognizes the OSL multiplication order as defined in `transform`.

resources/Materials/TestSuite/stdlib/color_management/ocio_color_management.mtlx
jstone-lucasfilm: @JGamache-autodesk Although I really like the idea of adding color charts to MaterialX in the future, ideally we'd want to express them directly as MaterialX graphs, rather than including more textures in the repository.  And for now, we can only access OpenEXR textures when building MaterialX with OpenImageIO, which isn't part of our default builds.

In the short term, what would you think about replacing each of these color charts with a single _color triple_, expressing the correct description of a color in the given color space.  This would remove the need to add textures as part of this PR, and this simple approach could be extended to full MaterialX-expressed color charts further in the future.

JGamache-autodesk: Agreed. I do have a 100% graph based solution using colorspace tags on "color3" inputs of cascaded switch nodes. I developed it for testing a very important USD fix where I use nanocolor to reduce the number of shader compilations.

I will convert that document back to pure MaterialX and submit it next week, thus removing the need for color charts.

JGamache-autodesk: With OCIO enabled, the new test should give:
Standard_Surface1_glsl
Without OCIO, the test will have banding:
Standard_Surface1_glsl

And, since it took little time, I also added a native test using a similar graph, but with color spaces known to MaterialX.

meshula: (@JGamache-autodesk I have a fix under way in OpenUSD for the naming issues you ran into with GfColor and nanocolor not being current to the same version of the CIF names)

source/MaterialXGenShader/Nodes/OpenColorIONode.cpp
```diff
@@ -0,0 +1,169 @@
+//
+// Copyright Contributors to the MaterialX Project
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#ifdef MATERIALX_BUILD_OCIO
+
+#include <MaterialXGenShader/Nodes/OpenColorIONode.h>
+#include <MaterialXGenShader/OpenColorIOManagementSystem.h>
```
jstone-lucasfilm, line 9: Although the wisdom of MaterialX filename conventions is fully open to debate, our current naming convention is to use the `Oiio` prefix for all files related to OpenImageIO, and we should probably follow suit by using the `Ocio` prefix for new files relating to OpenColorIO.

Here are two examples of files following the current naming convention for OpenImageIO:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXRender/OiioImageLoader.cpp
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXRender/OiioImageLoader.h

Following this approach, the new files in this changelist would be named `OcioNode.cpp`, `OcioNode.h`, `OcioColorManagementSystem.cpp`, `OcioColorManagementSystem.h`, and so forth.

Does that naming convention seem reasonable to you, @JGamache-autodesk?

JGamache-autodesk: Yes. That would improve consistency.

jstone-lucasfilm (approved):
Great work, @JGamache-autodesk, and I believe we can merge this feature for inclusion in MaterialX v1.39.4.

When time permits, we should add the new `MATERIALX_BUILD_OCIO` flag to our developer guide, so that teams are aware that this feature is available.

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/MainPage.md#building-materialx-c

PR #1922, opened 2024-07-08
Add a patch version to the MaterialX versioning.

dgovil:
This change adds a patch version to augment the MaterialX version triplet.

Traditionally, MaterialX has had three items in the triplet. The Major and Minor are taken by the document model, whereas the build version has been the one incrementing for every release.

This has made it somewhat difficult to discern when a change is significant or minor. Recently, we've had two changes to MaterialX to rectify build issues, but these have also incremented the same number as a version doing a major update within the same document model version.

By adding a patch version, we allow MaterialX to better delineate to consumers when a change is significant or minor.

For this, I do not modify any areas that may affect symbols or namespaces, as the patches should effectively be transparent to those.


jstone-lucasfilm:
@dgovil I don't want to pause discussion of this idea at all, but for the current 1.39 patch I'm thinking of moving forward with a 1.39.1 build, allowing us to preserve all of our existing versioning conventions in C++/Python API documentation, PyPI releases, and so forth.  Let's keep this discussion going, and whichever specific approach we end up choosing, I completely agree with the importance of supporting independent patch versions.

PR #1923, opened 2024-07-08
Fix missing include install

ld-kerley:
The `mx_add_library` refactor introduced in #1725, introduced a few bugs with installing files.

`Generated.h` was incorrectly being ignored, instead `Generated.h.in` was being installed, and also the mdl library files were missing, also some resources were also not being installed.

Should now be corrected in both monolithic  and non-monolithic builds.


jstone-lucasfilm (approved):
Thanks for this fix, @ld-kerley!

PR #1928, opened 2024-07-09
Fix include install - compatible with cmake 3.16

ld-kerley:
Related to #1923 - that lifted the minimum cmake version 3.23 - but using `FILE_SET`.

This PR should provide the same install functionality, but compatible with cmake 3.16 - albeit with more cmake code complexity.

Ultimately I think it would be nice to move the cmake version up to 3.23, but if we're not able to do that now, then this is an alternate path forwards.


jstone-lucasfilm:
Thanks for putting this build together, @ld-kerley, and after further discussion on the MaterialX Slack, it sounds like our best approach is to move forward with the upgrade to CMake 3.24 in #1929.

PR #1929, opened 2024-07-09
Upgrade required CMake to 3.24

dgovil:
This change updates the minimum required CMake to 3.24. It also removes a workaround needed for CMake 3.16 that required tricking CMake into thinking the `.mm` file was a CXX file instead of ObjC++.

With newer CMake versions, this can implicitly be handled as long as the `.mm` file doesn't share a name with an existing CXX file.


jstone-lucasfilm (approved):
This looks like a good improvement to me, thanks @dgovil!

Issue #1912, opened 2024-06-30
Graph Editor: Reproducible crash after renaming node


jstone-lucasfilm:
Repro steps:

- Launch the Graph Editor with its default marble material.
- Click on the `NG_marble1` node and rename it to `NewMarble`.
- Double click on the `NewMarble` node to view its graph.
- Navigate back to the original view by clicking on the left angle bracket.
- Connect the output of `NewMarble` to the `base_color` input of `SR_marble1`.

jsjtxietian:
Hi I'd like to tackle this one!

Since the crash is vertex out of bounds, a quick fix would be like this, if any node is invalid, just early out in `deleteLinkInfo`:
```C++
diff --git a/source/MaterialXGraphEditor/Graph.cpp b/source/MaterialXGraphEditor/Graph.cpp
index f2382e28..04ec66d8 100644
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -2753,6 +2753,10 @@ void Graph::deleteLinkInfo(int startAttr, int endAttr)
     int upNode = getNodeId(startAttr);
     int downNode = getNodeId(endAttr);
 
+    if (upNode == -1 || downNode == -1) {
+        return ;
+    }
+
     // Change input to default value
     if (_graphNodes[downNode]->getNode())
     {
```

But the root cause seems to be that when navigate back to the original view after renaming and viewing, some connections are lost.

jstone-lucasfilm:
Thanks for taking a look, @jsjtxietian, and a fix for that root cause would be much appreciated!

PR #1930, opened 2024-07-10
Update related input attribute when renaming a nodegraph

jsjtxietian:
Fixes https://github.com/AcademySoftwareFoundation/MaterialX/issues/1912

The problem is that after renaming, the input PortElement's `_attributeMap` still hold the old name, so when drawing the graph, `getConnectedOutput` failed.


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2753,6 +2753,10 @@ void Graph::deleteLinkInfo(int startAttr, int endAttr)
     int upNode = getNodeId(startAttr);
     int downNode = getNodeId(endAttr);
 
+    if (upNode == -1 || downNode == -1) {
```
jsjtxietian, line 2756: Should we keep this? Or just remove it and let it crash when bad things happen

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2753,6 +2753,10 @@ void Graph::deleteLinkInfo(int startAttr, int endAttr)
...
+        return ;
+    }
+
```
jsjtxietian, line 2759: Should we keep this? Or just remove it and let it crash when bad things happen

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3354,6 +3358,23 @@ void Graph::propertyEditor()
...
+                        {
+                            std::string inputName = inputs[i]->_name;
+                            UiNodePtr inputNode = node->getConnectedNode(inputName);
+                            if (inputNode != nullptr && inputNode->getName() == name && node->getNode())
+                            {
+                                node->getNode()->getInput(inputName)->setAttribute("nodegraph", name);
+                            }
+                        }
+                    }
+                }
```
jsjtxietian, line 3377: I copy pasted some other places's code to here since I'm new to the code base, not quite sure how to do this update elegantly. I appreciate your futther suggestion!

jstone-lucasfilm:
This looks very promising, thanks @jsjtxietian!  I'm CC'ing @lfl-eholthouser, who knows this code the best, for her thoughts and suggestions.

source/MaterialXGraphEditor/Graph.cpp
lfl-eholthouser: I think we should keep this in order to avoid crashing when possible.

jsjtxietian: Agreed! Should I add something like `std::err << ` to log the error instead of just return?

jstone-lucasfilm: Since this is an unexpected case, and the additional information would only be valuable to developers, I think it's fine to leave this as a silent handling of the error.

jstone-lucasfilm (approved):
Thanks for this fix, @jsjtxietian!

PR #1937, opened 2024-07-13
Make glTF PBR thickness input control thin-walledness

### LINKED ISSUE #1936: Update the glTF PBR implementation to support thin-walledness
As described in the initial glTF PBR PR #861, thin-walled support was not implemented due to technological limitations.

With the recent changes in MaterialX 1.39 (PR #1866), it seems that the functionality can now be implemented.

pablode:
Attempts to fix #1936. Needs some testing.


pablode:
Just a heads-up: looks good in MaterialXView, but I'm waiting for USD with MaterialX 1.39 so that this can be tested with a path tracer.

jstone-lucasfilm:
@pablode I believe it's now possible to test MaterialX 1.39 functionality in Omniverse Kit 106, if you'd like to give this a try.

pablode:
@jstone-lucasfilm I just tested this with MaterialX 1.39.1 and the latest MDL SDK. Unfortunately I'm getting an MDL compilation error:
```
[12:57:32.022] (ERROR) [MDL] <string>(80,121): C219 'mx_surface' : cannot convert argument 'mxp_thin_walled' from 'bool' to 'uniform bool': depends on parameter type
[12:57:32.022] (ERROR) [MDL] <string>(80,121): C219 'mx_surface' : cannot convert argument 'mxp_thin_walled' from 'bool' to 'uniform bool': depends on parameter type
```
At first glance, the problem seems to be that `thickness` is declared with [`uniform="false"`](https://github.com/AcademySoftwareFoundation/MaterialX/blob/a3a7744f3e7a290b4d029ea60452f4bab8a9496f/libraries/bxdf/gltf_pbr.mtlx#L27) in the gltf_pbr, while the surface node has `thin_walled` declared as [uniform="true"](https://github.com/AcademySoftwareFoundation/MaterialX/blob/a3a7744f3e7a290b4d029ea60452f4bab8a9496f/libraries/pbrlib/pbrlib_defs.mtlx#L244).

jstone-lucasfilm:
Thanks for testing this, @pablode, and I'm CC'ing @niklasharrysson for his thoughts on which direction we ought to go with the uniformity of thickness and thin-walled properties in MaterialX.

niklasharrysson:
Since the `thin_walled` input is uniform it doesn't work to connect a network with a varying / non-uniform parameter upstream. In order for this to work the `thickness` input must also be set as uniform.

`thin_walled` is set to uniform since a surface must be either a closed volume (thin_walled = false) or a thin doublesided sheet (thin_walled = true). It can't be both, depending on a varying variable.

It is MDL in particular that enforces this rule, as it has the concept of uniform variables built into the language. And `thin_walled` is a uniform in the MDL material model.

pablode:
It seems to me that there are two options on how to continue:
1) add a dedicated `thin_walled` input with `uniform="true"`
2) make `thickness` uniform.

As `thickness` is a [mutable attribute](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/ObjectModel.adoc#512-khr_materials_volume) according to the glTF spec, it could happen to be animated. Therefore, I think the first option is the way to go.

Going to close this PR for now, might open up a new one with the proposed change.

PR #1949, opened 2024-07-24
Initial translation from `open_pbr_surface` to `standard_surface`

mkuo-lucasfilm:
1. Authored an initial pass at the translation from OpenPBR to standard surface
- This would help to support OpenPBR in applications using earlier versions of MaterialX, which only have access to Standard Surface, but may want to participate in the OpenPBR ecosystem before they upgrade.

### DIFF
--- a/libraries/bxdf/translation/open_pbr_to_standard_surface.mtlx
+++ b/libraries/bxdf/translation/open_pbr_to_standard_surface.mtlx
@@ -96,7 +96,7 @@
 
   <nodegraph name="NG_open_pbr_surface_to_standard_surface" nodedef="ND_open_pbr_surface_to_standard_surface">
 
-    <!--Diffuse-->
+    <!--Base-->
     <dot name="base" type="float">
       <input name="in" type="float" interfacename="base_weight" />
     </dot>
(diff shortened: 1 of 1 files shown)


kwokcb:
Hi @mkuo-lucasfilm ,
I brought this up with @jstone-lucasfilm but wanted to ask if you dig into a couple of things.

1. First is if "chaining" translators will work now that you have actual nodes. 
There are already std-surface to glTF and USDPreviewSurface. The question is, with this new translator can a user connect a OpenPBR -> std-surface translator and then connect that to a std-surface -> X translator ?

    I know it's all node name based so not sure if it's actually possible. It would be nice if it does not to allow translations in 1 step since there is a bake that occurs for each step I believe so a 2-step translation may not be equivalent to single step one.

2. A second item is that as the version of OpenPBR / or std-surface changes since these translators are purely name based (without version information) how will this be handled ? e.g. even now how to you translate to the 2 versions of std surface?

Thanks.

jstone-lucasfilm:
Good questions, @kwokcb, and I can provide some initial thoughts:

1. We haven't yet tested the chaining of shader translation graphs on our side, but my guess is that we'd encounter minor issues in the shader translation code that would need to be addressed, after which this pathway could be tested in production situations.
2. We'll definitely need to resolve this in the future, and for now, the source and destination shaders are required to be at their default versions.  The simplest solution would be a convention for incorporating version numbers in translation node names, but it seems worthwhile to see if we can come up with a more elegant solution.

jstone-lucasfilm (approved):

PR #1953, opened 2024-07-28
Move proposed Specification changes to separate Proposals doc

dbsmythe:
Nodes, Elements and other things that were described in the MaterialX Specification but which have not actully been implemented yet are now moved out of tyhe Spec docs and into a separate MaterialX Proposals document.

Also adds table-of-content links in the PBR Spec for Khronos glTF PBR and OpenPBR Surface shading model examples, and removes duplicate descriptions of the Logical "and", "or" and "not" operators.

### DIFF
--- a/documents/Specification/MaterialX.Proposals.md
+++ b/documents/Specification/MaterialX.Proposals.md
@@ -1,0 +1,346 @@
+<!-----
+MaterialX Proposals v1.39
+----->
+
+
+# MaterialX: Proposed Additions and Changes
+
+**Proposals for Version 1.39**  
+July 20, 2024
+
+
+# Introduction
+
+The [MaterialX Specification](./MaterialX.Specification.md) has historically included descriptions of not just current functionality, but also forward-looking proposed functionality intended for eventual implementation.  This has resulted in a lack of clarity about which functionality is actually supported in the library, and which are proposed additions.
+
+To remedy this, those forward-looking proposals have been moved into this Proposed Additions and Changes document to be discussed and debated. These descriptions can then be migrated into the appropriate formal Specification document once actually implemented in the code base.  New proposals for changes and additions to MaterialX may be added to this document once a generally favorable consensus from the community is reached.
+
+
+
+## Table of Contents
+
+**[Introduction](#introduction)**  
+
+**[Proposals: General](#propose-general)**  
+
+**[Proposals: Elements](#propose-elements)**  
+
+**[Proposals: Stdlib Nodes](#propose-stdlib-nodes)**  
+
+**[Proposals: PBR Nodes](#propose-pbr-nodes)**  
+
+**[Proposals: NPR Nodes](#propose-npr-nodes)**  
+
+
+<p>&nbsp;<p><hr><p>
+
+# Proposals: General<a id="propose-general"></a>
+
+
+## Color Spaces
+
+When the OCIO NanoColor library (provide link) becomes available, MaterialX should support the official colorspace names in that spec, with the current MaterialX colorspace names supported as aliases.
+
+MaterialX should also support the following color spaces:
+* `lin_rec2020`
+* `g22_rec2020`
+
+
+
+<p>&nbsp;<p><hr><p>
+
+# Proposals: Elements<a id="propose-elements"></a>
+
+
+### AOV Output Elements
+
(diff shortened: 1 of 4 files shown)


documents/Specification/MaterialX.Proposals.md
```diff
@@ -0,0 +1,342 @@
...
+* **`separatecolor4`** (NG): output the RGB and alpha channels of a color4 as separate outputs.
+    * `in` (color4): the input value or nodename
+    * `outcolor` (output, color3): the RGB channel values.
+    * `outa` (output, float): the value of the alpha channel.
+
+
+
+### Convolution Nodes
+
+<a id="node-blur"> </a>
```
jstone-lucasfilm, line 324: Although our reference implementation of the `blur` node in MaterialX is not yet up to the quality level of the other MaterialX nodes, it does have initial implementations in GLSL/OSL, and these implementations are validated in our test suite.

Given that context, it may make more sense to leave this node in the official specification, and to clarify in the README of our data libraries which nodes (including `blur`) have implementations that are believed to require additional work to reach the level of quality we expect.

dbsmythe: Good point, change made.

documents/Specification/MaterialX.Specification.md
```diff
@@ -1148,19 +1123,9 @@ Applications may also reference other renderer-specific named spaces, at the exp
...
-```
-
 Standard Global nodes:
 
-<a id="node-ambientocclusion"> </a>
-
-* **`ambientocclusion`**: Compute the ambient occlusion at the current surface point, returning a scalar value between 0 and 1.  Ambient occlusion represents the accessibility of each surface point to ambient lighting, with larger values representing greater accessibility to light.  This node must be of type float.
-    * `coneangle` (float): the half-angle of a cone about the surface normal, within which geometric surface features are considered as potential occluders.  The unit for this input is degrees, and its default value is 90.0 (full hemisphere).
-    * `maxdistance` (float): the maximum distance from the surface point at which geometric surface features are considered as potential occluders.  Defaults to 1e38, e.g. "unlimited".
+There are currently no Global Nodes implemented in MaterialX.
```
jstone-lucasfilm, line 1128: Since our sole example of a Global Node has been moved to the Proposals document, perhaps we should take this opportunity to remove the section itself, and to consider `ambientocclusion` to be a proposal for the NPR Nodes specification?

dbsmythe: There are certainly use cases for ambient occlusion in pattern generation in a PBR render, e.g. for procedural dirt accumulation in crevices, but if moving this node over to the NPR document would make it easier or more likely for it to actually get implemented, I'm all for it.
I'm curious what others think.

jstone-lucasfilm: Good points, @dbsmythe, and I'm open to keeping the current Global Nodes category for this proposal.  Maybe we should delete the empty section from the main specification, though, making it a category that will come into existence when `ambientocclusion` or a similar node gets added?

documents/Specification/MaterialX.Specification.md
```diff
@@ -1363,23 +1323,6 @@ Math nodes have one or two spatially-varying inputs, and are used to perform a m
     * `in1` (float or color<em>N</em> or vector<em>N</em>): the first value or nodename
     * `in2` (same type as `in1` or float): the second value or nodename
 
-<a id="node-and"> </a>
```
ld-kerley, line 1366: https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/stdlib_defs.mtlx#L4766

These nodes are now part of the implementation, along with the additional `xor` for completeness.

documents/Specification/MaterialX.Specification.md
ld-kerley: +1 for removing the empty section, I think that when we get an Implementation of `ambientocclusion` that will be a good time to debate what section it belongs to, because we will perhaps have a little more context around that work.

For instance, in the OSL TSC, we sometimes debate the difference between ambient occlusion that is resolved at shade time, perhaps even implemented inside the shader, vs a deferred ambient occlusion accessed via a closure, all that to be said, the devil is in the details, and seeing the future is sometimes hard :)

documents/Specification/MaterialX.Specification.md
dbsmythe: Note that previously, the "and", "or" and "not" nodes were documented *twice* in the Spec, so I removed one set (from the "math nodes" section"), and kept the other (in a new Logical Operators section).

documents/Specification/MaterialX.Specification.md
dbsmythe: The empty Globals section (as well as any references to it) have been removed in a new commit.

jstone-lucasfilm (approved):
This looks great to me, @dbsmythe, and I think it's worthwhile to move forward with this improvement!

PR #1956, opened 2024-07-31
Update `standard_surface` to `open_pbr` translation graph  

mkuo-lucasfilm:
Replacing coat_affect_roughness with a constant value, since a coating always affects roughness in OpenPBR and is not dependent on the value of coat_darkening


jstone-lucasfilm (approved):

PR #1957, opened 2024-08-01
Fix swizzle edge cases in version upgrade

nadult:
This PR fixes 2 edge cases when upgrading materials from version 1.38 to 1.39 involving swizzle nodes:
- swizzle xyz -> xyz, rgb -> rgb wasn't handled properly, because convert nodedefs were missing for such cases
- swizzle xy -> xxx is incorrectly changed to convert node

Tests are also added to cover these cases.


source/MaterialXTest/MaterialXCore/Version.cpp
```diff
@@ -0,0 +1,88 @@
...
+//
+
+#include <MaterialXTest/External/Catch/catch.hpp>
+
+#include <MaterialXCore/Util.h>
+#include <MaterialXCore/Document.h>
+#include <MaterialXFormat/XmlIo.h>
+#include <MaterialXFormat/Util.h>
+
+// TODO: move to format/ ?
```
nadult, line 13: TODO: remove.

source/MaterialXTest/MaterialXCore/Version.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+namespace mx = MaterialX;
+
+const char* swizzleDoc1 = R"(
```
nadult, line 17: Rename to swizzleDoc.

nadult:
@jstone-lucasfilm sorry for the formatting errors. I actually thought that the code was properly formatted, because I'm using an automatic formatting plugin in Visual Studio based on clang-format. But it turns out, that I'm using a newer version (16) than the version which is used in MaterialX (13 I guess?). Unfortunately even with the same .clang-format configuration file, there are slight differences in formatting between different versions...

source/MaterialXCore/Version.cpp
```diff
@@ -1228,7 +1245,7 @@ void Document::upgradeVersion()
                             node->setInputValue("index", (int) CHANNEL_INDEX_MAP.at(channelString[0]));
                         }
                     }
-                    else if (CHANNEL_CONVERT_PATTERNS.count(channelString))
+                    else if (sourceType != destType && matchesConvertPattern(channelString, sourceChannelCount))
```
jstone-lucasfilm, line 1248: This looks like a great improvement, @nadult, though I wonder if it could be written even more efficiently using a `std::unordered_map<string, size_t>` containing the new data you define above.

If `CHANNEL_CONVERT_PATTERNS` were a `std::unordered_map<string, size_t>`, then this test could simply be written as:

```
else if (sourceType != destType && CHANNEL_CONVERT_PATTERNS.count(std::make_pair(channelString, sourceChannelCount)))
```

nadult: The code could be a couple lines shorter with std::set, but IMO it's not worth it. The number of elements is so low that brute force loop will be more performant than by using a red-black tree in std::set. Additionally each additional instantiation of such a complicated data structure involves additional bytecode which would have to be generated (probably around 50KB) and it would increase compilation time as well.

Adding a single instance of std::set<> won't matter much, but if you make many such decisions then the binary size and compilation time of your library might increase considerably without any real necessity.

nadult: I simplified this expression in a different way, so that now I hope that you don't see a need to use std::set<> here :)
Also I replaced std::vector<> with simple C arrays in two cases where vectors weren't needed at all. Just by doing this the size of release version of library MaterialXCore.lib decreased by 45KB :)

jstone-lucasfilm: I like this overall approach!  My one recommendation would be to use `std::array` instead of ANSI C arrays, so that we're aligned with the broader conventions of the MaterialX codebase, but this high-level idea of using fixed-length arrays instead of `std::vector` or `std::set` seems very appropriate here.

nadult: The only problem with std::array is that it's size has to be specified, which is painful:
https://stackoverflow.com/questions/26351587/how-to-create-stdarray-with-initialization-list-without-providing-size-directl

There is a solution to that issue in C++20: https://en.cppreference.com/w/cpp/container/array/to_array
But AFAIK MaterialX uses C++17, so for now I would suggest to stick to C arrays.

Alternatively I could implement to_array function in MaterialX until it switches to C++20. It's only several lines of code. Which solution would you prefer?

jstone-lucasfilm: Our current approach is to just specify sizes directly in `std::array` instances, since the additional overhead is not a major inconvenience:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/67702daa934d6bb397ec6e540b7dd41bc5c12650/source/MaterialXGenShader/Nodes/ConvolutionNode.cpp#L23

When we set C++20 as a minimum for MaterialX in the future, I like the idea of simplifying our arrays globally using `to_array`, but I would say that this improvement can wait until then.

nadult: Fixed.

jstone-lucasfilm (approved):
Thanks for this fix, @nadult!

PR #1958, opened 2024-08-02
Improved `open_pbr_surface` to `standard_surface` translation

mkuo-lucasfilm:
Integrated the logic of coat_darkening from OpenPBR into the calculations of base_color and subsurface_color

### DIFF
--- a/libraries/bxdf/translation/open_pbr_to_standard_surface.mtlx
+++ b/libraries/bxdf/translation/open_pbr_to_standard_surface.mtlx
@@ -95,17 +95,7 @@
 
   <nodegraph name="NG_open_pbr_surface_to_standard_surface" nodedef="ND_open_pbr_surface_to_standard_surface">
 
-    <!--Base-->
-    <dot name="base" type="float">
-      <input name="in" type="float" interfacename="base_weight" />
-    </dot>
-    <dot name="diffuseRoughness" type="float">
-      <input name="in" type="float" interfacename="base_diffuse_roughness" />
-    </dot>
-    <dot name="metalness" type="float">
-      <input name="in" type="float" interfacename="base_metalness" />
-    </dot>
-    <!-- Coat Darkening for Base Color  -->
+    <!-- Coat Darkening -->
     <subtract name="coat_ior_minus_one" type="float">
       <input name="in1" type="float" interfacename="coat_ior" />
       <input name="in2" type="float" value="1.0" />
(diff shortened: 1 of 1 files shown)


mkuo-lucasfilm:
Pearl comparisons:
1. OpenPBR
2. Standard Surface (original shader translation graph)
3. Standard Surface (latest shader translation graph)

Generated difference images between the Standard Surface image and the OpenPBR image for both versions. The original average difference was 0.92 (out of 255), and the new average difference is 0.57. 
open_pbr_pearl
open_pbr_pearl_ss
open_pbr_pearl_ss_3

mkuo-lucasfilm:
Carpaint comparisons:
1. OpenPBR
2. Standard Surface (original shader translation graph)
3. Standard Surface (latest shader translation graph)

Generated difference images between the Standard Surface image and the OpenPBR image for both versions. The original has an average difference of 1.57 (out of 255), and the new one has an average difference of 1.21.

open_pbr_carpaint
open_pbr_carpaint_ss
ments/assets/36a5747f-4322-4445-81cb-377ddd68b4e9)

mkuo-lucasfilm:
Carpaint comparisons:
1. OpenPBR
2. Standard Surface (original shader translation graph)
3. Standard Surface (latest shader translation graph)

Generated difference images between the Standard Surface image and the OpenPBR image for both versions. The original has an average difference of 1.57 (out of 255), and the new one has an average difference of 1.21.

open_pbr_carpaint
open_pbr_carpaint_ss
open_pbr_carpaint_ss_3

portsmouth:
I see, so this bakes in the darkening factor into the `base_color` (though `coat_affect_color` should be off I think, for that to work). That will work for an initial translation, though of course if the user decides to remove the coat afterwards (from the Standard Surface), the darkening persists.

The way we translate should depend on what the intended usage of the translated shader is. If it is assumed to be baked and immutable, then this kind of baking of the darkening is fine. But if the user might want to change the parameters (e.g. remove coat etc.), then it would be better to try to reproduce the structure as well as possible, e.g. keep the base color and just use the `coat_affect_color` parametrization to approximate the darkening.

I'm not sure what the more appropriate policy would be for these translation graphs.

jstone-lucasfilm:
Thanks for these notes, @portsmouth!

In my own experience, the most important quality of a shader translation graph is its _visual parity_, since the majority of the artistic work on the material usually occurs before shader translation was applied, and the receiving party is most interested in rendering the material as accurately as possible.

In situations where additional edits are needed, it's also possible to share the _unbaked_ translated graph, in which the original shading model interface of the material remains available for further editing, though in practice it's more common to bake out textures in the new shading model.

From this perspective, I would consider @mkuo-lucasfilm's latest work to be a meaningful improvement, since there's a small but measurable improvement to parity for coated materials, and no reduction of parity for other materials.

jstone-lucasfilm:
Let's move forward with this refinement to the shader translation graph, and there's still time for additional improvements before the release of MaterialX 1.39.1.

Thanks for your great work on this, @mkuo-lucasfilm!

jstone-lucasfilm (approved):

PR #1960, opened 2024-08-05
Fix incorrectly transposed matrix for displayp3 colorspaces

ld-kerley:
When originally authored the matrix was authored transposed.  Further internal testing revealed this to manifest as an undesired shift towards red. 

## Before
mtlx_p3_to_rec709_before

## After
mtlx_p3_to_rec709_fixed

### DIFF
--- a/libraries/cmlib/cmlib_ng.mtlx
+++ b/libraries/cmlib/cmlib_ng.mtlx
@@ -315,7 +315,8 @@
 
   <nodegraph name="NG_srgb_displayp3_to_lin_rec709_color3" nodedef="ND_srgb_displayp3_to_lin_rec709_color3">
     <constant name="mat" type="matrix33">
-      <input name="value" type="matrix33" value="1.22493029, -0.22492968, 0.00000006, -0.04205868,  1.04205894, -0.00000001, -0.01964128, -0.07864794, 1.09828925" />
+      <input name="value" type="matrix33"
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Great catch, thanks @ld-kerley!  I'm CC'ing @doug-walker and @carolalynn on this PR, to make sure we're remaining aligned with the upcoming NanoColor project.

dgovil:
Also just cc'ing @meshula about this too.

doug-walker:
@jstone-lucasfilm , here is what NanoColor would give for that conversion, based on the draft config that implements the ASWF Color Interop Forum color spaces. 
1.2249401763e+00, -4.2056954710e-02, -1.9637554590e-02],
       [-2.2494017628e-01,  1.0420569547e+00, -7.8636045551e-02],
       [ 2.2204460493e-16, -5.5511151231e-17,  1.0982736001e+00

jstone-lucasfilm:
Thanks @doug-walker, and in this context, I'd have no objections to moving forward with the matrix transpose in this pull request.  Further in the future, we're planning to import our color transform graphs directly from NanoColor, at which point any differences in precision assumptions should be resolved.

jstone-lucasfilm (approved):

meshula:
Since @dgovil summoned me, the matrix Doug wrote out matches the results of the nanocolor implementation currently underpinning GfColor in OpenUSD.

PR #1962, opened 2024-08-08
Fix Fresnel computation in GLSL transmission

jstone-lucasfilm:
This changelist fixes an error in the Fresnel computation for GLSL transmission, improving the visual parity of transmissive materials across shading languages.

With this fix in place, the graph definition of LamaGeneralizedSchlick no longer requires a manual inversion of its F0 and F90 inputs for transmission, as this inversion is handled correctly in GLSL as well as OSL and MDL.


jstone-lucasfilm:
Here are the latest reference renders in GLSL and OSL, leveraging the updates in this changelist:

[MaterialXRenderTests_08_07_2024.pdf](https://github.com/user-attachments/files/16551762/MaterialXRenderTests_08_07_2024.pdf)

PR #1963, opened 2024-08-08
Update documentation of OpenPBR and Standard Surface translations

mkuo-lucasfilm:
Clarified documentation where the motivation for the conversion might not be clear - both for the translation graph from OpenPBR to Standard Surface, and the translation graph from Standard Surface to OpenPBR.


jstone-lucasfilm (approved):
These updates look good to me, thanks @mkuo-lucasfilm!

PR #1968, opened 2024-08-12
Add Chiang hair BSDF

msuzuki-nvidia:
Add Chiang hair BSDF.

This PR provides Chiang hair BSDF model and related nodes proposed by #1973 

The nodes to add are:
- `<chiang_hair_bsdf>` : The BSDF
- `<chiang_hair_roughness>` : User friendly roughness mapping
- `<chiang_hair_absorption_from_color>` : Absorption coefficient mapping from user input color
- `<deon_hair_absorption_from_melanin>` : Absorption coefficient mapping from melanin parameters

It also include `simple_hair_default` material as a node graph example.


ld-kerley:
Just to recap concretely the request made in the TSC meeting - it would be great if this PR could also include updates to the PBR spec document reflecting the new nodes being added here

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.PBRSpec.md

msuzuki-nvidia:
Thanks for the recap @ld-kerley. I will update this PR including the PBR spec document sometime soon.

BrianSharpe:
I think the sqrtPiOver8 factor might be an internal BSDF implementation detail, rather than something that should be done externally.  For example, MDL already applies this factor internally.
https://github.com/NVIDIA/MDL-SDK/blob/master/src/mdl/jit/libbsdf/libbsdf_hair.h#L174

msuzuki-nvidia:
It can be internal. Although that part in MDL I'm not sure. MDL applies the $\sqrt{\pi/8}$ factor for both longitudinal and azimuthal roughness but the original paper (Appendix A. eq. 12) applies it only for azimuthal roughness, so I followed the paper. Other than that, it works for me.

BrianSharpe:
> MDL applies the π/8 factor for both longitudinal and azimuthal roughness

I don't think it does.
The MDL hair_prepare_roughness() function in that github link I posted only applies it to the Y component.

Given the π/8 factor is in the appendix, rather than in equation 8, makes me think it probably should be part of the internal BSDF implementation rather than in <hair_roughness>.  But happy to defer to others for that decision.

msuzuki-nvidia:
> The MDL hair_prepare_roughness() function in that github link I posted only applies it to the Y component.

You're right. The factor is applied to only azimuthal!

I also see where your suggestion came from, and I feel the same now. I will include the change in next update.

BrianSharpe:
I think <hair_roughness> might need some more changes.  In its current form, someone can't tweak the TT and TRT scaling factors.  Also, it's a little awkward to implement a scale on R longitudinal roughness (Chiang section 4.3).  So I'm thinking of two potential changes to accommodate this.

**possible change 1)** <hair_roughness> returns only a single "vec2 roughness" value (corresponding to roughness_R).  This means it's only implementing equations 7 and 8,  and any TT and TRT scaling should happen externally.
**or**
**possible change 2)** <hair_roughness> takes two additional inputs "tt_scale" and "trt_scale", which default to 0.25 and 4.0 (or should that be 0.5 and 2.0 with them being squared internally?)

In both cases, to implement a scale on R longitudinal roughness, it needs to be executed twice.  Option 1 has less redundancy, but is more work for the user.  Option 2 is easier to use, but has redundancy.  I'm fine with either option.

msuzuki-nvidia:
I would prefer the second option with pre-squared defaults, 0.5 and 2.0. Indeed these scaling values are somewhat arbitrary but most implementations follow the Marschner's paper so I think it makes sense that the scaling is built-in and exposed in the parameters.

I will make this change as well. Thank you for the suggestion!

documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -243,6 +243,21 @@ The PBS nodes also make use of the following standard MaterialX types:
     * `normal` (vector3): Normal vector of the surface. Defaults to world space normal.
     * `mode` (uniform string): Selects between `conty_kulla` and `zeltner` sheen models. Defaults to `conty_kulla`.
 
+<a id="node-hair-chiang-bsdf"> </a>
+
+* **`hair_chiang_bsdf`**: Constructs a hair BSDF based on the Chiang hair shading model[^Chiang2016]. This node does not support vertical layering.
```
jstone-lucasfilm, line 248: For consistency with existing MaterialX PBR nodes such as `burley_diffuse_bsdf`, perhaps the name of this new node should be `chiang_hair_bsdf`?

msuzuki-nvidia: I was wondering about that too, though other nodes related to hair are called `mx_hair_*` then I took the other consistency. Either works for me so I'm fine to follow your decision.

jstone-lucasfilm: If you're open to this convention, I'd definitely recommend `chiang_hair_bsdf` for consistency with other PBR nodes, and the helper functions starting with `mx_hair_chiang_*` could be updated to `mx_chiang_hair_*` to match.

msuzuki-nvidia: Sounds good. Technically, the method of `mx_hair_absorption_from_melanin` isn't from Chiang et al., so maybe we should rename those to:

- `mx_chiang_hair_bsdf`
- `mx_chiang_hair_roughness`
- `mx_chiang_hair_absorption_from_color`
- `mx_deon_hair_absorption_from_melanin`

How do they sound for you?

jstone-lucasfilm: Those names look good to me!

msuzuki-nvidia: Great, the changes submitted.

libraries/pbrlib/pbrlib_defs.mtlx
```diff
@@ -404,4 +423,40 @@
...
+  </nodedef>
+
+  <!--
+    Node: <chiang_hair_roughness>
+    Calculates hair roughness for R, TT and TRT component.
+  -->
+  <nodedef name="ND_chiang_hair_roughness" node="chiang_hair_roughness" nodegroup="pbr" doc="Calculates hair roughness for R, TT and TRT component.">
+    <input name="longitudinal" type="float" value="0.1" />
+    <input name="azimuthal" type="float" value="0.2" />
+    <input name="scale_TT" type="float" value="0.5" />
```
BrianSharpe, line 455: we could tag scale_TT + scale_TRT as "uiadvanced"
ie...
`<input name="scale_TT" type="float" value="0.5" uiadvanced="true"/>`
`<input name="scale_TRT" type="float" value="2.0" uiadvanced="true"/>`

documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -374,6 +389,27 @@ Note that the standard library includes definitions for [**`displacement`**](./M
     * `ior` (**output**, vector3): Computed index of refraction.
     * `extinction` (**output**, vector3): Computed extinction coefficient.
 
+<a id="node-chiang-hair-roughness"> </a>
+
+* **`chiang_hair_roughness`**: Converts the artistic parameterization hair roughness to roughness for R, TT and TRT lobes. Output type `multioutput`, `roughness_R`, `roughness_TT` and `roughness_TRT`, `vector2` type.
```
BrianSharpe, line 394: might be good to explicitly mention Chiang paper for this node too
ie
`Converts the artistic parameterization hair roughness to roughness for R, TT and TRT lobes, as described in [^Chiang2016]`

libraries/pbrlib/pbrlib_defs.mtlx
msuzuki-nvidia: sure, added those.

documents/Specification/MaterialX.PBRSpec.md
msuzuki-nvidia: sounds good. added.

jstone-lucasfilm (approved):
This looks great to me, thanks @msuzuki-nvidia, and I'm CC'ing @fpsunflower and @niklasharrysson for their thoughts from the OSL perspective.

msuzuki-nvidia:
Great to hear that, and thank you for correcting the spelling issues!

PR #1970, opened 2024-08-14
Refine MaterialX render module build configuration

ld-kerley:
Add more refined control over building MaterialX render modules. Specifically USD only needs the core render module in some build configurations (iOS).

Also add a guard to ensure that MaterialXGenGlsl is built if MaterialXGenMsl is built, because of the dependency in the data library.  USD uses `MaterialX::GlslShaderGenerator::TARGET` to locate those resources.


jstone-lucasfilm:
@ld-kerley I like the idea behind this proposal, though the split between `MATERIALX_BUILD_RENDER` and `MATERIALX_BUILD_RENDER_CORE` seems counterintuitive to me, as there's no corresponding module named MaterialXRenderCore.

As an alternative, what if we added a `MATERIALX_BUILD_RENDER_PLATFORMS` flag, defaulting to `ON`, with all platform-specific render modules gated by this flag?

This would allow USD developers to opt out of platform-specific render modules such as MaterialXRenderGlsl and MaterialXRenderOsl, without losing access to the MaterialXRender module that they still need.

jstone-lucasfilm:
Thanks for the updates, @ld-kerley, and this looks really promising!  I'll take a closer look in upcoming days, and let's see if we can get this change ready to include in MaterialX v1.39.2.

jstone-lucasfilm (approved):

PR #1971, opened 2024-08-14
Assign scope to OpenImageIO linking

jstone-lucasfilm:
This changelist assigns a required scope keyword to the linking step for OpenImageIO in MaterialXRender.

CMake requires scope keywords to be used in all linking steps after they are used in one, so this change is needed to match the new behavior of mx_add_library introduced in https://github.com/AcademySoftwareFoundation/MaterialX/pull/1944.


ld-kerley:
Looks reasonable to me

PR #1972, opened 2024-08-14
Update Lama Testing Suite

mkuo-lucasfilm:
Changed the default values for Lama nodes to match Pixar Renderman definitions


jstone-lucasfilm (approved):

Since these updates are so closely related to previous work on MaterialX Lama in 1.39.1, I think it makes sense to include them in the same release, so I'll go ahead and merge this.

PR #1975, opened 2024-08-16
Fix combine2 implementation in MDL

krohmerNV:
Fix a regression after this [commit](https://github.com/AcademySoftwareFoundation/MaterialX/commit/4713ec9ff9edcc33e6ec663e7d4c35e3ce68e385#diff-0bae9161e30f26f13ee964c3007c7f13d51d971da82c9fab4184cda13e26400c).

Add a custom implementation for combine2(color, float) because the index access is not allowed on the MDL color type.


jstone-lucasfilm (approved):
Thanks for this fix, @krohmerNV, and this looks worthwhile to include in 1.39.1.

krohmerNV:
Thanks for the fixes @jstone-lucasfilm
And yes, it would be nice to be included in the next release, thanks!

PR #1977, opened 2024-08-17
Fix banding artifacts in macOS viewer

shill-lucasfilm:
This addresses subtle banding on macOS with 10 bit / EDR displays.

I'm not a Metal expert and I suspect that the conversion code I added for screenshot saving could be improved (see commit comment), or ideally avoided (by saving to an HDR format such as EXR), but it gets the job done.

Before:
before

After:
after


jstone-lucasfilm:
Thanks for posting this fix, @shill-lucasfilm!  I'm CC'ing @dgovil, @lkerley, and @Morteeza for their expertise on the macOS side.

source/MaterialXView/Viewer.cpp
```diff
@@ -137,6 +137,12 @@ void applyModifiers(mx::DocumentPtr doc, const DocumentModifiers& modifiers)
     }
 }
 
+bool useFloatBuffer()
+{
+    auto [capability_10bit, capability_EDR] = ng::test_10bit_edr_support();
```
meshula, line 142: ng treats a request for a float buffer as a request for both a float buffer, and an EDR layer, however if the hardware can't do it, the float buffer will still work fine as the EDR status is just a request. Over unity values will get clamped in that case, but this viewer doesn't do HDR rendering, so you won't see the effect of that.

This ng function call tells you if *any* connected NSScreen is EDR or 10 bit capable, not necessarily the monitor your window is on...  Since the window manager can deal with 16 bit and edr frame buffers irrespective of the device capabilities, I would personally suggest just running a half frame buffer all the time.

shill-lucasfilm: Good info! It would certainly simplify the code to always use 16 bit buffers, so if everyone is fine with that then I can make that change.

shill-lucasfilm: Having said that, there would probably need to be changes elsewhere to support 16 bit across the board. Since this PR is focused on macOS, we could limit float buffers to that platform (perhaps via `MATERIALXVIEW_METAL_BACKEND`) for now.

jstone-lucasfilm:
@dgovil @ld-kerley @meshula Does the current change look like a good step forward, which we can build upon in future improvements, or are there changes that you'd recommend before we proceed?

dgovil:
In general, it looks good to me, and I especially agree with Nicks comments (and he's most familiar with it too). I haven't had a chance to build and try it so I'll defer to @ld-kerley on that front.

ld-kerley:
Looks good to me

shill-lucasfilm:
I updated the PR with the simplified version, in case anyone wants to take a final look (best viewed with "Show all changes").

jstone-lucasfilm (approved):
This looks good to me, @shill-lucasfilm, and if there are no objections from other reviewers, I'd recommend that we merge this after the release of MaterialX v1.39.1 next week.

PR #1980, opened 2024-08-19
Decouple GLSL and MSL "libraries" location.

ld-kerley:
Currently there is a dependency between the GLSL and MSL shader generation modules, because both targets share source code (which is great), the MSL shader generator uses files from the `genglsl` libraries locations.  

This becomes a small problem when building with USD, if the `MaterialXGenGLSL` module is not built, as the code uses `MaterialXGenGLSL::TARGET` to generate the `genglsl` component in the file path.

This PR aims to address this, by more correctly decoupling the MSL and GLSL shader generators, while still allowing the source code, and there only be a single copy of the code in the source repo, as well as make some of the cmake dependency tracking more robust.

All files shared between `genglsl` and `genmsl` have been moved from the `genglsl` location to a shared `_hwCommon` location.

The `libraries` CMakeLists.txt has been updated to "build" a staged folder from the source folder, where 
* Only the shader generators that are enabled will have their source code staged to the `libaries` folder.  This will result in smaller installations, where only one shader generator is desired.
* The `_hwCommon` files are correctly copied to both `genmsl` and `genglsl`.
* The OSL legacy closures file switch is also staged in this build location, this unifies this logic in one place.

This also slightly simplifies the testing logic as we don't have to replicate the OSL legacy closure logic, and instead can just copy the staged files to the location for testing.  The dependency tracking for copying the testing files is also fixed, so iterative edits to any of the files in `libraries` should be automatically installed to the testing location, previously this location would need to be deleted for cmake to copy the files.


jstone-lucasfilm:
@ld-kerley Since our thinking on the sharing of hardware shading code has evolved so much since this proposal was written, with our current focus being on meta-shading languages such as Slang, would it make sense to close this out?

ld-kerley:
Since this PR has been posted there has been a lot of work done to more tightly couple the MSL backend to the GLSL backend, and because of that it probably makes sense to close out this PR.

Slang in MaterialX still seems a fair way out and yet to even be labelled experimental - so I'd like to keep the idea presented open.  It would be ideal to be able to only install the minimal set of files required for a single target language.

jstone-lucasfilm:
That sounds good to me, thanks @ld-kerley!

Although it's a valid perspective to view our recent work as increasing the _coupling_ between hardware shading languages, my own perspective is that we're striving to maximize _code sharing_ across these languages, which takes us closer to a future where a single meta-language can be used to define all hardware shading code for MaterialX in a unified way.

Although Slang isn't quite ready to take on this role in the industry, it's my sense that it has very string potential to fill this role in the future, and improving our code sharing should put us in a good position to make this transition when the time comes.

Issue #1978, opened 2024-08-19
Smoothstep breaks GLSL compilation when multiple type signature nodes are used


HardCoreCodin:
When multiple signature variants of smoothstep participate in a graph, the viewport render goes blank and this shows in the output:
```
Error in compiling fragment shader:
0(493) : error C1013: function "mx_smoothstep_float" is already defined at 0(483)
```

This is likely because the function `mx_smoothstep_float` exists in multiple headers that get included in such cases. 

See attached example file.
[smoothstep_failure_mtlx.txt](https://github.com/user-attachments/files/16658780/smoothstep_failure_mtlx.txt)

kwokcb:
There are includes which I assume not being checked to see if that file has already been included in the non-float code implementations
```
#include "mx_smoothstep_float.glsl"
```
This breaks GLSL and Metal at least.

jstone-lucasfilm:
Good catch, @HardCoreCodin, and this sounds like a bug in our GLSL shader generation that should be fixed.

ld-kerley:
I think the fix is probably a pretty easy `#ifndef/#define` trick in the right place - I can investigate and put up a PR soon

kwokcb:
@ld-kerley , not sure if this is the place to look but I believe `#include`s are parsed and code is embedded -- but it should not try to include twice based on `ShaderStage::addInclude()`.

ld-kerley:
In this case the `#include` statement is in a file that is already being included - a quick local test and adding a standard include guard around the actual source does appear to work in both GLSL and MSL.

kwokcb:
I’ll just throw out one more option. Turn the vector versions of this into graphs that use the float version node instances — hence no include dependence exists. Would have to think some more on the other options, but adding the common include seems like a good idea if it is added on demand.

ld-kerley:
I like the nodegraph implementation idea, though if we applied this to all destination languages, then I think we would be constructing less optimal OSL code.

```
  <implementation name="IM_smoothstep_vector3_genosl" nodedef="ND_smoothstep_vector3" target="genosl" sourcecode="smoothstep({{low}}, {{high}}, {{in}})" />
```

OSL currently provides a `smoothstep()` that can take a `vector` type that will allow for more aggressive OSL optimizer optimizations. 

[...]

PR #1982, opened 2024-08-21
Add nodegraph implementations for color and vector smoothstep

ld-kerley:
Issue #1978 reports an issue with repeated code for the `smoothstep` node.  

This PR will resolve that issue (and also add a unit test that demonstrates the issue) by converting all the `smoothstep` nodes to nodegraph implementations.

This PR also simplifies the nodegraphs for `smoothstep` to allow follow the same pattern.
```
<separateX>  ->  <ND_smoothstep_float>  ->  <combineX>
```


jstone-lucasfilm (commented):
This looks like a good improvement to me, @ld-kerley!

In order to minimize duplication, would it make sense to implement the `color3` and `color4` variants in terms of the `vector3` and `vector4` versions, with `convert` nodes used for the type conversions?

ld-kerley:
I started out with that, and then opted for the version I pushed because I thought it would be easier to maintain because of the parity across all the different type implementations, and I also started to worry about the quality of the generated shader code with additional conversions.

I just briefly surveyed the existing nodegraph implementations and most (perhaps all) of the existing nodegraph implementations in `stdlib_ng.mtlx` that support the typical array of types all currently provide their own implementation for each type.  

And finally now I'm writing this, I'm wondering if the type conversion approach might complicate any overhaul of colorspace things in MaterialX in the future - having distinct implementations feels like it might be beneficial.

jstone-lucasfilm:
Those sound like good arguments to me, @ld-kerley, and perhaps we can consider this type of implementation sharing as an option for the future, where it would be applied globally if at all.

jstone-lucasfilm:
Since there's a parallel discussion about the correct behavior of `smoothstep` on the MaterialX channel of the ASWF Slack, it seems worthwhile to ask whether we should adopt the recommendations of that thread as part of this change, or whether these changes should be considered and committed independently:

https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1724060123942189

What are your thoughts?

ld-kerley:
I was just reading Larrys commit to OSL - https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/commit/67e886e9e24880f8618c07feeea8102926ad76ec

Which is the opposite of what we're currently doing for GLSL/MSL.

I think I would vote for a separate PR - which I can put up soon.  The change is easy, but we might want to debate/socialize a little that the change is coming with the wider community.

jstone-lucasfilm (approved):

PR #1985, opened 2024-08-21
Align smoothstep edge cases with OSL

ld-kerley:
Following a conversation on ASWF slack this PR was posted to OSL refining the edge cases for `smoothstep()`.

https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1851

This PR aligns the GLSL/MSL code with this change.

Before merging we need to decide...
* Does a similar change need to be applied to the MDL code?  I couldn't find any definitive direction on the edge case handling, other than the spec saying its undefined.
* Do we need to worry at all about some sort of graceful upgrade/backwards compatibility?  I'd vote for "no" and just class this as a bug-fix.

The MaterialX specification currently does not specify an opinion about the edges cases here, but I'm going to leave that to up coming specification refinement to add normative language to handle.


libraries/stdlib/genglsl/mx_smoothstep_float.glsl
```diff
@@ -1,9 +1,9 @@
 void mx_smoothstep_float(float val, float low, float high, out float result)
 {
-    if (val <= low)
-        result = 0.0;
-    else if (val >= high)
+    if (val >= high)
         result = 1.0;
+    else if (val <= low >)
```
pablode, line 5: There's a typo at the end of the line ;)

ld-kerley: good catch - but also - how did the tests even pass?!? is this even legal code?

jstone-lucasfilm (approved):
Thanks for this fix, @ld-kerley, and this looks good to me!

Issue #1981, opened 2024-08-20
Equivalence operator fails to handle vector string differences


kwokcb:
## Issue

The `operator==` operator seems to detect that a element differs when the format of the string used to represent
vectors differs. 

This seems very strange as the values are identical and would assume in memory that the comparison does not take into consideration the string formatting.

I cannot currently compare documents such as these.

### Example

Test file 1:
```xml
<?xml version="1.0"?>
<materialx version="1.39" colorspace="srgb_texture">
  <nodegraph name="patternGraph">
    <input name="ramp_right" type="color3" value="0.8392156862745098, 0.12156862745098039, 0.12156862745098039" />
    <input name="ramp_left" type="color3" value="0.08627450980392157, 0.44313725490196076, 0.7137254901960784" />
    <input name="hex_size" type="float" value="0.4" />
    <input name="hex_tiling" type="vector2" value="20.0, 10.0" />
    <output name="out" type="color3" nodename="multiply_color4" />
    <texcoord name="texcoord_vector3" type="vector2">
      <input name="index" type="integer" value="0" />
    </texcoord>
    <multiply name="multiply_color4" type="color3">
      <input name="in1" type="color3" nodename="tiledhexagons_color4" />
      <input name="in2" type="color3" nodename="ramplr_color4" />
    </multiply>
    <tiledhexagons name="tiledhexagons_color4" type="color3">
      <input name="texcoord" type="vector2" nodename="texcoord_vector3" />
      <input name="uvtiling" type="vector2" interfacename="hex_tiling" />
      <input name="uvoffset" type="vector2" value="0.0, 0.0" />
      <input name="size" type="float" interfacename="hex_size" />
      <input name="staggered" type="boolean" value="false" />
    </tiledhexagons>
    <ramplr name="ramplr_color4" type="color3">
      <input name="valuel" type="color3" interfacename="ramp_left" />
      <input name="valuer" type="color3" interfacename="ramp_right" />
      <input name="texcoord" type="vector2" nodename="texcoord_vector3" />
    </ramplr>
  </nodegraph>
  <gltf_pbr name="gltf_shader" type="surfaceshader">
    <input name="base_color" type="color3" nodegraph="patternGraph" />
  </gltf_pbr>
</materialx>
```

Test file 2:
```xml
<?xml version="1.0"?>
<materialx version="1.39" colorspace="srgb_texture">
  <nodegraph name="patternGraph">
[...]

PR #1988, opened 2024-08-23
Add utility to "normalize" numeric string values on a Document

kwokcb:
## Resolution

Fixes: #1981 

Allow for a user to "normalize" numeric string values on a document. This allows for 2 documents which may have differing formatting for these values to be comparable for equivalency if both are "normalized" before the comparison.

### Changes
- Add in string value normalization. `normalizeValueString()`. This supports all numeric types. For integers only leading and trailing spaces are removed. All vector / matrix values are separated using this string:  ", "
- Use in Document new value normalization interface:  `Document.normalizeValueStrings()`
- Add in unit tests for each.
- Add Pythong and JS bindings.

### DIFF
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -4,6 +4,7 @@
 //
 
 #include <MaterialXCore/Document.h>
+#include <MaterialXCore/Util.h>
 
 #include <mutex>
 
@@ -400,6 +401,33 @@ void Document::invalidateCache()
 {
     _cache->valid = false;
 }
+
+void Document::normalizeValueStrings(unsigned int numericPrecision)
+{
+    // For now this will only normalize numeric strings
+    StringSet numericType = { "float", "vector2", "vector3", "vector4", "color3", "color4" };
+    for (ElementPtr elem : traverseTree())
+    {
+        ValueElementPtr valueElem = elem->asA<ValueElement>();
+        if (!valueElem || numericType.end() == numericType.find(valueElem->getType()))
+        {
+            continue;
+        }
+
+        const string& originalString = valueElem->getValueString();
+        if (originalString.empty())
+        {
+            continue;
+        }
+
+        string normalizeString = normalizeNumericString(originalString, numericPrecision);
+        if (normalizeString != originalString)
+        {
+            valueElem->setValueString(normalizeString);
+        }
+    }
+}
+
 
 //
 // Deprecated methods
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -402,7 +402,7 @@ void Document::invalidateCache()
     _cache->valid = false;
 }
 
-void Document::normalizeValueStrings(unsigned int numericPrecision)
+void Document::normalizeValueStrings()
 {
     // For now this will only normalize numeric strings
     StringSet numericType = { "float", "vector2", "vector3", "vector4", "color3", "color4" };
@@ -420,7 +420,7 @@ void Document::normalizeValueStrings(unsigned int numericPrecision)
             continue;
         }
 
-        string normalizeString = normalizeNumericString(originalString, numericPrecision);
+        string normalizeString = normalizeNumericString(originalString);
         if (normalizeString != originalString)
         {
             valueElem->setValueString(normalizeString);
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -428,7 +428,6 @@ void Document::normalizeValueStrings(unsigned int numericPrecision)
     }
 }
 
-
 //
 // Deprecated methods
 //
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -404,12 +404,10 @@ void Document::invalidateCache()
 
 void Document::normalizeValueStrings()
 {
-    // For now this will only normalize numeric strings
-    StringSet numericType = { "float", "vector2", "vector3", "vector4", "color3", "color4" };
     for (ElementPtr elem : traverseTree())
     {
         ValueElementPtr valueElem = elem->asA<ValueElement>();
-        if (!valueElem || numericType.end() == numericType.find(valueElem->getType()))
+        if (!valueElem)
         {
             continue;
         }
@@ -420,7 +418,7 @@ void Document::normalizeValueStrings()
             continue;
         }
 
-        string normalizeString = normalizeNumericString(originalString);
+        string normalizeString = normalizeValueString(originalString, valueElem->getType());
         if (normalizeString != originalString)
         {
             valueElem->setValueString(normalizeString);
--- a/source/MaterialXCore/Util.cpp
+++ b/source/MaterialXCore/Util.cpp
@@ -181,6 +181,7 @@ string parentNamePath(const string& namePath)
     return EMPTY_STRING;
 }
 
+
 std::string normalizeValueString(const std::string& str, const std::string valueType)
 {
     // Currently supported value types.
@@ -189,6 +190,8 @@ std::string normalizeValueString(const std::string& str, const std::string value
     {
         return str;
     }
+
+    const string TOKEN_SEPARATOR = ", ";
 
     std::string result;
     std::string token;
@@ -210,37 +213,52 @@ std::string normalizeValueString(const std::string& str, const std::string value
             token.erase(0, token.find_first_not_of(' '));
             token.erase(token.find_last_not_of(' ') + 1);
 
-            // Remove leading zeros
-            token.erase(0, token.find_first_not_of('0'));
-
-            // Preserve 0 values
-            if (token.empty() || token[0] == '.')
+            // Skip if the token is empty. Note that this is
+            // invalid in MaterialX but can still be stored this way. 
+            if (!token.empty())
             {
-                token = "0" + token;
-            }
+                // Remove leading zeros
+                token.erase(0, token.find_first_not_of('0'));
 
-            // If the token contains a decimal point, remove trailing zeros
-            size_t decimalPos = token.find('.');
-            if (decimalPos != std::string::npos)
-            {
-                token.erase(token.find_last_not_of('0') + 1);
-
-                // If the token ends with a decimal point after removing trailing zeros, remove it
-                if (token.back() == '.')
+                // Preserve 0 values
+                if (token.empty() || token[0] == '.')
                 {
-                    token.pop_back();
+                    token = "0" + token;
+                }
+
+                // Check if string token has a 'e' character
+                // implying it is a floating point number in scientific notation.
+                // If it does not have an 'e' character, remove trailing zeros.
+                //
+                size_t ePos = token.find('e');
+                if (ePos == std::string::npos)
+                {
+                    // If the token contains a decimal point, remove trailing zeros
+                    size_t decimalPos = token.find('.');
+                    if (decimalPos != std::string::npos)
+                    {
+                        token.erase(token.find_last_not_of('0') + 1);
+
+                        // If the token ends with a decimal point after removing trailing zeros, remove it
+                        if (token.back() == '.')
+                        {
+                            token.pop_back();
+                        }
+                    }
                 }
             }
 
             // Append the formatted token to the result with a comma
-            result += token + ", ";
+            result += token + TOKEN_SEPARATOR;
         }
 
-        // Remove the last comma
+        // Remove the last separator
         if (!result.empty())
         {
-            result.pop_back();
-            result.pop_back();
+            for (size_t i = 0; i < TOKEN_SEPARATOR.size(); i++)
+            {
+                result.pop_back();
+            }
         }
     }
 
--- a/source/MaterialXCore/Util.cpp
+++ b/source/MaterialXCore/Util.cpp
@@ -6,6 +6,8 @@
 #include <MaterialXCore/Types.h>
 
 #include <cctype>
+#include <sstream>
+#include <iomanip>
 
 MATERIALX_NAMESPACE_BEGIN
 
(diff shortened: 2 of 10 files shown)


source/MaterialXCore/Util.h
```diff
@@ -69,6 +69,10 @@ MX_CORE_API string createNamePath(const StringVec& nameVec);
 /// Given a name path, return the parent name path
 MX_CORE_API string parentNamePath(const string& namePath);
 
+/// Normalize a string containing a numeric value. This can
```
kwokcb, line 72: I went with `normalizeNumericString` vs `normalizeValueString` since this specifically can only handle numeric values in a string. Also the precision allows for the conversion  of numbers to double to not be unbounded.

source/MaterialXTest/MaterialXCore/CoreUtil.cpp
```diff
@@ -36,6 +37,28 @@ TEST_CASE("String utilities", "[coreutil]")
     REQUIRE(!mx::stringStartsWith("testName", "Name"));
     REQUIRE(mx::stringEndsWith("testName", "Name"));
     REQUIRE(!mx::stringEndsWith("testName", "test"));
+
+    std::string inputScalar1 = "  00.1000  ";
```
kwokcb, line 41: This covers all the variations I could come up with.

source/MaterialXTest/MaterialXCore/Document.cpp
```diff
@@ -116,3 +118,48 @@ TEST_CASE("Document", "[document]")
...
+    {
+        mx::InputPtr input = doc2->addInput("input" + std::to_string(index), (*it).first);
+        input->setValueString((*it).second);
+        index++;
+    }
+
+    doc->normalizeValueStrings();
+    doc2->normalizeValueStrings();
+
+    // Note: do not check doc2 == doc as that is a pointer comparison
```
kwokcb, line 162: Not sure if there is a way to allow pointer comparison to be a value comparison, but left a note here since I ran into it.

source/MaterialXCore/Util.cpp
```diff
@@ -179,4 +181,47 @@ string parentNamePath(const string& namePath)
...
+    std::string token;
+
+    while (std::getline(ss, token, ','))
+    {
+        // Remove leading and trailing spaces
+        token.erase(0, token.find_first_not_of(' '));
+        token.erase(token.find_last_not_of(' ') + 1);
+
+        // Convert the string to a double (automatically removes leading zeros)
+        double number = std::stod(token);
```
jstone-lucasfilm, line 197: One important note is that we don't want to convert values strings to precision-limited numeric values in memory, and `normalizeNumericString` should be purely a string processing operation.  Given that framework, I wouldn't expect this function to take a `precision` argument, and the internal logic can be simplified to be purely string processing.

ld-kerley: If we don't do this conversion to a numeric type, then the string processing should probably also removing any leadings zeros as well, as well as extraneous `+` signs.  Converting to the concrete numeric type probably actually makes this code a little simpler, but at the expense of the precision conversion @jstone-lucasfilm points out.... 

Even if this function takes a purely string manipulation approach to the normalization, I can still imagine times where it might be useful to be able to normalize the precision of the numeric values in the document. I've seen documents where some automatic process has written out a value of say `53.9999999999` instead of `54`, and with a `precision` argument to the function, the document could be made more human readable again, but again if we're going to take the purely string manipulation approach here that will likely make the code yet more complex.

source/MaterialXCore/Document.cpp
```diff
@@ -401,6 +402,32 @@ void Document::invalidateCache()
     _cache->valid = false;
 }
 
+void Document::normalizeValueStrings(unsigned int numericPrecision)
+{
+    // For now this will only normalize numeric strings
+    StringSet numericType = { "float", "vector2", "vector3", "vector4", "color3", "color4" };
```
ld-kerley, line 408: I'm not sure if it's legal or not, but do we also want to consider removing unnecessary leading zeros from integer values? 

And we may also want to include matrix types here too.

source/MaterialXCore/Util.cpp
```diff
@@ -179,4 +181,47 @@ string parentNamePath(const string& namePath)
...
+        formattedNumber.erase(formattedNumber.find_last_not_of('0') + 1, std::string::npos);
+        if (formattedNumber.back() == '.')
+        {
+            formattedNumber.pop_back(); // Remove the trailing dot if there's no decimal part left
+        }
+
+        // Append the formatted number to the result
+        result.str(EMPTY_STRING); // Clear the stringstream for the next number
+        result.clear();
+        result << formattedNumber << ",";
```
ld-kerley: Spot checking a number of our existing test suite files, all of them I looked at include a space character after the comma.  So perhaps this could be `", "` ?

source/MaterialXCore/Document.cpp
```diff
@@ -401,6 +402,32 @@ void Document::invalidateCache()
...
+    for (ElementPtr elem : traverseTree())
+    {
+        ValueElementPtr valueElem = elem->asA<ValueElement>();
+        if (!valueElem || numericType.end() == numericType.find(valueElem->getType()))
```
ld-kerley, line 412: I think this check would be better done inside of `normalizeNumericString()` itself and just pass along the type name of the value being normalized.  Otherwise other callers of this function will likely have to replicate this logic elsewhere.

source/MaterialXCore/Util.h
ld-kerley: I think I might still vote to name the function `normalizeValueString` because we may find need in the future to normalize other types of values - perhaps removing leading/trailing white space from filenames? or other future types we don't yet have.  

Especially if we relocate the type filtering logic inside this function, I think the more generic name makes more sense, and is a little more forward looking from an API design perspective.

source/MaterialXCore/Util.cpp
kwokcb: Thanks for pointing this out. I added precision only because I was using double conversion.
Logic should be simpler now.

source/MaterialXTest/MaterialXCore/CoreUtil.cpp
ld-kerley: Looks like a pretty complete list to me. 

Though when I read it - I find myself wondering if MaterialX supports parsing scientific notation - but thats perhaps out of scope of this particular PR to address.

source/MaterialXTest/MaterialXCore/CoreUtil.cpp
```diff
@@ -36,6 +37,28 @@ TEST_CASE("String utilities", "[coreutil]")
...
+    std::string resultScalar2 = mx::normalizeNumericString(inputScalar2, 3);
+    REQUIRE(resultScalar2 == "0.123");
+    std::string resultScalar3 = mx::normalizeNumericString(inputScalar2, 12);
+    REQUIRE(resultScalar3 == "0.123456789");
+    std::string inputScalar3 = "0.12345678901234567890";
+    std::string resultScalar4 = mx::normalizeNumericString(inputScalar3, 9);
+    REQUIRE(resultScalar4 == "0.123456789");
+
+    std::string inputVector1 = "1.0, 2.0,  0000.231";
+    std::string inputVector2 = "0001.2000, 0000.00010";
```
ld-kerley, line 55: Nit. Can we move this line down two more lines next to its corresponding `normalizeNumericString()` call and test.

source/MaterialXCore/Util.cpp
kwokcb: There are ways to control what precision to access a number with and they could be used to overwrite existing values. I think it's better to keep that separate from this utility which should not change the value stored.

source/MaterialXTest/MaterialXCore/CoreUtil.cpp
kwokcb: I've put in a check for exponent ('e') and skip trailing 0 removal.

source/MaterialXCore/Util.cpp
```diff
@@ -179,4 +181,89 @@ string parentNamePath(const string& namePath)
...
+                if (token.empty() || token[0] == '.')
+                {
+                    token = "0" + token;
+                }
+
+                // Check if string token has a 'e' character
+                // implying it is a floating point number in scientific notation.
+                // If it does not have an 'e' character, remove trailing zeros.
+                //
+                size_t ePos = token.find('e');
```
kwokcb, line 233: This is heuristic that 'e' means there is an exponent specified.
This will "do no harm" as all that happens is that trailing 0 removal is skipped to avoid changing something like 1.2e-10 to 1.2e-1

source/MaterialXCore/Document.cpp
kwokcb: To be safe, I've only removed leading and trailing spaces for integers.

jstone-lucasfilm:
@kwokcb As a thought experiment, it seems worthwhile to consider the following line from the OpenPBR definition:

```
<input name="base_weight" type="float" value="1.0" uimin="0.0" uimax="1.0" uiname="Base Weight" uifolder="Base" />
```

After the proposed normalization logic, would it become the following?

```
<input name="base_weight" type="float" value="1" uimin="0.0" uimax="1.0" uiname="Base Weight" uifolder="Base" />
```

If so, then I wonder if this approach will really achieve our goal of normalizing numeric value strings across a document, allowing for more straightforward comparisons.  In a sense, the transformation above would make the definition of OpenPBR less consistent rather than more so!

kwokcb:
> @kwokcb As a thought experiment, it seems worthwhile to consider the following line from the OpenPBR definition:
> 
> ```
> <input name="base_weight" type="float" value="1.0" uimin="0.0" uimax="1.0" uiname="Base Weight" uifolder="Base" />
> ```
> 
> After the proposed normalization logic, would it become the following?
> 
> ```
> <input name="base_weight" type="float" value="1" uimin="0.0" uimax="1.0" uiname="Base Weight" uifolder="Base" />
> ```
> 
> If so, then I wonder if this approach will really achieve our goal of normalizing numeric value strings across a document, allowing for more straightforward comparisons. In a sense, the transformation above would make the definition of OpenPBR less consistent rather than more so!

Hi @jstone-lucasfilm, 

If it's desired then a `.0` can be preserved. There is logic to remove the `.` if all trailing zeros is removed as the trailing zeros provide no additional precision. This can instead leave a `.0` instead. The same way a `0.` is preserved if all leading zeros are removed. 

I don't think this utility should be used to enforce format consistency as it's intent was for comparison consistency and "normalization" semantics were added . That said, the above can be done but is basically preserving a 1 decimal place precision format when none is required.

jstone-lucasfilm:
@kwokcb In this case, it's not the choice of formatting conventions that I'm thinking of, but rather that these conventions aren't being applied consistently across the document.  As a user, it would seem very surprising that normalization would be applied to the `value` attribute, but not to the `uimin` or `uimax` attribute, leaving the three of them in a seemingly inconsistent state.

I wonder if we may be pursuing the wrong approach here, and it may be worthwhile to go back to the initial motivation for this proposed feature.  If the formatting of numeric strings is changing in a roundtrip of MaterialX documents through USD, is there potentially some way to _preserve_ this formatting in the interchange steps, rather than attempting to apply a standardized string formatting after the fact?

kwokcb:
1. I don't see anyway to enforce a format for any translation process or it could just be hand-written.
2.  Yes I see your point. It's not affecting all numeric values (but ui meta-data isn't structured as actual `Value` data)

    I think I have a simpler and more consistent implementation. This goes back to actually converting to a typed data vs a generic string.

    - Instead of writing bespoke text formatting, use the existing string<->value conversion interfaces. 
      - All public API paths eventually reach `stringToData()`  when reading and `dataToString()` for writing. 
    - This handles all current types (including numeric vales) so would be more future-proof as well.
    - This would give you "consistent" results in terms of what is currently already being applied via these interfaces. 
    
    - `Value` elements can be handled by just calling `Value::setValue(Value::getValue())`
    - Non-`Value` elements such as ui meta-data can use `Value::createValueFromString()` (which is what get's called from `Value::getValue()`).  This is currently being used to for UI data for MaterialXViewer + MaterialXGraphEditor.

3. The alternative which could be much simpler is to just add a new function which compares typed values
instead of strings. This would never mutate data, it would get strings this way:
  ```python
  value = valueElemen.getValue()
  valueToCompare = value.getValueString()
  ```
   - Interestingly there is no `Value::operator==` method which could hide the string conversion logic.
   
@ld-kerley and @jstone-lucasfilm let me know your thoughts.

ld-kerley:
In general I'm less of a fan of having to mutate the document to perform a comparison on it.

I like the idea of having explicit some sort of comparison function for all the MaterialX elements, and that comparison function accepting arguments to control the criteria for comparison.  This could account for the difference between "these two documents are going to render the same" vs "these two documents are going to present the same UI", and potentially also allow us to provide "these two documents are within a specified floating point tolerance of each other" by providing an error threshold.

dbsmythe:
Yeah, I was never expecting that the **document** would ever get changed.  I thought this was strictly to be used as a way to check "are these two values the same" when values are expressed as strings without introducing potential floating point conversion errors.

jstone-lucasfilm:
Earlier in this discussion, I recall a suggestion to add an `Element::isEquivalent` method, and that may actually be the best path forward.  This would avoid the potential pitfalls of document normalization, which isn't a concept that we've had in MaterialX previously, and may not actually be a well-posed process.

If we took the `Element::isEquivalent` path, we'd be matching the nomenclature used in our epsilon-based matrix comparisons, which are distinct from our raw matrix comparisons that demand exact matches:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/18c06ea9077d26fc707d58a6637712b2508b9d4c/source/MaterialXCore/Types.h#L410

kwokcb:
This sounds good as we've basically come full circle back :).
BTW, the mutation only occurs if you take the result and reassign it back to the value. 

I'd just suggest closing this PR and suggest 
- a `Element::isEquivalent` which is generic +
- a `ValueElement::isEquivalent` specialization

jstone-lucasfilm:
That approach sounds good to me, @kwokcb, and thanks for working through this tricky set of issues with us!  :)

kwokcb:
A new PR is up ##2003. I will leave this for the time being if a comparison is needed for testing.
Otherwise this PR can be closed.

jstone-lucasfilm:
Thanks @kwokcb, and #2003 looks like a good approach to me, so let's close this original PR for now.

PR #1990, opened 2024-08-23
Reinstate logic to provide a path for OSL shader includes

ld-kerley:
A previous PR #1868 removed the ability to define the location for the OSL standard includes.  This is still required for Windows builds.


jstone-lucasfilm (approved):

PR #2003, opened 2024-09-04
Add Element Equivalence Interfaces

kwokcb:
## Issue

Addresses issues: #1988 and #1981 dealing with comparison of element values by adding in a new "functionally" equivalent interface, such that functional equivalence means that the 2 elements perform the same computation. This differs from the existing `operator==` interface which checks for syntactical equivalence.

The key points for functional equivalence include:
- The order of attributes specified on an `Element` does not matter.
- The order of children Elements does not matter as long as these are immediate children of a `Document` or a `Compound NodeGraph`.
- Equivalence can be  performed based on runtime values as opposed to the string value used for text storage. For floating point numeric values equivalence can be performed within a given precision tolerance and/or numeric format.
- Non-functional attributes may be explicitly skipped, such as UI formatting. UI numeric meta-data follows the same equivalence criteria as for input/output values (e.g. `uimax` or `uimin`). `Name` and `Category` equivalence cannot be skipped.

## Changes

- Adds a new `isEquivalent` method to `Element`. 
- Adds a new `isAttributeEquivalent` method for `Element` and override for `ValueElement` to allow for value vs string comparisons on the latter.
- Adds equivalence options class: `ElementEquivalanceOptions` which can be passed in as an argument to `isEquivalent`
- Adds an optional results / feedback class: `ElementEquivalenceResult` which can be passed in as an argument to `isEquivalent` and `isAttributeEquivalent`.

The following equivalence  options are available:
- Floating point format and precision for a float `ValueElement`. This is the same precision mechanism used for all value<->string conversions including values for code generation and file I/O.
- The ability to skip a user provided list of attributes. e.g. skipping `{ xpos, ypos }`.
- Perform raw string compares for values (as `operator==`) currently does instead of value comparisons for a `ValueElement`.

### Implementation Notes
- Note that value comparisons convert from string to a concrete runtime type which will remove any
extraneous string formatting such as leading and trailing spaces, `0` and leading `+` sign.
- The the `Value`'s string representation is used to perform the actual comparison allow for support of precision options.

### Tests

A new unit test has been added with various numeric variations to test values, ordering, and precision.
The document pair used for testing looks like this:
```xml
<materialx version="1.39">
  <input name="input0" type="boolean" value="false">
  <input name="input1" type="color3" value="  1.0,   +2.0,  3.0   ">
  <input name="input2" type="color4" value="1.0,   2.00, 0.3000, -4">
  <input name="input3" type="filename" value="filename1">
  <input name="input4" type="float" uimin="  0.0100 " uimax="  01.0100 " value="  1.2e-10  ">
  <input name="input5" type="float" uimin="  0.0100 " uimax="  01.0100 " value="  00.1000  ">
  <input name="input6" type="integer" value="  12 ">
  <input name="input7" type="matrix33" value="01.0,         2.0,  0000.2310,    01.0,         2.0,  0000.2310, 01.0,         2.0,  0000.2310       ">
  <input name="input8" type="matrix44" value="01.0,         2.0,  0000.2310, 0.100, 01.0,         2.0,  0000.2310, 0.100, 01.0,         2.0,  0000.2310, 0.100, 01.0,         2.0,  0000.2310, 0.100">
  <input name="input9" type="string" value="mystring">
  <input name="input10" type="vector2" value="1.0,   0.012345608">
  <input name="input11" type="vector3" value="  1.0,   +2.0,  3.0   ">
  <input name="input12" type="vector4" value="1.0,   2.00, 0.3000, -4">
</materialx>
```
and
```xml
<materialx version="1.39">
  <input name="input0" type="boolean" value="false">
  <input name="input1" type="color3" value="1, 2, 3">
  <input name="input2" type="color4" value="1, 2, 0.3, -4">
  <input name="input3" type="filename" value="filename1">
  <input name="input4" type="float" value="1.2e-10" uimin="  0.01" uimax="  1.01">
  <input name="input5" type="float" value="0.1" uimin="  0.01" uimax="  1.01">
  <input name="input6" type="integer" value="12">
  <input name="input7" type="matrix33" value="1, 2, 0.231,  1, 2, 0.231,  1, 2, 0.231,  1, 2, 0.231">
  <input name="input8" type="matrix44" value="1, 2, 0.231, 0.1, 1, 2, 0.231, 0.1, 1, 2, 0.231, 0.1, 1, 2, 0.231, 0.1">
  <input name="input9" type="string" value="mystring">
  <input name="input10" type="vector2" value="1, 0.012345611">
  <input name="input11" type="vector3" value="1, 2, 3">
</materialx>
```

This command script was used to test the Python API:

```python
import argparse
import sys, os

import MaterialX as mx

def main():
    parser = argparse.ArgumentParser(description="Test if two documents are functionally equivalent.")
    parser.add_argument(dest="inputFilename", help="Filename of the input document.")
    parser.add_argument(dest="inputFilename2", help="Filename of the input document to compare against.")
    parser.add_argument('-sa', '--skipAttributes', nargs='+', help="List of attributes to exclude from comparisons.")
    parser.add_argument('-sv', '--skipValueComparisons', action='store_true', help="Skip value comparisons. Default is False.")    
    parser.add_argument('-p', '--precision', type=int, default=None, help="Specify the precision for floating-point comparisons.", )

    opts = parser.parse_args()

    # Check if both files exist
    if not os.path.isfile(opts.inputFilename):
        print(f"File {(opts.inputFilename)} does not exist.")
        sys.exit(0)
    if not os.path.isfile(opts.inputFilename2):
        print(f"File {(opts.inputFilename2)} does not exist.")
        sys.exit(0)

    doc = mx.createDocument()
    try:
        mx.readFromXmlFile(doc, opts.inputFilename)
    except mx.ExceptionFileMissing as err:
        print(err)
        sys.exit(0)

    doc2 = mx.createDocument()
    try:
        mx.readFromXmlFile(doc2, opts.inputFilename2)
    except mx.ExceptionFileMissing as err:
        print(err)
        sys.exit(0)

    print(f'Version: {mx.getVersionString()}')    

    equivalence_opts = mx.ElementEquivalenceOptions()
    if opts.skipAttributes:
        for attr in opts.skipAttributes:
            equivalence_opts.skipAttributes.add(attr)
    if opts.skipValueComparisons:
        equivalence_opts.skipValueComparisons = True
    if opts.precision:
        equivalence_opts.precision = opts.precision

    equivalent, results = doc.isEquivalent(doc2, equivalence_opts)
    if equivalent:
        print(f"Documents are equivalent")
    else:
        print(results)
        print(f"Documents are not equivalent: {len(results)} differences found")
        for i in range(0, len(results)):
            difference = results[i]
            print(f"- Difference[{i}] : Path: '{difference.path1}' vs path: '{difference.path2}'. Difference Type: '{difference.differenceType}'"
                  + (f". Attribute: '{difference.attributeName}'" if difference.attributeName else "")) 
    
if __name__ == '__main__':
    main()
```


source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +81,69 @@ bool Element::operator!=(const Element& rhs) const
...
+    {
+        if (getAttribute(attr) != rhs->getAttribute(attr))
+        {
+            options.status = "Attribute " + attr + " differs for elements: '" + getNamePath() + "', '" + rhs->getNamePath() + "'";
+            return false;
+        }
+    }
+
+    // Compare children.
+    const vector<ElementPtr>& c1 = getChildren();
```
kwokcb, line 132: I would like to add an option to ignore child-order for `GraphElements` (nodegraphs / documents) only but want to get this up for initial review first.

source/MaterialXTest/MaterialXCore/Document.cpp
```diff
@@ -116,3 +119,146 @@ TEST_CASE("Document", "[document]")
     // Validate the combined document.
     REQUIRE(doc->validate());
 }
+
+TEST_CASE("Document equivalence", "[document]")
```
kwokcb, line 123: This is very similar to the tests in PR #1988 but with more option test cases.

source/MaterialXCore/Element.h
```diff
@@ -71,6 +71,52 @@ using ElementMap = std::unordered_map<string, ElementPtr>;
...
+    /// are specified on an element. Default is to ignore order.
+    bool ignoreAttributeOrder;
+
+    /// Do not perform any value comparisions. Instead perform exact string comparisons for attributes
+    /// Default is false. The operator==() method can be used instead as it always performs
+    /// a strict comparison. Default is false.
+    bool skipValueComparisons;
+
+    /// Status string for first difference found.
+    string status;
```
ld-kerley, line 117: I really like having `ElementEquivalenceOptions`, its going to allow us to expand on this functionality much easier in the future.

I'm not sure `status` really feels like it belongs in the equivalence options object. I can imagine creating a single version of this object and wanting to re-use it for many comparisons. 

Perhaps instead we could have a `ElementEquivalenceResult` object that is returned.  Which stores both the `true/false` success of the check, but optionally the status as well.  I can imagine cases where perhaps we might want the equivalence to run as fast as possible and not be constructing strings, if we don't care about why.  I can also imagine cases where we might not want the equivalence to exist with `false` at the first difference, but instead run over the entire document and report as many differences are is possible, then we might want a `vector<string>`. This `ElementEquivalenceResult` could have a bool operator defined, so it could still be treated like the `true/false` result that it is most likely to be used for. 

Alternately we could keep the boolean return value and just carry the status result as a separate argument.

source/MaterialXCore/Element.cpp
```diff
@@ -482,6 +545,107 @@ TypeDefPtr TypedElement::getTypeDef() const
 // ValueElement methods
 //
 
+bool ValueElement::isEquivalent(ConstElementPtr rhs, ElementEquivalenceOptions& options) const
+{
+    if (getCategory() != rhs->getCategory() ||
```
ld-kerley, line 550: Would be nice if we could find the right way to have less code duplication between these two functions.

source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +81,69 @@ bool Element::operator!=(const Element& rhs) const
     return !(*this == rhs);
 }
 
+bool Element::isEquivalent(ConstElementPtr rhs, ElementEquivalenceOptions& options) const
+{
+    if (getCategory() != rhs->getCategory() ||
```
ld-kerley, line 86: I'm curious why category and name are specialized here - won't they also be covered by the attribute comparison below?

kwokcb: No, they are stored in different class member data.

source/MaterialXCore/Element.cpp
kwokcb: I think a virtual `isAttributeEquivalent(elementPtr, attrbuteString)` can work here so there is one `Element::isEquivalent()`. The no-op child element check is a small overhead that is part of the ValueElement compare but it's not more than what already occurs with `operator==` and will keep it generic in case ValueElements ever have children.

source/MaterialXCore/Element.h
kwokcb: I think it's more consistent and convenient to return value a `boolean`. Then like the `validate(string *message)` method status can be returned if the argument is specified. 

Instead of a string it can be a actual struct to allow for future changes without a signature. `ElementEquivalenceResult` sounds good to me. It can contain a list of "differences". For now it will return the first difference.

source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +81,84 @@ bool Element::operator!=(const Element& rhs) const
...
+        attributeNames.erase(std::remove_if(attributeNames.begin(), attributeNames.end(),
+            [&skipAttributes](const string& attr) { return skipAttributes.find(attr) != skipAttributes.end(); }),
+            attributeNames.end());
+        rhsAttributeNames.erase(std::remove_if(rhsAttributeNames.begin(), rhsAttributeNames.end(),
+            [&skipAttributes](const string& attr) { return skipAttributes.find(attr) != skipAttributes.end(); }),
+            rhsAttributeNames.end());
+    }    
+
+    // Ignore ordering if option specified.
+    if (options.ignoreAttributeOrder)
```
jstone-lucasfilm, line 112: Since attribute order has no _functional_ effect on a document, I think it's reasonable to simplify this method and assume that attribute order is ignored when testing equivalence.

source/MaterialXCore/Element.h
ld-kerley: I like that, it would be nice to also have the ability to not require the explanation at all - as an optimization if its just a computer doing an equivalence check and a human doesn't need to know why.

source/MaterialXCore/Element.cpp
kwokcb: I mostly added this in as @ld-kerley mentioned it would be good to support both syntactical as well as functional logic. I'm fine with removing it.

source/MaterialXCore/Element.h
kwokcb: See if what I've stuck in looks good or not. Only if the optional argument is specified it will log info otherwise it does nothing.

source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +81,84 @@ bool Element::operator!=(const Element& rhs) const
...
+    {
+        // Sort the string arrays
+        std::sort(attributeNames.begin(), attributeNames.end());
+        std::sort(rhsAttributeNames.begin(), rhsAttributeNames.end());
+    }
+    if (attributeNames != rhsAttributeNames)
+    {
+        if (result)
+            result->addMessage("Attribute names differ for elements: '" + getNamePath() + "', '" + rhs->getNamePath() + "'");
```
jstone-lucasfilm, line 121: Just in the interest of keeping this feature as simple as possible, I'm curious about the situation in which these messages would be useful to a developer or an artist.  Would these messages be displayed in some fashion, allowing a developer or artist to respond to the differences that were found?

kwokcb: That is my thought at the moment -- in the similar vein of `validate()` also providing feedback on where validation fails this is where equivalence has failed, can be corrected and the test re-run again. This is what I found frustrating with `operator==` where I had to modify the code to find out what differed.

source/MaterialXCore/Element.cpp
ld-kerley: @jstone-lucasfilm - do we know that no application that renders a UI for MaterialX relies on the order of the attributes defined in the document?  

If we are taking the opinion that attribute order on an element is irrelevant, that should probably be explicitly stated in the specification somewhere, if it isn't already (I didn't check).

jstone-lucasfilm: I'd thought that the goal of `isEquivalent` was to test functional, rather than syntactical equivalence.  After all, we already have an equality test that is designed to test syntactical equivalence!

For functional/rendering equivalence, there's no difference produced when reordering child elements or reordering attributes, but of course both operations affect syntactical equivalence.

jstone-lucasfilm: To @ld-kerley's point, any syntactical change can affect the display of a MaterialX document in an application UI, even the introduction of a space in a value string or the swapping of the order of two elements in the document.  Neither of these can affect rendering, however, so it's important that we choose which type of equivalence we're testing here.

source/MaterialXTest/MaterialXCore/Document.cpp
```diff
@@ -116,3 +119,193 @@ TEST_CASE("Document", "[document]")
...
+    for (size_t i=0; i<differenceCount; ++i)
+    {
+        mx::StringVec message = result.getDifference(i);
+        std::cout << label << ": " << "Element: " << message[0] << 
+            ", Element: " << message[1] << ", Difference Type: " << message[2] 
+            << ", Value: " << message[3] << std::endl;
+    }
+}
+
+TEST_CASE("Document equivalence", "[document]")
```
jstone-lucasfilm, line 135: Looking at the current build errors, they seem to be platform-specific differences in the behavior of this new unit test!

source/MaterialXCore/Element.cpp
kwokcb: I'm perfectly fine with make this a functional test only and removing unneeded options. `operator==` already handles all syntactical differences. Thus it's clear that one interface is purely for functional and one is purely for syntactical.

For syntactical differences affecting display, can we spawn a new thread -- where I'd suggest looking at *explicit* display meta-data  to address ambiguity for things like implied syntactical order.  Where possible, the documentation should denote where ordering matters (e.g. node input / output signature ordering) and where it may affect display (attribute characteristics).

jstone-lucasfilm: That sounds like a good approach to me, @kwokcb.

For document display in an application UI, I would note that virtually all syntactic changes to a document will affect its display for _some_ application, even the most trivial change to a value string, so a differentiation between exact document equality and UI equality is not likely to be universal or valuable.

For functional/render equivalence, I would say that this is what we're defining in your proposed isEquivalent method, and documenting these same rules in a future version of the MaterialX specification makes good sense.

source/MaterialXCore/Element.h
```diff
@@ -71,6 +71,92 @@ using ElementMap = std::unordered_map<string, ElementPtr>;
 /// A standard function taking an ElementPtr and returning a boolean.
 using ElementPredicate = std::function<bool(ConstElementPtr)>;
 
+/// @class ElemenEquivalenceResult
+/// The results of comparing for equivalence.
+class MX_CORE_API ElementEquivalenceResult
+{
+  public:
+      ElementEquivalenceResult() = default;
```
jstone-lucasfilm, line 79: You may want to double-check the formatting of this new class, as I'm seeing some unusual indentation here.

source/MaterialXCore/Element.cpp
```diff
@@ -482,6 +591,70 @@ TypeDefPtr TypedElement::getTypeDef() const
 // ValueElement methods
 //
 
+bool ValueElement::isAttributeEquivalent(ConstElementPtr rhs, const string& attributeName, 
+                                         ElementEquivalenceOptions& options, ElementEquivalenceResult* result) const
+{    
+    bool perforDefaultCompare = true;
```
jstone-lucasfilm, line 597: What's the meaning of "perfor" that you're thinking of here?  Should we perhaps simplify this variable name to "defaultCompare"?

ld-kerley (commented):
Looking pretty good to me - just left a couple of very minor code style things.

source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +88,108 @@ bool Element::operator!=(const Element& rhs) const
...
+    {
+        if (!isAttributeEquivalent(rhs, attr, options, result))
+        {
+            return false;
+        }
+    }
+
+    // Compare children.
+    const vector<ElementPtr>& c1 = getChildren();
+    const vector<ElementPtr>& c2 = rhs->getChildren();
```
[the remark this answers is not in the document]
kwokcb: Sounds good.

source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +88,108 @@ bool Element::operator!=(const Element& rhs) const
...
+            result->addDifference(getNamePath(), rhs->getNamePath(), ElementEquivalenceResult::CHILD_COUNT);
+        return false;
+    }
+    for (size_t i = 0; i < c1.size(); i++)
+    {
+        ElementPtr c2Element = c2[i];
+        // Handle unordered children if parent is a compound graph (NodeGraph, Document).
+        // (Functional graphs have a "nodedef" reference and define node interfaces
+        // so require strict interface ordering.)
+        GraphElementPtr graph = this->getSelfNonConst()->asA<GraphElement>();
```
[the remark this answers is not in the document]
kwokcb: I need to do this to be able to downcast (asA<>). I'll check if anything can be always const.

source/MaterialXCore/Element.cpp
kwokcb: Bad keyboarding. Should be `performDefaultCompare`. Could be `defaultCompare` as well.

kwokcb: I've just nixed the variable was uneeded.

source/MaterialXCore/Element.cpp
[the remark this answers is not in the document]
kwokcb: I've made this all const now.

ld-kerley (approved):

source/MaterialXCore/Element.cpp
```diff
@@ -64,13 +71,13 @@ bool Element::operator==(const Element& rhs) const
     }
 
     // Compare children.
-    const vector<ElementPtr>& c1 = getChildren();
```
jstone-lucasfilm, line 67: @kwokcb Does this PR contain any functional change to our `Element::operator==` method, or is this just a renaming of its variables to follow a new convention?  If it's the latter case, then I'd recommend that we omit this renaming work from the current PR, so that it's clear that this change is focused on new functionality, and doesn't change the behavior of existing methods.  After that, it should be fine to propose new naming conventions in a separate, non-functional PR.

source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +88,108 @@ bool Element::operator!=(const Element& rhs) const
...
+    const vector<ElementPtr>& rhsChildren = rhs->getChildren();
+    if (children.size() != rhsChildren.size())
+    {
+        if (result)
+            result->addDifference(getNamePath(), rhs->getNamePath(), ElementEquivalenceResult::CHILD_COUNT);
+        return false;
+    }
+    for (size_t i = 0; i < children.size(); i++)
+    {
+        ElementPtr c2Element = rhsChildren[i];
```
jstone-lucasfilm, line 153: This naming convention now seems very confusing, as there's no longer any `c2` from which the `c2Element` is derived!

ld-kerley: I'd suggest we rename `c2Element` to `rhsElement`.

source/MaterialXCore/Element.cpp
kwokcb: Sounds good. It was a search+replace change. I'll revert.

kwokcb:
I added in Python bindings to avoid another PR. I have an equivalence script, but that's not necessary to add in unless you think it's worthwhile to ship with it

```python
import argparse
import sys, os

import MaterialX as mx

def main():
    parser = argparse.ArgumentParser(description="Test if two documents are functionally equivalent.")
    parser.add_argument(dest="inputFilename", help="Filename of the input document.")
    parser.add_argument(dest="inputFilename2", help="Filename of the input document to compare against.")
    parser.add_argument('-sa', '--skipAttributes', nargs='+', help="List of attributes to exclude from comparisons.")
    parser.add_argument('-sv', '--skipValueComparisons', action='store_true', help="Skip value comparisons. Default is False.")    
    parser.add_argument('-p', '--precision', type=int, default=None, help="Specify the precision for floating-point comparisons.", )

    opts = parser.parse_args()

    # Check if both files exist
    if not os.path.isfile(opts.inputFilename):
        print(f"File {(opts.inputFilename)} does not exist.")
        sys.exit(0)
    if not os.path.isfile(opts.inputFilename2):
        print(f"File {(opts.inputFilename2)} does not exist.")
        sys.exit(0)

    doc = mx.createDocument()
    try:
        mx.readFromXmlFile(doc, opts.inputFilename)
    except mx.ExceptionFileMissing as err:
        print(err)
        sys.exit(0)

    doc2 = mx.createDocument()
    try:
        mx.readFromXmlFile(doc2, opts.inputFilename2)
    except mx.ExceptionFileMissing as err:
        print(err)
        sys.exit(0)

    equivalence_opts = mx.ElementEquivalenceOptions()
    if opts.skipAttributes:
        for attr in opts.skipAttributes:
            equivalence_opts.skipAttributes.add(attr)
    if opts.skipValueComparisons:
        equivalence_opts.skipValueComparisons = True
    if opts.precision:
        equivalence_opts.precision = opts.precision

    results = mx.ElementEquivalenceResult()
    equivalent = doc.isEquivalent(doc2, equivalence_opts)
    equivalent = doc.isEquivalent(doc2, equivalence_opts, results)
    if equivalent:
        print(f"Documents are equivalent")
    else:
        print(f"Documents are not equivalent: {results.differenceCount()} differences found")
        for i in range(0, results.differenceCount()):
            difference = results.getDifference(i)
            print(f"- Difference[{i}] : Path: {difference[0]} vs path: {difference[1]}. Difference Type: {difference[2]}"
                  + (f" attribute: {difference[3]}" if difference[3] else "")) 
    
if __name__ == '__main__':
    main()
```

source/MaterialXCore/Element.h
```diff
@@ -71,6 +71,92 @@ using ElementMap = std::unordered_map<string, ElementPtr>;
 /// A standard function taking an ElementPtr and returning a boolean.
 using ElementPredicate = std::function<bool(ConstElementPtr)>;
 
+/// @class ElemenEquivalenceResult
+/// The results of comparing for equivalence.
+class MX_CORE_API ElementEquivalenceResult
```
jstone-lucasfilm, line 76: I'd recommend declaring `ElementEquivalenceResult` and `ElementEquivalenceOptions` _after_ the core `Element` class, since the understanding of these classes depends on understanding Element.  This is more of a consideration for people reading through the code, and hopefully should not impact the compilation of these classes!

source/MaterialXTest/MaterialXCore/Document.cpp
```diff
@@ -10,6 +10,9 @@
 #include <MaterialXFormat/Util.h>
 #include <MaterialXFormat/XmlIo.h>
 
+#include <unordered_map>
```
jstone-lucasfilm, line 13: Isn't `unordered_map` already included by `Library.h`?

source/MaterialXCore/Element.h
```diff
@@ -71,6 +71,92 @@ using ElementMap = std::unordered_map<string, ElementPtr>;
...
+    static const string ATTRIBUTE;
+    static const string ATTRIBUTE_NAMES;
+    static const string CHILD_COUNT;
+    static const string CHILD_NAME;
+    static const string NAME;
+    static const string CATEGORY;
+
+  private:
+    /// A list of  differences
+    vector<StringVec> differences;
```
jstone-lucasfilm, line 120: This seems like an unusual choice, making the `differences` variable a vector of _vectors_.  Would it be more logical to define the inner vector as a custom class containing exactly four member variables?

source/MaterialXCore/Element.h
```diff
@@ -71,6 +71,92 @@ using ElementMap = std::unordered_map<string, ElementPtr>;
 /// A standard function taking an ElementPtr and returning a boolean.
 using ElementPredicate = std::function<bool(ConstElementPtr)>;
 
+/// @class ElemenEquivalenceResult
```
jstone-lucasfilm, line 74: Minor: spelling of "Element"

source/MaterialXCore/Element.h
```diff
@@ -71,6 +71,92 @@ using ElementMap = std::unordered_map<string, ElementPtr>;
...
+    static const string CHILD_NAME;
+    static const string NAME;
+    static const string CATEGORY;
+
+  private:
+    /// A list of  differences
+    vector<StringVec> differences;
+};
+
+/// @class ElemenEquivalenceOptions
```
jstone-lucasfilm, line 123: Minor: spelling of "Element"

source/MaterialXTest/MaterialXCore/Document.cpp
```diff
@@ -116,3 +119,215 @@ TEST_CASE("Document", "[document]")
     // Validate the combined document.
     REQUIRE(doc->validate());
 }
+
+void printDifferences(mx::ElementEquivalenceResult& result, const std::string& label)
```
jstone-lucasfilm, line 123: Rather than baking in the use of `std::cout` in this helper method, it might be better to follow the example of `prettyPrint` and return the results as a standard string, allowing the user to send them to any logging system that they prefer:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Element.cpp#L666

source/MaterialXCore/Element.h
jstone-lucasfilm, line 76: Ideally we should declare these classes _after_ the core `Element` class, since a new developer reading through our code needs to understand the `Element` class before they can understand these.

Since `ElementEquivalenceResult` objects are always returned via pointer, it should be possible to forward declare this class in the header, and it may be worthwhile to find a way to forward declare `ElementEquivalenceOptions` as well.

source/MaterialXCore/Element.h
kwokcb: I've change this to a class for a single result and added a vector class wrapper.

source/MaterialXCore/Element.cpp
```diff
@@ -81,6 +87,108 @@ bool Element::operator!=(const Element& rhs) const
     return !(*this == rhs);
 }
 
+bool Element::isEquivalent(ConstElementPtr rhs, const ElementEquivalenceOptions& options,
+                           ElementEquivalenceResults* result) const
```
jstone-lucasfilm, line 91: I really like this latest change, where the passed object is simply a vector of `ElementEquivalenceResult` objects.

For the name of this vector, I might suggest `ElementEquivalenceResultVec`, which seems more clear than just pluralizing the name of the underlying class.

Although we don't often name vector types in MaterialX, I believe we've used the `Vec` suffix in cases where we do:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Library.h#L55

kwokcb: Yes, this is better / more consistent, I'll make the change.

source/PyMaterialX/PyMaterialXCore/PyElement.cpp
```diff
@@ -26,9 +26,36 @@ namespace mx = MaterialX;
...
     py::class_<mx::Element, mx::ElementPtr>(mod, "Element")
         .def(py::self == py::self)
         .def(py::self != py::self)
+        .def("isEquivalent", [](const mx::Element& elem, mx::ConstElementPtr& rhs, const mx::ElementEquivalenceOptions& options)
+        {
+            mx::ElementEquivalenceResultVec results;
+            bool res = elem.isEquivalent(rhs, options, &results);
+            return std::pair<bool, mx::ElementEquivalenceResultVec>(res, results);
+        })        
+        .def("", &mx::Element::isEquivalent,
```
jstone-lucasfilm, line 57: What is the motivation for providing _two_ interfaces for this method, rather than using the unified approach seen below in `Element::validate`?  Would it make sense to adopt the `validate` approach here as well?

kwokcb: I think I had a bad merge. This second declaration should be deleted.

jstone-lucasfilm (approved):

PR #2016, opened 2024-09-15
Fix the call to anisotropic_vdf closure in OSL

niklasharrysson:
Update the implementation of `anisotropic_vdf` node to call the corresponding closure in OSL.

### DIFF
--- a/libraries/pbrlib/genosl/mx_anisotropic_vdf.osl
+++ b/libraries/pbrlib/genosl/mx_anisotropic_vdf.osl
@@ -1,6 +1,8 @@
-void mx_anisotropic_vdf(vector absorption, vector scattering, float anisotropy, output VDF vdf)
+void mx_anisotropic_vdf(color absorption, color scattering, float anisotropy, output VDF vdf)
 {
-    // TODO: Need to remap parameters to match the new closure, 
-    // or change the MaterialX spec to OSL parameterization.
-    vdf = 0;
+    // Convert from absorption and scattering coefficients to
+    // extinction coefficient and single-scattering albedo.
+    color extinction = absorption + scattering;
+    color albedo = scattering / extinction;
+    vdf = anisotropic_vdf(albedo, extinction, anisotropy);
 }


jstone-lucasfilm (approved):

jstone-lucasfilm:
@niklasharrysson I just ran new GLSL/OSL render tests with this code, and there's a new difference that seem worthwhile to highlight, just to make sure it's expected:

Here's the GLSL/OSL comparison for LamaDielectric before this change:

LamaDielectric_08_24_2024

And here's the same GLSL/OSL comparison after this change:

LamaDielectric_09_16_2024

LamaDielectric is one of the nodes that makes use of a BSDF/VDF layering operation, so it's worthwhile to consider whether this change would be expected with the latest code.

niklasharrysson:
@jstone-lucasfilm That change in the OSL result was not expected. Since volumetric rendering is not supported yet in OSL `testrender` I would expect this change on MaterialX side to have no effect at all on our test suite results.

And even if it was supported this result looks wrong.. It looks like the refraction is changed. Perhaps there is a bug in testrender where somehow the IOR of the top closure is modified when there is a vdf closure layered as base?

jstone-lucasfilm:
I'm CC'ing @fpsunflower, in case he has thoughts on why this visual change might occur on the OSL side.

fpsunflower:
That closure is ignored by testrender, so there must be something else going on here.

fpsunflower:
It could also be argued that the way `testrender` parses the closure tree here could be improved. Maybe `anisotropic_vdf` should leave the `refraction_ior` field alone? This way the microfacet closure would "win" when both are present.

niklasharrysson:
Yes, since `anisotropic_vdf` doesn't have a separate parameter for the ior of the medium, I think it should be assumed that the medium ior is the same as the ior set on the transmission BSDF (the BSDF set as top layer above the VDF).

jstone-lucasfilm:
@fpsunflower @niklasharrysson That proposed change on the OSL side sounds like a good resolution to me.

For now, should we proceed with this update to OSL shader generation in MaterialX, with the expectation that `testrender` will give us unexpected results in the short term?

fpsunflower:
Yes, I will submit a PR to fix the OSL side.

jstone-lucasfilm:
Thanks @fpsunflower, and let's go ahead and merge this change from @niklasharrysson.

fpsunflower:
OSL fix: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1870

PR #2019, opened 2024-09-17
Add `parseStructValueString` function

ld-kerley:
As requested by @jstone-lucasfilm, breaking apart #1831.  

This is the first part.

Adding a simple function that tokenize a c++ style list initializer string, which will be used to initialize struct values.

### DIFF
--- a/source/MaterialXCore/Util.cpp
+++ b/source/MaterialXCore/Util.cpp
@@ -101,60 +101,6 @@ string joinStrings(const StringVec& stringVec, const string& sep)
     return res;
 }
 
-StringVec splitListInitializer(const string& str)
-{
-    const char SEPARATOR = ';';
-    const char OPEN_BRACE = '{';
-    const char CLOSE_BRACE = '}';
-
-    if (str.empty())
-        return StringVec();
-
-    // validate the string is correctly formatted - must be at least 2 characters long and start and end with braces
-    if (str.size() < 2 || (str[0] != OPEN_BRACE || str[str.size()-1] != CLOSE_BRACE))
-    {
-        return StringVec();
-    }
-
-    StringVec split;
-
-    // strip off the surrounding braces
-    string substring = str.substr(1, str.size()-2);
-
-    // sequentially examine each character to parse the list initializer.
-    string part = "";
-    int braceDepth = 0;
-    for (const char c : substring)
-    {
-        if (c == OPEN_BRACE)
-        {
-            // we've already trimmed the starting brace, so any additional braces indicate members that are themselves list initializers.
-            // we will just return this as a string of the list initializer.
-            braceDepth += 1;
-        }
-        if (braceDepth > 0 && c == CLOSE_BRACE)
-        {
-            braceDepth -= 1;
-        }
-
-        if (braceDepth == 0 && c == SEPARATOR)
-        {
-            // when we hit a separator we store the currently accumlated part, and clear to start collecting the next.
-            split.emplace_back(part);
-            part = "";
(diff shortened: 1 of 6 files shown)


source/MaterialXCore/Util.cpp
```diff
@@ -101,6 +101,60 @@ string joinStrings(const StringVec& stringVec, const string& sep)
     return res;
 }
 
+StringVec splitListInitializer(const string& str)
```
jstone-lucasfilm, line 104: From the text description of this changelist, I see that it's focused on the custom struct syntax for value strings in MaterialX, but this isn't yet reflected in the name of the new function or its documentation.

Should we consider giving this function a name along the lines of `parseStructValueString`?  And would it make more sense as a helper method in `Value.h` and `Value.cpp`, rather than as a core string manipulation method in `Util.h`?

ld-kerley: I put the function in the string manipulation code because it's only operating on the strings, it felt like a very close sibling to `splitString`.  In fact at one point I even considered making it a "mode" of `splitString` but the code felt clearer once it was separated.

I'm open to moving it to `value.cpp`, but I would worry that in the future developers would think it only relates to struct types, where as the actual function has no ties to structs at all, it just happens to be that we have defined in the specification that the structs are initialized with strings that look like C++ initializer lists.  It feels more discoverable to me if it lives with the other string manipulation code.

jstone-lucasfilm: I could be wrong, but the syntax rules in this helper function seem very closely tied to our custom struct syntax, especially in the use of the semicolon to separate struct members.  I wouldn't expect a developer find this function useful in parsing true C++ initializer lists, for example, and I can't think of other uses in the MaterialX context beyond parsing custom structs.

Maybe I'm missing other examples, though, where this same function would be useful to a MaterialX developer?

source/MaterialXCore/Value.cpp
```diff
@@ -254,6 +254,60 @@ ScopedFloatFormatting::~ScopedFloatFormatting()
     Value::setFloatPrecision(_precision);
 }
 
+StringVec parseStructValueString(const string& str)
```
jstone-lucasfilm, line 257: We should move this function definition into the `Global functions` block above, in order to make it more straightforward to find.

source/MaterialXCore/Value.cpp
```diff
@@ -254,6 +254,60 @@ ScopedFloatFormatting::~ScopedFloatFormatting()
...
+{
+    const char SEPARATOR = ';';
```
jstone-lucasfilm, line 259: Should we leave these constant character definitions inside the function, or move them up to the anonymous namespace above, where they would become part of the static initialization of `Value.cpp`, and could be assigned more unique names?  I don't have a strong opinion on this, and wanted to mention it as an option for consideration.

jstone-lucasfilm (commented):
This looks really promising, thanks @ld-kerley, and I had just a couple of minor notes.

source/MaterialXCore/Value.cpp
ld-kerley: done

source/MaterialXCore/Value.cpp
ld-kerley: I made declared them `static` so they'll get initialized the first time the function gets called - but it feels easier to read if we keep them defined close to where they are used.

ld-kerley:
Looks good to me - thanks for the tidy up @jstone-lucasfilm

Issue #1609, opened 2023-11-29
Add a combined version define in MaterialX C++


pablode:
Currently, checking against the MaterialX version in C++ code can be code-heavy:
```c++
#if (MATERIALX_MAJOR_VERSION > 1) || \
    (MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION > 38) || \
    (MATERIALX_MAJOR_VERSION == 1 && MATERIALX_MINOR_VERSION == 38 && MATERIALX_BUILD_VERSION > 8)
    // ...some new feature
#endif
```

I propose that a define such as `MATERIALX_VERSION` should be added, just like it's done in USD:
https://github.com/PixarAnimationStudios/OpenUSD/blob/0b18ad3f840c24eb25e16b795a5b0821cf05126e/cmake/defaults/Version.cmake#L29

The above check can then be simplified to
```c++
#if MATERIALX_VERSION > 103808
  // ...
#endif
```

jstone-lucasfilm:
This is a good suggestion, @pablode, though I wonder if we can do better than the combined integer approach for preprocessor comparisons.

For context, here is the pattern we currently use to compare MaterialX versions in Python:

```
if (mx.getVersionIntegers() > (1, 38, 8))
```

And here is the equivalent comparison in C++:

```
if (mx::getVersionIntegers() > std::tuple<int, int, int>(1, 38, 8))
```

Could there potentially be an equivalent approach for preprocessor comparisons, or are combined integers truly our best bet?

pablode:
Indeed, the combined integer could be constructed with a macro; see f.i. Vulkan’s [VK_MAKE_VERSION](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_MAKE_VERSION.html).

A MaterialX equivalent would then look like this:
```c++
#if MATERIALX_VERSION > MATERIALX_MAKE_VERSION(1, 38, 8)
```

ruiyang2021:
Hi can I try solving this one for Dev Days 2024?

jstone-lucasfilm:
Definitely, and welcome @ruiyang2021!

ruiyang2021:
Hi folks, I take a closer look today. I'm still a beginner learning from the code repo, may have some shallow understanding. 

I see that tuples are compared position by position, it is great we can use ">" and "<" operators to compare the versions.

Then here's a doubt:
[...]

PR #2020, opened 2024-09-17
Add support for Apple framework builds

dgovil:
This PR makes a few changes that have been derived from a corresponding PR to USD https://github.com/PixarAnimationStudios/OpenUSD/pull/2969

The changes are:
1. Deprecates `MATERIALX_BUILD_IOS` in favor of using `CMAKE_SYSTEM_NAME`. This allows for better support for iPhone derived targets like visionOS. Earlier this would force it to iOS which can cause subtle issues when compiled against other SDKs.  See  https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-visionos-or-watchos

2. Changes `TARGET_OS_IOS` define to `TARGET_OS_IPHONE` which is more correct to support iPhone derivatives. See     https://chaosinmotion.com/2021/08/02/things-to-remember-compiler-conditionals-for-macos-ios-etc/ . But the gist is `TARGET_OS_IOS` refers specifically to the modern iOS/iPadOS SDK, whereas `TARGET_OS_IPHONE` refers to all Apple SDKs that derived from the original iPhone. 

3. Add support for building as a Framework. This creates a special directory structure that streamlines embedding of MaterialX within Apps on Apple platforms to just dragging the project into the app in Xcode. No other linker and compiler configuration is necessary. See Framework notes below.

4. Adds a new  `getSharedLibraryPath` function on Apple platforms that finds the path that contains the dylib, instead of just the executable that is ultimately linking to the dylib. I didn't implement this for Windows/Linux since there's only one the one need for it right now. This is derived from USD's method to do the same.

5. Additionally, I've added a `MATERIALX_BUILD_IDENTIFIER` that mirrors how USD generates a single integer build number for its namespace. I only use this within the Plist for the framework since it requires the identifier version to be an incrementing integer.

## Framework notes

I have opted to use a zsh to keep this consistent with the PR for USD, rather than use CMake's built in Framework generation. This is also because, either way, I need to do some processing to the headers (see below) , fixing of linkage etc so it's simpler to just do it all in one spot.

Zsh is always guaranteed to exist on any macOS in the last several years.

The zsh script also modifies the Headers because Clang will automatically add frameworks to the include search path, as long as the include starts with the frameworks name. This means that `#include <MaterialXCore/Document.h>` now needs to be `#include <MaterialX/MaterialXCore/Document.h>`. However if your code requires the unprefixed version, you can still manually add the Header directory in the framework and it'll work as it normally would.

One other advantage to using the Zsh script as a post process is that you still get a standard build of the libraries too in case you want to use them otherwise, and you also can fix issues in the Framework without doing a full rebuild.


jstone-lucasfilm:
@dgovil Should this changelist include corresponding updates to our GitHub CI, so that the new, recommended pattern for iOS builds is validated for current and future commits?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/.github/workflows/main.yml#L94

dgovil:
Ah good catch, I'd missed that hidden folder.
Changed it over.

CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(MATERIALX_MAJOR_VERSION 1)
 set(MATERIALX_MINOR_VERSION 39)
 set(MATERIALX_BUILD_VERSION 2)
 set(MATERIALX_LIBRARY_VERSION ${MATERIALX_MAJOR_VERSION}.${MATERIALX_MINOR_VERSION}.${MATERIALX_BUILD_VERSION})
+math(EXPR MATERIALX_BUILD_IDENTIFIER "${MATERIALX_MAJOR_VERSION} * 10000 + ${MATERIALX_MINOR_VERSION} * 100 + ${MATERIALX_BUILD_VERSION}")
```
jstone-lucasfilm, line 6: Since there are many different strategies for combining three version numbers into a monotonically-increasing "version index", we may want to keep this new logic into the corresponding `AppleFrameworkBuild.zsh.in` file for now, so that we're not officially stating that this is the approach that MaterialX will apply universally in the future.

Here's some additional discussion of different options for this combining logic, including the approaches used in Vulkan and USD, with the most recent posts leaning towards the Vulkan approach:

https://github.com/AcademySoftwareFoundation/MaterialX/issues/1609
`

dgovil: I need it in the CMake, since the CMake is what generates the final `plist` which is where its used.
I'll move it into the block that generates the plist though so its not a global like it is now.

source/MaterialXFormat/File.h
```diff
@@ -207,6 +207,11 @@ class MX_FORMAT_API FilePath
     /// Return the directory containing the executable module.
     static FilePath getModulePath();
 
+    #if defined(__APPLE__)
+    /// Return the directory containing the shared library.
+    static FilePath getSharedLibraryPath();
```
jstone-lucasfilm, line 212: This ends up being an unusual addition to the `FilePath` class, where a new public method is only defined on a _single platform_, and I wonder if we can improve upon this.

Could the logic within `getSharedLibraryPath` instead be moved directly into `getDefaultDataSearchPath`, as a block of code that is included within a check of `BUILD_APPLE_FRAMEWORK`?

Further in the future, if this logic needs to be shared between `getDefaultDataSearchPath` and other methods, we can design a good approach for making it available more broadly, but I feel that the current approach is not quite what we want here.

dgovil: That's fair. I inlined the method as requested.

jstone-lucasfilm (approved):
This looks good to me, thanks @dgovil!  I'm CC'ing @ld-kerley in case he has additional thoughts and recommendations.

dgovil:
I'd leave the CMAKE_OSX_ARCHITECTURE as whatever your host system is (so x86_64 if you're on an Intel Mac, arm64 if you're on Apple silicon ) and yes, setting CMAKE_OSX_SYSROOT to the name of the sdk to iPhoneSimulator should work, though I prefer setting the full path to the SDK myself. 

As for why there are no prebuilt binaries, I think it's just a matter of the number of builds to provide. In theory there could be iOS/visionOS/watchOS/tvOS plus simulators for each arch. It just adds up quickly, though maybe an xcframework could solve that.

PR #2022, opened 2024-09-18
Fix version upgrade for ND_normalmap_vector2

rafalSFX:
The upgrade code for the <normalmap> shader was indiscriminately renaming all node definitions to "ND_normalmap_float". This included "ND_normalmap_vecto2" nodes, whose nodedef should not be renamed.

This commit fixes that, limiting the renaming only to "ND_normalmap", based on the type of the "scale" input, which is the only differentiating factor between the two definitions.


jstone-lucasfilm (approved):
Thanks for this fix, @rafalSFX!

Issue #1599, opened 2023-11-15
Shader Translation: Support transmission in Standard Surface to glTF PBR


kwokcb:
# Issue

Unlike standard surface and openpbr, base color on gltf pbr affects transmission. This seems
to be the way it's supposed to work but is inconsistent with the other shading models.

For example if you take the [glass example](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/Examples/StandardSurface/standard_surface_glass.mtlx) which has a `base` value if 0, this renders as a "clear" material using standard surface and openpbr, but renders opaque black with gltf pbr.

# Test

This is the glTF after translation, but can just create a material and set baseColorFactor to 0 and transmissionFactor to non-0.

```JSON
"materials": [
    {
      "extensions": {
        "KHR_materials_transmission": {
          "transmissionFactor": 1
        },
        "KHR_materials_ior": {
          "ior": 1.5
        }
      },
      "pbrMetallicRoughness": {
        "baseColorFactor": [
          0,
          0,
          0,
          1
        ],
        "metallicFactor": 0,
        "roughnessFactor": 0
      },
      "emissiveFactor": [
        0,
        0,
        0
      ],
      "doubleSided": true,
      "name": "SR_glass"
    }
  ],
```

Result in model-viewer:
![image](https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/3c4262a6-dc0e-47d4-9489-15125033b0b8)

Result in MaterialX Graph Editor:
![image](https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/7963dbcd-3092-4166-b644-7325ca2101b2)

# Workaround
Modify the base color to be white.

CC'ing @proog128 , @emackey as a possible item to check for future [OpenPBR](https://github.com/AcademySoftwareFoundation/OpenPBR/blob/main/reference/open_pbr_surface.mtlx)  compatibility.

# Repro Steps:
1. Load the `glass` material into `MaterialXVew` from the installed `resources/Materials/Examples/StandardSurface` folder
2. Open up the advanced settings and set the target shader model to be `glt_pbr`:
![image](https://github.com/user-attachments/assets/17de9861-3d28-49b3-8588-ad7dcf8b9bd1)
3. Close the dialog and press 't' (to translate) which will save a new file to disk in the same folder.
4. Load in the new file. See if it's black :).

emackey:
[...]

PR #2027, opened 2024-09-26
Support transmission in translation from Standard Surface to glTF PBR

charo-m:
This corrects the translation of transmission parameters from Standard Surface to glTF PBR.
Issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/1599

Here's the change in a more readable pseudocode:
```
if (transmission > 0) {
    if (transmission_depth > 0) {
        attenuation_color = transmission_color
        base_color = white
    } else {
       attenuation color = white
       base_color = transmission_color
    }
} else {
    attenuation_color = white (default)
    base_color = no_transmission_base_color (hasCoatColor ? mixedBaseColor : scaledBaseColor)
}

```
Before ->
BeforeGraph

After ->
AfterGraph


jstone-lucasfilm:
@charo-m This looks really promising, and to resolve the EasyCLA issue, I'd recommend following up through the support link that the Linux Foundation provides:

charo-m:
thanks @jstone-lucasfilm ! I had a mismatching email address, just managed to get my CLA signed.

jstone-lucasfilm:
@charo-m Excellent!  Sometimes the LFX needs to bump the EasyCLA check so that it updates to your new email address, so feel free to check with them if it doesn't automatically turn from red to green.

kwokcb:
As a suggestion, it might might be better test to have the graph with the translation node in it to be able to do visual compares as part of the render test suite. Perhaps it be added to the test suite in a new folder `resources/Materials/TestSuite/pbrlib/translation`. ( @jstone-lucasfilm, I don't see any other translation examples anywhere else ?  Others could be added in as desired).

libraries/bxdf/translation/standard_surface_to_gltf_pbr.mtlx
```diff
@@ -76,12 +76,18 @@
     </dotproduct>
 
     <!-- Metallic roughness -->
-    <ifequal name="base_color" type="color3">
+    <ifequal name="no_transmission_base_color" type="color3">
```
jstone-lucasfilm, line 79: For consistency with other example graphs, I would recommend using camel case names for nodes representing intermediate variables, with snake case used only for the final nodes that match the naming conventions in glTF PBR.

So this variable would be named `noTransmissionBaseColor`, and similarly the new intermediate variables below would be named `transmissionInEffectAttenuationColor` and so forth.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -141,6 +141,43 @@ TEST_CASE("GenShader: Shader Translation", "[translate]")
     }
 }
 
+TEST_CASE("GenShader: Gltf Transmission Translation", "[translate]")
```
jstone-lucasfilm, line 144: For now, I would recommend omitting this new unit test, as what we really want is _render validation_ for a wider set of shader translation use cases, including the new case for which you've added support.

At the moment, we have just one shader translation test in our GitHub Actions suite, as we're currently using CPU renders and the overhead of each new test is quite high, but this is the location where we'll want to add new tests when GPU renders are supported:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/.github/workflows/main.yml#L265

jstone-lucasfilm (commented):
This looks really good, thanks @charo-m, and I had just two recommendations for improvements.

jstone-lucasfilm:
Thanks for these updates, @charo-m!

For now, I think it's fine to omit the new `standard_surface_to_gltf_pbr_glass.mtlx` unit test, which is an interesting suggestion for a future improvement to our translation system, but not an update that we need as part of your Dev Days project.

jstone-lucasfilm (approved):
Great work on this improvement, thanks @charo-m!

Issue #1984, opened 2024-08-21
Graph Editor: Mouse UI action orbits camera in render view


HardCoreCodin:
1. When the scrollbar appears in the property editor, grabbing it with the mouse to scroll and dragging the scrollbar up next to the render view, treats it as a mouse drag in the render view and orbits the camera.
2. Using the File menu and clicking a menu option and dragging the mouse also does the same and orbits the camera as it is over the render view.

kwokcb:
There are bunch of UI items [here](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1398). I think your case  is similar to another item mentioned.

LeaRe:
Hi! I would like to tackle this issue for the ASWF Dev Days 2024.

LeaRe:
While working on this issue I also noticed another case of unintentional camera orbiting. The previous logic to detect the cursor in the render view did not consider that the render view moves up when scrolling. As a result, the camera would
still orbit when clicking parts of the Node Property Editor that were previously occupied by the render view. I added a fix for this behaviour as well in my changes for this ticket (PR will be out soon).

jstone-lucasfilm:
Thanks to @HardCoreCodin for this original report, and to @LeaRe for the fix in #2032!

PR #2032, opened 2024-09-27
Fix Graph Editor unintentional camera orbit in render view

LeaRe:
This fixes the logic that detects whether the cursor is in the render view. Previously, there were three known issues that are fixed with this change:
- The camera would orbit when grabbing the scrollbar located next to the render view.
- The camera would orbit when clicking a menu option in the menu dropdown overlapping the render view.
- The previous logic to detect the cursor in the render view did not consider that the render view moves up when scrolling. As a result, the camera would still orbit when clicking parts of the Node Property Editor that were previously occupied by the render view.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3240,6 +3240,15 @@ void Graph::graphButtons()
     ImGui::BeginChild("Selection", ImVec2(paneWidth, 0), false, windowFlags);
     ImVec2 windowPos = ImGui::GetWindowPos();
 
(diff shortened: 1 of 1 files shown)


LeaRe:
This change fixes issue #1984 and another similar issue I noticed during this work (see comment in issue).

jstone-lucasfilm:
This looks really promising, @LeaRe, and I'm CC'ing @lfl-eholthouser for her expertise with this code.

lfl-eholthouser:
This looks great, @LeaRe! Thanks

jstone-lucasfilm (approved):
Great work on this project, thanks @LeaRe!

PR #2034, opened 2024-09-27
Added attribute strings to the Element class in the Python API

### LINKED ISSUE #1545: Attribute strings for various Element classes missing from Python API
## Issue

- There are a number of classes which do not expose the static attribute enumeration strings.
- This includes the Element class and various derived classes.

## Effect

- Without these exposed, users need to find out what these are called and hard-code them. 
- If the attribute names change in the future then their code may break due to hard-coded string usage.
- This could help with performance as well by encouraging string reuse.

## Proposal

- Scan through and add these to the Python API.

rherrera-ilm:
This change fixes #1545


jstone-lucasfilm:
@rherrera-ilm This looks really promising, and to resolve the EasyCLA issue, I'd recommend following up through the support link that the Linux Foundation provides:

jstone-lucasfilm:
Thanks @rherrera-ilm, and that's a great first step!

It looks like you have two commits from two _different_ email addresses, and I'd recommend posting an LFX support ticket for assistance to get that resolved:

They're usually very quick to respond to this type of issue, and can resolve it much more expertly than I can.

Issue #1209, opened 2023-01-31
Color inputs in MaterialXView and Editor handle display space


kwokcb:
Both of these editors do not handle conversion to / from display space properly currently.

See the thread found here:
https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1674221447235719

Seems the fix is to change to use built in conversions in MaterialX to go to / from display space to the space specified for the value.

sgmq0:
Would like to work on this issue for ASWF Dev Days 2024!

jstone-lucasfilm:
I'll clear out the assignment on this task, so that it's available for our next Dev Days in May 2025, but feel free to join us then, @sgmq0!

mialana:
Hey, I'm interested in picking this up for the upcoming Dev Days (May 2025). I will be setting up my MaterialX build today and tomorrow, and will try to familiarize myself with the codebase as much as possible for the 15th. Looking forward to working on this issue!

Also BTW, @sgmq0 and I are actually both in the CG undergrad department at UPenn, and I've confirmed with him that I can take a turn at tackling this.

ld-kerley:
Welcome @mialana - happy to have you on board for dev days 2025.  Please feel free to reach out if you have any questions or problems getting the build setup, either here or on the #materialx ASWF slack channel.

mialana:
Hello. Are there any more available repositories of example `.mtlx` Material files, and specifically any that use **_color4_** fields?

mialana:
To elaborate, for this issue, it is my understanding we also need to add color space transformation in the "Color4" case (please correct me if this is wrong). 

I see that there are some node definitions such as `ND_UsdUVTexture` that have a field `<output name="rgba" type="color4" />`. However, from what I can tell, none of the example materials in the `resources/Materials/` directory of the MaterialX source repo seem to use those nodes. 

I'm not familiar if there are any other community-available resources that would be good for testing with? If not, no worries, I can take a stab at creating a Material that uses one of those nodes myself! Just wanted to double-check first...

ld-kerley:
Regarding example `.mtlx` files - there are a large number of examples in the `resources` folder in this repository, a lot of which are actually the basis for our unit tests.  If you can't find something you need in there, I'd be happy to help you construct what you need to test with - just ping me on slack - or else you can use the MaterialX GraphEditor itself to create mtlx files interactively too. 

[...]

PR #2035, opened 2024-09-28
Convert from sRGB to linear in color picker

sgmq0:
This addresses issue #1209 

Let me know if I'm missing anything, I'm not very sure about the naming conventions.


jstone-lucasfilm:
This looks like a great start, thanks @sgmq0!  Two initial thoughts that come to mind:

- Should we add the reverse transform to the _display_ of each color in the Viewer, so that the color displayed to the artist in a color picker is sRGB encoded rather than linear?
- Should we add the same color transforms to the Graph Editor, so that the artist experience is consistent across these two applications?

kwokcb:
Checking with `nanogui` (Viewer) and `imgui` (Editor) both appear to assume sRGB for display colors so looks like the reverse transform is required.

sgmq0:
School got busy and I ended up not having the bandwidth to finish in a timely manner, but thank you for the reminder! I can try to address both requests within a week or so.

sgmq0:
Hi @jstone-lucasfilm !
Just to clarify, is the goal to transform the color picker to sRGB to match the color space of the viewer?

jstone-lucasfilm:
That's exactly right, @sgmq0, and you should be able to use the built-in `linearToSrgb` function to implement this transform.

This should effectively transform colors from the linear space of the renderer (`lin_rec709`) to the srgb-encoded space of the color picker (`srgb_texture`).

jstone-lucasfilm:
@sgmq0 Just checking in on this proposal, to see if you might have time to finish it up!

jstone-lucasfilm:
@sgmq0 Would it make sense for us to close out this PR, and you can join us at the next Dev Days in May if you'd like to continue this project?

https://www.aswf.io/dev-days/

sgmq0:
@jstone-lucasfilm 
Hey! Sorry for the late reply, I've been busy with school and other things. Yeah, that would make a lot of sense! I will work on this at the next Dev Days.

jstone-lucasfilm:
No problem, @sgmq0, and feel free to join us in May!

PR #2043, opened 2024-09-30
Fix memory leaks in viewer property editor

hridizaroy:
Fixed memory leaks in Viewer by making sure that _gridLayout3 in PropertyEditor gets connected to the top-level NanoGUI parent

### DIFF
--- a/source/MaterialXView/Editor.cpp
+++ b/source/MaterialXView/Editor.cpp
@@ -631,6 +631,8 @@ void PropertyEditor::updateContents(Viewer* viewer)
         {
             ng::Widget* twoColumns = new ng::Widget(_container);
             twoColumns->set_layout(_gridLayout2);
+            ng::Widget* threeColumns = new ng::Widget(_container);
+            threeColumns->set_layout(_gridLayout3);
             ng::Label* modelLabel = new ng::Label(twoColumns, "Shading Model");
             modelLabel->set_font_size(20);
             modelLabel->set_font("sans-bold");


source/MaterialXView/Editor.cpp
```diff
@@ -631,6 +631,8 @@ void PropertyEditor::updateContents(Viewer* viewer)
         {
             ng::Widget* twoColumns = new ng::Widget(_container);
             twoColumns->set_layout(_gridLayout2);
+            ng::Widget* threeColumns = new ng::Widget(_container);
```
jstone-lucasfilm, line 634: Should we expect this new `threeColumns` widget to be referenced elsewhere in the UI layout, as the `twoColumns` variable is used in the construction of `Label` objects?  Or is the connection of `threeColumns` to `_gridLayout3` all that we need here, and no additional references to the widget are needed?

jstone-lucasfilm (commented):
This looks very promising, thanks @hridizaroy, and I had just one question about the expected role of the new variable.

source/MaterialXView/Editor.cpp
hridizaroy: Hello! Thanks for the question. The connection of `threeColumns` to `_gridLayout3` is all we need here. The main purpose of this variable is to guarantee that `_gridLayout3` gets connected to a parent, so that it gets automatically destroyed once the program ends.

jstone-lucasfilm: Sounds good, @hridizaroy, and thanks for the explanation!

This change looks very worthwhile as it stands, though I should note that this will only affect memory leaks in the [MaterialX Viewer](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/Viewer.md), and shouldn't be expected to have an impact on the [MaterialX Graph Editor](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/GraphEditor.md), which is a fully independent code base.

I don't believe that should be considered any kind of blocker for your fix, though you may want to update the description of your change to clarify that it only impacts the MaterialX Viewer.

And if you'd like to address similar issues for the MaterialX Graph Editor in a second PR, we'd appreciate that work as well, though of course you're not obliged to take this on!

hridizaroy: You're so right - sorry for the oversight! I updated the description and commit message! I'll be working on the graph editor and submit a second PR.

jstone-lucasfilm: Thanks @hridizaroy, and a second PR for the Graph Editor would be much appreciated!

jstone-lucasfilm (approved):
Great work, thanks @hridizaroy!

PR #2046, opened 2024-10-01
Allow for choice of font in Graph Editor

kwokcb:
## Issue

The current default font is not a true-type font and does not look very good or scale well.

## Propose Change

Allow the user to specify a true-type font file and size from the command line. 

New options are:
- `--font` : desired font file to use
- `--fontSize` : desired font size. Minimum size clamp to 12.

If the font is not found or fails to load then the current built-in default font and size is used.
Font files are not packaged with the application.

## Examples

- Default:
image
- Roboto 18: (included with `imgui`)
image
- Papyrus 36.
image


jstone-lucasfilm:
This is a very cool idea, @kwokcb!  I'm CC'ing @lfl-eholthouser for her thoughts on this, and I'd be interested in what the consensus choice for a default font would be.

kwokcb:
I got tired of looking at the pixelated font :). 

I'd suggest adding a new pre-packaged "default" font is perhaps beyond the scope of this PR, as it would require packaging a font with the application which I think is different for different platforms ?

lfl-eholthouser:
This is really exciting to see and looks great @kwokcb! I'm not super familiar with packaging fonts but with your comment about that are you saying that the process to package the font is different per application or that certain fonts would only be available on specific applications?

kwokcb:
Hi @lfl-eholthouser,
I am just avoiding the problem of packaging / platforms altogether :). 
This is just a tiny change to add the logic to allow switching the font at runtime. Where the font comes from is up to the user. 

I think a different PR could tackle an enhancement actually package additional fonts with the executable. I haven't looked into what this means and if it may be affected by platform etc. I "think" font files can be used across platform but don't know at this point.

jstone-lucasfilm (approved):

Issue #2018, opened 2024-09-16
Group nodes to create node graph in the Graph Editor


lfl-eholthouser:
In order to create a node graph with the MaterialX Graph Editor the user needs to create a node graph node, double click on it to dive into that node and then add nodes inside that node graph node. Nodes that already exist can only be added to a node graph node by copying or cutting and pasting. I think it would be better for the user experience if there was a hot key that would create a node graph node and add all the selected nodes to it.

WaffleBoyTom:
Hey, this is Thomas, I will be trying to tackle this issue during Dev Days :)

kwokcb:
Two notes on this:

1. Nested node graphs are disallowed in the editor currently due to lack of support. You'll want to check that the selected nodes do not include a nodegraph otherwise this will create unsupported nesting.

2. @lfl-eholthouser, does this also include preserving connections ? I assume to keep it simple it it may not for dev-days , but good to have clarify what the desired logic is here.

lfl-eholthouser:
@kwokcb I was imagining it would preserve connections That way if someone had created a whole graph at the same level as the surface shader, but then realized they actually wanted it in its own nodegraph that work would not have to be redone. This could be two separate issues as well. If the initial phase is tackled during dev days, a new issue can be creating for adding support for connections in the future.

camilleramseur:
Hello! Can I work on this issue for May 2025 DevDays?

camilleramseur:
Will be taking a look at this for Dev Days September 2025

jstone-lucasfilm:
@camilleramseur Since we're approaching our next [ASWF Dev Days event](https://www.aswf.io/dev-days/) in May, I wanted to check in on the status of this project, to see if you'd like to continue working on it.  Let us know!

jstone-lucasfilm:
@camilleramseur I'll unassign you from this issue for now, so that it's available for our upcoming ASWF Dev Days event, but feel free to request it again if you are interested.

minhpham1810:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll start by reproducing the problem locally this week. Any pointers on where to start in the codebase would be appreciated!

[...]

PR #2048, opened 2024-10-02
Add Shift + C hotkey to create a nodegraph from selected nodes

WaffleBoyTom:
Addresses #2018. 
Shift + C will create nodegraph containing selected nodes.
Connections are not preserved.


jstone-lucasfilm:
@WaffleBoyTom This looks like a great start, and to resolve the EasyCLA issue, I'd recommend following up through the support link that the Linux Foundation provides:

jstone-lucasfilm:
In parallel with the CLA process, I'm CC'ing @lfl-eholthouser for her thoughts on the implementation of the feature.

kwokcb (commented):
- Great to see this worked out.  
- I've left some comments about added some more conditional checking and code cleanup. Thanks.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4030,6 +4032,134 @@ void Graph::drawGraph(ImVec2 mousePos)
             }
         }
 
+        // shift + C is pressed
+        // build nodegraph from selected node
+        if (ImGui::IsKeyReleased(ImGuiKey_C) && 
+            io2.KeyShift &&
+            !io2.KeyCtrl) 
+        {
```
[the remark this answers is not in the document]
WaffleBoyTom: I'll fix all of these issues ! Thanks :)

jstone-lucasfilm:
I wanted to follow up on this proposal, @WaffleBoyTom, to see if you might have the bandwidth to address the requests above.  Let us know either way!

WaffleBoyTom:
> I wanted to follow up on this proposal, @WaffleBoyTom, to see if you might have the bandwidth to address the requests above. Let us know either way!

Yes ! Sorry about the radio silence. I was planning on finishing this up this weekend hopefully !

source/MaterialXGraphEditor/Graph.cpp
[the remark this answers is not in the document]
WaffleBoyTom: fixed most of these but not entirely sure how to check that I'm not already within a nodegraph : >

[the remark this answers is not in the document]
WaffleBoyTom: I've updated my code, it should take into account your notes

jstone-lucasfilm:
This looks very promising, thanks @WaffleBoyTom!

Make sure to resolve the EasyCLA issue above, following the link provided by the Linux Foundation, so that we can merge your work once it's approved by the MaterialX team.

WaffleBoyTom:
> This looks very promising, thanks @WaffleBoyTom!
> 
> Make sure to resolve the EasyCLA issue above, following the link provided by the Linux Foundation, so that we can merge your work once it's approved by the MaterialX team.

Hey ! I followed the link and signed the CLA so I think that should be okay ? I might be missing something

jstone-lucasfilm:
@WaffleBoyTom It looks like the EasyCLA system is still warning about commits that haven't been approved:

WaffleBoyTom:
> @WaffleBoyTom It looks like the EasyCLA system still needs approvals for some of the commits above, so perhaps you can follow up by clicking their support link?
> 
> > For further assistance with EasyCLA, [please submit a support request ticket] [jira.linuxfoundation.org/servicedesk/customer/portal/4](https://jira.linuxfoundation.org/servicedesk/customer/portal/4)).

I think we did it :o

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4029,6 +4031,106 @@ void Graph::drawGraph(ImVec2 mousePos)
...
+            
+                for (std::map<UiNodePtr, UiNodePtr>::iterator iter = _copiedNodes.begin(); iter != _copiedNodes.end(); iter++)
+                {
+                    copyUiNode(iter->first);
+                }
+                _addNewNode = true;
+            }
+
+        }
+
```
kwokcb, line 4133: A small thing here. I think you want to clear `_copiedNodes` if not performing the copy to make the action fail in an atomic manner. 

In fact maybe you want to clear it all the time since your just using copy+paste as an internal implementation convenience.

kwokcb (commented):
Thanks for the updates @WaffleBoyTom .  Looks good. 

I just have one small addition item which is to clear the copy "buffer" after the action to it's not accidently pasted erroneously at some later time.

source/MaterialXGraphEditor/Graph.cpp
WaffleBoyTom: hey ! Thanks ! 
I should able to make those changes this weekend ! :)

WaffleBoyTom: I added the call to clear the copied nodes. I also cleaned up a sneaky #include iostream that I had accidently left in there..

kwokcb (approved):
Thanks for the cleanup @WaffleBoyTom. Looks good to go!

jstone-lucasfilm:
@WaffleBoyTom I've been giving this new feature a try in my local build, but so far it's not behaving as I would expect, though I may be misunderstanding the intended design.

I'm clearing the contents of the Graph Editor, creating a small number of nodes at root scope, then selecting them all, as seen in the following screenshot:

image

I then hit the `SHIFT-C` key combination, but nothing happens in the editor, and the new logic seems to be hitting this early exit because `isNodeGraph` is `true`:

```
if (!_copiedNodes.empty() && !isNodeGraph)
```

Just to make sure I didn't accidentally break anything with my own merges and formatting fixes, I tried the same process again with the codebase at the time of your most recent commit, but I get the same results.

What might I be missing?

WaffleBoyTom:
> I then hit the `SHIFT-C` key combination, but nothing happens in the editor, and the new logic seems to be hitting this early exit because `isNodeGraph` is `true`:
> 
> ```
> if (!_copiedNodes.empty() && !isNodeGraph)
> ```
> 
> Just to make sure I didn't accidentally break anything with my own merges and formatting fixes, I tried the same process again with the codebase at the time of your most recent commit, but I get the same results.
> 
> What might I be missing?

Hey Jonathan ! 
It seems you're right.. `!isNodeGraph` was added to prevent nested node graphs but that means that the code fails in your example because `tiledhexagons_color3` is a nodegraph. 
I struggled to find what the right checks should be to prevent nested nodegraphs and this clearly isn't it unfortunately.
Would you happen to know how I can correct this ? 
Thanks :)

kwokcb:
@WaffleBoyTom, @jstone-lucasfilm : I'll take a quick look. The current logic does need some changes. It should be:
1. Are we currently inside a nodegraph: ` _isNodeGraph` 
2. Have we selected any nodes which are nodegraphs: `selectedNode->getCategory() == "nodegraph"`.

WaffleBoyTom:
Thank you, I should be able to make these changes this weekend

WaffleBoyTom:
Thanks !

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4054,6 +4057,89 @@ void Graph::drawGraph(ImVec2 mousePos)
...
+                }
+            }
+
+            // Delete nodes we just copied
+            if (!_copiedNodes.empty() && !isNodeGraph)
+            {
+                for (std::map<UiNodePtr, UiNodePtr>::iterator iter = _copiedNodes.begin(); iter != _copiedNodes.end(); iter++)
+                {
+                    UiNodePtr node = iter->first;
+                    deleteNode(node);
```
kwokcb, line 4102: This seems to break the links between nodes so even though they are copied the links are not copied over? 

Tried moving this later after the paste but it crashes with some shared ptr de-refencing error.  Even with this as is, I've seen hang in refresh if you try and select the surfaceshader in the default marble graph and try this operation then try and go back to the parent.

@lfl-eholthouser pinging you as you know this code much better than I.  If you don't delete then a nodegraph is created properly with linked nodes and have not experienced any side-effects.

WaffleBoyTom: @kwokcb this works great, thank you ! 
I am noticing something though.
If I create an `add_float` then create another one by copy and pasting the first one, then select the copy and hit shift + C to create the subgraph, the node does not get cut (but also appears inside the subgraph). Clicking on the node then crashes the editor. The copy also is collapsed and doesn't show its inputs. 

This seems to be an issue with how pasting and cutting works. If I put down an `add_float` then copy it. Then select both and use ctrl+x to cut, the copy does not get cut.

jstone-lucasfilm:
@WaffleBoyTom Just checking in on this proposal, to see where things currently stand!  Are there outstanding improvements that still need to be made?

jstone-lucasfilm:
@WaffleBoyTom Would it make sense for us to close out this PR, and you can join us at the next Dev Days in May if you'd like to continue this project?

https://www.aswf.io/dev-days/

jstone-lucasfilm:
@WaffleBoyTom I'll close out this PR for now, and feel free to pick up the thread of this work at Dev Days in May!

PR #1567, opened 2023-10-13
Sphinx Python Documentation

StefanHabel:
Work in progress.

This PR adds a new build target named `MaterialXDocsPython`, which generates Python API documentation using Sphinx.

The existing developer guide contents are incorporated into the new HTML documentation, which lives side-by-side to the existing Doxygen-generated C++ API documentation.

Docstrings were added to all modules, functions, classes, methods, and attributes.

All parameters were named. None should appear as `arg0`, `arg1`, `arg2` etc anymore.

Statistics about the Python API have been added to the Sphinx build process.

The Sphinx build output currently ends with the following messages, providing an overview of the state of the API docs:
```
The MaterialX Python API consists of:
    * 11 modules
    * 48 functions
    * 139 classes
    * 1180 methods
    * 6 exception types

build succeeded.
```

Docstring example:
```
$ python3 -c 'import PyMaterialXFormat; help(PyMaterialXFormat.flattenFilenames)'
```
```
Help on built-in function flattenFilenames in module PyMaterialXFormat:

flattenFilenames(...) method of builtins.PyCapsule instance
    flattenFilenames(doc: PyMaterialXCore.Document, searchPath: PyMaterialXFormat.FileSearchPath = mx.FileSearchPath(), customResolver: PyMaterialXCore.StringResolver = None) -> None

    Flatten all filenames in the given document, applying string resolvers at
    the scope of each element and removing all fileprefix attributes.

    :param doc: The document to modify.
    :type doc: Document
    :param searchPath: An optional search path for relative to absolute path
        conversion.
    :type searchPath: FileSearchPath
    :param customResolver: An optional custom resolver to apply.
    :type customResolver: StringResolver
```

Preview of generated HTML pages:

Screenshot 2023-10-21 at 14 14 41

<hr>

Screenshot 2023-10-21 at 14 28 03

<hr>

Development environment:

- Apple clang version 15.0.0 (clang-1500.0.40.1)
- Python 3.12.0 (v3.12.0:0fb18b02c8, Oct  2 2023, 09:45:56) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
- Sphinx 7.2.6
- cmake version 3.27.6

Update #342

Issue #342, opened 2020-01-15
[Lucasfilm] Sphinx Python Documentation


jstone-lucasfilm:
Add support for the generation of Sphinx documentation for MaterialX Python, using the approach described here:

https://pybind11.readthedocs.io/en/stable/advanced/misc.html#generating-documentation-using-sphinx

StefanHabel:
I'd like to take a stab at this for the [ASWF Dev Days 2023](https://sites.google.com/view/aswfdevdays/home).

If I’m understanding this right, the work here roughly consists of two parts:

1. Integrating a Sphinx build of the MaterialX Python API documentation into the MaterialX CMake build
2. Adding/extending docstrings for MaterialX Python bindings using pybind11

While it may not be possible/feasible to add/curate docstrings for all MaterialX Python bindings during Dev Days, it should be possible to lay the foundation for building API documentation using Sphinx on the day.

Proposed approach/outcome:

- An optional extra build target similar to the existing `cmake --build . --target MaterialXDocs` (which uses `doxygen` to generate C++ API documentation), possibly named `MaterialXDocsPython`.
- Documentation in HTML format that is generated in a sibling folder next to the existing C++ API documentation (as opposed to living in the same `html` folder).

Proposed artifacts:

1. A new CMake module at `cmake/modules/FindSphinx.cmake` to be able to find the `sphinx-build` executable (Windows: `sphinx-build.exe`) from `documents/CMakeLists.txt` using `find_package(Sphinx)`.
    - cf. https://eb2.co/blog/2012/03/sphinx-and-cmake-beautiful-documentation-for-c-projects/
    - cf. https://www.vortech.nl/en/integrating-sphinx-in-cmake/
2. Modifications to `documents/CMakeLists.txt` to create a new build target named `MaterialXDocsPython`, which would run `sphinx-build` to generate Python API developer documentation, in a directory that is separate to the existing `doxygen`-generated `html` documentation, e.g. named `html-sphinx`.
    - https://www.sphinx-doc.org/en/master/tutorial/getting-started.html
4. A Sphinx configuration template file (e.g. `documents/sphinx-conf.py.in`) that would be filled with values of configuration variables like `MATERIALX_LIBRARY_VERSION` as part of the build, in order to generate the `conf.py` file that `sphinx-build` uses.
[...]

PR #2051, opened 2024-10-07
Docstrings for PyMaterialXCore classes.

StefanHabel:
Mostly copied from corresponding C++ header files, but using backticks to automatically link to relevant sections in the Python API documentation, e.g. `Document` to link to `PyMaterialXCore.Document.html`.

Using `mod.attr("<name>").doc() = ...` syntax so that this patch only adds new lines, rather than modifying existing lines.

Using `:see:` fields to link to the corresponding C++ API docs page.

The docstrings can be retrieved using `help()`, e.g.
```bash
$ python3 -c 'import PyMaterialXCore; help(PyMaterialXCore.NodeDef)'
```
```
Help on class NodeDef in module PyMaterialXCore:

class NodeDef(InterfaceElement)
 |  A node definition element within a `Document`.
 |
 |  A `NodeDef` provides the declaration of a node interface, which may then
 |  be instantiated as a `Node`.
 |
 |  :see: https://materialx.org/docs/api/class_node_def.html
 |
 |  Method resolution order:
 |      NodeDef
 |      InterfaceElement
 |      TypedElement
 |      Element
 |      pybind11_builtins.pybind11_object
 |      builtins.object
 |
 |  Methods defined here:
 |
 |  __init__(self, /, *args, **kwargs)
 |      Initialize self.  See help(type(self)) for accurate signature.
 |
[...]
```

Demo of generated HTML page available here:
https://stefanhabel.github.io/generated/PyMaterialXCore.html

Here's how the above docstring of the `NodeDef` class appears in the Python API documentation:
https://stefanhabel.github.io/generated/PyMaterialXCore.NodeDef.html

| Before | After |
|-|-|
| Screenshot 2024-10-06 at 13 11 00 | Screenshot 2024-10-06 at 17 13 54 |

Split from #1567.

Update #342.


source/PyMaterialX/PyMaterialXCore/PyValue.cpp
```diff
@@ -39,4 +43,20 @@ void bindPyValue(py::module& mod)
...
+    mod.attr("TypedValue_floatarray").doc() = "A `Value` object that stores a `list` of `float` values.";
+    mod.attr("TypedValue_integer").doc() = "A `Value` object that stores a value of type `int`.";
+    mod.attr("TypedValue_integerarray").doc() = "A `Value` object that stores a `list` of `int` values.";
+    mod.attr("TypedValue_matrix33").doc() = "A `Value` object that stores a value of type `Matrix33`.";
+    mod.attr("TypedValue_matrix44").doc() = "A `Value` object that stores a value of type `Matrix44`.";
+    mod.attr("TypedValue_string").doc() = "A `Value` object that stores a value of type `str`.";
+    mod.attr("TypedValue_stringarray").doc() = "A `Value` object that stores a `list` of `str` values.";
+    mod.attr("TypedValue_vector2").doc() = "A `Value` object that stores a value of type `Vector2`.";
+    mod.attr("TypedValue_vector3").doc() = "A `Value` object that stores a value of type `Vector3`.";
+    mod.attr("TypedValue_vector4").doc() = "A `Value` object that stores a value of type `Vector4`.";
```
StefanHabel, line 61: We may want to simplify these, e.g.
```diff
-    mod.attr("TypedValue_boolean").doc() = "A `Value` object that stores a value of type `bool`.";
+    mod.attr("TypedValue_boolean").doc() = "A `Value` object that stores a `bool` value.";
```
Or even more radical:
```diff
-    mod.attr("TypedValue_boolean").doc() = "A `Value` object that stores a value of type `bool`.";
+    mod.attr("TypedValue_boolean").doc() = "A `bool` value.";
```

StefanHabel: Simplified in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2051/commits/056a7efcd5a276074dd9871f6e922c806f2a6291

source/PyMaterialX/PyMaterialXCore/PyDefinition.cpp
```diff
@@ -35,6 +35,13 @@ void bindPyDefinition(py::module& mod)
         .def_readonly_static("CHANNEL_NODE_GROUP", &mx::NodeDef::CHANNEL_NODE_GROUP)
         .def_readonly_static("ORGANIZATION_NODE_GROUP", &mx::NodeDef::ORGANIZATION_NODE_GROUP)
         .def_readonly_static("TRANSLATION_NODE_GROUP", &mx::NodeDef::TRANSLATION_NODE_GROUP);
+    mod.attr("NodeDef").doc() = R"docstring(
+    A node definition element within a `Document`.
+
+    A `NodeDef` provides the declaration of a node interface, which may then
```
jstone-lucasfilm, line 41: My main concern here is introducing _duplicate_ documentation, which would need to be kept in sync with the corresponding Doxygen comments in the C++ codebase:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Definition.h#L79

Is there any potential solution where we can inherit these Python documentation strings from the existing Doxygen comments, or does the functionality of the Sphinx documentation system not support this?

StefanHabel: yeah, that is a concern 🫤

Alternatively, following the _Don't Repeat Yourself_ principle would require engineering to extract the docstrings from header files and assign them dynamically to Python objects as the Sphinx docs are generated.

This could potentially be achieved by running Doxygen and extracting the header documentation in a format that could be read in a Python script as part of generating the Sphinx docs.

On the other hand, a benefit of curating the docstrings separately in the way proposed here would be to accommodate language-specific documentation, e.g. to account for differences in the Python API compared to the C++ API, e.g. when it comes to mutable parameters in C++ vs. return values of functions in Python.

ld-kerley: In the past I have had some success using https://github.com/pybind/pybind11_mkdoc - which extracts the doxygen doc strings in a format that can be consumed by pybind11.  Its been a while, but maybe its worth a look?

jstone-lucasfilm:
After further discussion with @StefanHabel, we're going to leave the remaining step of publishing a full MaterialX Python API as work for the future, and future developers should feel free to pick up where Stefan left off.

See https://github.com/AcademySoftwareFoundation/MaterialX/pull/2051#discussion_r2040710113 for more details on the remaining steps for this project.

PR #2053, opened 2024-10-08
Fix issue with framework builds on latest Xcode with embedded platforms

dgovil:
I discovered an issue with my [previous frameworks PR](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2020) that was showing up on recent Xcode versions where the `Resources` folder was acting as a reserved name on iOS, but is a required name on macOS. I also took the opportunity to clean up the linker code here.


jstone-lucasfilm (approved):
This looks great to me, thanks @dgovil, and I'm CC'ing @ld-kerley in case he has additional thoughts.

PR #2054, opened 2024-10-08
Add support for data library referencing

ashwinbhat:
### Background
MaterialX documents are required to be complete self contained with all required node definitions. This is done by using the `mx::importLibrary` API. While processing large number of documents the Import and Merge workflows can cause performance and memory challenges. 
### Proposal 
Since the MaterialX library or the Data library hosts definitions, we propose that these definitions could be referenced by a document instead of importing them. This concept of referencing the Data library may have some workflow impact such as, do new definitions to into Data library or the document or can a document have local definitions? The answer depends on the specific artist or studio workflow.

To maintain compatibility with existing usage patterns, we have introduced a new set of methods **that allow registration of data library for a document by setting a document reference to it**.
1. setDataLibrary(ConstDocumentPtr dataLibrary);
2. getDataLibrary()
3. hasDataLibrary()

### Using this new Data library registration, two usage patterns are possible
**Build library and import**
This is the current usage pattern
- use `importlibrary `to build your datalibrary
- use `importLibrary `to import the datalibrary to the working document.

**Build library and register**
Proposed usage pattern
- use `importlibrary` to build your datalibrary
- use `setDataLibrary `to assign the datalibrary to the working document.
This can be extended were 3rd party defs can be imported into the datalibrary

**Other changes include**
- Updates to `getChildOfType` and `getChildrenOfType` to support Data Library.
- Updates to redirect _Document::get*_ calls to process local document content and Data Library.

### Results

Benchmark results from GLSL Performance Test. This test loads, validates and shadergen about 50 MaterialX files. 100 samples are generated thus that simulating 5k documents. 

**Current uses ImportLibrary**
```
benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
Load documents, validate and                                                   
generate shader                                100             1     12.4211 m 
                                         8.94671 s     7.51572 s     14.8634 s 
                                         12.0468 s    132.829 ms     28.2088 s 

906.537 s: GenShader: GLSL Performance Test
```
**Proposed uses Datalibrary**
```
benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
Load documents, validate and                                                   
generate shader                                100             1     29.8943 s 
                                        253.171 ms    251.445 ms    255.226 ms 
                                        9.59824 ms    7.90013 ms    12.2619 ms 
                                                                              
30.054 s: GenShader: GLSL Performance Test


```


source/PyMaterialX/PyMaterialXCore/PyElement.cpp
```diff
@@ -14,8 +14,6 @@
 
 #define BIND_ELEMENT_FUNC_INSTANCE(T)                                                                           \
 .def("_addChild" #T, &mx::Element::addChild<mx::T>)                                                             \
-.def("_getChildOfType" #T, &mx::Element::getChildOfType<mx::T>)                                                 \
-.def("_getChildrenOfType" #T, &mx::Element::getChildrenOfType<mx::T>, py::arg("category") = mx::EMPTY_STRING)   \
```
ashwinbhat, line 18: @bernardkwok @jstone-lucasfilm I need some guidance on these binding api 
`template <class T> shared_ptr<T> getChildOfType(ConstDocumentPtr datalibrary, const string& name) const`
`template <class T> shared_ptr<T> getChildOfType(const string& name) const`

source/MaterialXCore/Document.cpp
```diff
@@ -357,6 +357,14 @@ vector<OutputPtr> Document::getMaterialOutputs() const
 
 vector<NodeDefPtr> Document::getMatchingNodeDefs(const string& nodeName) const
 {
+    // Return all nodedefs from datalibrary if available
```
jstone-lucasfilm, line 360: @ashwinbhat Wouldn't we also need to handle the case of a content document that contains _its own_ custom nodes, as might be packaged and shared between studios?  To handle that case, we'd want to first check for a matching nodedef in the data libraries, and then fall back to checking for a matching nodedef in the content document.

Following that pattern, perhaps we should take the same approach with the other updated methods in this pull request, including `Document::getMatchingImplementation` below.

source/PyMaterialX/PyMaterialXCore/PyElement.cpp
jstone-lucasfilm: @ashwinbhat Initially I think it should be fine to omit these new Python bindings, while we're working out the details of the new API, and we can add these once the C++ API is locked down.

source/MaterialXCore/Document.cpp
ashwinbhat: @jstone-lucasfilm in the code below, we check in the datalibrary if available and then fallback to checking in local.

jstone-lucasfilm: I see what you mean, but couldn't there be nodedefs for the given node name in _both_ the data library and the content document?  For example, the data library might have a node named `multiply`, but only the content document would have the specific version  of `multiply` that the caller is looking for.

ashwinbhat: The proposed workflow, if the caller chooses to Register a datalibrary then the data library gets precedence. The caller can still continue to use the import mechanism. 

We can update getMatchingNodeDefs so that it checks both data library and content document. Would that be preferred?

jstone-lucasfilm: @ashwinbhat I'm imagining that we want the behavior to exactly match what a user would see if they called `importLibrary` instead of `registerDataLibrary`, and I believe that would require us to check both the data library and content document, merging the matches from both of them.

Let me know your thoughts, though, and that's just one opinion on the matter!

source/MaterialXCore/Document.h
```diff
@@ -680,6 +708,8 @@ class MX_CORE_API Document : public GraphElement
   private:
     class Cache;
     std::unique_ptr<Cache> _cache;
+    // Data library for the document
+    ConstDocumentPtr _dataLibrary;
```
ld-kerley, line 712: I think other methods like `initialize()` and `copy()` probably also need to be updated to handle this extra member variable.

source/MaterialXCore/Element.cpp
```diff
@@ -564,7 +564,10 @@ bool ValueElement::validate(string* message) const
         const string& unittype = getUnitType();
         if (!unittype.empty())
         {
-            unitTypeDef = getDocument()->getUnitTypeDef(unittype);
+
+            unitTypeDef = getDocument()->hasDataLibrary() ?
```
ld-kerley, line 568: Is there a reason this logic isn' just in `Document::getUnitTypeDef()` like the other document query functions?

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -202,10 +202,15 @@ void ShaderGraph::addDefaultGeomNode(ShaderInput* input, const GeomPropDef& geom
         // input here and ignore the type of the geomprop. They are required to have the same type.
         string geomNodeDefName = "ND_" + geomprop.getGeomProp() + "_" + input->getType().getName();
         NodeDefPtr geomNodeDef = _document->getNodeDef(geomNodeDefName);
-        if (!geomNodeDef)
+        if (!geomNodeDef && _document->hasDataLibrary())
```
ld-kerley, line 205: This code here looks to me like we're inspecting the local document for a nodeDef first, and then falling back to the dataLibrary.

But then isn't Document::getNodeDef() above already checking the dataLibrary?  so perhaps this additional secondary check isn't necessary?

source/MaterialXCore/Document.cpp
ld-kerley: I think this is an interesting point to discuss - should `registerDataLibrary` match the `importLibrary` behavior?  

I can certainly imagine circumstances where I might even want to not fallback to any local node definition in the local document and want to be able to validate that my document only uses nodes from the dataLibrary provided. 

Perhaps `registerDataLibrary()` has some concept of strictness?  or some input that could control its "fallback" behavior?

ashwinbhat: I've updated this so that both the data library and content document are now checked. @ld-kerley I think once the datalibrary concept is adopted we could start introducing strictness. The current goal is keep it similar to importLibrary workflow.

source/JsMaterialX/JsMaterialXCore/JsElement.cpp
```diff
@@ -22,10 +22,6 @@ namespace mx = MaterialX;
     BIND_MEMBER_FUNC("setValue" #NAME, mx::ValueElement, setValue<T>, 1, 2, const T&, stRef)
 
 #define BIND_ELEMENT_CHILD_FUNC_INSTANCE(NAME, T)                                                 \
-    BIND_MEMBER_FUNC("addChild" #NAME, mx::Element, addChild<T>, 0, 1, stRef)                     \
```
jstone-lucasfilm, line 25: @ashwinbhat I see that the one build failure is happening in our JavaScript test suite, and I wonder if it would address the issue if we restored these lines in `JsElement.cpp` while we're working out the details of the C++ API.

source/MaterialXCore/Node.cpp
```diff
@@ -714,6 +722,18 @@ NodeDefPtr NodeGraph::getNodeDef() const
             }
         }
     }
+    // Check datalibrary if available
```
kwokcb, line 725: Again should this not come first for consistency ?

source/JsMaterialX/JsMaterialXCore/JsElement.cpp
kwokcb: Maybe just comment it out ? I could try to build locally to help resolve wrapper issues.

kwokcb:
Just wanted to check ordering precedence:

A. If you do this:
------------------
1. importLibrary definitions on working doc
2. registerLibrary definitions on working doc 

Then registerLibrary defs have precedence for nodedefs, impls, typedefs, unitdefs etc. (or we could flip this order)

B. If you do this:
------------------
1. importLibrary definitions on working doc
2. registerLibrary definitions on working doc
3. load defs from another doc into working doc

Then registerLibrary defs still have precedence since 3 is merged into 1 (into the same document).

C. If you do this:
------------------
1. importLibrary definitions on working doc
2. registerLibrary definitions on working doc
3. extract defs from another doc into the registered library (*)

Then registerLibrary defs + local defs will have precedence , since 3 and 2 are in the same document.
(*) This has the side-effect that all documents sharing the same library will be  "polluted" by local defs which 
I don't believe is currently possible to do.

D. If you have a 3rd party library you have these options
--------------------------------------------------------------
1. importLibrary 3rd party definitions into working doc, and/or  2a or 2c.
2a. importLibrary 3rd party defs into shared library and/or 1 or 2c.
2b. registerLibrary(shared library)
2c. importLibrary 3rd party defs into shared library and/or 1 or 2a

In all cases, you cannot keep 3rd party defs separated (order) as is the case with local document defs.

source/MaterialXCore/Document.cpp
```diff
@@ -357,34 +357,39 @@ vector<OutputPtr> Document::getMaterialOutputs() const
...
+    // Gather all nodedefs from datalibrary if available
+    vector<NodeDefPtr> matchingNodeDefs = hasDataLibrary() ? 
+                                          getRegisteredDataLibrary()->getMatchingNodeDefs(nodeName) :
+                                          vector<NodeDefPtr>();
+
     // Refresh the cache.
     _cache->refresh();
 
     // Return all nodedefs matching the given node name.
     if (_cache->nodeDefMap.count(nodeName))
```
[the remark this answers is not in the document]
ashwinbhat: @kwokcb what is the current workflow if a user does `loadlibraries` and performs an `importLibrary` into a document that has duplicates. Which ones get rejected?

source/MaterialXCore/Document.cpp
```diff
@@ -399,6 +404,7 @@ bool Document::validate(string* message) const
 void Document::invalidateCache()
 {
     _cache->valid = false;
+    _dataLibrary = nullptr;
```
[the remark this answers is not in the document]
ashwinbhat: Good point, I will correct this.

source/MaterialXCore/Document.h
```diff
@@ -380,13 +384,13 @@ class MX_CORE_API Document : public GraphElement
     /// Return the AttributeDef, if any, with the given name.
     AttributeDefPtr getAttributeDef(const string& name) const
     {
-        return getChildOfType<AttributeDef>(name);
+        return hasDataLibrary() ? getChildOfType<AttributeDef>(getRegisteredDataLibrary(), name) : getChildOfType<AttributeDef>(name);
```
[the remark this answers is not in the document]
ashwinbhat: We can try it out in future.

source/MaterialXCore/Document.h
```diff
@@ -665,6 +669,32 @@ class MX_CORE_API Document : public GraphElement
...
+    /// @{
+
+    /// Register the given document as MaterialX data library for document
+    /// The MaterialX data library can be created using the loadLibraries utility
+    /// For improved performance it is recommended the data library 
+    /// is on the document instead of importing it
+    /// @param Data Library document to register.
+    void registerDataLibrary(ConstDocumentPtr dataLibrary)
+    {
+        _dataLibrary = dataLibrary;
```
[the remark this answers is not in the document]
ashwinbhat: Good catch, I'll add a clearRegisteredDataLibary

source/MaterialXCore/Node.cpp
ashwinbhat: When do an import aren't the library nodes added before the current document? In that case we are keeping the same logic checking the library nodes and then local. Let me know if this should be still reversed.

source/JsMaterialX/JsMaterialXCore/JsElement.cpp
ashwinbhat: @kwokcb I'll restore the current state

ashwinbhat:
> Just wanted to check ordering precedence:
> 
> ## A. If you do this:
> 1. importLibrary definitions on working doc
> 2. registerLibrary definitions on working doc
> 
> Then registerLibrary defs have precedence for nodedefs, impls, typedefs, unitdefs etc. (or we could flip this order)
> 
> ## B. If you do this:
> 1. importLibrary definitions on working doc
> 2. registerLibrary definitions on working doc
> 3. load defs from another doc into working doc
> 
> Then registerLibrary defs still have precedence since 3 is merged into 1 (into the same document).
> 
> ## C. If you do this:
> 1. importLibrary definitions on working doc
> 2. registerLibrary definitions on working doc
> 3. extract defs from another doc into the registered library (*)
> 
> Then registerLibrary defs + local defs will have precedence , since 3 and 2 are in the same document. (*) This has the side-effect that all documents sharing the same library will be "polluted" by local defs which I don't believe is currently possible to do.
> 
> ## D. If you have a 3rd party library you have these options
> 1. importLibrary 3rd party definitions into working doc, and/or  2a or 2c.
>    2a. importLibrary 3rd party defs into shared library and/or 1 or 2c.
>    2b. registerLibrary(shared library)
>    2c. importLibrary 3rd party defs into shared library and/or 1 or 2a
> 
> In all cases, you cannot keep 3rd party defs separated (order) as is the case with local document defs.

@kwokcb I don't think importlibrary and registerLibrary are supposed to be used in conjunction. 
The recommended patterns are
## A: Build library and register
1. use importlibrary  to build your datalibrary
2. use registerLibrary to assign the datalibrary to the working document.
This can be extended were 3rd party defs can be imported into the datalibrary

## B: Build library and import
1. use importlibrary  to build your datalibrary
2. use importLibrary to import the datalibrary to the working document.

source/PyMaterialX/PyMaterialXCore/PyElement.cpp
```diff
@@ -14,8 +14,8 @@
...
+/*.def("_getChildOfType" #T, &mx::Element::getChildOfType<mx::T>)                                             */\
+/*.def("_getChildrenOfType" #T, &mx::Element::getChildrenOfType<mx::T>, py::arg("category") = mx::EMPTY_STRING) */\
```
ashwinbhat, line 18: @jstone-lucasfilm @kwokcb Does this need new python implementation here
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/python/MaterialX/main.py#L39

jstone-lucasfilm: Perhaps we can just start with the proposed C++ API changes, and once these are locked down, we can move on to the Python and JavaScript bindings?

source/MaterialXCore/Element.h
```diff
@@ -445,6 +445,20 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
         return child ? child->asA<T>() : shared_ptr<T>();
     }
 
+    /// Return the child element from data library , if any, with the given name and subclass.
+    /// If a child with the given name exists, but belongs to a different
+    /// subclass, then an empty shared pointer is returned.
+    template <class T> shared_ptr<T> getChildOfType(ConstElementPtr datalibrary, const string& name) const
```
jstone-lucasfilm, line 451: @ashwinbhat I'd recommend leaving the interface of these core methods, `getChildOfType` and `getChildrenOfType` unchanged in your PR, and creating a brand new name for the methods that you'd like to add.  This should allow us to focus on just the C++ API in this first phase of review, leaving the authoring of Python and JavaScript bindings to a future phase.

jstone-lucasfilm: @ashwinbhat That sounds fine to start with, and I suspect many of these helper methods will evolve during the review and discussion process!

source/MaterialXCore/Document.h
```diff
@@ -665,6 +669,42 @@ class MX_CORE_API Document : public GraphElement
...
 
+    /// @name MaterialX data library
+    /// @{
+
+    /// Register the given document as MaterialX data library for document
+    /// The MaterialX data library can be created using the loadLibraries utility
+    /// For improved performance it is recommended the data library 
+    /// is on the document instead of importing it
+    /// @param Data Library document to register.
+    void registerDataLibrary(ConstDocumentPtr dataLibrary)
```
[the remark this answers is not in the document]
ashwinbhat: @jstone-lucasfilm I agree it does line up with current Document api pattern.

source/MaterialXCore/Document.h
```diff
@@ -78,13 +79,13 @@ class MX_CORE_API Document : public GraphElement
     /// Return the NodeGraph, if any, with the given name.
     NodeGraphPtr getNodeGraph(const string& name) const
     {
-        return getChildOfType<NodeGraph>(name);
+        return hasDataLibrary() ? getChildofTypeFromLibrary<NodeGraph>(getDataLibrary(), name) : getChildOfType<NodeGraph>(name);
```
jstone-lucasfilm, line 82: @ashwinbhat For this category of updated method implementation (e.g. `getNodeGraph`, `getNodeDef`, `getAttributeDef`), I believe we actually need to check the content document when the data library _has no matching child_, not merely in the case when the data library is not present at all.

Does that sound right to you as well?

ashwinbhat: @jstone-lucasfilm I think the implementation of `template <class T> shared_ptr<T> getChildofTypeFromLibrary` will fallback to get this behavior you are suggesting.

jstone-lucasfilm: Ah, thanks for that clarification, @ashwinbhat!

Now I see the pattern that you're proposing here, and one suggestion that I would make would be to handle the `nullptr` data library case _inside_ of `getChildofTypeFromLibrary`, allowing you to write the external call in a clearer way:

```
return getChildofTypeFromLibrary<NodeGraph>(getDataLibrary(), name);
```

We might even consider renaming the function from `getChildofTypeFromLibrary` to `getChildOfTypeWithLibrary`, in order to make this pattern as clear as possible to readers.

source/MaterialXCore/Document.h
```diff
@@ -78,13 +79,13 @@ class MX_CORE_API Document : public GraphElement
     /// Return the NodeGraph, if any, with the given name.
     NodeGraphPtr getNodeGraph(const string& name) const
     {
-        return getChildOfType<NodeGraph>(name);
+        return getChildOfTypeWithLibrary<NodeGraph>(getDataLibrary(), name);
```
jstone-lucasfilm, line 82: This is really starting to come together, thanks @ashwinbhat!

Looking at the latest proposal, one thought that comes to mind is that we can now merge the new functionality of `getChildOfTypeWithLibrary` directly into the existing `getChildOfType`, simply making `dataLibrary` an optional second argument rather than a mandatory first argument.

This same pattern could potentially be applied to similar helper methods in the future, where an optional `dataLibrary` argument can be passed in order to request the checking of the data library in addition to the content document.

Does that proposed simplification sound reasonable to you?

ashwinbhat: This is what my initial intent was but I was running into some pybind issues and it was touch a lot of files. Let me try it out. I'll post an update if it works.

jstone-lucasfilm: With all of the latest simplifications to the proposal, I think we're now in a great place to make that change, and hopefully some combination of myself, @bernardkwok, and @ld-kerley can help to work through the Python and JS bindings!

ashwinbhat: @jstone-lucasfilm I've simplified it. I think something like datalibrary should be a required argument but since we are trying to keep the API it should be fine. Maybe we can refine it in future.

jstone-lucasfilm: @ashwinbhat Here's the specific API and implementation that I was thinking we would use, and let me know if this seems like a logical approach to you:

```
    template <class T> shared_ptr<T> getChildOfType(const string& name, ConstDocumentPtr dataLibrary = nullptr) const
    {
        ElementPtr child = dataLibrary ? dataLibrary ->getChild(name) : nullptr;
        if (!child)
        {
            child = getChild(name);
        }
        return child ? child->asA<T>() : shared_ptr<T>();
    }
```

ashwinbhat: Thanks that's what I have. I've updated the py and js bindings as well.

source/MaterialXCore/Definition.cpp
```diff
@@ -186,13 +186,28 @@ StringVec TargetDef::getMatchingTargets() const
 vector<UnitDefPtr> UnitTypeDef::getUnitDefs() const
 {
     vector<UnitDefPtr> unitDefs;
+    
+    // Gather from data library.
+    if (getDocument()->hasDataLibrary())
```
jstone-lucasfilm, line 191: With the latest `dataLibrary` argument to `getChildrenOfType`, can this update to `getUnitDefs` now be simplified to use it as well?

I'm imagining something along the lines of the following:

```
vector<UnitDefPtr> UnitTypeDef::getUnitDefs() const
{
    vector<UnitDefPtr> unitDefs;

    for (UnitDefPtr unitDef : getDocument()->getChildrenOfType<UnitDef>(EMPTY_STRING, getDocument()->getDataLibrary()))
    {
        if (unitDef->getUnitType() == _name)
        {
            unitDefs.push_back(unitDef);
        }
    }

    return unitDefs;
}
```

Essentially we would now just be adding the new `dataLibrary` argument to the existing call, and making no other changes to the method.

source/MaterialXCore/Definition.cpp
```diff
@@ -186,13 +186,28 @@ StringVec TargetDef::getMatchingTargets() const
 vector<UnitDefPtr> UnitTypeDef::getUnitDefs() const
```
jstone-lucasfilm, line 186: With the latest `dataLibrary` argument to `getChildrenOfType`, can this update to `getUnitDefs` now be simplified to use it as well?

I'm imagining something along the lines of the following:

```
vector<UnitDefPtr> UnitTypeDef::getUnitDefs() const
{
    vector<UnitDefPtr> unitDefs;

    for (UnitDefPtr unitDef : getDocument()->getChildrenOfType<UnitDef>(EMPTY_STRING, getDocument()->getDataLibrary()))
    {
        if (unitDef->getUnitType() == _name)
        {
            unitDefs.push_back(unitDef);
        }
    }

    return unitDefs;
}
```

Essentially we would now just be adding the new `dataLibrary` argument to the existing call, and making no other changes to the method.

source/MaterialXCore/Definition.cpp
```diff
@@ -186,13 +186,28 @@ StringVec TargetDef::getMatchingTargets() const
...
 {
     vector<UnitDefPtr> unitDefs;
+    
+    // Gather from data library.
+    if (getDocument()->hasDataLibrary())
+    {
+        for (UnitDefPtr unitDef : getDocument()->getDataLibrary()->getChildrenOfType<UnitDef>())
```
jstone-lucasfilm, line 193: With the latest `dataLibrary` argument to `getChildrenOfType`, can this update to `getUnitDefs` now be simplified to use it as well?

I'm imagining something along the lines of the following:

```
vector<UnitDefPtr> UnitTypeDef::getUnitDefs() const
{
    vector<UnitDefPtr> unitDefs;

    for (UnitDefPtr unitDef : getDocument()->getChildrenOfType<UnitDef>(EMPTY_STRING, getDocument()->getDataLibrary()))
    {
        if (unitDef->getUnitType() == _name)
        {
            unitDefs.push_back(unitDef);
        }
    }

    return unitDefs;
}
```

Essentially we would now just be adding the new `dataLibrary` argument to the existing call, and making no other changes to the method.

source/MaterialXCore/Definition.cpp
```diff
@@ -186,7 +186,7 @@ StringVec TargetDef::getMatchingTargets() const
 vector<UnitDefPtr> UnitTypeDef::getUnitDefs() const
 {
     vector<UnitDefPtr> unitDefs;
-    for (UnitDefPtr unitDef : getDocument()->getChildrenOfType<UnitDef>())
+    for (UnitDefPtr unitDef : getDocument()->getChildrenOfType<UnitDef>(EMPTY_STRING, getDocument()->getDataLibrary()))
```
jstone-lucasfilm, line 189: @ashwinbhat This latest work looks really close to the mark, and my one remaining question is whether we should _always_ include the data library when calling `getChildOfType` or `getChildrenOfType` on a `Document`.  Although there are a few edge cases where we omit the data library for `Document` calls, they actually look like _bugs_ that we'll need to fix in the future, rather than intentional choices in the current proposal.

If my sense is correct, and we actually want to provide the data library universally when calling these methods on a `Document`, then we can greatly simplify this proposed change in the following way:

1. Remove the `dataLibrary` argument from both `getChildOfType` and `getChildrenOfType`.
2. Within each of these methods, check whether the current element is a `Document`.  If we find that it's a `Document`, then the data library referenced by the document is included; otherwise, no data library is included.

I think this would cover all cases that the current proposal does, but in a more universal, simplified, and bulletproof way.  Let me know your thoughts!

jstone-lucasfilm, line 189: @ashwinbhat This latest work looks really close to the mark, and my one remaining question is whether we should always include the data library when calling getChildOfType or getChildrenOfType on a Document. Although there are a few edge cases where we omit the data library for Document calls, they actually look like bugs that we'll need to fix in the future, rather than intentional choices in the current proposal.

If my sense is correct, and we actually want to provide the data library universally when calling these methods on a Document, then we can greatly simplify this proposed change in the following way:

Remove the dataLibrary argument from both getChildOfType and getChildrenOfType.
Within each of these methods, check whether the current element is a Document. If we find that it's a Document, then the data library referenced by the document is included in the search; otherwise, no data library is included.
I think this would cover all cases that the current proposal does, but in a more universal, simplified, and bulletproof way. Let me know your thoughts!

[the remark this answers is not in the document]
ashwinbhat: @jstone-lucasfilm IIRC there is a cyclic dependency between `Element` and `Document` 
both getChildOfType and getChildrenOfType use the passed datalibrary as a ConstElementPtr 
In order to implement what you propose I think we might need to move datalibrary into the Element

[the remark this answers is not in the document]
jstone-lucasfilm: @ashwinbhat Good point, and another option would be to add overrides to the `Document` class for `getChildOfType` and `getChildrenOfType`, and only these overrides would check the associated data library and search through it for matching children.

That might be a simpler approach for readers to understand and maintain, and I believe it should be fine to remove both the Python and JavaScript bindings for `getChildOfType` and `getChildrenOfType` for now, so that we don't need to solve these at the same time as the C++ API.

source/MaterialXCore/Document.h
```diff
@@ -98,6 +121,33 @@ class MX_CORE_API Document : public GraphElement
     /// nodes, and include both Input and Output elements.
     vector<PortElementPtr> getMatchingPorts(const string& nodeName) const;
 
+    /// Return the child element, if any, with the given document and data library
+    template <class T> shared_ptr<T> getChildOfType(const string& name) const
```
jstone-lucasfilm, line 125: This latest version looks excellent, @ashwinbhat!

For consistency, though, wouldn't `getChildOfType` and `getChildrenOfType` need to be virtual methods, with the `Document` class overriding the behavior of its base `Element` class?

In addition to providing callers with more consistent behavior, this would allow the `resolveNameReference` method below to be implemented without a `dataLibrary` argument, and its implementation would require only a single call to `getChildOfType`.

ashwinbhat: @jstone-lucasfilm Just realized, since `getChildOfType` and `getChildrenOfType` are templated, they can't be `virtual`
https://christiandaley.github.io/2024/02/18/virtual-function-templates-with-stateful-metaprogramming-in-c++-20.html

jstone-lucasfilm: Ah, that's a great point!

As an alternative, what if we were to move the implementations for these templated methods into `Element.cpp`, where they would have access to the `Document` class?

That would allow us to use the approach proposed earlier in the thread, where the implementations of `getChildOfType` and `getChildrenOfType` would check whether the calling element is a `Document`, and handle the data library of the `Document` for that case.

That would have the same net effect, where the behavior of `getChildOfType` and `getChildrenOfType` would be consistent, regardless of the type of the calling pointer, and we could remove the last `dataLibrary` argument from `resolveNameReference`.

ashwinbhat: I think moving them to the cpp will require us to explicitly instantiate each type for export. Right now we depend on the compiler and header includes so anyone who has custom derived types are automatically resolved. I think moving it the cpp might cause some additional integrations issues for some teams.
I think `resolveNameReference` is the only outlier, in future we might want to reconsider moving these.

jstone-lucasfilm: That's a good point, @ashwinbhat, though I'm torn about the API churn that would be created by adding a `dataLibrary` argument to `resolveNameReference`, and then very likely returning to our original API in a subsequent change.

When I'm working on significant changes like your current PR, one approach that I'll often use is to implement any required refactoring steps in a _separate branch_, where it's straightforward to validate that the refactor has no unexpected impact on other workflows.

Using that approach, either you or I could propose the move of `getChildOfType` and `getChildrenOfType` into `Element.cpp`, independent of your current work, and we could validate that this step is itself a safe one to make.

I'd be happy to help with this if it's too time-consuming to take such a step on your side, and if it works out well, we would merge it into your current PR, allowing it to be merged without any associated API churn.

ashwinbhat: @jstone-lucasfilm any thoughts about add an API to document to resolve name reference? e.g. `resolveNameFromDataLibrary` 

Which would essentially do something this:
```
ElementPtr resolveNameFromDataLibrary(const string& name, ConstElementPtr parent = nullptr) const
{
    ElementPtr child = hasDataLibrary() ? getDataLibrary()->getChild(name) : nullptr;
    return child;
}
```

jstone-lucasfilm: That's a clever idea!  I can see that handling the case of `getChildOfType` very elegantly, but it's not clear how it would handle the case of `getChildrenOfType`, which needs to resolve more than a single name.

How are you imagining it would handle that second case?

ashwinbhat: I was thinking about how to remove the use of datalibrary in `resolveNameReference` thus keeping all datalibrary usage in Document. I think the main caller is  `Node::getNodeDef` that needs to either `resolveNameReference` if `nodedef` attribute is in use otherwise falls to use `Document::getMatchingNodeDefs` 
https://github.com/AcademySoftwareFoundation/MaterialX/blob/b058d9cbba16d06ed858c53f39f2af7b9e41b550/source/MaterialXCore/Node.cpp#L71
We could introduce `Document::getMatchingNodeDef` as counterpart to `Document::getMatchingNodeDefs` 
Since `resolveNameReference` is not a public `Node::getNodeDef` could continue to use that, but use  `Document::getMatchingNodeDef` to resolve node definition from datalibrary.

jstone-lucasfilm: That sounds very reasonable, and let's see how things go using that approach!

I'm still open to moving `getChildOfType` and `getChildrenOfType` to `Element.cpp` if that's a useful adjunct to your change, and just let me know if that step might be needed here.

jstone-lucasfilm, line 125: @ashwinbhat I've had a chance to implement the move of `getChildOfType` and `getChildrenOfType` from Element.h to `Element.cpp`, which would allow these two non-virtual overloads to be removed:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/2069

Although non-virtual overloaded methods are allowed in C++, we strongly avoid this pattern in MaterialX because of the hazards for unwary developers, and static analysis packages like PVS-Studio emit warnings when non-virtual overloads are detected in the code!

What are your thoughts?  If I went ahead and merged my change linked above, would you be open to updating your change to take advantage of it?

ashwinbhat: @jstone-lucasfilm this is great. Please go ahead and merge, I'll update my PR to take advantage of this.

jstone-lucasfilm: @ashwinbhat Done!

source/MaterialXCore/Node.cpp
```diff
@@ -72,12 +72,30 @@ NodeDefPtr Node::getNodeDef(const string& target, bool allowRoughMatch) const
 {
     if (hasNodeDefString())
     {
-        return resolveNameReference<NodeDef>(getNodeDefString());
```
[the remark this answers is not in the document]
ashwinbhat: Yes, It I've pushed an update. Thanks you @jstone-lucasfilm for the updates and refinements.

source/MaterialXCore/Node.cpp
```diff
@@ -76,8 +76,18 @@ NodeDefPtr Node::getNodeDef(const string& target, bool allowRoughMatch) const
     }
     vector<NodeDefPtr> nodeDefs = getDocument()->getMatchingNodeDefs(getQualifiedName(getCategory()));
     vector<NodeDefPtr> secondary = getDocument()->getMatchingNodeDefs(getCategory());
-    vector<NodeDefPtr> roughMatches;
     nodeDefs.insert(nodeDefs.end(), secondary.begin(), secondary.end());
+
+    // Recurse to data library if present.
+    if (getDocument()->hasDataLibrary())
```
jstone-lucasfilm, line 82: Is this additional code block still needed here, or will it now be handled by the data library checks within the two `getMatchingNodeDefs` calls above?

ashwinbhat: Good catch, updated them.

jstone-lucasfilm (approved):
Thanks for this great work, @ashwinbhat, and I think this looks ready to merge!

Issue #2056, opened 2024-10-09
chiang_hair_bsdf node compile error on MacOS


ashwinbhat:
The MSL shaders generated for chiang_hair_bsdf node produce compilation errors on MacOS.

Steps: 
build MaterialX viewer on MacOS
load simple_hair_default.mtlx 

Errors:
program_source:1364:28: error: 'angles' declared as array of references of type 'float2 &'
        thread vec2& angles[4]
                           ^
program_source:1383:62: error: 'Ap' declared as array of references of type 'float3 &'
    void mx_hair_attenuation(float f, vec3 T, thread vec3& Ap[4])  // Ap
                                                             ^
program_source:1448:24: error: unexpected type name 'float3': expected expression
        vec3 tint[4] = vec3[](tint_R, tint_TT, tint_TRT, tint_TRT);
                       ^
program_source:7:14: note: expanded from macro 'vec3'
#define vec3 float3
             ^
program_source:1448:29: error: expected expression
        vec3 tint[4] = vec3[](tint_R, tint_TT, tint_TRT, tint_TRT);
                            ^
program_source:1453:22: error: unexpected type name 'float2': expected expression
        vec2 vs[4] = vec2[](roughness_R, roughness_TT, roughness_TRT, roughness_TRT);
                     ^
program_source:6:14: note: expanded from macro 'vec2'
#define vec2 float2
             ^
program_source:1453:27: error: expected expression
        vec2 vs[4] = vec2[](roughness_R, roughness_TT, roughness_TRT, roughness_TRT);
                          ^
program_source:1541:14: warning: unused variable 'dirAlbedo' [-Wunused-variable]
        vec3 dirAlbedo = mx_ggx_dir_albedo(NdotV, avgAlpha, F0, 1.0) * comp;
             ^
program_source:1565:20: warning: unused variable 'N' [-Wunused-variable]
            float3 N = normalize(vd.normalWorld);
                   ^

Shader source attached:
[simplehair.msl.txt](https://github.com/user-attachments/files/17319308/simplehair.msl.txt)

It appears to be related to two issues related to MSL and needs further investigation
1. Arrays need a size.
2. Arrays can't be passed by reference (?)

ld-kerley:
Poking around a little, I think this boils down to two different situations where the GLSL syntax and MSL syntax are incompatible.

### Passing arrays to functions by reference
GLSL
```
[...]

PR #2057, opened 2024-10-10
Fix Chiang hair model for Metal Shading language

ld-kerley:
The current approach for converting GLSL code to MSL doesn't work for the recent Chiang hair PR #1968 from @masuosuzuki. 

The details of the conversion failures are outlines in #2056 reported by @ashwinbhat.

I'm just putting up this PR as one option for how we could resolve this.


jstone-lucasfilm:
This looks good to me, @ld-kerley, and if local render test suites now pass on MacOS, I'd say that we should move forward with this approach.

As a follow-up, we might consider adding a Chiang Hair BSDF example to our MSL shader validation tests on GitHub Actions:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/.github/workflows/main.yml#L234

jstone-lucasfilm (approved):

PR #2058, opened 2024-10-11
Simplify mx_math files

jstone-lucasfilm:
This changelist simplifies mx_math.glsl and mx_math.metal to reduce code repetition and clarify the differences between dedicated helper functions and simple keyword replacement.

Additionally it addresses an omitted update from atan to mx_atan in the implementation of chiang_hair_bsdf, and adds an update step to Linux CI for robustness.


jstone-lucasfilm:
I'm definitely interested in more discussion on this general approach, but we should likely move forward with this initial change for now, in order to allow future work to build upon it.

PR #2068, opened 2024-10-14
Reduce duplication in MSL implementations

jstone-lucasfilm:
This changelist removes a number of MSL implementations that duplicate their counterparts in GLSL, instead using the inheritance relationship between these two shader generators to handle code sharing.


jstone-lucasfilm:
This change isn't likely to be controversial, but I'd be interested in thoughts from @ld-kerley and @dgovil!

It's also worthwhile to highlight the new validation tests for MSL shader generation in the GitHub CI for MaterialX, which is expressly designed to improve coverage for changes like the current PR:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/.github/workflows/main.yml#L239

PR #2069, opened 2024-10-14
Move templated accessors to Element.cpp

jstone-lucasfilm:
This changelist moves the templated accessors `getChildOfType` and `getChildrenOfType` from Element.h to Element.cpp, allowing upcoming features to be implemented in a more straightforward fashion.


ashwinbhat:
@jstone-lucasfilm  This is great. It really simplifies the template code

PR #2070, opened 2024-10-15
Harmonize MSL and GLSL implementations

jstone-lucasfilm:
This changelist addresses a handful of remaining cases where our MSL and GLSL code diverged, harmonizing on shading code that works in both contexts.


libraries/stdlib/genglsl/mx_normalmap.glsl
```diff
@@ -1,9 +1,9 @@
 void mx_normalmap_vector2(vec3 value, vec2 normal_scale, vec3 N, vec3 T, vec3 B, out vec3 result)
 {
-    value = (value == vec3(0.0)) ? vec3(0.0, 0.0, 1.0) : value * 2.0 - 1.0;
-
-    value = T * value.x * normal_scale.x + B * value.y * normal_scale.y + N * value.z;
-
+    value = (dot(value, value) == 0.0) ? vec3(0.0, 0.0, 1.0) : value * 2.0 - 1.0;
```
ld-kerley, line 3: Curious why you chose to use `dot(value, value)==0` here instead of just comparing each component.  Wouldn't the dot product here be more expensive?  And I think that comparing each component here would be clearer to understand. 

I also wonder if this might even be simplified even more to something like this... (or similar)
```
if (value.x != 0 || value.y != 0 || value.z != 0) {
  value =  value * 2.0 - 1.0;
  value = T * value.x * normal_scale.x +
            B * value.y * normal_scale.y +
            N * value.z;
    result = normalize(value);
} else {
  result = normalize(N);
}
```
This also nicely shows that if a zero valued input is passed we just return the normalized N.

libraries/stdlib/genmsl/stdlib_genmsl_impl.mtlx
```diff
@@ -41,10 +41,6 @@
     <input name="default" type="vector4" implname="default_value" />
   </implementation>
 
-  <!-- <normalmap> -->
```
ld-kerley, line 44: Instead of removing these implementations, don't they just need to be redirected to the GLSL implementation? 

Otherwise for a `genmsl` target - how does it find the correct implementation at shader generation time?

ld-kerley (commented):
I think this looks good to me - just left a couple of questions below.

libraries/stdlib/genmsl/stdlib_genmsl_impl.mtlx
jstone-lucasfilm: Our MSL shader generator already inherits from the GLSL shader generator, so this is handled automatically!

ld-kerley: I didn't realize this... maybe that means there are other lines in this file that aren't necessary - as they are just referencing GLSL implementations.

jstone-lucasfilm: I've been working through much of the unneeded MSL/GLSL duplication this week, though additional improvements can likely still be found!

For reference, here's where the inheritance of genmsl from genglsl is declared:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/targets/genmsl.mtlx

PR #2072, opened 2024-10-16
Add support `closest` texture lookup mode in MDL backend

msuzuki-nvidia:
Add support `closest` texture lookup mode in MDL backend.

This change affects in all type of `<image>` nodes.


jstone-lucasfilm (approved):
This looks good to me, thanks @msuzuki-nvidia!

PR #2076, opened 2024-10-18
Add defaultgeomprop support for nodegraph inputs

feldstj:
This PR makes it so that defaultgeomprop values can be used with nodegraph inputs as opposed to solely used on nodedef inputs. Includes a test file that is a nodegraph version of a checkerboard nodegraph implementation. The nodegraph exposes a defaultgeomprop value on the nodegraph's texcoord input of "UV0". Attached is what the rendered test result looks like. Without the fix it would be black.
defaultgeomprop_glsl copy


jstone-lucasfilm (approved):
This looks good to me, thanks @feldstj!

dgovil:
Hey folks, just catching up on these great PRs but I was wondering if there was an expectation of how this would work with OpenUSD?
I don't believe there's a concept for this in USD implicitly, so would usdMtlx require updating?

JGamache-autodesk:
Hi @dgovil, this information is not converted by usdMtlx. Even if we were to add it as metadata on the input of the USD NodeGraph, it would not survive the translation to HdMaterial, which means it could not be restored when generating the MaterialX document back. This is indeed a hard problem.

dgovil:
Thanks @JGamache-autodesk ! That's good info to have so we can start maintaining a list of what translates over.

PR #2077, opened 2024-10-18
CMake updates for Apple builds

jstone-lucasfilm:
- Mark the CMAKE_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY flag as Apple-specific, removing it from CMake interfaces on other platforms.
- Simplify comment blocks to highlight key points.
- Reorder the order of option declarations for clarity.


jstone-lucasfilm:
Although this is a minor update, I'm CC'ing @dgovil just to make sure I didn't miss an edge case that our GitHub CI doesn't catch.

dgovil:
I think this change should be fine.

Issue #2044, opened 2024-09-30
UsdPreviewSurface `opacityThreshold` implementation does not match the specification


ld-kerley:
The specification for `UsdPreviewSurface` states the following...

> * opacity - float - 1.0
When opacity is 1.0 then the gprim is fully opaque, if it is smaller than 1.0 then the prim is translucent, when it is 0 the gprim is transparent. Note that even a fully transparent object still receives lighting as, for example, perfectly clear glass still has a specular response.
> * opacityThreshold - float - 0.0
The opacityThreshold input is useful for creating geometric cut-outs based on the opacity input. A value of 0.0 indicates that no masking is applied to the opacity input, while a value greater than 0.0 indicates that rendering of the surface is limited to the areas where the opacity is greater or equal to that value. A classic use of opacityThreshold is to create a leaf from an opacity input texture, in that case the threshold determines the parts of the opacity texture that will be fully transparent and not receive lighting. Note that when opacityThreshold is greater than zero, the opacity values less than the opacityThreshold will not be rendered, and the opacity values greater than or equal to the opacityThreshold will be fully visible. Thus, the opacityThreshold serves as a switch for how the opacity input is interpreted; this “translucent or masked” behavior is common in engines and renderers, and makes the UsdPreviewSurface easier to interchange. It does imply, however, that it is not possible to faithfully recreate a glassy/translucent material that also provides an opacity-based mask… so no single-polygon glass leaves.

Currently the MaterialX `UsdPreviewSurface` implementation always blends the `diffuse_bsdf` and `transmission_bsdf` nodes based on the value of `opacity`.

```
    <mix name="transmission_mix" type="BSDF">
      <input name="fg" type="BSDF" nodename="diffuse_bsdf" />
      <input name="bg" type="BSDF" nodename="transmission_bsdf" />
      <input name="mix" type="float" interfacename="opacity" />
    </mix>
```

Which is incorrect.  If `opacityThreshold > 0`, then the material should be in "masking" mode and only cut-out the surface, not blend through it. 

I believe the fix would look something like this....

```
    <ifgreater name="transmission_mix_amount" type="float">
      <input name="value1" type="float" interfacename="opacityThreshold" />
      <input name="value2" type="float" value="0" />
      <input name="in1" type="float" value="1" />
      <input name="in2" type="float" interfacename="opacity" />
    </ifgreater>
    <mix name="transmission_mix" type="BSDF">
[...]

PR #2081, opened 2024-10-21
Update UsdPreviewSurface Nodegraph

klucknav:
This PR fixes the cutout issue described in gitIssue #2044. It also updates the nodegraph to reflect the recently updated PreviewSurface v2.6 which removes specular highlights for fully transparent materials.

### DIFF
--- a/libraries/bxdf/usd_preview_surface.mtlx
+++ b/libraries/bxdf/usd_preview_surface.mtlx
@@ -6,7 +6,7 @@
   <!-- ======================================================================== -->
 
   <!-- Node: UsdPreviewSurface -->
-  <nodedef name="ND_UsdPreviewSurface_surfaceshader" node="UsdPreviewSurface" nodegroup="pbr" doc="USD preview surface shader" version="2.3" isdefaultversion="true">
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm (approved):
This looks good to me, @klucknav, and thanks for the updates!

ld-kerley:
Do we need to retain the previous version as `version=2.3` and add this as a `version=2.6`?  

I think I recall @crydalch mentioning that he's seen people using `UsdPreviewSurface` for rendering glass by setting `opacity=0` and relying on the specular response remaining.

I'm in favor of this change in the `UsdPreviewSurface` spec, as it aligns its behavior with other shading models, but I wonder if MaterialX consumers might get caught out by the change, and retaining the older version would allow people to ensure an older look would still be achievable?

jstone-lucasfilm:
@ld-kerley Karen and I were just discussing this, but in my local testing, glass materials with zero opacity (i.e. maximum transmissive strength) render in the same fashion as before.  That seems consistent with other industry shading models such as glTF PBR and Standard Surface, but if it's not the intent for UsdPreviewSurface, then additional updates may be needed.

PR #2085, opened 2024-10-24
Fix hair nodes in MDL backend

msuzuki-nvidia:
Fix the issue where the generated MDL code with hair nodes does not compile.

Also address warnings for unused parameters.


jstone-lucasfilm (approved):
Thanks for these fixes, @msuzuki-nvidia!

PR #2091, opened 2024-10-29
Reduce verbosity of unit tests

jstone-lucasfilm:
This changelist reduces the verbosity of our unit tests for element equivalence, focusing on REQUIRE statements to provide coverage.

### DIFF
--- a/source/MaterialXTest/MaterialXCore/Document.cpp
+++ b/source/MaterialXTest/MaterialXCore/Document.cpp
@@ -10,8 +10,6 @@
 #include <MaterialXFormat/Util.h>
 #include <MaterialXFormat/XmlIo.h>
 
-#include <iostream>
-
 namespace mx = MaterialX;
 
 TEST_CASE("Document", "[document]")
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
I like the idea of adding a verbosity mode in the future, and this quick fix was focused on unblocking our render test GLSL/OSL comparisons, where the unusually verbose output from these unit tests was hiding real warnings in the test suite.

ld-kerley:
maybe its at least worth re-instating the code and just commenting it out for now - and adding a TODO to add a verbosity flag or build time option.  I don't want to forget this good work exists - git history is not the easiest to discover old code in.

jstone-lucasfilm:
@ld-kerley In this case, my sense is that we should have omitted this debug output in the original PR, since it's streaming verbose text to our standard output stream, which is not an approach that we ever use in our unit tests.  That's fully on me, because I didn't catch this in my original testing of the PR, but had I done so I would have recommended that we omit this logic entirely.

PR #2092, opened 2024-10-29
Add support for geompropvalueuniform node

crydalch:
Adds a string-/filename-friendly primvar/geompropvalue node, as was proposed for 1.39.


crydalch:
This is till a WIP PR of course, but a quick summary of the changes in-progress:

- Adds `geompropvalueuniform`
- Removed string signature from `geompropvalue`
- Adds uniform string/filename signatures to `UsdPrimvarReader`

Thanks in advance for feedback, and especially any help with the codegen and/or various shading targets.

libraries/stdlib/genmsl/stdlib_genmsl_impl.mtlx
```diff
@@ -77,14 +77,17 @@
...
   <implementation name="IM_geompropvalue_float_genmsl" nodedef="ND_geompropvalue_float" function="mx_geompropvalue_float" target="genmsl" />
   <implementation name="IM_geompropvalue_color3_genmsl" nodedef="ND_geompropvalue_color3" function="mx_geompropvalue_color" target="genmsl" />
   <implementation name="IM_geompropvalue_color4_genmsl" nodedef="ND_geompropvalue_color4" function="mx_geompropvalue_color4" target="genmsl" />
   <implementation name="IM_geompropvalue_vector2_genmsl" nodedef="ND_geompropvalue_vector2" function="mx_geompropvalue_vector2" target="genmsl" />
   <implementation name="IM_geompropvalue_vector3_genmsl" nodedef="ND_geompropvalue_vector3" function="mx_geompropvalue_vector" target="genmsl" />
   <implementation name="IM_geompropvalue_vector4_genmsl" nodedef="ND_geompropvalue_vector4" function="mx_geompropvalue_vector4" target="genmsl" />
 
+  <!-- <geompropvalueuniform> -->
+  <implementation name="IM_geompropvalue_string_genmsl" nodedef="ND_geompropvalueuniform_string" function="mx_geompropvalueuniform_string" target="genmsl" />
```
ld-kerley, line 88: I think the intention here is to use the existing `mx_geompropvalue_string` function, as you're doing below for the "filename" case?  Otherwise I'm not sure where `mx_geompropvalueuniform_string` is defined.

libraries/bxdf/usd_preview_surface.mtlx
```diff
@@ -363,7 +368,14 @@
       <input name="geomprop" type="string" interfacename="varname" />
       <input name="default" type="string" interfacename="fallback" />
     </geompropvalue>
-    <output name="out" type="string" nodename="primvar" />
+    <output name="out" type="string" nodename="primvar" uniform="true" />
+  </nodegraph>
+  <nodegraph name="IMP_UsdPrimvarReader_filename" nodedef="ND_UsdPrimvarReader_filename">
+    <geompropvalue name="primvar" type="string">
```
ld-kerley, line 374: I think because we're introducing an entirely new node `geompropvalueuniform` to handle strings - I think this node XML tag needs to match that new name - `geompropvalueuniform`. 

I also wonder if there are other cases elsewhere in the test suite where the existing `geompropvalue` node is being used for a `string` or `filename`, that would also need to be updated.

crydalch:
Fixed a few more mistakes/issues (thanks @ld-kerley for looking it over!), down to 8 failures (on my end at least).

crydalch:
geompropvalueuniform_blank

Okay CI issues are resolved, so build is "clean", but no luck getting it to work so far, in MaterialXViewer.

Maybe some MacOS-specific quirk or something else causing it to fail? I'm not seeing any messages/shell output :(

```<?xml version="1.0"?>
<materialx version="1.39">
  <surfacematerial name="mtlxsurfacematerial1" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="mtlxstandard_surface" />
  </surfacematerial>
  <standard_surface name="mtlxstandard_surface" type="surfaceshader">
    <input name="base_color" type="color3" nodename="mtlximage1" />
  </standard_surface>
  <image name="mtlximage1" type="color3">
    <input name="default" type="color3" value="0, 1, 0" />
    <input name="file" type="filename" nodename="mtlxgeompropvalue1" />
  </image>
  <geompropvalueuniform name="mtlxgeompropvalue1" type="filename">
    <input name="geomprop" type="string" value="my_string_primvar" />
    <input name="default" type="filename" value="grid.jpg" />
  </geompropvalueuniform>
</materialx>
```

grid

crydalch:
Was given tip to use MaterialXGraphEditor (which uses GLSL even on MacOS), and it's showing this error:

```Error in compiling fragment shader:
ERROR: 0:2590: Samplers not allowed except in non-buffer uniforms and in-parameters
ERROR: 0:2594: Use of undeclared identifier 'geompropvalueuniform_filename_out'
```

I suspect the second item is because the property isn't authored on the generic shaderball geometry (it's not clear if/how that's even possible); so that might just be a red herring.

krohmerNV:
Since @jstone-lucasfilm asked for more discussions ;)
I'm wondering if filenames in primvars is the only use case here. I'm no fan of such a feature. I know, it's very common in some offline rendering workflows, but it does not apply to real-time rendering and it hides important material properties elsewhere in the scene. If you see other useful application, I would be interested.

crydalch:
Thanks @krohmerNV! Yes filenames are the main use case, it's been a long-standing request for path tracing workflows; the primvars setting the texture map paths are authored in the scene, but I get that real-time renderers don't like or use it (or don't use it much, at least). There could be other cases for uniform strings, but those would just be hypotheticals at this point.

FWIW, the current `geompropvalue` has strings, which also don't work (hence this PR would also remove those signatures from `geompropvalue`). We included strings and filenames just so they have a home together in `geompropvalueuniform`; string properties couldn't work anyway, because all string/filename inputs are uniform anyway.

krohmerNV:
As I said, I know that these workflows exist. If they are good and you want to use them you need to decide on your own. ;)

But aside from that, having uniform geometry properties seems to be valuable. It might help constant folding even for real-time renderers.

jstone-lucasfilm:
This looks very reasonable to me, thanks @crydalch, and I'm CC'ing @ld-kerley and @kwokcb for their thoughts.

jstone-lucasfilm (approved):
This change looks good to me, thanks @crydalch!

Issue #2093, opened 2024-10-30
Proposal: Add hextiling nodes


msuzuki-nvidia:
# MaterialX Hextiling Node Proposal

## Summary

Hextiling provides tiling a texture image without an obvious repeating pattern. This document proposes two new MaterialX hextiling nodes based on the article "Practical Real-Time Hex-Tiling" (Morten S. et al, 2022) \[1\].

* `<hextiledimage>`
* `<hextilednormalmap>`

The abstract of the article is as follows:

> To provide a convenient, easy-to-adopt approach to randomly tiled textures in the context of real-time graphics, we propose an adaptation of the by-example noise algorithm of Heitz and Neyret. The original method preserves contrast using a histogram-preserving method that requires a precomputation step to convert the source texture into a transform and inverse transform texture, which must both be sampled in the shader rather than the original source texture. Thus deep integration into the application is required for this to appear opaque to the author of the shader and material. In our adaptation we omit histogram preservation and replace it with a novel blending method that allows us to sample the original source texture. This omission is particularly sensible for a normal map as it represents the partial derivatives of a height map. In order to diffuse the transition between hex tiles, we introduce a simple metric to adjust the blending weights. For a texture of color, we reduce loss of contrast by applying a contrast function directly to the blending weights. Though our method works for color, we emphasize the use case of normal maps in our work because non-repetitive noise is ideal for mimicking surface detail by perturbing normals.

Unlike standard image tiling, hextiling involves image blending. As described in the abstract above, simple image blending is not suitable for normal maps; therefore, separate nodes for color images and normal maps are necessary.

### `<hextiledimage>`

| Parameter            |   Type   |  Range   |     Default     | Description                                   |
| -------------------- | :------: | :------: | :-------------: | --------------------------------------------- |
| **file**             | filename |          |                 | Image file path                               |
| **default**          |  color3  |          | (0.0, 0.0, 0.0) | Default color                                 |
| **texcoord**         | vector2  |          |      `UV0`      | Texture coordinate                            |
| **tiling**           | vector2  |          |   (1.0, 1.0)    | Tile repeat                                   |
| **rotation**         |  float   | \[0, 1\] |       1.0       | Rotation randomness                           |
| **rotation_range**   | vector2  |          |  (0.0, 360.0)   | Range of values used for randomizing rotation |
| **scale**            |  float   | \[0, 1\] |       1.0       | Scale randomness                              |
[...]

PR #2094, opened 2024-10-30
Add GLSL hextiling node implementation

msuzuki-nvidia:
This PR adds GLSL implementation for hextiling nodes proposed by #2093.

Nodes to add:
- `<hextiledimage>` (`color3` and `color4`)
- `<hextilednormalmap>` (`vector3`)


crydalch:
We had some of our folks take a look internally the implementation looks great. Thank you @msuzuki-nvidia!

Are there any outstanding issues, or is this ready enough to consider for 1.39.2 perhaps?

msuzuki-nvidia:
@crydalch Thank you for your review!

I would leave the timing of the merge and release to @jstone-lucasfilm

libraries/stdlib/genglsl/mx_hextiledimage.glsl
```diff
@@ -0,0 +1,102 @@
...
+    vec2 tiling,
+    float rotation,
+    vec2 rotation_range,
+    float scale,
+    vec2 scale_range,
+    float offset,
+    vec2 offset_range,
+    float falloff,
+    float falloff_contrast,
+    vec3 lumacoeffs,
```
jstone-lucasfilm, line 20: @msuzuki-nvidia Since we're planning to refactor our `luminance` node in a future MaterialX version, harmonizing it with our approach to color space management, we should probably avoid having an explicit `lumacoeffs` input to this new node, so that we don't create unnecessary churn in the node signature.

For now, I would recommend replacing this input with a `LUMA_COEFFS` constant within the implementation of the node, and we can add support for the new `luminance` approach once it has been proposed and implemented in MaterialX.

msuzuki-nvidia: I think we shouldn't make the lumacoeffs constant for now so that it can be set with a color space where it's consumed until the refactoring is completed. I'm happy to update this code when we start the refactoring.

jstone-lucasfilm: Ah, I see your point, @msuzuki-nvidia.  Ok, let's leave this alone for now, and I'll continue reading through the code with this in mind.

jstone-lucasfilm (approved):
This change looks good to me, thanks for the great work, @msuzuki-nvidia.

Two key next steps will be implementing support for these nodes in OSL and MDL, but those improvements make the most sense in separate pull requests.

PR #2095, opened 2024-10-30
Simplify the MetalTexture abstraction

ld-kerley:
`mx_texture.metal` is used to provide an abstraction to the texture code when being called from GLSL code.

There are a few functions in the interface of this abstraction that do not appear to be used anywhere.

This PR proposes removing them, as a precursor to other upcoming texture work.  I'm proposing this as a separate PR to make things easier to review.

All of the MSL unit tests pass.


msuzuki-nvidia:
Could you add `textureGrad()` support as well? It is used in hextiling code but  found that it isn't supported in MSL backend.

ld-kerley:
@msuzuki-nvidia - I pulled your branch down - there are a few other changes that are necessary to get it working for MSL, so its probably better to integrate all of those changes in to your PR - I'll ping you on slack to coordinate.

jstone-lucasfilm (approved):

PR #2099, opened 2024-10-30
Removing MDL struct test function

ld-kerley:
This function was incorrectly included  in #1831.


jstone-lucasfilm (approved):
Looks good to me, thanks @ld-kerley!

PR #2101, opened 2024-10-31
Improvements to the MaterialX Viewer

jstone-lucasfilm:
- Replace raw pointers with NanoGUI references to improve application robustness.
- Allow UI names to be displayed without a corresponding UI folder.

### DIFF
--- a/source/MaterialXView/Editor.cpp
+++ b/source/MaterialXView/Editor.cpp
@@ -18,11 +18,11 @@ namespace
 class EditorColorPicker : public ng::ColorPicker
 {
   public:
-    EditorColorPicker(ng::Widget* parent, const ng::Color& color) :
+    EditorColorPicker(ng::ref<ng::Widget> parent, const ng::Color& color) :
         ng::ColorPicker(parent, color)
     {
-        ng::Popup* popup = this->popup();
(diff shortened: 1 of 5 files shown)


source/MaterialXView/Editor.cpp
```diff
@@ -18,11 +18,11 @@ namespace
...
   public:
-    EditorColorPicker(ng::Widget* parent, const ng::Color& color) :
+    EditorColorPicker(ng::ref<ng::Widget> parent, const ng::Color& color) :
         ng::ColorPicker(parent, color)
     {
-        ng::Popup* popup = this->popup();
-        ng::Widget* floatGroup = new ng::Widget(popup);
+        ng::ref<ng::Popup> popup = this->popup();
+        ng::ref<ng::Widget> floatGroup = new ng::Widget(popup);
         auto layout = new ng::GridLayout(ng::Orientation::Horizontal, 2,
```
ld-kerley, line 26: I think this `auto` here is going to be a raw pointer too - maybe we should change it as well for consistency?

jstone-lucasfilm: I noticed that one, too, and I think you're right that we should clean up these auto pointers in a subsequent change.

PR #2104, opened 2024-11-02
Extend Chiang tests across languages

jstone-lucasfilm:
This changelist extends testing of Chiang Hair BSDF nodes across shading languages, allowing render comparisons to include the complete Examples and TestSuite folders.

In order to avoid shader compilation errors, placeholder OSL implementations of these nodes have been added, and these should be replaced with accurate implementations once they are supported.


jstone-lucasfilm:
Here's a new GLSL/OSL comparison with the changes in this PR:

[MaterialXRenderTests_11_02_2024.pdf](https://github.com/user-attachments/files/17608480/MaterialXRenderTests_11_02_2024.pdf)

jstone-lucasfilm:
@ld-kerley That's the approach we've taken with previous MaterialX nodes, where we allow our render test suite to run without failures, even when individual features aren't yet implemented in every language that we're testing (e.g. thin-film interference in OSL, surface displacement in GLSL, and so forth).

It's worth discussing whether we can propose a better approach going forward, but this is the approach that we've used historically when developing new MaterialX nodes, where implementations across languages are usually staggered in time.

PR #2105, opened 2024-11-05
Consider subsurface color in dielectric albedo

JGamache-autodesk:
I am experimenting with OpenPBR to UsdPreviewSurface export code in Maya and tried chaining the two translator graphs already provided in MaterialX. Results are quite good, except when the subsurface color is extremely different from the base color.

This can be fixed by computing a dielectric base albedo that mixes the scaled base color and the subsurface color. With this change in place, the results are quite interesting:
ConverterTestRig
Without this fix, the UsdPreviewSurface sphere would be bright orange instead of purple.
The test rig is very simple:
[ConverterTestRig.zip](https://github.com/user-attachments/files/17636035/ConverterTestRig.zip)


jstone-lucasfilm (approved):
Thanks for this improvement, @JGamache-autodesk!

PR #2110, opened 2024-11-11
Skip comments in functional equivalence tests

kwokcb:
## Change

- Skip `<comment>` blocks when checking for functional equivalency as they should not affect the comparison.
- Add "comment" category string for reuse.

## Validation

- Update equivalency test to add in comments for documents to compare.


ld-kerley (approved):
Looks good to me - thanks @kwokcb

source/MaterialXCore/Types.cpp
```diff
@@ -23,6 +23,7 @@ const string SURFACE_MATERIAL_NODE_STRING = "surfacematerial";
 const string VOLUME_MATERIAL_NODE_STRING = "volumematerial";
 const string MULTI_OUTPUT_TYPE_STRING = "multioutput";
 const string NONE_TYPE_STRING = "none";
+const string COMMENT_TYPE_STRING = "comment";
```
jstone-lucasfilm, line 26: Since this string represents a _node category_ rather than a data type, maybe the name of this new string constant should be `COMMENT_ELEMENT_CATEGORY`?

Alternatively, maybe we could just use the existing string constant `CommentElement::CATEGORY`?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Element.h#L1199

jstone-lucasfilm (commented):
This looks very promising, thanks @kwokcb, though I had a question about the name of the new string constant.

source/MaterialXCore/Types.cpp
kwokcb: I thought that CommentElement::CATEGORY was not accessible for some reason. I've removed the new string.

jstone-lucasfilm (approved):

PR #2116, opened 2024-11-15
Add API to allow input reordering on node graphs

kwokcb:
## Workflow Motivation

In order for users to specify interface ordering for node graphs they need to re-order the storage order of child inputs. Additionally, when creating new definitions (`nodedef`) they may which to specify the ordering of child inputs as part of definition publishing.

## Proposal

Provide a simple atomic API to allow for inputs on node graphs to be re-ordered without affecting non-input child ordering.
The proposed API name is: `NodeGraph::setInputOrdering()` where a list all input names specifies the desired ordering.

### DIFF
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -700,6 +700,41 @@ void NodeGraph::modifyInterfaceName(const string& inputPath, const string& inter
     }
 }
 
+void NodeGraph::setInputOrdering(const StringVec& inputNames)
+{
+    vector<InputPtr> inputs;
+    StringSet childInputNames;
+    std::vector<unsigned int> childInputIndicies;
+        
+    for (ElementPtr child : _childOrder)
+    {
+        InputPtr input = child->asA<Input>();
+        if (!input)
+            continue;
+
+        inputs.push_back(input);
+        const string& inputName = input->getName();
+        childInputNames.insert(inputName);
+        childInputIndicies.push_back(getChildIndex(inputName));
+    }
+
+    // Check that all names exist in the inputs
+    StringSet orderedNameSet(inputNames.begin(), inputNames.end());
+    if (childInputNames != orderedNameSet)
+    {
+        throw Exception("The set of names provided does not match the existing set of input names");
+    }
+
+    // Use the existing input child indices to set the locations for the
+    // named list. Pre-sort to add from lowest index to highest
+    std::sort(childInputIndicies.begin(), childInputIndicies.end());
+    for (size_t i = 0; i < childInputIndicies.size(); i++)
+    {
+        unsigned int newLocation = childInputIndicies[i];
+        setChildIndex(inputNames[i], newLocation);
+    }
+}
+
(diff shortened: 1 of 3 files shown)


meshula:
Would this need some work on the usd side to capture the ordering somewhere? Perhaps UsdUI or something?

kwokcb:
Hi @meshula, 
This doesn't add any additional meta-data to specify the UI ordering. It's just a way to re-order inputs on a nodegraph. 

Does USD already have separate for UI ordering (e.g. is this this stacking order in [UsdUITokenType](https://openusd.org/dev/api/struct_usd_u_i_tokens_type.html) ?
I'm guessing from your comment it does not, but If it does then it might be worthwhile to look at the equivalent support for MaterialX as a related but separate change.

kwokcb:
Thanks for the explanation about stacking order. 

To confirm your summary of `childOrder`:
- Yes this is construction order but can serve double duty as "interface order" for inputs (not nodes).  
- This user ordering is optional for a compound graphs but from the various DCCs I've seen it's pretty well what folks are using for the UI. I've seen that Maya additionally allows for input re-ordering to modify user ordering.
- When it comes to creating a new definitions from such a graph the input ordering is taken as a definition's input ordering so becomes "fixed" at that time -- i.e. if you change the ordering the definition is actually a different one.

Thus this utility to change construction ordering is being proposed to help support both the above workflows for user ordering.

---

That all said there is nothing required on the usd side. If at some point in the future definitions or graphs have additional meta-data to specify input ordering than this can be brought up.

jstone-lucasfilm:
@kwokcb Rather than making this a core method of the `NodeGraph` class in MaterialXCore, perhaps this would be better as a MaterialX Python API example, showing developers how they can build upon existing API methods such as `Element::setChildIndex`?

ld-kerley:
@jstone-lucasfilm - Do you believe that this operation would only be used by people authoring python code?  It seems like there might be use cases outside of python, and providing it as part of the C++ API would allow everyone to benefit from a common implementation.   

Maybe there are other reasons other than using it as a python API example that you think it shouldn't be part of the C++ API?

ld-kerley:
I guess I think things like input validation are exactly the sort of things it's good to share as part of the library, instead of leaving it to the users to implement themselves. 

If the hesitation is to increase the API surface of the Nodegraph object itself - perhaps we could start creating free function that perform these helper tasks, but still include them as part of the library?

jstone-lucasfilm:
Yes, I was thinking that @kwokcb's new `setInputOrdering` represents a useful _pattern_ for MaterialX developers, but that each use case is likely to be slightly different, e.g. ordering the inputs of an interface, ordering the nodes of a graph, ordering the top-level elements of a document, and so forth.

That might point to the idea of making this a "best practices" example somewhere in the MaterialX codebase, allowing developers to derive their own custom logic from this _pattern_, rather than making this a permanent feature of the NodeGraph API.

I'm definitely open to ideas on this, and we don't yet have an obvious home for best practices C++ examples in MaterialX, though I think having one would potentially be a real benefit to users.

PR #2122, opened 2024-11-26
Fix mxvalidate to check for stdlib usage

kwokcb:
## Change

Patch `mxvalidate.py` script to avoid trying to access an undefined object `stdlib`
if not specified from the command line (default).

```
doc.setDataLibrary(stdlib)
                       ^^^^^^
UnboundLocalError: cannot access local variable 'stdlib' where it is not associated with a value
```


jstone-lucasfilm (approved):
Good catch, thanks @kwokcb!

PR #2124, opened 2024-11-27
Fix handling of missing scenes in the Web Viewer

samip:
JsViewerX attempted to use a cube when the selected GLTF file contained no model, but the code never added it to the scene.

### DIFF
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -90,14 +90,14 @@ export class Scene
         }
 
         this.#_rootNode = null;
-        const model = gltfData.scene;
+        let model = gltfData.scene;
         if (!model)
         {
             const geometry = new THREE.BoxGeometry(1, 1, 1);
             const material = new THREE.MeshBasicMaterial({ color: 0xdddddd });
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This looks very promising, @samip!  You should be able to resolve the EasyCLA issue by clicking on one of the two links in the Linux Foundation comment above, and let us know if you need additional assistance.

javascript/MaterialXView/source/viewer.js
```diff
@@ -90,14 +90,14 @@ export class Scene
...
+        let model = gltfData.scene;
         if (!model)
         {
             const geometry = new THREE.BoxGeometry(1, 1, 1);
             const material = new THREE.MeshBasicMaterial({ color: 0xdddddd });
             const cube = new THREE.Mesh(geometry, material);
-            obj = new Group();
-            obj.add(geometry);
+            model = new Group();
+            model.add(cube);
```
kwokcb, line 100: Thanks for the fix here. Curious that `obj` was not caught as being undeclared -- may it's declared somewhere else.

jstone-lucasfilm (approved):
Thanks for this improvement, @samip!

PR #2126, opened 2024-11-27
Add JavaScript bindings for element equivalence

kwokcb:
## Change
- Expose the `Element::isEquivalent()` + options structure API in Javascript.

### Tests
- New unit test was added which performs a basic error check.

### DIFF
--- a/javascript/MaterialXTest/element.spec.js
+++ b/javascript/MaterialXTest/element.spec.js
@@ -263,49 +263,9 @@ describe('Equivalence', () =>
         options.performValueComparisons = false;
         let result = doc.isEquivalent(doc2, options, differences);
         expect(result).to.be.false;
-        console.log(differences.message);
+        expect(differences.message).to.not.be.empty;
 
-        options.performValueComparisons = true;
-        result = doc.isEquivalent(doc2, options, differences);
-        expect(result).to.be.true;
-
-        let currentPrecision = mx.Value.getFloatPrecision();
-        options.floatPrecision = 8;
-        result = doc.isEquivalent(doc2, options, differences);
-        expect(result).to.be.false;
-        options.floatPrecision = currentPrecision;
-
-        options.setAttributeExclusionList([mx.ValueElement.UI_MIN_ATTRIBUTE, mx.ValueElement.UI_MAX_ATTRIBUTE]);
-        floatInputs.forEach(input => {
-            input.setAttribute(mx.ValueElement.UI_MIN_ATTRIBUTE, "0.9");
-            input.setAttribute(mx.ValueElement.UI_MAX_ATTRIBUTE, "100.0");
-        });
-        result = doc.isEquivalent(doc2, options, differences);
-        expect(result).to.be.true;
-        floatInputs.forEach(input => {
-            input.setAttribute(mx.ValueElement.UI_MIN_ATTRIBUTE, "  0.01");
-            input.setAttribute(mx.ValueElement.UI_MAX_ATTRIBUTE, "  1.01");
-        });
-
-        let mismatchElement = doc.getDescendant("mygraph/input_color4");
-        let previousName = mismatchElement.getName();
-        mismatchElement.setName("mismatch_color4");
-        result = doc.isEquivalent(doc2, options, differences);
-        expect(result).to.be.false;
-
-        mismatchElement.setName(previousName);
-        result = doc.isEquivalent(doc2, options, differences);
-        expect(result).to.be.true;
-
-        let nodeGraph = doc.getNodeGraph("mygraph");
-        expect(nodeGraph).to.exist;
-        doc.addNodeDef("ND_mygraph");
-        nodeGraph.setNodeDefString("ND_mygraph");
-        let nodeGraph2 = doc2.getNodeGraph("mygraph");
-        expect(nodeGraph2).to.exist;
-        doc2.addNodeDef("ND_mygraph");
-        nodeGraph2.setNodeDefString("ND_mygraph");
-        result = doc.isEquivalent(doc2, options, differences);
+        result = doc.isEquivalent(doc2, options, undefined);
         expect(result).to.be.false;
     });
 });
--- a/javascript/MaterialXTest/element.spec.js
+++ b/javascript/MaterialXTest/element.spec.js
@@ -164,3 +164,146 @@ describe('Element', () =>
         });
     });
 });
+
+describe('Equivalence', () =>
+{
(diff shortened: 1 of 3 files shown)


source/JsMaterialX/JsMaterialXCore/JsElement.cpp
```diff
@@ -35,11 +35,46 @@ namespace mx = MaterialX;
...
+            self.attributeExclusionList = std::set<std::string>(exclusionList.begin(), exclusionList.end());
+        }));
+
     ems::class_<mx::Element>("Element")
         .smart_ptr<std::shared_ptr<mx::Element>>("Element")
         .smart_ptr<std::shared_ptr<const mx::Element>>("Element") // mx::ConstElementPtr
         .function("equals", ems::optional_override([](mx::Element& self, const mx::Element& rhs) { return self == rhs; }))
         .function("notEquals", ems::optional_override([](mx::Element& self, const mx::Element& rhs) { return self != rhs; }))
+        .function("isEquivalent", ems::optional_override([](mx::Element &self, const mx::Element& rhs, const mx::ElementEquivalenceOptions& options, 
+                                                            ems::val message)
```
jstone-lucasfilm, line 54: Since the `isEquivalent` interface is still pre-release, it seems worth double-checking that we don't want it to return a simple message string instead of a vector of descriptors, matching the behavior of this JavaScript wrapper.

This wouldn't preclude us from adding the vector of descriptors further in the future, and it might be a more natural interface for users to initially work with, matching the behavior of our `validate` methods.

What are your thoughts?

jstone-lucasfilm, line 54: Since the C++ interface for `isEquivalent` is still pre-release, it seems worth double-checking that we don't want it to return a simple message string instead of a vector of descriptors, matching the behavior of this JavaScript wrapper.

This wouldn't preclude us from adding the vector of descriptors further in the future, and it might be a more natural interface for users to initially work with, matching the behavior of our `validate` methods.

What are your thoughts?

kwokcb: I agree it's overly complicated currently since it currently always stops after the first difference so the list degenerates to a single message. I'll look at the other PR.

jstone-lucasfilm: Thanks for the review, @kwokcb, and the simplification is now merged!

kwokcb: @jstone-lucasfilm, I've merged the latest changes and updated the JS wrappers.

javascript/MaterialXTest/element.spec.js
```diff
@@ -164,3 +164,148 @@ describe('Element', () =>
...
+describe('Equivalence', () =>
+{
+    let mx, doc, doc2, inputMap, inputMap2, floatInputs;
+
+    before(async () => {
+        mx = await Module();
+
+        doc = mx.createDocument();
+        inputMap = new Map([
+            ["color3", "  1.0,   +2.0,  3.0   "],
```
jstone-lucasfilm, line 177: Since these JavaScript bindings are simple wrappers around the underlying C++ methods, perhaps we don't need these additional unit tests in JavaScript, which mirror the same setup as our C++ tests?

My sense is that the new bindings in `JsElement.cpp` are sufficient here, and we can let our existing C++ unit tests do the heavy lifting!

source/JsMaterialX/JsMaterialXCore/JsElement.cpp
```diff
@@ -35,11 +35,34 @@ namespace mx = MaterialX;
...
     ems::class_<mx::Element>("Element")
         .smart_ptr<std::shared_ptr<mx::Element>>("Element")
         .smart_ptr<std::shared_ptr<const mx::Element>>("Element") // mx::ConstElementPtr
         .function("equals", ems::optional_override([](mx::Element& self, const mx::Element& rhs) { return self == rhs; }))
         .function("notEquals", ems::optional_override([](mx::Element& self, const mx::Element& rhs) { return self != rhs; }))
+        .function("isEquivalent", ems::optional_override([](mx::Element &self, const mx::Element& rhs, const mx::ElementEquivalenceOptions& options, 
+                                                            ems::val message) 
+        {
+            mx::ConstElementPtr rhsPtr = rhs.getSelf();
+            std::string differences;
```
jstone-lucasfilm, line 57: Now that our equivalence tests return a `message` string, can we potentially remove the duplication between `message` and `differences` strings in this logic?

javascript/MaterialXTest/element.spec.js
kwokcb: I've simplified it to just test where there are C++ logic for wrapper for `isEquivalent()`

source/JsMaterialX/JsMaterialXCore/JsElement.cpp
kwokcb: I've fixed this up so it is have a proper variadic binding return message argument. The ems message is an object argument and the return type from C++ is a string so is mapped to the object's `.message` member (the same as is done for `validate()`).

javascript/MaterialXTest/element.spec.js
```diff
@@ -164,3 +164,108 @@ describe('Element', () =>
...
+    {
+        let options = new mx.ElementEquivalenceOptions();
+
+        let differences = {};
+        options.performValueComparisons = false;
+        let result = doc.isEquivalent(doc2, options, differences);
+        expect(result).to.be.false;
+        expect(differences.message).to.not.be.empty;
+
+        result = doc.isEquivalent(doc2, options, undefined);
```
kwokcb, line 268: I left this as a single JS wrapper since optional arguments aren't really native for JS. 
The alternative is to add 2 internal wrappers but doesn't seem worth it `validate` also follows this argument convention.

javascript/MaterialXTest/element.spec.js
jstone-lucasfilm: Maybe I'm missing something important here, but I was imagining that this entire setup (e.g. the extensive `inputMap` and `inputMap2`, the testing of various difference categories, and so forth) could be left in C++, and that our JavaScript test would be as simple as verifying that the JavaScript binding works at all -- e.g. that the `isEquivalent` method can be called without triggering a JavaScript error message.

kwokcb: You're correct. I forgot to remove the complex setup.

jstone-lucasfilm (approved):

PR #2132, opened 2024-12-05
Add imagearray node

ld-kerley:
As previously discussed, this PR adds a new `<imagearray>` node.  It behaves very similarly to the `<image>` node but  accepts an array of filenames, instead of a single file.  The `index` integer input is used to select which filename from the array of filenames is used.

This PR adds shader gen and render support for both MSL and GLSL.  Both MSL and GLSL use texture arrays to implement the `<imagearray>` node.

I attempted an OSL implementation, but the texture filename in OSL uses a struct (`textureresource`).

```
struct textureresource { string filename; string colorspace; }
```

In order to support `<imagearray>` in OSL we would need an array of this `texture resource` struct.  Unfortunately arrays of structs are not supported as OSL parameters. 

I believe we should be able to refactor how filenames and colorspaces are managed in OSL, and thus allow `<imagearray>` to be supported in OSL, but I'm opting to leave this to a follow up PR, as its likely to be a more invasive change.


crydalch:
This is a neat idea! One question/concern I have, would this require having array-flavors of every current/future texture-map related shader (triplanar, hextiled)?

What if there was a shader node, that takes an array of filepaths, and has an integer input to select one of the filepath outputs? This could then be connected to <image>, <triplanar> or any other texture node. Or does that cause problems with the various shading languages/rendering targets?

ld-kerley:
@lgritz - I posted [this issue](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/issues/1910) to the OSL project, as a response to your comment above.  If there's a good reason things are the way they are feel free to close. 

@crydalch - This is not an array of filenames, but instead a MaterialX realization of the hardware shading concept of texture arrays, where a number of images are bound to a single texture unit and then selected by index.  So you read the texel from an <imagearray> by providing a UV value and an integer index, and then get back the color(?) value. 

I guess we could provide a nodegraph fallback implementation that works the way you describe - but for the hardware languages I think we would likely want to directly use the hardware support.  This nodegraph idea might work around the current complexities in the OSL implementation. 

I don't think this proposal necessarily infers array versions of other nodes, at least not yet, and certainly not necessarily.  Though I can see a world where one might want to add an image-array version of something like hextile for instance, and then randomly selecting an index to use to increase variation in the generated texture.

jstone-lucasfilm:
@ld-kerley Taking a step back from this specific implementation, it seems worthwhile to consider the alternative, where the new `index` input is simply added to the `image` node, and all of the existing functionality of image nodes (e.g. filter modes, address modes, default color behavior, texture coordinate overrides) would automatically extend to image arrays as well.

I don't have a definitive answer as to which of these two approaches is best, but adding an `index` input is certainly _simpler_ and more straightforward to extend to `tiledimage` and other image-like nodes in the future.  In addition, it would provide artists with a straightforward mechanism to maintain array support in custom graphs that leverage `image` under the hood.

jstone-lucasfilm:
Thanks for all the work here, @ld-kerley, and apologies that this one has sat for a while.  I think it raises a genuinely useful question that's worth settling as a community before we pick an implementation.

Taking a step back, I'd suggest the key decision is what `imagearray` means, and there are really two different interpretations bundled together in this PR:

1. A GPU texture array, where N same-size images are bound to one texture unit and indexed at sample time.  This is what the GLSL/MSL path implements.
2. An image selected from a set, where a list of files is provided, an index picks one, and it's then sampled exactly like `image` (same filter/address modes, default, colorspace, texcoord behavior).

Our usual design philosophy is that a node should be meaningful across all of our targets: GLSL, MDL, OSL, as well as custom renderers such as Karma.  Interpretation (1) doesn't really meet that bar, since texture arrays are a hardware-rasterizer resource-binding concept, and OSL/MDL/Karma reference textures by filename through their own texture systems, with no notion of binding a fixed set of slices to a unit and indexing them.  That's why the OSL side here needs the `string filename[100]` workaround, related to the arrays-of-structs limitation that @lgritz questioned earlier.

Interpretation (2), on the other hand, is meaningful everywhere by construction, since every target already samples images by filename.  It also makes the secondary questions go away on their own.  Per-file colorspace (@dbsmythe's point) just works, the same-size constraint disappears, and all of the existing image behavior comes along for free instead of being re-specified per backend.  GLSL/MSL would remain free to implement it as a texture array under the hood as an optimization, but as an implementation detail rather than part of the contract.

If that framing holds, then I think the cleanest realization is similar to what I proposed earlier: extend the `image` family with an array-of-files input plus a uniform `index` input, rather than a parallel `imagearray` node that would eventually want array-flavored `tiledimage`, `hextiledimage`, and similar nodes.  We'd also need to clarify out-of-range `index` behavior in the specification rather than leaving it undefined.

I'm curious as to what others on the thread think.  Does interpretation (2) cover the use cases you had in mind, or is dynamic per-fragment indexing (which is the one thing it can't portably guarantee) actually a requirement here?

PR #2135, opened 2024-12-07
Static analysis fixes

jstone-lucasfilm:
This changelist addresses a handful of static analysis warnings flagged by PVS-Studio:

- Validate `downstream` before dereferencing in `ShaderGraph::createConnectedNodes`.
- Pass a string argument by const reference in `StructTypeDesc::addMember`.
- Avoid repeated calls to `Value::asA` in `Graph::setConstant`.


jstone-lucasfilm:
Although these changes seem straightforward, I'm CC'ing @ld-kerley and @lfl-eholthouser for their thoughts, just in case they spot issues that I've missed.

PR #2137, opened 2024-12-09
Use setDataLibrary in Graph Editor

jstone-lucasfilm:
This changelist updates the Graph Editor to use `Document::setDataLibrary` rather than `Document::importLibrary`, improving the efficiency of document loading.

Additionally, it removes an unused method and member variable from the `RenderView` class.


jstone-lucasfilm:
I'm CC'ing @lfl-eholthouser and @ashwinbhat on this proposed change, just in case they spot any issues that I've missed.

lfl-eholthouser:
Looks great to me! @jstone-lucasfilm

PR #2138, opened 2024-12-10
Use setDataLibrary in Web Viewer

jstone-lucasfilm:
This changelist updates the Web Viewer to use Document::setDataLibrary rather than Document::importLibrary, improving the efficiency of document loading.

### DIFF
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -606,7 +606,7 @@ export class Material
         // Re-initialize document
         var startDocTime = performance.now();
         var doc = mx.createDocument();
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
I'm CC'ing @kwokcb and @ashwinbhat on this proposed change, just in case they spot any issues that I've missed.

kwokcb:
Don't see any issues with this change. Looks good to me.

jstone-lucasfilm:
Thanks for the review, @kwokcb!

I took a look at updating our JavaScript unit tests to this pattern as well, but it looks like we'd need access to the `loadLibraries` helper function in JavaScript, so that remains a future step for now.

PR #2139, opened 2024-12-10
Simplify shader generator exclusions

jstone-lucasfilm:
This changelist simplifies the shader generator exclusion lists used in unit testing, removing legacy exclusions that are no longer needed with the current data libraries.


jstone-lucasfilm:
I'm CC'ing @kwokcb and @niklasharrysson on this proposed change, just in case they spot any issues that I've missed.

PR #2121, opened 2024-11-25
Remove support for legacy OSL closures

jstone-lucasfilm:
This changelist removes support for the legacy (i.e. pre-v1.12) OSL closures, focusing exclusively on the MaterialX-synchronized OSL closures that were introduced to OSL in October 2022.

PR #2142, opened 2024-12-13
Additional cleanup from OSL legacy closure removal

ld-kerley:
#2121 removed the OSL legacy closures - but we can simplify the data library installation a little more now they are gone.


jstone-lucasfilm (approved):
Good catch, thanks @ld-kerley!

PR #2145, opened 2024-12-14
Fix edge case with editor pasting

jstone-lucasfilm:
This changelist fixes the edge case of pasting with an empty copy buffer in the Graph Editor.

Previously, this action would have an unintended effect on node positions, and it now has no impact on the graph.


jstone-lucasfilm:
I'm CC'ing @lfl-eholthouser and @kwokcb on this proposed change, just in case they spot any issues that I've missed.

kwokcb:
Looks good to me.

PR #2147, opened 2024-12-15
Add null checks to UiPin methods

jstone-lucasfilm:
This changelist adds null checks to the UiPin::addConnection and UiPin::deleteConnection methods in the Graph Editor, providing additional protection against crashes.


lfl-eholthouser:
Looks great to me!

Issue #1782, opened 2024-04-19
Graph Editor: Off-by-one error in visual updates


zicher3d:
Please see videos below.
The attached test material should be copied into the libraries/stdlib folder. It will add a node called "test_update_38ten".

When connected to the output node, the materials correctly compiles and display, but changing any of the properties does not update.
This was working in 1.38.8.

[test_update_38ten.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/15044112/test_update_38ten.zip)

Behavior in 1.30.10:
https://github.com/AcademySoftwareFoundation/MaterialX/assets/7254958/de790e34-dbad-4554-9bb6-ecb4507df348

Behavior in 1.30.8:
https://github.com/AcademySoftwareFoundation/MaterialX/assets/7254958/144eb4f7-8faf-45ba-8b5d-9dafa12c52a9

jstone-lucasfilm:
Good catch, @zicher3d, and I think this bug is a special case of an off-by-one error for visual updates in the Graph Editor.  At the moment, the user needs to first select _away_ from their newly-added node, and then select _back_ to it, after which the render view updates as expected.

This issue still occurs in MaterialX v1.39.2, and the built-in `disney_principled` node can be used in place of the provided custom node.

This seems like an important issue to fix in the Graph Editor, and I'm CC'ing @lfl-eholthouser and @kwokcb for their thoughts and advice.

jstone-lucasfilm:
Thanks for this report, @zicher3d, and this should be resolved by #2150, which will be included in MaterialX v1.39.2.

PR #2150, opened 2024-12-16
Add fallbacks for findUniform methods

jstone-lucasfilm:
This changelist adds fallbacks to the GlslMaterial::findUniform and MslMaterial::findUniform methods, allowing them to find matches for uniforms that don't yet have a full path.

This addresses issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/1782.

### DIFF
--- a/source/MaterialXRenderGlsl/GlslMaterial.cpp
+++ b/source/MaterialXRenderGlsl/GlslMaterial.cpp
@@ -331,6 +331,13 @@ ShaderPort* GlslMaterial::findUniform(const std::string& path) const
         {
             return (port && stringEndsWith(port->getPath(), path));
         });
+        if (!port)
+        {
+            port = publicUniforms->find([path](ShaderPort* port)
+            {
+                return (port && stringEndsWith(path, port->getName()));
+            });
+        }
 
         // Check if the uniform exists in the shader program
         if (port && !_glProgram->getUniformsList().count(port->getVariable()))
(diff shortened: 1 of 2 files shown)


source/MaterialXRenderGlsl/GlslMaterial.cpp
```diff
@@ -331,6 +331,13 @@ ShaderPort* GlslMaterial::findUniform(const std::string& path) const
         {
             return (port && stringEndsWith(port->getPath(), path));
         });
+        if (!port)
+        {
+            port = publicUniforms->find([path](ShaderPort* port)
+            {
+                return (port && stringEndsWith(path, port->getName()));
```
ld-kerley, line 338: Just wondering if this might be cleaner (and more efficient?) if we folded this check inside the one above?

```
        port = publicUniforms->find([path](ShaderPort* port)
        {
            return (port && (stringEndsWith(port->getPath(), path) || stringEndsWith(path, port->getName()));
        });
```

jstone-lucasfilm: That's a good instinct from an efficiency perspective, though in this case it would alter the _results_ of the function in a way that we don't intend, allowing the fallback logic to affect searches that already successfully found matches.

As this logic is currently written, if the first search succeeds, then the fallback logic is entirely skipped, making it so that only failure cases for the original function are extended by the fallback behavior.

ld-kerley: makes sense - thanks for the clarification.

I wonder if it's helpful to add a code comment to that effect - to avoid someone thinking there is a potential optimization here in the future?

PR #2151, opened 2024-12-18
Improve specification of geometric spaces

jstone-lucasfilm:
- Centralize the list of supported geometric spaces in a new `Geometric Spaces` section.
- Remove `tangent` and `common` from the list of geometric spaces, as they are not supported in the codebase, and more design discussion would be needed to propose their addition.
- Remove the `space` input from `heighttonormal`, as it's not supported in the codebase, and more design discussion would be needed to propose its addition.


ld-kerley:
I might also propose removing `model` as from what I can tell there is no difference in current implementation to `object`.  In a recent meeting conversation there was some confusion around the difference between `object` and `model` if I recall - and it doesn't appear that the code-base treats the two spaces differently at all. 

If I'm understanding the intention behind `model` then it would require the renderer to have access to the pre-deformed geometry.  For an offline rendering system like the renderer we used at Imageworks using OSL, that would require passing the T-pose (or other reference pose) positions as additional prim-vars - In renderman I believe the conventional name would be `Pref`.

Perhaps we can move `model` space to the proposals document?

jstone-lucasfilm:
That's a good suggestion, @ld-kerley, though the `model` geometric space is in a different category from `tangent` or `common`, as it's already one of the enumerated values that we promote in our interfaces for geometric nodes:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/stdlib_defs.mtlx#L1283

That would make it a breaking change to formally remove support for `model`, and we should hold a discussion with the various rendering teams before proposing this.

From the perspective of our GLSL, OSL, and MDL code generators, the main reason that we don't currently make a distinction between `object` and `model` is that we don't have access to this information in our simple reference renderers (e.g. MaterialXView), and our hope is that renderers with support for animations will use the distinction in a visually-meaningful way.

If we determine MaterialX TSC discussions that the `object` versus `model` distinction is unwanted, though, then we can remove this feature in the future, maintaining backwards compatibility with our version upgrade system.

jstone-lucasfilm:
Just to provide an external reference point, here's the documentation for the `position` node in Houdini/Solaris, where they include the MaterialX language on the behavior of the `model`, `object`, and `world` spaces:

https://www.sidefx.com/docs/houdini/nodes/vop/mtlxposition.html

ld-kerley:
One option might be for us to retain code support for reading `model` as `object` space, for backwards compatibility (at least until 1.40), but just move the language in the spec in to the proposals document until its fully implemented.

jstone-lucasfilm:
That sounds like a great discussion topic for an upcoming MaterialX TSC meeting, since we'd want to get the opinions of teams with existing MaterialX integrations before changing the published interfaces of our geometry nodes, which currently expose all three of `model`, `object`, and `world` as enumerated options.

My guess is that SideFX and other teams already support the distinction between `model` and `object` in their MaterialX integrations, since they note this distinction so clearly in their own documentation, but if we learn that they would prefer to remove one or more of these options, then we can propose that change for v1.40.

jstone-lucasfilm:
I'll move this proposal forward for now, and let's keep the discussion moving forward with additional improvements.  I'd be open to a discussion of the present and future role of `model` if there's interest in including this in an upcoming MaterialX TSC.

PR #2153, opened 2024-12-20
Updates for OSL shader generation

jstone-lucasfilm:
This changelist updates OSL shader generation to account for recent feature contributions (e.g. worleynoise, geompropvalueuniform), allowing render comparisons to complete successfully.

### DIFF
--- a/libraries/stdlib/genosl/mx_worleynoise2d_float.osl
+++ b/libraries/stdlib/genosl/mx_worleynoise2d_float.osl
@@ -1,4 +1,4 @@
-void mx_worleynoise2d_float(vector2 texcoord, float jitter, output float result)
+void mx_worleynoise2d_float(vector2 texcoord, float jitter, int style, output float result)
 {
     result = mx_worley_noise_float(texcoord, jitter, 0);
 }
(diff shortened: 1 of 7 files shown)


ld-kerley:
This appears to add the OSL Worley node implementation interface, but not the actual code.  

I understand wanting to prevent the tests from failing so violently. However, this change introduces a subtle bug. With this PR merged, an artist using an OSL backend can modify the new `style` parameter, expecting a change, but it won’t work. We can improve this to help artists and future developers.

Possible options:
* Take a stance that we don’t merge the original PR until all backend languages are supported (or the shortfall is documented in the spec). This may increase the burden on the original developer, who may lack experience in all backend languages.
* Accept the original PR and then:
  * Leave the tests failing until full support is added (or documentation updated to reflect the shortfall)
  * Make a PR to allow the tests to pass (albeit with incorrect picture), but also add documentation reflecting the missing functionality.

We need to find an appropriate way to document functionality supported only in a subset of backend languages. Once we create that documentation, we’ll likely find other similar cases that slipped through the cracks.

jstone-lucasfilm:
@ld-kerley I really like the idea of documenting missing language implementations for visibility, and this seems like a more effective approach than pausing new contributions until they support all languages.

We have the very beginnings of this documentation here, and it seems worthwhile to extend this over time, recommending that contributors add to this page when they propose new features that don't initially cover all languages:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/README.md#target-support

ld-kerley:
This page seems like a good start - but this document is not very discoverable for artists.  I think it would be a good idea for us to to try and find a place that would be more user-facing to convey this. 

(In this instance I also posted #2155 - which will hopefully help fill this hole here - but I suspect there are others).

Issue #1542, opened 2023-09-26
Graph Editor: Improved Add Node workflow


lfl-eholthouser:
Now the only way to add a new node is with the tab button. In order to help user workflows, add a new node popup when dragging a link off of an pin. Filter the nodes in the popup to only show ones that can be connected to the current pin.

GregoryD2017:
gonna try this one out for dev days! 🥳 🥳

jstone-lucasfilm:
I'm clearing out the assignment on this task, and marking it as a Good First Issue for future developers to tackle.

beersandrew:
Hello! I'm going to start taking a closer look at this issue and post here if I have any questions. Let me know if this isn't needed or someone else is working on it. Thanks!

beersandrew:
@jstone-lucasfilm I'm unfortunately unable to continue work on this due to taking on a new role and not having clear guidance on Open Source contributions at this time. If my open PR (https://github.com/AcademySoftwareFoundation/MaterialX/pull/2157) is of help to anyone, then great, however there are some questions in the PR that were unclear at the time. Feel free to remove me as the owner so someone else can either start from scratch or build off of my Draft PR. Thanks! And apologies for not being able to finish this.

kiyoyukinakagaki-fn:
Hi! I'd like to work on this issue for ASWF Dev Days 2025 please.

I'm going to start off by picking up beersandrew's PR. It would be great if anyone can answer beersandrew's open questions to clarify the goal of the feature.

Question 1) beersandrew's implementation filters the candidate nodes by their output types. Do we want to filter candidate nodes by input types so that all the candidates nodes that have at least one matching input type to the type of the pin from which the link is drawn from?

Question 2) If the answer to the question 1 is "Yes", do we want the link to automatically connect to the first input with the matching type of the newly created node if it has multiple inputs with the matching type?

I'm assuming the answers are "Yes" and "Yes", but confirmation would be greatly appreciated before getting onto the work.

ld-kerley:
Welcome @kiyoyukinakagaki-fn, glad to have you onboard for dev days (and hopefully beyond :)).  I'm not familiar with the history here - but I will try and find sometime today to poke around and get you some answers.

kiyoyukinakagaki-fn:
As I work through this issue, I think I have answers to those questions myself.

[...]

PR #2157, opened 2024-12-24
Graph Editor: Improve Add Node Workflow

beersandrew:
Addressing #1542 . This PR is a draft as I have some open questions remaining, listed below

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3677,7 +3677,7 @@ void Graph::showHelp() const
 
 void Graph::addNodePopup(bool cursor)
 {
-    bool open_AddPopup = ImGui::IsWindowFocused(ImGuiFocusedFlags_RootWindow) && ImGui::IsKeyReleased(ImGuiKey_Tab) ||
+    bool open_AddPopup = (ImGui::IsWindowFocused(ImGuiFocusedFlags_RootWindow) && ImGui::IsKeyReleased(ImGuiKey_Tab)) ||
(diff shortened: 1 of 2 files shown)


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3718,7 +3720,8 @@ void Graph::addNodePopup(bool cursor)
                 // Allow spaces to be used to search for node names
                 std::replace(subs.begin(), subs.end(), ' ', '_');
 
-                if (str.find(subs) != std::string::npos)
+                if ((subs.size() == 0 || str.find(subs) != std::string::npos) &&
+                    (_menuFilterType == mx::EMPTY_STRING || node.getType() == _menuFilterType))
```
beersandrew, line 3724: Is this the correct filtering logic? It does filter properly however I'm not sure this is the right way to filter. Given I'm pulling a pin from a `color3` this would:
1. Get the `color3` type from the `_pinFilterType`
2. Assign that type to `_menuFilterType`
3. Filter nodes by that type

This doesn't feel quite right to me since these types seem to be output types rather than input types. 

It's unclear to me how to address this as I can imagine, using the same example, you could have many nodes that would have `color3` inputs, but what would be the desired behavior in how to link them up, I imagine the first input of the pin type could be matched. 

I'll continue down this path to try to filter based on input pin types rather than output types

jstone-lucasfilm:
Thanks again to @beersandrew for getting this project started last year, and I believe we can close out this original draft, which @kiyoyukinakagaki-fn has completed in #2383.

Issue #1525, opened 2023-09-09
The `closest` filtertype on images isn't respected in viewers


expenses:
Hi, I'm running into an issue where `image` nodes that are set to closest/nearest-neighbour texture filtering still use bilinear filtering in `MaterialXView`, `MaterialXGraphEditor` and `usdview` when USD is compiled with MaterialX support. 

![20230909_15h41m56s_grim](https://github.com/AcademySoftwareFoundation/MaterialX/assets/13566135/328fd871-aadd-4bb4-8d0c-21a6b3d9bdd8)

I'm not sure if this is the same issue for both MaterialX and USD that can be solved by a fix to MaterialX or if they're seperate.

jstone-lucasfilm:
Good catch, @expenses, and this looks like an oversight that needs to be addressed in MaterialXRender and MaterialXRenderGlsl.

mapetikea:
I think that the bug is also present in the OSL codegen. It doesn't seem to respect filters either?

jstone-lucasfilm:
Thanks for this report, @expenses, and both the GLSL and OSL cases should now be addressed in the main branch.

PR #2158, opened 2024-12-29
Add filtering properties to OSL textures

jstone-lucasfilm:
This changelist adds filtering properties to texture calls in OSL shader generation, improving the visual consistency between generated OSL and other shading languages.

Previously, the "interp" property was omitted from texture calls in generated OSL, causing the default "smartcubic" filtering mode to be used.


jstone-lucasfilm:
This change addresses the OSL component of https://github.com/AcademySoftwareFoundation/MaterialX/issues/1525, and it may prove useful in OSL-based texture baking workflows.

I'm CC'ing @lfl-eholthouser for visibility, as she may be interested in integrating this change into her work.

PR #2161, opened 2024-12-30
Increase OSL samples in reference renders

jstone-lucasfilm:
This changelist increases the number of OSL samples in reference-quality renders, improving the visual parity between OSL and other languages in MaterialX render tests.

Additionally, it pares down the list of document paths for MaterialX render tests, reducing the resulting PDF document to a size that can be directly shared on Slack.


jstone-lucasfilm:
Attaching a reference render comparison between GLSL and OSL, using the latest codebase in this changelist:

[MaterialXRenderTests_12_30_2024_GitHub.pdf](https://github.com/user-attachments/files/18277437/MaterialXRenderTests_12_30_2024_GitHub.pdf)

PR #2165, opened 2025-01-02
Improvements to heighttonormal node

jstone-lucasfilm:
- Improve the specification of the `heighttonormal` node, clarifying that its output contains tangent-space vectors encoded in the [0-1] range.
- Improve the OSL implementation of the `heighttonormal` node, leveraging the built-in partial derivative instructions to compute the gradient at each surface point.
- Simplify example material filenames to improve layout in render tests.
- Add convolution nodes to the list of standard render tests.

### DIFF
--- a/documents/Specification/MaterialX.Specification.md
+++ b/documents/Specification/MaterialX.Specification.md
@@ -1776,7 +1776,7 @@ Convolution nodes have one input named "in", and apply a defined convolution fun
 
 <a id="node-heighttonormal"> </a>
 
-* **`heighttonormal`**: convert a scalar height map to a normal map of type vector3.
+* **`heighttonormal`**: convert a scalar height map to a tangent-space normal map of type vector3.  The output normal map is encoded with all channels in the [0-1] range, enabling its storage in unsigned image formats.
     * `in` (float): the input value or nodename
     * `scale` (float): the scale of normal map deflections relative to the gradient of the height map.  Default is 1.0.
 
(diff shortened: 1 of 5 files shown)


jstone-lucasfilm:
Attaching a reference render comparison between GLSL and OSL using the latest code in this PR:

[MaterialXRenderTests_01_02_2025_GitHub.pdf](https://github.com/user-attachments/files/18294844/MaterialXRenderTests_01_02_2025_GitHub.pdf)

jstone-lucasfilm:
Here's a close-up of the latest render comparison for the `heighttonormal` node.  There's still room for further improvement, but this is significantly closer to the mark than our previous results:

HeightToNormal_GLSL_vs_OSL

jstone-lucasfilm:
I'm CC'ing @dbsmythe, @anderslanglands, @ld-kerley, and @niklasharrysson for their thoughts and guidance, in case they spot improvements that we could make to this changelist.

documents/Specification/MaterialX.Specification.md
```diff
@@ -1776,7 +1776,7 @@ Convolution nodes have one input named "in", and apply a defined convolution fun
 
 <a id="node-heighttonormal"> </a>
 
-* **`heighttonormal`**: convert a scalar height map to a normal map of type vector3.
+* **`heighttonormal`**: convert a scalar height map to a tangent-space normal map of type vector3.  The output normal map is encoded with all channels in the [0-1] range, enabling its storage in unsigned image formats.
```
ld-kerley, line 1779: Can we be explicit here about what each component of the output tangent-space vector3 correspond to?  

ie. is it (normal, tangent, bitangent) (N,T,B) or (T,B,N) or something else etc.... the code is already making an explicit choice here - would be good to have the language in the spec reflect that choice. 

I believe Ole Gulbrandsen was asking for this language to be explicit somewhere.

PR #2166, opened 2025-01-03
Simplify plane.obj geometry

pablode:
I noticed that the plane geometry has nearly 1000 vertices. This new simplified plane has identical extents, normals and texcoords.


pablode:
Realized that this is probably in order to support displacement without runtime tessellation. Closing.

PR #2168, opened 2025-01-04
Add <latlongimage> node

pablode:
This PR proposes an initial version of a node that samples equiangular images.

Question: should the image framerange/frameoffset/frameendaction inputs be forwarded?

### DIFF
--- a/documents/Specification/MaterialX.Specification.md
+++ b/documents/Specification/MaterialX.Specification.md
@@ -708,6 +708,14 @@ The type of the &lt;image> node determines the number of channels output, which 
     * `realworldtilesize` (vector2): the real-world size of a single square 0-1 UV tile, with unittype "distance".  A `unit` attribute may be provided to indicate the units that `realworldtilesize` is expressed in.
     * `filtertype` (uniform string): the type of texture filtering to use; standard values include "closest" (nearest-neighbor single-sample), "linear", and "cubic".  If not specified, an application may use its own default texture filtering method.
 
+<a id="node-latlongimage"> </a>
+
+* **`latlongimage`** (NG): samples an equiangular map along a view direction with adjustable latitudinal offset.
+    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](#filename-substitutions) section above.
+    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;_geometry token_>, [_interface token_] or {_hostattr_} is included in the filename but no substitution value or default is defined, or if the resolved `file` URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
+    * `viewdir` (vector3): the view direction determining the value sampled from the projected equiangular map.
+    * `rotation` (float): the latitudinal sampling offset.
+
 <a id="node-triplanarprojection"> </a>
 
 * **`triplanarprojection`** (NG): samples data from three images (or layers within multi-layer images), and projects a tiled representation of the images along each of the three respective coordinate axes, computing a weighted blend of the three samples using the geometric normal.
--- a/libraries/stdlib/stdlib_ng.mtlx
+++ b/libraries/stdlib/stdlib_ng.mtlx
@@ -196,6 +196,59 @@
   </nodegraph>
 
   <!--
+    Node: <latlongimage>
+  -->
+  <nodegraph name="NG_latlongimage" nodedef="ND_latlongimage">
+    <separate3 name="viewDirChannels" type="multioutput">
+      <input name="in" type="vector3" interfacename="viewdir" />
+    </separate3>
+    <atan2 name="angleXZ" type="float">
+      <input name="iny" type="float" nodename="viewDirChannels" output="outx" />
+      <input name="inx" type="float" nodename="viewDirChannels" output="outz" />
+    </atan2>
+    <multiply name="scaleXZ" type="float">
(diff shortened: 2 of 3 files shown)


jstone-lucasfilm:
This looks like a great step forward to me, @pablode, and I believe the testing errors just require the same fix as you implemented for `fract`, where each type variant of the new node has a single usage in our `Materials/TestSuite` folder.

pablode:
Not sure about that, it seems rather an issue with the Unit system not being set up correctly in the test runner. (See log messages and 2nd commit succeeding.)

jstone-lucasfilm:
@pablode Ah, good catch, and I was making assumptions about these errors being the same as the ones in `fract`.

As a starting point, perhaps we should omit the new `unit` attribute in the `latlongimage` node, and we can log that as a future improvement that would be linked to updates to our test suite?

pablode:
Sure, reverted the commit. Maybe in the future, this can be an 'idiomatic' MaterialX practice ;)

documents/Specification/MaterialX.Specification.md
```diff
@@ -708,6 +708,14 @@ The type of the &lt;image> node determines the number of channels output, which
...
 
+<a id="node-latlongimage"> </a>
+
+* **`latlongimage`** (NG): samples an equiangular map along a view direction with adjustable latitudinal offset.
+    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](#filename-substitutions) section above.
+    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;_geometry token_>, [_interface token_] or {_hostattr_} is included in the filename but no substitution value or default is defined, or if the resolved `file` URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
+    * `viewdir` (vector3): the view direction determining the value sampled from the projected equiangular map.
+    * `rotation` (float): the latitudinal sampling offset.
```
jstone-lucasfilm, line 717: Should this language read "longitudinal sampling offset", since the rotation amount is added to the longitude in the graph definition?

pablode: Good catch! Fixed the typo.

meshula:
My experience working with latlong images for video suggests that a specification might also provide guidance on sampling, as realtime latlong implementations are in particular difficult to implement well, given the fact that the first and last row of pixels you need to take into account several entire rows merely to generate good values at nadir and zenith. hardware samplers only work as designed at the equator!

meshula:
I also suggest that the node specify a min and max latitude, with a default color for values outside that range. in some sense pixels above and below 15 degrees are wasted. when used as a sky dome a scene will typically cover the bottom with a ground plane, and may use a blended cap for the sky. the wastage goes beyond occlusion though, the top and bottom 15 degrees consume a vastly disproportionate amount of memory versus their solid angular extent. it's very common in the 360 video world therefore to use a restricted latitude band and supply caps which are planar projections. this tweak further allows you to get away with unadorned texture sampling by the hardware for the latlong texture.

pablode:
Good notes @meshula, though I'm not sure how the guidance on sampling would look like specifically. Regarding the min and max latitude, do you suggest to add inputs so that the node can be implemented without a node graph? Otherwise, IMO it sound like something that can be refined in the future

meshula:
Yes, I propose that min and max latitude are foundational, and explicitly inputs on the node. Based on having shipped several 360 video products, I propose that the limits are in radians, positive toward zenith, zero at equator, and default to +pi and -pi. 

Second, I would also propose a nadir cap input and zenith cap input. Such a cap would either be a constant color or a secondary texture input defined as the projection of an image such that the 0,0 uv coordinate in the latlong map (assuming 0,0 is top left) corresponds to 0.5, 0 in the zenith cap, and the 0,1 uv coordinate in latlong corresponds to 0.5, 1.0 in the nadir cap. 0.5, 0.5 corresponds to the nadir, or zenith. The projection on the cap is from the inside of the sphere.

meshula:
As to sampling advice, that's merely informative, and probably belongs in a programming guide, not in the specification.

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -195,6 +195,59 @@
...
+    </atan2>
+    <multiply name="scaleXZ" type="float">
+      <input name="in1" type="float" nodename="angleXZ" />
+      <input name="in2" type="float" value="-0.15915494" />
+    </multiply>
+    <add name="longitude" type="float">
+      <input name="in1" type="float" nodename="scaleXZ" />
+      <input name="in2" type="float" value="0.5" />
+    </add>
+    <multiply name="rotationDegrees" type="float">
```
jstone-lucasfilm: This pull request looks really close to the mark, but in my local validation it seems to remove the ability to rotate the environment background using the `Light Rotation` slider in the MaterialX Viewer.

The fix might be as simple as updating this logic that updates the appropriate shader uniform in RenderPipelineGL.cpp:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXView/RenderPipelineGL.cpp#L371

Additionally, you may want to verify that the conversion between angle units (in the location of this comment) is behaving as you expect, as it's not currently being tested in the MaterialX Viewer.

pablode: Thanks, good points! I believe to have addressed them. Although I have to admit I do not understand where the original 0.5 offset in the light rotation logic comes from, and have therefore omitted it.

Main branch
Screenshot 2025-06-07 at 16 32 11
This PR
Screenshot 2025-06-07 at 16 33 24

jstone-lucasfilm (approved):
This looks good to me, thanks @pablode!

PR #2119, opened 2024-11-21
Enhance Worley noise with solid cells option

crydalch:
worley3d_float

Adds a new control to `worley2d` and `worley3d`, which allows artists to choose between a distance for each cell (the default) and solid values for each cell. This allows for more patterns to be expressed in materials (i.e. flakes).

https://github.com/user-attachments/assets/5e97248d-4893-4135-8f60-0d4bced9d4a0

Here is an example document, which can be opened with `MaterialXGraphEditor --material test_worley_styles_openpbr.mtlx`:

<details>
  <summary>Click to see 'test_worley_styles_openpbr.mtlx'</summary>

  ```<?xml version="1.0"?>
  <materialx version="1.39" colorspace="lin_rec709">
    <surfacematerial name="TestNewWorley_3D" type="material" xpos="17.804348" ypos="0.077586">
      <input name="surfaceshader" type="surfaceshader" nodename="open_pbr_surface_surfaceshader" />
    </surfacematerial>
    <worleynoise2d name="worleynoise2d_float" type="float" xpos="3.673913" ypos="-5.094828">
      <input name="texcoord" type="vector2" nodename="multiply_vector2FA" />
      <input name="jitter" type="float" nodename="change_jitter" />
      <input name="style" type="integer" nodename="change_styles" />
    </worleynoise2d>
    <mix name="mix_color3" type="color3" xpos="5.304348" ypos="-4.931035">
      <input name="mix" type="float" nodename="worleynoise2d_float" />
      <input name="fg" type="color3" value="1, 1, 1" />
      <input name="bg" type="color3" value="0, 0, 0" />
    </mix>
    <multiply name="multiply_vector2FA" type="vector2" xpos="-0.144928" ypos="0.422414">
      <input name="in1" type="vector2" nodename="texcoord_vector2" />
      <input name="in2" type="float" value="25" />
    </multiply>
    <texcoord name="texcoord_vector2" type="vector2" xpos="-1.782609" ypos="0.620690" />
    <switch name="switch_color3" type="color3" xpos="10.615942" ypos="-0.724138">
      <input name="in1" type="color3" nodename="mix_color3" />
      <input name="in2" type="color3" nodename="convert_vector2_color3" />
      <input name="in3" type="color3" nodename="convert_vector3_color3" />
      <input name="which" type="float" value="5" />
      <input name="in4" type="color3" nodename="mix_color4" />
      <input name="in5" type="color3" nodename="convert_vector2_color4" />
[...]

PR #2177, opened 2025-01-09
Promote Worley style control to Unified Noise nodes

crydalch:
In PR #2119 I'd forgotten to include Worley's new cell styles on the Unified Noise node graphs.

image


jstone-lucasfilm (approved):
Thanks for this update, @crydalch, and this seems like a safe change to include in MaterialX v1.39.2 without warranting a new release candidate.

PR #2180, opened 2025-01-13
AOUSD example node definitions

anderslanglands:
Hello, as part of the effort under AOUSD to write an ISO standard for MaterialX in OpenUSD, we would like to suggest some updates to the node definitions in the MaterialX spec. This Draft PR contains a couple of example files to aid discussion.

In the PR, [example.md](https://github.com/AcademySoftwareFoundation/MaterialX/compare/main...anderslanglands:MaterialX:aousd-node-specs?expand=1#diff-3307ded09fb9306c75ec63520420714868e244416410244fc8fd41b0ec925489) contains some introductory descriptions of the specification structure as well as a dummy node example showing how that looks. [math.md](https://github.com/AcademySoftwareFoundation/MaterialX/compare/main...anderslanglands:MaterialX:aousd-node-specs?expand=1#diff-25796bc63c6828abf7ab42ba187da4327fdea9caad3896ede627562794162604) is a (nearly) complete update of the math nodes to this format, to show how it would look rendered.

The main motiviation here is to provide a concise, exact, normative definition of each node's behavior such that a developer could implement the nodes using the definition alone, that will be suitable for direct referencing in the ISO specification. We also believe the consistent formatting and depth of information will be useful for users as well.

We look forward to your comments and discussion.

### DIFF
--- a/documents/Specification/example.md
+++ b/documents/Specification/example.md
@@ -1,0 +1,135 @@
+# Front matter
+
+## Input and Output Types
+
+Unless explicitly specified in the Signature and Parameter tables, each node has a single output, named `out`.
+
+Where an input type is specified as __multi__, the exact type is dependent on the specific node overload. See the Signature table for a complete list of all overload types for each node. 
+
+__zero__ is defined to be the additive identity for each type, that is:
+
+|Type       |Value                 |
+|-----------|----------------------|
+|float      | 0.0                  |
+|integer    | 0                    |
+|boolean    | false                |
+|vector2    | [0.0, 0.0]           |
+|vector3    | [0.0, 0.0, 0.0]      |
+|vector4    | [0.0, 0.0, 0.0, 0.0] |
+|color3     | [0.0, 0.0, 0.0]      |
+|color4     | [0.0, 0.0, 0.0, 0.0] |
+|matrix33   | [0.0, ... 0.0]       |
+|matrix44   | [0.0, ... 0.0]       |
+|string     | ""                   |
+|filename   | ""                   |
+|array      | empty array          |
+
+__one__ is defined to be the multiplicative identity for each type, that is:
+
+|Type       |Value                 |
+|-----------|----------------------|
+|float      | 1.0                  |
+|integer    | 1                    |
+|boolean    | true                 |
+|vector2    | [1.0, 1.0]           |
+|vector3    | [1.0, 1.0, 1.0]      |
+|vector4    | [1.0, 1.0, 1.0, 1.0] |
+|color3     | [1.0, 1.0, 1.0]      |
+|color4     | [1.0, 1.0, 1.0, 1.0] |
+|matrix33   | identity matrix      |
+|matrix44   | identity matrix      |
+|string     | undefined            |
+|filename   | undefined            |
+|array      | undefined            |
+
+
+## Value Ranges
+
+The Signature table has a column, "Accepted values" which contains the range of values that the node is expected to generate a correct result for. Values outside of this range generate undefined behavior. 
+
+An empty "Accepted values" field indicates that any value representable in the underlying data type is valid.
+
+## Equations
+
+The Details section of each node definition contains equations where appropriate. In these equations, node inputs are written in $\mathtt{monospaced \space font}$, and may be aliased to single-letter variables where it helps for legibility in the equations, for example:
+
+```math
+\begin{align}
+
+\mathtt{out} &= \omega \cdot N \notag \\
+\omega &= \mathtt{input} \notag \\
+N &= \mathtt{normal} \notag 
+
+\end{align}
+```
+
+Many nodes describe component-wise operations where the sample equation must be applied individually to each component of a vector or matrix, or directly to a scalar value. These are represented using subscript notation for the component indices, with indices counting from 0.  
+
+Since many nodes are polymorphic over input types, equations are written for the highest-rank type and when reading for lower-rank types, the higher dimensions should be ignored. For instance, `add` is defined as:
+
+```math
+\mathtt{out}_{i,j} = \mathtt{in1}_{i,j} + \mathtt{in2}_{i,j}
+```
+
+which is appropriate for the `matrix33` and `matrix44` types, and should be read as:
+
+```math
+\mathtt{out}_{i} = \mathtt{in1}_{i} + \mathtt{in2}_{i}
+```
+
+for `vectorN` and `colorN` types, and:
+
+```math
+\mathtt{out} = \mathtt{in1} + \mathtt{in2}
+```
+
+for scalars.
+
+
+## Example
+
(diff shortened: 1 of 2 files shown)


kwokcb:
Hi @anderslanglands,

Some testing  deployment notes:

**Syntax (Minor)**
- I tried to see compatibility with html output and to let you know it can all be pre-processed thus far to give you something that can be consumed by Web math formatters: `MathJax` and `Katex`.  (see below)
- `\lparan`, and `\rparan` may not be supported well so may be better to use `\left`, `\right`
- items like subscripts: `_{i}` for compatibility need to be escaped these using `\_{i}`
- `\aligned` vs `\align` may be more compatible depending on library

**Maintainability**

- I'm wondering if it's better to pull information off of nodedefs to build docs. Currently you can have the documentation in 2 places: here and in the `nodedefs` -- or in the current case no documentation on `nodedefs` or places in a 3rd place as comments in the XML file.

- If deriving from `nodedefs` a possible option is to clean up the Python script called `mxdoc` which scans `nodedefs` to build docs.
- Especially things like the listing arg type variations could become cumbersome to maintain.

- If using information here, the alternative suggestion is to inject this documentation into `nodedefs`. If there was some id tag
for each category of node then it should be pretty easy to do. e.g. Instead of:
```md
## add
```
use
```html
<h2 id="add">add</h2> 
```

**"LaTex" / Markdown / HTML Processing**

<details>

Here the example preprocessing for Web I have worked out.
Place `math.md` in the same folder as these files, start a local http server,
and load in the html file below. 

```html
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">

  <!-- Use KaTex -->
  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.20/dist/katex.min.css">

  <script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>

  <!-- Uncomment to use MathJax
  <script type="text/javascript" async>
    window.MathJax = {
      tex: {
        packages: {'[+]': ['mathtools']}, // Load the mathtools package
        inlineMath: [['$', '$']], // Enable single dollar signs for inline math
        displayMath: [['$$', '$$']] // Double dollar signs for block math
      },
      options: {
        skipHtmlTags: ['script', 'noscript', 'style', 'textarea', 'pre'], // Avoid parsing code blocks
      }
    };
  </script>
  <script src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
  -->

  <!-- Custom styling as desired -->
  <link rel="stylesheet" href="style.css">   
</head>

<body>
  <h1>MaterialX Node Library</h1>
  <div id="content">Loading...</div>
  <textarea id="html_result" rows="20" style="width: 100%"></textarea>

  <!-- Include KaTeX JS and auto-render extension -->
  <script src="https://cdn.jsdelivr.net/npm/katex@0.16.20/dist/katex.min.js"></script>
  <script src="https://cdn.jsdelivr.net/npm/katex@0.16.20/dist/contrib/auto-render.min.js" onload="renderMathInElement(document.body, {
      delimiters: [
        { left: '$$', right: '$$', display: true },  // Block math
        { left: '$', right: '$', display: false }    // Inline math
      ]
    });">
    </script>

  <script id="render_page">

    const replaceMath = (match, content) => {
      // Split the content into lines
      // Filter out empty lines (lines with only whitespace)
      // Join the non-empty lines back together with \n
      const lines = content.split('\n');
      const nonEmptyLines = lines.filter(line => line.trim() !== '');
      let modifiedContent = nonEmptyLines.join('\n');

      // Replace \\ with \\\\ to escape backslashes
      modifiedContent = modifiedContent.replace(/\\\\/g, '\\\\\\\\');

      // Perform if \align not handled
      let replaceAlign = false
      if (replaceAlign) {
        // Replace \begin{align} with \begin{aligned} and \end{align} with \end{aligned}
        modifiedContent = modifiedContent      
          .replace(/\\begin\{align\}/g, '\\begin{aligned}')
          .replace(/\\end\{align\}/g, '\\end{aligned}');
      }

      //console.log('Replace:', content, '\nwith:', `\n$$${modifiedContent}$$\n`)

      // Wrap the content with $$ and preserve the triple backticks
      return `\n$$${modifiedContent}$$\n`;
    };

    // Load the Markdown file
    fetch('math.md')
      .then(response => response.text())
      .then(text => {

        const pattern = /```math\s*\n([\s\S]*?)\n\s*```/g
        text = text.replace(pattern, replaceMath);

        //console.log(text)

        // Convert Markdown to HTML using marked.js
        html = marked.parse(text);

        // Insert the HTML into the content div
        document.getElementById('content').innerHTML = html;

        // Output the converted page before rendering.
        const clonedDocument = document.documentElement.cloneNode(true);
        // - Remove the script with id="render_page" from the cloned document
        const scriptToRemove = clonedDocument.querySelector('#render_page');
        if (scriptToRemove) {
          scriptToRemove.remove();
        }
        // - Remove the html result area
        const htmlResultRemove = clonedDocument.querySelector('#html_result');
        if (htmlResultRemove) {
          htmlResultRemove.remove();
        }

        // Re-render formulas
        setTimeout(() => {
          // Use KaTex
          renderMathInElement(document.getElementById('content'),
            {
              delimiters: [
                { left: '$$', right: '$$', display: true },  // Block math
                { left: '$', right: '$', display: false }    // Inline math
              ]
            });

          // OR Use MathJax for the new content
          //MathJax.typesetPromise();

          // Get the outerHTML of the modified clone
          const renderedPageContent = clonedDocument.outerHTML;
          //console.log('Fully evaluated page content:', renderedPageContent);
          document.getElementById('html_result').value = "\<!DOCTYPE html\>\n" + renderedPageContent;
        }, 100); // Adjust the delay as needed

      })
      .catch(error => {
        console.error('Error loading the Markdown file:', error);
        document.getElementById('content').textContent = 'Error loading the Markdown file.';
      });
  </script>
</body>

</html>
```
CSS file `style.css`:
```css
body {
    font-family: Arial, sans-serif;
    line-height: 1.6;
    /* max-width: 800px; */
    margin: 20px auto;
    padding: 40px;
}

pre {
    background: #f4f4f4;
    padding: 10px;
    border-radius: 5px;
    overflow-x: auto;
}

table {
    border-collapse: collapse;
    width: 100%;
    max-height: 300px;   /* Set a maximum height for the table */
    overflow-y: auto;    /* Enable vertical scrolling */
    display: block
}

th,
td {
    width: 100%;
    border: 1px solid black;
    padding: 4px;
    text-align: left;
}

th {
    background-color: rgb(0, 98, 133); /* Black background */
    color: white;           /* White text */
    font-weight: bold;      /* Bold text for headers */
    padding: 8px;           /* Padding for better spacing */
    border: 1px solid black; /* Ensure borders are consistent */
}

```

</details>

PR #2183, opened 2025-01-14
Add `fractal2d` node

ld-kerley:
Mirroring the existing behavior for `<fractal3d>`, we add the corresponding `<fractal2d> node that uses the `texcoord` input as its manifold.

Add support for OSL, MSL, GLSL, MDL (note : MDL not tested).

Updated specification documents to add `<fractal2d>` to the specification and remove it from the proposals document.

Also remove `period` inputs for the noises that don't have those inputs in the specification.  This is already documented in the proposals document.

Note : in the GLSL noise library code I renamed the existing `mx_fractal_noise_xxx()` functions to `mx_fractal3d_noise_xxx()` to make the differentiation cleaner with the new `mx_fractal2d_noise_xxx()`.  But I retained wrapper functions with the old name that just call through to the new name for backwards compatibility.  I was unsure how much custom customer code might be referencing those functions.  We can remove the shim functions when we have another breaking API release.


libraries/stdlib/genglsl/lib/mx_noise.glsl
```diff
@@ -424,19 +424,79 @@ vec3 mx_fractal_noise_vec3(vec3 p, int octaves, float lacunarity, float diminish
...
+vec4 mx_fractal3d_noise_vec4(vec3 p, int octaves, float lacunarity, float diminish)
 {
-    vec3  c = mx_fractal_noise_vec3(p, octaves, lacunarity, diminish);
-    float f = mx_fractal_noise_float(p+vec3(19, 193, 17), octaves, lacunarity, diminish);
+    vec3  c = mx_fractal3d_noise_vec3(p, octaves, lacunarity, diminish);
+    float f = mx_fractal3d_noise_float(p+vec3(19, 193, 17), octaves, lacunarity, diminish);
     return vec4(c, f);
 }
 
+// These wrappers are provided as a backwards compatible shim after renaming mx_fractal_noise_float() to
```
jstone-lucasfilm, line 479: This is a good instinct for backwards compatibility, but we don't traditionally maintain API compatibility for internal GLSL/OSL/MDL functions in MaterialX, as this would require too many boilerplate wrappers in our active shader codebase.  I think it's fine to omit these and just move forward with your new function names!

ld-kerley: I'll remove.

But for what its worth - I think this shader code does make up the public API for the project - and that we should consider how we can ensure we keep things API stable moving forwards if we want to provide a project that can be relied upon to be stable and robust.

source/MaterialXGenMdl/mdl/materialx/noise.mdl
```diff
@@ -482,6 +482,77 @@ export float mx_fractal_noise_float(
...
+    anno::noinline()
+]]
+{
+
+    float3 c = mx_fractal2d_noise_float3(mxp_p, mxp_octaves, mxp_lacunarity, mxp_diminish);
+    float a = mx_fractal2d_noise_float(mxp_p+float2(19, 193), mxp_octaves, mxp_lacunarity, mxp_diminish);
+    return float4(c.x, c.y, c.z, a);
+}
+
+// TODO - do we want to rename these to mx_fractal3D_noise_float - if so where else might things need to be updated
```
jstone-lucasfilm, line 533: That rename sounds logical to me, and I'm CC'ing @krohmerNV for his thoughts from the MDL side.

krohmerNV:
[MaterialX-mdl-fractal2d-node.patch](https://github.com/user-attachments/files/18559513/MaterialX-mdl-fractal2d-node.patch)
given the time zone difference and the scope of the patch, you can apply it easily yourself.

ld-kerley:
I applied the patch provided by @krohmerNV.  I didn't realize that we have no MDL evaluation tests in the CI - we should try and work towards fixing that. 

On the topic of reusing the 3D noise calls to provide the 2D fractal node - If I'm understanding the code - the 3D version is more expensive - so I think maybe it does make sense to keep separate 2D and 3D versions.  Certainly I would propose we leave that work to a follow up PR where someone can profile the changes across all the languages and noise types - and in some cases it might be a win - I'm not sure.

jstone-lucasfilm:
@ld-kerley On MDL code validation using `mdlc`, this has long been a goal for the MaterialX project, and I believe @krohmerNV has been steadily working towards this goal!

On the proposed change to 3D noise functions, I think Kai is referring to the `TODO` above, which suggests that `mx_fractal_noise_float` would be renamed to `mx_fractal3D_noise_float` for consistency with the functions that you've added here.

I think that's a great idea, if you still support the idea of this rename, and it would make this changelist more complete and self-consistent.

krohmerNV:
I only looked into the MDL implementation of the fractal3D and fractal2D functions and the only difference I saw was a +17 floating point add. What I was suggesting is to have one e.g.  `fractalXD_float2(float3 pos, ...)` and:  
`fractal2D_float2`  calls factalXD_float2(pos=(texcoord.xy, 0)  
`fractal3D_float2`  calls factalXD_float2(pos=(hitpoint.xyz, 0)  
on the MDL level i could remove of the functions in noise.mdl when the only difference would be removed:  
the 2D functions has no +17: [noise.mdl#L515](https://github.com/ld-kerley/MaterialX/blob/41543c2660197f59fa9b001bfe496c48ca1f7f5a/source/MaterialXGenMdl/mdl/materialx/noise.mdl#L515)  
the 3D functions has a +17: [noise.mdl#L586](https://github.com/ld-kerley/MaterialX/blob/41543c2660197f59fa9b001bfe496c48ca1f7f5a/source/MaterialXGenMdl/mdl/materialx/noise.mdl#L586 ) 
without the difference we could remove the new 2D one again, and just call the general one from stdlib_1_6.mdl

apart from MDL, if you say 3D is more expensive, then we should keep two separate implementations. Otherwise, the pattern above could be implemented as a graph.

krohmerNV:
Saw the renaming. Thanks. This looks more consistent now.
About unifying the 2D and 3D code I was asking for. It's not a blocker. If you don't want to align those two, I'm fine.

ld-kerley:
I think the implementation behind the 2D noise functions does less work than the 3D version - at least thats how I'm reading things.  So it seems to make sense to leave them separate - if I'm incorrect here - or we would prefer less code at the expense of slightly slower 2D noise - I'm happy to merge them together.  I don't really have a strong opinion on MDL side of things and happily defer to you.

krohmerNV:
tracked down the [HLSL implementations](https://github.com/AcademySoftwareFoundation/MaterialX/blob/6032c4ce1b3c32ff782a3fb51d3a0a9c5a17beb5/libraries/stdlib/genglsl/lib/mx_noise.glsl#L249) and you are right, there is a difference. 2D is using bilinear interpolation, 3D trilinear. There is a difference. I volunteer to change add this optimization to the MDL nodes.

krohmerNV:
[MaterialX-fractal3d-rename.patch](https://github.com/user-attachments/files/18767750/MaterialX-fractal3d-rename.patch)
you missed some calls while renaming

jstone-lucasfilm:
@ld-kerley Now that the Standard Nodes have been moved to their own document, in support of upcoming specification work at the AOUSD, this PR will need a minor adjustment to take the new location into account:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.StandardNodes.md

jstone-lucasfilm (commented):
The code in this change looks good to me, thanks @ld-kerley.

One valuable next step would be to run the Render Test Suite at reference quality, so that we can confirm that materials based on `fractal3d` (including the classic [standard_surface_marble_solid](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/Examples/StandardSurface/standard_surface_marble_solid.mtlx) example) remain visually unchanged in both GLSL and OSL.

ld-kerley:
Do you have documentation for running the render tests at reference quality and how to compose the pdf for sharing? 
I couldn't find the documentation for these steps anywhere.

jstone-lucasfilm:
@ld-kerley Here's our documentation for the MaterialX Render Test Suite, including the full set of steps for generating output images across languages, and generating the final HTML page for comparisons.  Please ask questions if you run into trouble with these instructions, and we know they can be improved!

https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/source/MaterialXTest#3-shader-generation-and-render-test-suite

ld-kerley:
I had read that page - but missed any mention of "reference quality" - does the CTest run of the render tests (which I have enabled locally by default) run at the required quality level by default?  If so then I believe I'm generating the images already.

Then I think the only part I'm missing are instructions on how you create the pdf in the correct format for sharing here.  Are those steps documented anywhere?

jstone-lucasfilm:
Oh, good questions, @ld-kerley.

So for the reference quality setting, that's found [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/_options.mtlx#L72).  You're completely correct that we don't yet document this, and we definitely should.  For finding any unintended visual changes in a single commit, you could use normal-quality renders as well, as long as you have a normal-quality render from before and after your changes.  I mainly focus on the high-quality renders so that I can keep a running history of our Render Test Suite output, looking for progress and/or regressions over a longer time period.

For the PDF generation, that's completely optional, and you can always just screenshot the output HTML page as needed.  My own approach for generating the PDF is to load the HTML file into Chrome, and then right-click on the page to "Print" it to a PDF, but I don't believe this is necessarily the only (or even best) approach.

ld-kerley:
I ran the render tests - the test_to_html wasn't working for so just posting the before after pictures of the test here



|        | `main`                                                                                                  | `PR`                                                                                                   |
|--------|---------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|
| `GLSL` | Marble_3D_glsl_main | Marble_3D_glsl_new |
| `OSL`  | Marble_3D_osl_main  | Marble_3D_osl_new  |



Appears that this PR doesn't do any harm to the existing fractal code.

jstone-lucasfilm (approved):

PR #2184, opened 2025-01-15
Spelling and spacing fixes

rex4539:
This changelist implements spelling and spacing fixes across the MaterialX repository, leveraging the `codespell` project and manual inspection.


jstone-lucasfilm:
This looks very promising, thanks @rex4539, and let's aim to review and merge these improvements after the upcoming MaterialX v1.39.2 release.

One issue that we'll need to resolve is the spelling fixes to _library dependencies_ such as `stb_image` and `glfw`.  Usually we leave library dependencies unmodified, unless there's a serious bug requiring a MaterialX-side fix, in order to streamline the process of upgrading to newer versions of these libraries.

From a higher level, I'd also be interested in the approach that you used to find and correct these typos across the codebase.  If there's a package that you've found to work well in this role, we might consider integrating it into our CI process in the future.

rex4539:
Reverted the external stuff.

Regarding the tool, I used `codespell`. It catches most of the typos but not all. It needs some tweaking to avoid false positives but works pretty well in general and I know it has been integrated in many CI workflows.

source/MaterialXGenShader/UnitSystem.cpp
```diff
@@ -116,7 +116,7 @@ UnitTransform::UnitTransform(const string& ss, const string& ts, TypeDesc t, con
     }
 }
 
-const string UnitSystem::UNITSYTEM_NAME = "default_unit_system";
+const string UnitSystem::UNITSYSTEM_NAME = "default_unit_system";
```
ld-kerley, line 119: This change is going to change the name of a C++ symbol - so we might want to defer this to a time when we're breaking API.

jstone-lucasfilm (approved):
These changes look good to me, thanks @rex4539, and I believe they're safe to include in the upcoming v1.39.2 release.

ld-kerley (commented):
This really looks like fantastic work! 

I would be very exciting to see the project adopt using a tool like this in the CI to improve the quality of the source.

Could you add an example of the command you're running to correct the source here?  So we can investigate further how to automatically integrate this into our CI? 

There was one spot where a C++ symbol changes where I believe we should wait - for compatibility reasons.

rex4539:
I used the following command.

```
codespell -L keypair,keypairs,crate,wit,hist,parm,parms,iif,gost,wee,ser,unser,nin,nd,ba,fo,te,inout,upto,uptodate,optin,alloced,homogenous,cachable,crypted,tolen,inflight,toword,implementor,implementors,afterall
```

Basically, what you want is to try and avoid as much false positives as possible. Depending on the codebase, you can add more keywords to the command so the results are clearer.

source/MaterialXGenShader/UnitSystem.cpp
jstone-lucasfilm: This is a good point, @ld-kerley, though it's worth asking whether our shader generator modules should be held to the same API compatibility standards as core modules such as MaterialXCore and MaterialXFormat.

So far, we've been allowing API changes to our shader generator modules without providing backwards compatibility (e.g. https://github.com/AcademySoftwareFoundation/MaterialX/pull/1854, https://github.com/AcademySoftwareFoundation/MaterialX/pull/1867), but I'm open to the idea that we might impose additional rules to limit this in the future.

PR #2185, opened 2025-01-16
Make closure addition a data-driven node

ld-kerley:
Inspired by recent TSC conversations I decided to try and move one of the current C++ backed nodes to be data driven.

I believe this PR is mergable - but I encourage discussion around the direction taken.  If the community is in alignment with this approach I believe there are a number of other nodes that could be given a similar treatment - bringing us closer to being a data driven format.


jstone-lucasfilm (approved):
This looks like a great improvement, thanks @ld-kerley, and let's plan to merge this after the upcoming v1.39.2 release.

PR #2189, opened 2025-01-21
Update scikit-build-core

JeanChristopheMorinPerso:
We received an automated from PyPI:

>This email is notifying you of an upcoming deprecation that we have determined may affect you as a result of your recent upload to 'MaterialX'.
>
>In the future, PyPI will require all newly uploaded source distribution filenames to comply with [PEP 625](https://peps.python.org/pep-0625/). Any source distributions already uploaded will remain in place as-is and do not need to be updated.
>
>Specifically, your recent upload of 'MaterialX-1.39.2.tar.gz' is incompatible with PEP 625 because it does not contain the normalized project name 'materialx'.
>
>In most cases, this can be resolved by upgrading the version of your build tooling to a later version that supports PEP 625 and produces compliant filenames.
>
>If you have questions, you can email [admin@pypi.org](mailto:admin@pypi.org) to communicate with the PyPI admin@pypi.org to communicate with the PyPI administrators.

So this PR does exactly what this email is asking about.

While updating, I took the decision to drop the custom plugin and instead rely on a built-in regex plugin to extract the version from cmake. This is much faster and won't tie the project to a specific scikit-build-core version.


jstone-lucasfilm (approved):
Thanks for these updates, @JeanChristopheMorinPerso!

PR #2191, opened 2025-01-23
Make closure `layer`, `mix`, and `multiply` data-driven

ld-kerley:
Following on from #2185 - move the remaining closure C++ nodes to be data driven.


jstone-lucasfilm:
This looks like a great step forward, thanks @ld-kerley!

jstone-lucasfilm (approved):

Issue #1874, opened 2024-06-07
HwTimeNode hardcodes frame rate in generated code instead of using the fps uniform


JGamache-autodesk:
To test: connect a `ND_time_float` node to the base color of a standard_surface.

Looking at generated code, I see:
```C++
    float time_out = u_frame / 24.000000;
```
where I should instead see:
```C++
    float time_out = u_frame / time_fps;
```

This means changing the fps value is a topological change that requires recompiling the shader. 

Is this per specification (which means there is a missing `uniform="true"` metadata in the fps input of the time NodeDef), or is it something that should be fixed in [HwTimeNode.cpp](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/Nodes/HwTimeNode.cpp#L32) to use `shadergen.emitInput(fpsInput, context, stage);` instead of `shadergen.emitString(fps, stage);`?

ld-kerley:
I'm curious in what cases it's helpful to be able to vary the `fps` input.  I think I would expect it to generally be constant in a typical application.  Which would suggest that tagging the input as `uniform` would make the most sense to me?

jstone-lucasfilm:
I'd still be interested in the answer to @ld-kerley's question above, but I'll close out this task as resolved by #2192.  Thanks @JGamache-autodesk and @ld-kerley!

PR #2192, opened 2025-01-24
Implement `time` as a nodegraph in HW languages

ld-kerley:
Implement the `time` node as a nodegraph rather than custom C++ code in hardware shading languages.


jstone-lucasfilm:
This looks very promising, @ld-kerley!

I'm linking related issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/1874 from @JGamache-autodesk, as it's possible that your proposed simplification addresses this issue as well.

JGamache-autodesk:
This does indeed fix #1874. Thanks!

libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
```diff
@@ -176,7 +176,14 @@
   <implementation name="IM_frame_float_genglsl" nodedef="ND_frame_float" function="mx_frame_float" target="genglsl" />
 
   <!-- <time> -->
-  <implementation name="IM_time_float_genglsl" nodedef="ND_time_float" function="mx_time_float" target="genglsl" />
+  <nodegraph name="NG_time_float_genglsl" nodedef="ND_time_float" target="genglsl">
```
jstone-lucasfilm, line 179: @ld-kerley Since this implementation is now cross-platform, perhaps we should move it to this location in `stdlib_ng.mtlx`?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/stdlib_ng.mtlx#L2907

ld-kerley: The PR as it stands is not making things cross-platform.  The nodegraph here is only for the `genglsl` target to replace the `HwTime.cpp` c++ node.  I left the OSL/MDL implementations alone.

If you'd prefer I can try and update those other implementations as well - but I wasn't sure if that was a clear win here or not. Especially as MDL doesn't explicitly follow the same math - it appears to use an `animation_time()` function call.

jstone-lucasfilm: Ah, that makes sense!  In that case we may just want to update the description of the PR to clarify that it only applies to hardware shading languages.

jstone-lucasfilm (approved):

PR #2195, opened 2025-01-24
Remove ClosureContext params storage

ld-kerley:
This does not look to be used anywhere in the current code base.

Anyone who is relying on this external to MaterialX - please could you provide an example of the use-case, and ideally a MaterialX unit-test to demonstrate such.


jstone-lucasfilm:
This looks like a good simplification to me, @ld-kerley, and the only references that I see to `ScopedSetClosureParams` on GitHub are from earlier versions of the MaterialX codebase.

CC'ing @niklasharrysson just to make sure we're not missing anything important related to these classes.

PR #2196, opened 2025-01-25
Add bitangent input to `bump` node

pablode:
Starting with MaterialX 1.39, the `normalmap` node now has a bitangent input. It needs to be exposed for the bump node, because otherwise the tangent frame can't be set correctly.


jstone-lucasfilm (approved):
Good catch, @pablode, and this change looks correct to me.

ahamedy6 (commented):
ahamed

PR #2197, opened 2025-01-25
Remove legacy OSL helper files

jstone-lucasfilm:
This changelist removes helper files that were only referenced in the legacy OSL rendering path, which was removed in MaterialX v1.39.2.

As of OSL v1.12, the computation of directional albedo, energy compensation, and related terms is handled internally by the OSL implementations of the MaterialX-synchronized closures.


jstone-lucasfilm:

[MaterialXRenderTests_01_25_2025_GitHub.pdf](https://github.com/user-attachments/files/18548020/MaterialXRenderTests_01_25_2025_GitHub.pdf)

PR #2199, opened 2025-01-26
Add `Value::isEqual` to check for data equality

pablode:
Found the need to compare two mx::ValuePtrs and noticed that there is no straightforward way. Did I overlook something? If not, I'm proposing this helper function.


ld-kerley:
Seems like a nice addition to the API to me.

source/MaterialXCore/Value.cpp
```diff
@@ -333,6 +333,34 @@ string AggregateValue::getValueString() const
     return result;
 }
 
+bool AggregateValue::isEqual(ConstValuePtr other) const
+{
+  if (!other->isA<AggregateValue>())
```
jstone-lucasfilm, line 338: This change looks good to me from a high level, and the one issue that I notice is the use of 2-space indentation instead of the 4-space indentation that's standard in our C++ codebase.

jstone-lucasfilm (approved):

PR #2201, opened 2025-01-27
Fix MDL implementation of `fract`

krohmerNV:
Fix the MDL implementation to use the correct intrinsic. Amends #2169.


jstone-lucasfilm (approved):
Good catch, thanks @krohmerNV!

Issue #2202, opened 2025-01-27
Support abbe number in dielectric_bsdf


AdrienHerubel:
OpenPBR and Standard Surface define a [dispersion parametrization](https://academysoftwarefoundation.github.io/OpenPBR/#model/basesubstrate/translucentbase) based on the abbe number. The parametrization exists on the reference MaterialX graphs (https://github.com/AcademySoftwareFoundation/OpenPBR/blob/b212785f1e35462853c7d455ae5a41f8ce71bf98/reference/open_pbr_surface.mtlx#L38), but is not used or connected to any node making it difficult for implementers to use.

The abbe number should be exposed on the dielectric BSDF and implemented in each target.

PR #2203, opened 2025-01-27
Add abbe number to dielectric BSDF

AdrienHerubel:
This PR addresses the limitation described in #2202 and makes it possible to implement support for dispersion in the OpenPBR and Standard Surface node graphs.

### DIFF
--- a/libraries/pbrlib/pbrlib_defs.mtlx
+++ b/libraries/pbrlib/pbrlib_defs.mtlx
@@ -70,6 +70,7 @@
     <input name="tangent" type="vector3" defaultgeomprop="Tworld" />
     <input name="distribution" type="string" value="ggx" enum="ggx" uniform="true" />
     <input name="scatter_mode" type="string" value="R" enum="R,T,RT" uniform="true" />
+    <input name="abbe_number" type="float" value="0" uimin="0.0" uisoftmin="9.0" uisoftmax="91.0" />
     <output name="out" type="BSDF" />
   </nodedef>
 


portsmouth:
Things to note:
- The name `abbe_number` seems sensible.
- The definition of what it means requires some formulas. Those can be stated, or referenced, or alternatively, just leave it to the implementor to make sense of it (similar to the rest of the current spec).
- The default of 0 is taken to mean "no dispersion", i.e. technically it corresponds to infinite Abbe number..  Since we presumably want the default to be no dispersion, and 0 is the most obvious sentinel value. In OpenPBR we make the default 20, and have another scale parameter which divides that to generate the Abbe number, where the scale defaults to zero. For MaterialX, I think it makes more sense to just have the physical Abbe number, with the 0 sentinel meaning infinite Abbe i.e. zero dispersion.
- The implementation in `testrender` requires some more thought, as there are various ways to account for refractive dispersion (which differ depending on whether pathtracing or rasterizing). I take it this initial PR can just be for the change to the spec, and a `testrender` implementation can come later.

portsmouth:
In the MaterialX meeting, we noted:
- the Abbe number would make sense to add to the "generalized Schlick" BSDF as well. In this case the refractive index would map to the F0 of the Schlick model. (The input itself would be the same presumably).
- it would be good to loop in the OSL and MDL teams, to double check that this parametrization works for them.
- it would be good to outline here (and give some references) how implementations of refractive dispersion in various renderers might be done (pathtracing versus rasterization), to inform how this can be integrated into the various targets for testrender.
-

jstone-lucasfilm:
@AdrienHerubel @portsmouth I wanted to ping this discussion, as there was interest in moving forward with this feature at today's MaterialX TSC meeting.

As a starting point, should we simply extend Adrien's PR to include `generalized_schlick_bsdf` as well as `dielectric_bsdf`, so that we have a complete pull request to review with the community?

Additionally, I'm looping in @fpsunflower and @masuosuzuki for thoughts from the OSL and MDL perspectives.

PR #2204, opened 2025-01-29
Updates for OSL v1.14

jstone-lucasfilm:
- Call the new subsurface_bssrdf closure in OSL v1.14 and above, falling back to the original closure signature in earlier versions.
- Call the new chiang_hair_bsdf closure in OSL v1.14 and above, falling back to a dielectric_bsdf approximation in earlier versions.
- Inline the OSL implementations of displacement nodes for simplicity.
- Update the minimum OSL and MDL versions in libraries documentation.

### DIFF
--- a/libraries/README.md
+++ b/libraries/README.md
@@ -69,8 +69,8 @@ This folder contains the standard data libraries for MaterialX, providing declar
 
 ### Target Support
 - GLSL target support is for version 4.0 or higher.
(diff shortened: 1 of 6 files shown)


jstone-lucasfilm:

[MaterialXRenderTests_01_28_2025_GitHub.pdf](https://github.com/user-attachments/files/18581146/MaterialXRenderTests_01_28_2025_GitHub.pdf)

PR #2205, opened 2025-01-31
Simplify HW shader generation for closures

ld-kerley:
Currently all the GLSL BSDF functions are declared as separate functions per closure context (reflection, transmission, indirect, etc.).  This leads to some repetition in the GLSL code and some significant complexity in the HW Shader Generation code.

This PR refactors the HW Shader Generation to avoid the need for separate `ClosureContext` objects for each illumination stage, in favor of passing a single `ClosureData` object to each GLSL BSDF function, with a `closureType` enum indicating which illumination stage is being executed.

This allows the GLSL code to be refactored from something like 
```
void mx_burley_diffuse_bsdf_reflection(vec3 L, vec3 V, vec3 P, float occlusion, float weight, vec3 color, float roughness, vec3 normal, inout BSDF bsdf)
{
    bsdf.throughput = vec3(0.0);

    if (weight < M_FLOAT_EPS)
    {
        return;
    }

    normal = mx_forward_facing_normal(normal, V);

    float NdotV = clamp(dot(normal, V), M_FLOAT_EPS, 1.0);
    float NdotL = clamp(dot(normal, L), M_FLOAT_EPS, 1.0);
    float LdotH = clamp(dot(L, normalize(L + V)), M_FLOAT_EPS, 1.0);

    bsdf.response = color * occlusion * weight * NdotL * M_PI_INV;
    bsdf.response *= mx_burley_diffuse(NdotV, NdotL, LdotH, roughness);
}

void mx_burley_diffuse_bsdf_indirect(vec3 V, float weight, vec3 color, float roughness, vec3 normal, inout BSDF bsdf)
{
    bsdf.throughput = vec3(0.0);

    if (weight < M_FLOAT_EPS)
    {
        return;
    }

    normal = mx_forward_facing_normal(normal, V);

    float NdotV = clamp(dot(normal, V), M_FLOAT_EPS, 1.0);

    vec3 Li = mx_environment_irradiance(normal) *
              mx_burley_diffuse_dir_albedo(NdotV, roughness);
    bsdf.response = Li * color * weight;
}
```

to

```
void mx_burley_diffuse_bsdf(ClosureData closureData, float weight, vec3 color, float roughness, vec3 normal, inout BSDF bsdf)
{
    vec3 V = closureData.V;
    vec3 L = closureData.L;
    float occlusion = closureData.occlusion;

    bsdf.throughput = vec3(0.0);

    if (weight < M_FLOAT_EPS)
    {
        return;
    }

    normal = mx_forward_facing_normal(normal, V);

    float NdotV = clamp(dot(normal, V), M_FLOAT_EPS, 1.0);

    if (closureData.closureType == CLOSURE_TYPE_REFLECTION)
    {
        float NdotL = clamp(dot(normal, L), M_FLOAT_EPS, 1.0);
        float LdotH = clamp(dot(L, normalize(L + V)), M_FLOAT_EPS, 1.0);

        bsdf.response = color * occlusion * weight * NdotL * M_PI_INV;
        bsdf.response *= mx_burley_diffuse(NdotV, NdotL, LdotH, roughness);
    }
    else if (closureData.closureType == CLOSURE_TYPE_INDIRECT)
    {
        vec3 Li = mx_environment_irradiance(normal) *
                  mx_burley_diffuse_dir_albedo(NdotV, roughness);
        bsdf.response = Li * color * weight;
    }
}
```

Note : I have currently only performed this code consolidation for the burley diffuse file - but it should be trivial to apply to the other files once this is merged.

The other side effect of this, is a significant simplification of how the shader code is generated, bringing us a little closer again to being a data driven library.


source/MaterialXGenShader/HwShaderGenerator.cpp
```diff
@@ -153,6 +153,7 @@ const string PUBLIC_UNIFORMS                  = "PublicUniforms";
 const string LIGHT_DATA                       = "LightData";
 const string PIXEL_OUTPUTS                    = "PixelOutputs";
 const string DIR_N                            = "N";
+const string CLOSURE_DATA                     = "closureData";
```
niklasharrysson, line 156: Nitpick: Our convention has been to use upper-case for first letter in struct types.

source/MaterialXGenShader/HwShaderGenerator.cpp
```diff
@@ -494,52 +472,17 @@ ShaderPtr HwShaderGenerator::createShader(const string& name, ElementPtr element
 
 void HwShaderGenerator::emitFunctionCall(const ShaderNode& node, GenContext& context, ShaderStage& stage) const
 {
+    // TODO - revist later if we feel we need to keep the comment below - or if we can
+    // just use the similar implementation in ShaderGenerator::emitFunctionCall()
```
niklasharrysson, line 476: I think it's fine to omit that code comment generated here. So this whole method can be removed.

source/MaterialXGenShader/TypeDesc.cpp
```diff
@@ -133,6 +133,7 @@ TYPEDESC_REGISTER_TYPE(FILENAME, "filename")
 TYPEDESC_REGISTER_TYPE(BSDF, "BSDF")
 TYPEDESC_REGISTER_TYPE(EDF, "EDF")
 TYPEDESC_REGISTER_TYPE(VDF, "VDF")
+TYPEDESC_REGISTER_TYPE(CLOSUREDATA, "ClosureData")
```
niklasharrysson, line 136: I don't think it's necessary to register ClosureData as a standard/public data type. As far as I can see it's only being used by the HW shader generators, so it's an implementation detail of those generators.

niklasharrysson (commented):
This looks like an excellent change! 

Very nice simplification of the code, and a great move towards having more data driven code generation.

I only had a few minor comments above. When this has been merged I think it would be possible to simplify further and remove the ClosureContext altogether.

Thanks @ld-kerley!

source/MaterialXGenShader/HwShaderGenerator.cpp
ld-kerley: Here this string is used as the variable name in the shader code - not the struct type name - so I think the lower-case first letter is the convention here - if not - let me know I'll change it.

source/MaterialXGenShader/HwShaderGenerator.cpp
ld-kerley: removed

source/MaterialXGenShader/TypeDesc.cpp
ld-kerley: Moved to inside HwShaderGenerator code.

ld-kerley:
Thanks @niklasharrysson.  There are few more optimizations we can do after this PR lands for sure.  I wasn't quite sure the cleanest way to handle adding the Arguments to the function call site - but if you have suggestions I'd love to hear them.  I think removing the ClosureContext completely would be a nice win.

I also want to go through the GLSL code and refactor it all the same way as is done for the burley diffuse here - but wanted to get feedback from people on this change in principle before laboring through all that code.

jstone-lucasfilm:
This looks great to me, and I'll plan to run a new GLSL/OSL render comparison with these changes in place, in order to confirm that we're maintaining visual parity.

jstone-lucasfilm:
@ld-kerley Purely reflective materials look great in the latest render comparisons, but I'm seeing an unintended change in _transmissive_ materials, and I'll post some examples.

jstone-lucasfilm:
Here's the transmissive pawn top in our most recent GLSL/OSL comparison on January 28th:

image

jstone-lucasfilm:
And here's the same comparison with the changes in PR 2205:

image

jstone-lucasfilm:
Similarly, here's our GLSL/OSL comparison for Standard Surface Glass on January 28th:

image

jstone-lucasfilm:
And here's the same comparison with PR 2205:

image

jstone-lucasfilm:
Similarly, here's the GLSL/OSL comparison for Standard Surface Glass on January 28th:

image

jstone-lucasfilm:
And here's the same comparison with PR 2205:

image

jstone-lucasfilm:
It looks like we may be reusing one of our transmissive terms in reflections or vice versa, in a fashion that moves us further away from our reference renders in OSL.

ld-kerley:
@jstone-lucasfilm - thanks for running the comparisons - yes it turned out that there was a small visual regression.  

I was no longer correctly filtering out the respective closure categorization calls for BSDF nodes accurately.

I believe this is now fixed - at least I was able to a diff between the GLSL renders from `main` and this branch - and didn't see any visible diff at all. 

Note - this moved the handling of the `scatter_mode` input out of the shader generation C++ code, and in to the shader source - making it data-driven (another small win). 

I don't want to add yet more changes here - but I think it should be possible to similarly move the handling of the `bsdf` attribute on the noddefs to the shader source as well.

source/MaterialXGenShader/TypeDesc.cpp
niklasharrysson: My thinking was that you would not need a TypeDesc at all for the ClosureData, since it's an internal implementation detail of the HW generators.

But I see now that it is being used by the ClosureContext::Argument feature, so yes we need this for now, and it's good that you moved it inside the HwShaderGenerator code.

Once this PR is merged, we can look at removing the ClosureContext completely, and then this TypeDesc can also be removed. But let's look at that after this PR is merged to not complicate this change list too much.

source/MaterialXGenShader/HwShaderGenerator.cpp
niklasharrysson: Ah sorry! I thought this was a type name and not a variable name.
(This could also be removed later if we can simplify the argument handling and get rid of the ClosureContext)

jstone-lucasfilm:
Thanks for the updates, @ld-kerley, and the latest render comparison looks consistent with the main branch:

[MaterialXRenderTests_02_01_2025_GitHub.pdf](https://github.com/user-attachments/files/18630395/MaterialXRenderTests_02_01_2025_GitHub.pdf)

jstone-lucasfilm (approved):
This looks excellent, thanks @ld-kerley!

PR #2206, opened 2025-01-31
Unify HW shading implementations of geometric/convolution nodes

ld-kerley:
I noticed that the C++ implementations of these nodes were either identical, or nearly identical, so I think it makes sense to merge them together and provide a single HW shader version of these nodes.

GeomPropValueNode, GeomPropValueNodeAsUniform and GeomColorNode, the C++ implementation code is identical.

HeightToNormal and Blur both `#include` a separate shader source file that is language specific - so here the implementation registration uses a lambda function to wrap the differing filename in the `create()` call registered.

Note : Fix metal sobel sampling function to work with height to normal node, this provides a non-constant memory address space variable to the function.


jstone-lucasfilm:
This looks excellent, thanks @ld-kerley, and the only step I'd want to take before merging is to validate that our GLSL/OSL render comparisons remain consistent with this change.

jstone-lucasfilm:
Here's a new GLSL/OSL comparison with this PR, and everything looks consistent with previous comparisons:

[MaterialXRenderTests_PR_2206_GitHub.pdf](https://github.com/user-attachments/files/18626926/MaterialXRenderTests_PR_2206_GitHub.pdf)

jstone-lucasfilm (approved):

PR #2207, opened 2025-01-31
Simplify greysphere example materials

jstone-lucasfilm:
- Remove default-valued inputs from the Standard Surface Greysphere example materials, leaving only those inputs that have a visual impact.
- Decouple a JavaScript unit test from the specific element counts of example materials.


jstone-lucasfilm:

[MaterialXRenderTests_01_31_2025_GitHub.pdf](https://github.com/user-attachments/files/18625866/MaterialXRenderTests_01_31_2025_GitHub.pdf)

PR #2213, opened 2025-02-06
Add None check to RMS output in test report

krohmerNV:
Add None check to RMS output in the html test page script.
In case one of the two images for a diff is not present, the rms value is None and causes the next string concatenation to crash.


jstone-lucasfilm (approved):
Thanks for the fix, @krohmerNV!

PR #2214, opened 2025-02-06
Add support for MDL 1.10

krohmerNV:
Update the MDL code generator to make use of MDL 1.10 features that. This includes:
- taking the color82 of the generalized Schlick BSDF into account
- implement the Zeltner mode for the Sheen BSDF
- add energy conservation to the Oren-Nayar diffuse BSDF


jstone-lucasfilm (approved):

PR #2215, opened 2025-02-06
WIP: Open OPR - rewrite thin_film layering for compatibility with MDL

krohmerNV:
In the MDL code generator we currently can't handle layering constructs like this:
`layer(mix(bsdf1, bsdf2, weight), bsdf3)`
so I propose a workaround that should not change the result but uses an alternative implementation:
`mix(layer(bsdf1, bsdf3), layer(bsdf2, bsdf3), weight)`

### DIFF
--- a/libraries/bxdf/open_pbr_surface.mtlx
+++ b/libraries/bxdf/open_pbr_surface.mtlx
@@ -409,16 +409,19 @@
       <input name="thinfilm_thickness" type="float" nodename="thin_film_thickness_nm" />
       <input name="thinfilm_ior" type="float" interfacename="thin_film_ior" />
     </dielectric_bsdf>
-    <mix name="dielectric_reflection_tf_mix" type="BSDF">
-      <input name="fg" type="BSDF" nodename="dielectric_reflection_tf" />
-      <input name="bg" type="BSDF" nodename="dielectric_reflection" />
-      <input name="mix" type="float" interfacename="thin_film_weight" />
-    </mix>
-    <layer name="dielectric_base" type="BSDF">
-      <input name="top" type="BSDF" nodename="dielectric_reflection_tf_mix" />
+    <layer name="dielectric_base_tf" type="BSDF">
+      <input name="top" type="BSDF" nodename="dielectric_reflection_tf" />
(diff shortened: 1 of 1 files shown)


libraries/bxdf/open_pbr_surface.mtlx
```diff
@@ -409,16 +409,19 @@
...
       <input name="thinfilm_ior" type="float" interfacename="thin_film_ior" />
     </dielectric_bsdf>
-    <mix name="dielectric_reflection_tf_mix" type="BSDF">
-      <input name="fg" type="BSDF" nodename="dielectric_reflection_tf" />
-      <input name="bg" type="BSDF" nodename="dielectric_reflection" />
-      <input name="mix" type="float" interfacename="thin_film_weight" />
-    </mix>
-    <layer name="dielectric_base" type="BSDF">
-      <input name="top" type="BSDF" nodename="dielectric_reflection_tf_mix" />
+    <layer name="dielectric_base_tf" type="BSDF">
```
ld-kerley, line 412: Might be good to leave a comment in here as to why this is the way it is - so that someone else doesn't accidentally refactor it back to the existing simpler representation.

krohmerNV: Sure, a comment is a good idea.
Question still is if we want to change it like I proposed.
Do you see alternatives? Do we have some functionality to do such a transformation in the beginning of the code gen?

jstone-lucasfilm: @krohmerNV I like your suggestion to make this an optional transformation that we would add to the MaterialX API, and which the MDL shader generator would leverage just before the usual generation step.  This would allow the original definition of OpenPBR to stand unchanged, and the transformation step could even be removed as the feature set of MDL evolves in the future.

krohmerNV: I applied the same pattern you did for #2539 and it seems to work well (as an intermediate solution as well).

jstone-lucasfilm: That's a great idea, @krohmerNV, and with this latest refactoring, I think it makes good sense to include this refinement in 1.39.4.

PR #2217, opened 2025-02-07
Make ClosureData struct and CLOSURE_TYPE defines data-driven

ld-kerley:
While testing the latest dev branch with the latest USD branch I realized that the recent PR #2205 doesn't work without similar changes to the HdStorm shader generator.  Taking this data driven approach should mean it works without any changes on the HdStorm side.


jstone-lucasfilm (approved):
This change looks good to me, thanks @ld-kerley, and it should be fine to merge this once all tests pass.

PR #2219, opened 2025-02-07
Refactor to remove the ClosureContext class

ld-kerley:
Following on from the previous work to simplify how closures are handled in HW shader generators.

This PR removes the `ClosureContext` class completely.

The creation of the `ClosureData` parameter and passing the argument to the closure HW functions is hardcoded in the cases it's needed. 

@niklasharrysson - would certainly appreciate your eyes on this. Specifically the handling of the HW light nodes - as I'm not sure we have solid test coverage here.


source/MaterialXGenShader/Nodes/ClosureCompoundNode.cpp
```diff
@@ -26,114 +26,76 @@ void ClosureCompoundNode::emitFunctionDefinition(const ShaderNode& node, GenCont
...
 
-    // Begin function signature
-    shadergen.emitLineBegin(stage);
-    if (cct)
-    {
         shadergen.emitString("void " + _functionName  + "(", stage);
 
-        // Add any extra argument inputs first
-        for (const ClosureContext::Argument& arg : cct->getArguments())
+        // Check if we have a closure context to modify the function call.
```
niklasharrysson, line 41: Should this comment be changed?

source/MaterialXGenShader/Nodes/ClosureSourceCodeNode.cpp
```diff
@@ -39,15 +40,9 @@ void ClosureSourceCodeNode::emitFunctionCall(const ShaderNode& node, GenContext&
             shadergen.emitString(_functionName + "(", stage);
 
             // Check if we have a closure context to modify the function call.
```
niklasharrysson, line 42: Same code comment here.

niklasharrysson (approved):
This looks good to me. Nice cleanup!

I only left a few minor remarks.

As for test coverage or light nodes, as far as I know there are always light nodes included in the test suite GLSL rendering. The test suite options has a setting of which light rig to use when rendering, see here:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/_options.mtlx#L20

jstone-lucasfilm (approved):

PR #2218, opened 2025-02-07
Simplify additional HW closures

jstone-lucasfilm:
Building upon recent work in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2205, this changelist simplifies the implementations of the `conductor_bsdf`, `dielectric_bsdf`, and `oren_nayar_diffuse_bsdf` closures in hardware shading languages, using a single function to handle all closure types.

Additionally, it harmonizes on the name `N` for normal inputs in HW closures, making it more straightforward to compare code across implementations.

PR #2221, opened 2025-02-08
Simplify remaining HW closures

jstone-lucasfilm:
Building upon recent work in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2218, this changelist simplifies the implementations of the `chiang_hair_bsdf`, `generalized_schlick_bsdf`, `sheen_bsdf`, `subsurface_bsdf`, and `translucent_bsdf` closures in hardware shading languages, using a single function to handle all closure types.


jstone-lucasfilm:
Attaching a new set of GLSL/OSL render comparisons using the codebase in this PR, along with the latest energy conservation improvements in OSL:

[MaterialXRenderTests_02_08_2025_GitHub.pdf](https://github.com/user-attachments/files/18723207/MaterialXRenderTests_02_08_2025_GitHub.pdf)

libraries/pbrlib/genglsl/mx_translucent_bsdf.glsl
```diff
@@ -25,19 +9,20 @@ void mx_translucent_bsdf_indirect(vec3 V, float weight, vec3 color, vec3 normal,
...
     }
 
+    vec3 V = closureData.V;
+    vec3 L = closureData.L;
+
     // Invert normal since we're transmitting light from the other side
-    vec3 Li = mx_environment_irradiance(-normal);
-    bsdf.response = Li * color * weight;
-}
+    N = -mx_forward_facing_normal(N, V);
```
ld-kerley, line 16: In the previous code the reflection component was not using the `mx_face_forward_normal()` call but instead just inverting the normal - The normal is provided as a user input - so could this change lead to an image change if a user is providing an unusual normal input?

libraries/pbrlib/genglsl/mx_translucent_bsdf.glsl
```diff
@@ -25,19 +9,20 @@ void mx_translucent_bsdf_indirect(vec3 V, float weight, vec3 color, vec3 normal,
...
 
-void mx_translucent_bsdf(ClosureData closureData, float weight, vec3 color, vec3 normal, inout BSDF bsdf)
-{
     if (closureData.closureType == CLOSURE_TYPE_REFLECTION)
     {
-        mx_translucent_bsdf_reflection(closureData.L, closureData.V, closureData.P, closureData.occlusion, weight, color, normal, bsdf);
+        float NdotL = clamp(dot(N, L), M_FLOAT_EPS, 1.0);
```
ld-kerley, line 20: Is this clamp the same as the guard used previously?  if the normal input vector isn't normalized then perhaps its not?  

I think this code probably is better - but maybe it's better to land this refactor without changing behavior and making the behavior change in a separate PR for clarity?

libraries/pbrlib/genglsl/mx_translucent_bsdf.glsl
jstone-lucasfilm: This is a great observation, @ld-kerley, and I gave a good deal of thought to this as well.

As far as I can tell, the missing call to `mx_face_forward_normal` was just an oversight in the original implementation of the reflection term of `mx_translucent_bsdf`, and it wouldn't make sense to ignore this step when rendering this lobe in hardware shading languages.

Since thin-walled materials are relatively rare in MaterialX material libraries, it's likely that this oversight was never caught before, and is only being noticed now.

Before moving ahead with this change, though, I'll put together a thin-walled test material, just to make sure it behaves as expected with this pull request.

libraries/pbrlib/genglsl/mx_translucent_bsdf.glsl
jstone-lucasfilm: You're correct that this simplification goes a bit beyond the scope of the other changes here, in part because `mx_translucent_bsdf` was written less consistently than the other BSDF nodes in hardware shading languages.

My thinking is that this is a good opportunity to bring `mx_translucent_bsdf` into closer line with the other nodes, and I can add some thin-walled example materials as part of this pull request, so that we have a better baseline for thin-walled materials in the future.

libraries/pbrlib/genglsl/mx_translucent_bsdf.glsl
jstone-lucasfilm: After taking a close look at our `subsurface.mtlx` example file, which contains an example of thin-walled rendering, I think your original take is correct, and we should leave the interpretation of the inverted normal alone in this change.

I've removed the `mx_forward_facing_normal` call from this function, and the visual interpretation now appears to exactly match the original.

libraries/pbrlib/genglsl/mx_translucent_bsdf.glsl
jstone-lucasfilm: Since this change doesn't have a visual impact in my tests of `subsurface.mtlx`, and it brings the subsurface node into closer alignment with other BSDF nodes, my thinking is that we can keep this minor change.

As noted above, I've reverted the addition of `mx_forward_facing_normal`, since this has a visual impact on subsurface materials, and we don't want this specific PR to change visuals.

ld-kerley: I'm still wondering if the change to `clamp(dot(N,L)....` here instead of 
```
    float NdotL = dot(L, -normal);
     if (NdotL <= 0.0 || weight < M_FLOAT_EPS)
     {
         return;
     }
```

is the same in all cases.

Doesn't the clamp leave us with a tiny amount of `bsdf.response` if the light is facing away from the normal?  Which would seem odd - I think?

jstone-lucasfilm: Good point, @ld-kerley, and here I'm following the pattern for other dot products in our GLSL codebase, though in this specific case there's no mathematical motivation for clamping to epsilon rather than zero.

I'll go ahead and change this from epsilon to zero here, and we can consider other adjustments in future changes.

PR #2222, opened 2025-02-10
Add CONFIGURE_DEPENDS to CMake GLOB

ld-kerley:
Add CONFIGURE_DEPENDS to glob statements so they update incrementally, this means while actively developing if a file is added the cmake configure stage doesn't need to be re-run.  CMake recommendation is not to use GLOB to collect source files, but given that the MaterialX project does, this feels like a good compromise to make life a little easier for the developers. 

Also simplify the Metal source inclusion for MaterialXViewer


jstone-lucasfilm (approved):

Issue #2025, opened 2024-09-25
Graph Editor: Walk to connected nodes via hotkeys


lfl-eholthouser:
With large node graphs and overlapping links it can be difficult to quickly determine which nodes are actually connected. I think the ability to find connected nodes based off a selected link would be beneficial. With a link selected clicking the right arrow key could select the downstream node while the left arrow key could select the upstream node. This could also be expanded to include the ability to walk the entire node graph though it would need to be a specific patter for nodes that have multiple output connections.

jstone-lucasfilm:
Just noting here that Camille Ramseur from Adobe plans to take a look at this issue for Dev Days 2024, and we'll assign the task to her once we have her GitHub ID.

camilleramseur:
Thanks Jonathan. I'm having trouble using my Adobe GitHub account to log in here so I'm just going to use my personal account. I was told it was ok.

camilleramseur:
I don't think I will have this finished by the end of dev days but I would still like to work on this if possible. I'm still new to MaterialX. Do you have any recommendations on how to go about and start this? I was looking in source/MaterialXGraphEditor and was trying to find the keyboard shortcuts mentioned here to see the cod and patterns that were done previous here: https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/GraphEditor.md

It looks like I might potentially have to me do this in the graph.cpp file ? Any advice is appreciated.

lfl-eholthouser:
Hi @camilleramseur ! I think working in the graph.cpp makes sense. Some of the different keyboard shortcuts can be found here: https://github.com/AcademySoftwareFoundation/MaterialX/blob/70cb28de8b313371462eb9f0f3e4ec31e903e4a7/source/MaterialXGraphEditor/Graph.cpp#L3113 
Then here is the hotkey for focusing on a node https://github.com/AcademySoftwareFoundation/MaterialX/blob/70cb28de8b313371462eb9f0f3e4ec31e903e4a7/source/MaterialXGraphEditor/Graph.cpp#L3737.

I think something starting with something like the above, checking if a link is selected and then a right or left arrow key has been released. Is that has been clicked then selecting that node and navigating to that selection.

kwokcb:
Note that I wrote the logic for traversal in this PR. I did not hook it up to any hot-keys.
https://github.com/AcademySoftwareFoundation/MaterialX/pull/2225

[...]

PR #2225, opened 2025-02-12
Node editor traversal

kwokcb:
## Proposal

- Add in clickable links to traverse through pin connections (inputs / outputs)
- This gives an easy way to move through a graph and also provide information on what is being routed in and out of a node
- Useful for diagnosis (see bugs section below)
- Options: Allow for output connections to be displayed with a per node toggle.
- Did not add hotkeys as selection of what to traverse to up / downstream is ambiguous for multiple input / output connections on a node. Can consider later and hook to keys like left/right arrow for instance.

- Aside: Add in background graph color option as it's pretty hard to see selected outlines, and better contrast with graph elements, grid etc. (`--graphBackground "r,g,b")

### Details
- Currently pin connections show the type which gives no information as to what is connected to it. You can hover over the pins in the graph but this just gives you pin names w/o node names
- Instead add in clickable buttons to traverse between pins.
- Traversal includes selecting the centering around selected node
- There are some existing bugs found in the code:
  - Deletion does not refresh connections properly
  - Inputs to graphs are not properly updating.
- Prevent a possible crash when deleting. Didn't fix the issue
- Fix positioning of "Show all inputs" for `nodegraph`s. (missed from a previous submission)
  
### Results 

https://github.com/user-attachments/assets/9b3f157e-dbca-44f4-a156-2706a2d0928e

1. Fan-out from an output:
image

2. Upstream connections:
image

3. Hide output connections: 
image

---
Pinging @lfl-eholthouser  and @jstone-lucasfilm for comments.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3494,8 +3494,39 @@ void Graph::propertyEditor()
                             }
                             else
                             {
-                                std::string typeText = " [" + input->_input->getType() + "]";
-                                ImGui::Text("%s", typeText.c_str());
+                                std::string displayString = input->_input->getType();
+
+                                const std::vector<UiPinPtr>& connections = input->getConnections();
+                                std::shared_ptr<UiNode> pinNode = nullptr;
+                                if (!connections.empty())
+                                {
+                                    UiPinPtr pin = connections[0];
+                                    std::string pinName = std::string(pin->_name);
+
+                                    pinNode = pin->_pinNode;
+                                    if (pinNode)
+                                    {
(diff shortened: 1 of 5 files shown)


jstone-lucasfilm:
@kwokcb Like so many folks on the MaterialX Slack, I'm really struck by the proposed color change, and I wonder if you could separate this out into its own PR, so that we can evaluate these two distinct proposals separately.

kwokcb:
This no longer has color background logic. I does do the same traversal discovery required for #2025 though different UI interface.

kwokcb:
This has been sitting here for a very long time. Is there interest in adding this in. If not I will close this as I already have this in my own editor.

PR #2228, opened 2025-02-14
Add null check in getShaderNodes

jstone-lucasfilm:
This changelist adds a missing null check in getShaderNodes, handling the edge case of an invalid output string.  Previously, this edge case would trigger a crash in MaterialXCore, and now it correctly generates validation warnings and proceeds.

### DIFF
--- a/source/MaterialXCore/Material.cpp
+++ b/source/MaterialXCore/Material.cpp
@@ -50,7 +50,11 @@ vector<NodePtr> getShaderNodes(NodePtr materialNode, const string& nodeType, con
             vector<OutputPtr> outputs;
             if (input->hasOutputString())
             {
-                outputs.push_back(nodeGraph->getOutput(input->getOutputString()));
+                OutputPtr connectedOutput = nodeGraph->getOutput(input->getOutputString());
+                if (connectedOutput)
+                {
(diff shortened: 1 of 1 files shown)


ld-kerley:
In cases like this where we find a bug I think we should add a unit test to exercise the fix at the same time as submitting the fix - that way we will try and ensure that we don't regress.   Do you think we can add a small unit test here to simulate the bad input data?

jstone-lucasfilm:
I was thinking along the same lines, @ld-kerley, though we'll need to create a bit more infrastructure to allow for "invalid" documents in our TestSuite.  Two potential paths forward come to mind:

1. We could create a new "invalid" folder in our TestSuite, which would be loaded along with other examples during testing, but would not require a passing result in validation.  As of today, we require all of our TestSuite examples to pass validation, so adding even a single invalid file would be a breaking change.
2. We could extend our [fuzz testing](https://github.com/AcademySoftwareFoundation/MaterialX/blob/7ac1c71de5187dc29793292b5a8dc6d784192ecf/source/MaterialXTest/MaterialXFormat/XmlIo.cpp#L265) system so that it consistently generates this category of error procedurally.  This would require more work, but would potentially catch a much wider set of error conditions, not just the invalid output string that inspired this fix.

PR #2226, opened 2025-02-12
Improve Data Library incremental build

ld-kerley:
This PR updates how cmake handles the data library files. Previously they were just being copied during the INSTALL stage, which meant that incremental updates would not always accurately be respected.  This PR changes that to BUILD the data library files and then install that built folder when requested.

PR #2235, opened 2025-02-17
Update CMake minimum version to 3.26

ld-kerley:
When working on #2226 I inadvertently introduce the use of `copy_directory_if_different` which doesn't exist until CMake 3.26.  An alternate implementation failed the Windows CI - so bumping the minimum version seems like the easiest path forwards.


jstone-lucasfilm (commented):
This looks good to me, thanks @ld-kerley, though I'd like to bring this up on the MaterialX Slack just to double-check that teams are OK with the change.

jstone-lucasfilm (approved):
This change looks good to me, @ld-kerley, and we have confirmation from our OpenUSD colleagues that it works for them as well.

PR #2237, opened 2025-02-18
Refactor use of  `copy_directory_if_different`

ld-kerley:
While working on #2226 I inadvertently introduced the use of `copy_directory_if_different` to our CMake.  This feature is only present in CMake 3.26, and our current minimum CMake version is 3.24.

This PR refactors the data library build to avoid the use of `copy_directory_if_different`, so we can stick with our current CMake minimum version of 3.24.


jstone-lucasfilm:
Thanks for providing this alternative implementation, @ld-kerley, and this should now be resolved with the merge of https://github.com/AcademySoftwareFoundation/MaterialX/pull/2235.

PR #2241, opened 2025-02-20
Remove references to CONFIGURE_DEPENDS

jstone-lucasfilm:
This changelist removes references to CONFIGURE_DEPENDS in our CMake configurations, as this feature is not well-implemented across our target platforms.

From the CMake documentation:

> The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.


ld-kerley:
I'm on board with this change.  It is generally well supported on linux/Mac using Ninja as the generator, windows support appears to incur a performance hit.

I will note that I originally added this option to address the use of `file(GLOB)` to gather the sources. This is not advisable best practice according to the CMake documentation, and instead its recommended to list the source files.  If `CONFIGURE_DEPENDS` is problematic for some target platforms, I do think we should consider listing the sources and bringing the projects CMake structure up to recommended best practices.

ld-kerley (approved):

ld-kerley:
Another option might be to provide a wrapper around `file(GLOB)` that can optionally enable `CONFIGURE_DEPENDS` for people who are working platforms where it does have successful support.

PR #2243, opened 2025-02-22
Static analyis fixes

jstone-lucasfilm:
This changelist addresses two static analysis warnings flagged by PVS-Studio:

- Remove an unneeded const modifier from the return argument of `TypeDesc::getStructMembers`.
- Remove an unused intermediate value in `Graph::buildGroupNode`.


ld-kerley (commented):
Looks good to me

PR #2245, opened 2025-02-23
Re-add OSL normalmap tangent frame orthogonalization

pablode:
This PR reverts the removal of OSL tangent frame orthogonalization (caused by PR #1911) and adds an explanatory comment.

### DIFF
--- a/libraries/stdlib/genosl/mx_normalmap.osl
+++ b/libraries/stdlib/genosl/mx_normalmap.osl
@@ -13,7 +13,7 @@ void mx_normalmap_vector2(vector value, vector2 normal_scale, vector N, vector T
         //
         vector v = value * 2.0 - 1.0;
         vector Tn = normalize(T - dot(T, N) * N);
-        vector Bn = normalize(cross(N, Tn));
+        vector Bn = normalize(B - dot(B, N) * N - dot(B, Tn) * Tn);
         result = normalize(Tn * v[0] * normal_scale.x + Bn * v[1] * normal_scale.y + N * v[2]);
     }
 }


libraries/stdlib/genosl/mx_normalmap.osl
```diff
@@ -1,19 +1,21 @@
...
+        vector T = normalize(U - dot(U, N) * N);
+        vector B = normalize(cross(N, T));
+        result = normalize(T * v[0] * normal_scale.x + B * v[1] * normal_scale.y + N * v[2]);
     }
-
-    result = normalize(T * decodedValue[0] * normal_scale.x + B * decodedValue[1] * normal_scale.y + N * decodedValue[2]);
 }
 
-void mx_normalmap_float(vector value, float normal_scale, vector N, vector T, vector B, output vector result)
+void mx_normalmap_float(vector value, float normal_scale, vector N, vector U, vector B, output vector result)
```
klucknav, line 18: Just as we renamed T -> U should we also rename B -> V?

jstone-lucasfilm: My own sense is that we should leave the input names unchanged, as Pablo's existing documentation provides clarity on what differs in this implementation, and what improvements remain for the future.

libraries/stdlib/genosl/mx_normalmap.osl
```diff
@@ -1,16 +1,21 @@
...
     {
-        decodedValue = value * 2.0 - 1.0;
+        // The OSL backend uses dPdu and dPdv for tangents and bitangents, but these vectors are not
+        // guaranteed to be orthonormal.
+        //
+        // Orthogonalize the tangent frame using Gram-Schmidt, unlike in the other backends.
+        //
+        vector v = value * 2.0 - 1.0;
+        vector Tn = normalize(T - dot(T, N) * N);
+        vector Bn = normalize(cross(N, Tn));
```
pablode, line 16: Just realized that this line needs to be adjusted to accommodate the explicit bitangent..

jstone-lucasfilm: Ah, that makes sense.  Did you want to make that additional modification, since you have the best visibility on the explicit bitangent use case?

pablode: (Would appreciate if you could commit/test this, since building OSL is a nightmare on macOS/Windows..)

jstone-lucasfilm: No problem, @pablode, and I'll propose that change from here.  Once we have a proposal that looks correct to everyone, I can run the full suite of GLSL/OSL comparisons for validation.

jstone-lucasfilm: @pablode I've heard from @klucknav at Pixar that this latest change breaks one of the examples in their test suite, though I'm not fully looped in on the details.

jstone-lucasfilm: Scratch that last comment, and I've heard from @klucknav that this latest version works in Pixar tests.  Let's go ahead and merge this for the release candidate tomorrow.

PR #2246, opened 2025-02-24
Restore CMake exclusion to data install

jstone-lucasfilm:
This changelist restores the exclusion of CMakeLists.txt to the data library installation process, aligning the root folder of the installation with that of MaterialX v1.39.2.


ld-kerley (approved):
I will note that one of the benefits of building the library is that we simplify the cmake install stage - because building gives us the chance to more appropriately filter the files.

PR #2248, opened 2025-02-25
Add bitangent input check to `normalmap` upgrade

klucknav:
This PR adds a check to the normalmap upgrade function to make sure that we don't try to add a bitangent input if one is already set.


jstone-lucasfilm (approved):
Thanks for the fix, @klucknav, and this change looks good to me.

PR #2258, opened 2025-03-01
Update example material for 1.39

jstone-lucasfilm:
This changelist updates the `usd_normal_map.mtlx` example for MaterialX 1.39, replacing a legacy reference to object-space `normalmap` with explicit decoding math.


jstone-lucasfilm:
GLSL/OSL render comparison of the updated example material:

RenderComparison_UsdNormalMap

PR #2259, opened 2025-03-02
Use relative directories in CMake

DeclanRussell:
TLDR: This PR switches the `CMAKE_BINARY_DIR` and `CMAKE_SOURCE_DIR` variables to `MaterialX_BINARY_DIR` and `MaterialX_SOURCE_DIR` respectably.

The motivation behind this is that I want to build MaterialX as a sub directory of another CMake project (i.e. `add_subdirectory(contrib/MaterialX EXCLUDE_FROM_ALL)`). As I understand, using the non-MaterialX versions of these variables means that the variable directory values get set from the original CMake invocation directory, which in my case, it a parent project. Therefore, it could break some paths in the MaterialX build. As I understand, using the MaterialX variants of these variables should guard against this. I'm not a CMake expert though, so happy to admit if I've misunderstood something here.

I've submitted the CLA request (I work at Autodesk), I'm just waiting for approval.


ld-kerley:
I think this is a change for the better - thanks for putting up the PR.

Poking around a few online recommendations, it appears that convention suggests using `PROJECT_SOURCE_DIR` and `PROJECT_BINARY_DIR` where possible instead of referencing the project name directly.  I guess perhaps thats a little more maintainable moving forwards if the project name ever changed.  If that works for your sub-module use-case it might be a little cleaner.

jstone-lucasfilm:
Thanks for this proposal, @DeclanRussell, and there are details in the EasyCLA comment above on resolving the remaining warning:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/2259#issuecomment-2692799511

DeclanRussell:
> it appears that convention suggests using `PROJECT_SOURCE_DIR` and `PROJECT_BINARY_DIR` where possible instead of referencing the project name directly.

Good idea. This also works for the submodule case. I've made the change in 77d5642.

DeclanRussell:
> there are details in the EasyCLA comment above on resolving the remaining warning:

Ahh thanks for the heads up, all done 🙂

jstone-lucasfilm (approved):

PR #2261, opened 2025-03-05
Override the aggregated struct printing for MDL code gen

krohmerNV:
Add a custom StructTypeSyntax implementation for MDL.


jstone-lucasfilm:
Thanks for this fix, @krohmerNV, and this change looks good to me.

Since this change is scoped to MDL shader generation, and it was written by the domain expert in this system, I would be inclined to see this as safe to include in our upcoming v1.39.3 release.

What are your thoughts, @ld-kerley?

jstone-lucasfilm:
@krohmerNV It looks as if we've lost your EasyCLA approval over the last day, and let me know if you run into any trouble restoring it!

krohmerNV:
@jstone-lucasfilm is there anything you can do about CLA? I requested access as compony contributor again. Maybe I need to reach out again internally.

jstone-lucasfilm:
@krohmerNV Alas, I don't have any direct influence over EasyCLA approvals, but the Linux Foundation usually responds quickly when you fill out a support ticket, as described in the comment from their automated system:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/2261#issuecomment-2702109835

jstone-lucasfilm (approved):

Issue #2127, opened 2024-12-02
Broken OSL texture colorspace in 1.38.10


AdrienHerubel:
It seems that after #1757 the color space of a file is not passed to the shadernode input anymore. This breaks OSL renderers relying on having the color space attribute passed to the texture call for fully-featured OCIO transforms and filtering aware color conversion. 

The following simple patch restores the functionality but might cause a double conversion if a user is using emitted transforms on top of the OSL texture color transform. Maybe it would be better to not optionally not emit color transform for files when using the OSL generator by default ?

```
diff --git a/source/MaterialXGenShader/ShaderGraph.cpp b/source/MaterialXGenShader/ShaderGraph.cpp
index b95a9471..c2d73755 100644
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -654,6 +654,8 @@ void ShaderGraph::applyInputTransforms(ConstNodePtr node, ShaderNodePtr shaderNo
                     }
                 }
 
+                ShaderInput* shaderInput = shaderNode->getInput(input->getName());
+                shaderInput->setColorSpace(sourceColorSpace);
                 ShaderOutput* shaderOutput = shaderNode->getOutput();
                 populateColorTransformMap(colorManagementSystem, shaderOutput, sourceColorSpace, targetColorSpace, false);
                 populateUnitTransformMap(unitSystem, shaderOutput, input, targetDistanceUnit, false);
```

kwokcb:
From what I recall it is either the transform is inserted into the shader code or the "colorspace" metadata is added to the `texture` call. 

If it's possible to get double transforms than I agree that by default only one should be allowed with "colorspace" being emitted in the `texture` call by default. 

I think when this was added OSL had not added support for this and this only worked for Arnold, thus the case may have not been testable with `testrender` as part of unit tests.

boberfly:
Hi,

Just to note the "colorspace" optional argument call to both Cycles 4.4 and Renderman 26.1's implementations of `texture()` causes a warning in the former and an error in the latter. Is this something that could be a config toggle before generating the OSL code?

I am using a few commits ahead of 1.39.2 (notably https://github.com/AcademySoftwareFoundation/MaterialX/commit/fc388e41b0cc10db197c949bc7ff2eb1c6be1966 )

Cheers

jstone-lucasfilm:
[...]

PR #2263, opened 2025-03-07
Added oslFileTextureColorspace GenOptions.

boberfly:
Added oslFileTextureColorspace GenOptions to omit the colorspace optional argument to the texture function.

I commented in https://github.com/AcademySoftwareFoundation/MaterialX/issues/2127#issuecomment-2705692762 about it, but I thought I'd figure it out myself.

I couldn't think of another way to approach it but this seems to allow generating the nodes without the optional argument to texture() which causes errors in some renderers.

eg. ERROR : RenderMan : S20006 OSL: Unknown texture optional argument: "colorspace", <string> (shaders\\__mtlx\\__ND_image_color3.osl:35)

Kind regards
-Alex

### DIFF
--- a/source/MaterialXGenShader/GenOptions.h
+++ b/source/MaterialXGenShader/GenOptions.h
@@ -93,7 +93,8 @@ class MX_GENSHADER_API GenOptions
         hwNormalizeUdimTexCoords(false),
         hwWriteAlbedoTable(false),
         hwWriteEnvPrefilter(false),
-        hwImplicitBitangents(true)
+        hwImplicitBitangents(true),
+        oslFileTextureColorspace(true)
     {
     }
     virtual ~GenOptions() { }
@@ -191,6 +192,10 @@ class MX_GENSHADER_API GenOptions
     /// Calculate fallback bitangents from existing normals and tangents
     /// inside the bitangent node.
     bool hwImplicitBitangents;
+
+    /// OSL-specific override to enable/disable the optional argument
+    /// to OSL's texture() function.
+    bool oslFileTextureColorspace;
 };
 
 MATERIALX_NAMESPACE_END
--- a/libraries/stdlib/genosl/mx_image_color3.osl
+++ b/libraries/stdlib/genosl/mx_image_color3.osl
@@ -15,6 +15,6 @@ void mx_image_color3(textureresource file, string layer, color default_value, ve
     out = texture(file.filename, st.x, st.y,
                   "subimage", layer, "interp", filtertype,
                   "missingcolor", missingColor,
-                  "swrap", uaddressmode, "twrap", vaddressmode,
-                  "colorspace", file.colorspace);
+                  "swrap", uaddressmode, "twrap", vaddressmode
+                  $oslFileTextureColorspace);
 }
--- a/libraries/stdlib/genosl/mx_image_color4.osl
+++ b/libraries/stdlib/genosl/mx_image_color4.osl
@@ -17,8 +17,8 @@ void mx_image_color4(textureresource file, string layer, color4 default_value, v
     color rgb = texture(file.filename, st.x, st.y, "alpha", alpha,
                         "subimage", layer, "interp", filtertype,
                         "missingcolor", missingColor, "missingalpha", missingAlpha,
-                        "swrap", uaddressmode, "twrap", vaddressmode,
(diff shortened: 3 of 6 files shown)


jstone-lucasfilm:
@boberfly This looks like an important fix to implement in our OSL shader generation, but I wonder if it might be better to leverage the built-in `OSL_VERSION_MAJOR` and `OSL_VERSION_MINOR` defines to achieve this.

An example of this approach can be seen in this earlier pull request, which allows MaterialX to support OSL 1.14 without losing support for earlier versions:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/2204

boberfly:
Hey @jstone-lucasfilm 

I was thinking this, but I was looking through the source code and documentation of OSL and couldn't find `colorspace` as a valid optional argument for `texture()`, it might just be an Arnold-specific addition? Maybe future renderers will adopt it I'm not sure though.

kwokcb:
This was indeed introduced as a Arnold request, but the proposal was to make this part of the OSL spec itself. I remember it was something being looked at but don't know what happened here. Adding @lgritz for comment.

lgritz:
I think maybe we added it to the texture system but didn't wire up all of the OSL side? I'll check on it.

jstone-lucasfilm:
@boberfly @lgritz It's true that I don't see the colorspace input documented in the specification for OSL 1.13.10, and if it's not yet a standard feature, then I'd support @boberfly's suggestion to enable it with a shader generator option.  If it's perhaps a _hidden_ feature, though, then we could enable it based on the detected OSL version defines.

jstone-lucasfilm:
From additional discussions with @lgritz, it sounds as if this proposal is indeed the right solution, as colorspace support isn't yet guaranteed above a specific version of OSL.

@boberfly Let's continue focusing on this PR as the recommended path forward, and I'll take a closer look at the code when time permits.

source/MaterialXGenShader/GenOptions.h
```diff
@@ -191,6 +192,10 @@ class MX_GENSHADER_API GenOptions
     /// Calculate fallback bitangents from existing normals and tangents
     /// inside the bitangent node.
     bool hwImplicitBitangents;
+
+    /// OSL-specific override to enable/disable the optional argument
+    /// to OSL's texture() function.
+    bool oslFileTextureColorspace;
```
jstone-lucasfilm, line 198: In order to make this new generator option as clear as possible, what would you think about naming it `oslTextureColorspace`?  The corresponding C++ constant could then be named `T_OSL_TEXTURE_COLORSPACE` for consistency.

ld-kerley:
Instead of adding a GenContext option, perhaps we should just use the OSL_VERSION guard @jstone-lucasfilm suggests above, but just set it to the next version of OSL that isn't released yet?  I was in the OSL TSC meeting yesterday and @lgritz said he hoped to have this keyword arg wired up in pretty soon, so it should drop in to `main` quickly, and anyone building from there will get the new version.

jstone-lucasfilm:
I'd be fine with that suggestion as well, @ld-kerley, as long as it doesn't make it too challenging for developers to access these colorspace attributes in their proprietary renderers.

By the time developers are working with MaterialX v1.39.4, perhaps it's reasonable to ask them to update to a newer version of OSL as well.

dal:
At Pixar we're watching this PR closely because we've hit the OSL compile error trying to deploy mtlx at the studio. Both approaches seem fine but I think we favor the OSL_VERSION guard because it would involve minimal code changes on our end.

boberfly:
Ok I see OSL `1.14.5.0` has the optional argument now from @lgritz 's [changelog](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/releases/tag/v1.14.5.0).  I'll go the `OSL_VERSION` approach and re-push this to this PR soon and simplify things.

@dal another approach that I've taken short-term in Renderman 26.1 is to just swap all texture-related MaterialX OSL nodes to `PxrTexture` and just map to the input equivalents. Not perfect though but it's functional...

jstone-lucasfilm:
This fix looks good to me, thanks @boberfly.

I'm additionally CC'ing @lfl-eholthouser for her thoughts, as this change potentially benefits upcoming work on OSL shader translation.

jstone-lucasfilm (approved):

PR #2264, opened 2025-03-07
MDL: apply subsurface_bsdf weight in MDL implementation and add tests

krohmerNV:
This PR is mostly about the added tests materials.
Attached is a comparison between GLSL, OSL, and MDL.

- MDL should be correct as far as I can tell. SSS is implemented using a random walk in the path tracer (which will be added to #2254) 
- For the GLSL implementation the radius seems to not have a big impact which might be okay due to the real-time approximation. But we are wondering about  the `max(mfp, 0.1)` [here ](https://github.com/AcademySoftwareFoundation/MaterialX/blob/0e6b5bfb46c7d3c4ad804e4394c8333d740e6d01/libraries/pbrlib/genglsl/lib/mx_microfacet_diffuse.glsl#L175). This means at least 10cm radius if the scene is modelled in meters, right?
- OSL matches GLSL but I haven't checked the source.

Screenshot_7-3-2025_153024_

### DIFF
--- a/resources/Materials/TestSuite/pbrlib/bsdf/subsurface_bsdf.mtlx
+++ b/resources/Materials/TestSuite/pbrlib/bsdf/subsurface_bsdf.mtlx
@@ -1,0 +1,104 @@
+<?xml version="1.0"?>
+<materialx version="1.39">
+  <nodegraph name="subsurface_1_common">
+    <subsurface_bsdf name="subsurface_bsdf1" type="BSDF">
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm (approved):
These look like good changes to me, thanks @krohmerNV.

Our render support for `subsurface_bsdf` in GLSL and OSL `testrender` are very early, so I wouldn't yet expect a match with the more mature implementation in MDL, but these new tests should help to align future implementations across all languages.

Issue #2271, opened 2025-03-11
GenVK: Generated shader is invalid


ashwinbhat:
Run:
python python\Scripts\generateshader.py --target vulkan Examples\StandardSurface\standard_surface_default.mtlx

The generated Default.vulkan.vert fails validation. 

```
// Inputs block: VertexInputs
layout (location = 0) in vec3 i_position;
layout (location = 1) in vec3 i_normal;
layout (location = 2) in vec3 i_tangent;

layout (location = 0) out vec3 normalWorld;
layout (location = 1) out vec3 tangentWorld;
layout (location = 2) out vec3 positionWorld;

void main()
{
    vec4 hPositionWorld = u_worldMatrix * vec4(i_position, 1.0);
    gl_Position = u_viewProjectionMatrix * hPositionWorld;
    vd.normalWorld = normalize((u_worldInverseTransposeMatrix * vec4(i_normal, 0.0)).xyz);
    vd.tangentWorld = normalize((u_worldMatrix * vec4(i_tangent, 0.0)).xyz);
    vd.positionWorld = hPositionWorld.xyz;
}
```

The `vd.` is unexpected for Vulkan GLSL.

[Default.vulkan.vert.txt](https://github.com/user-attachments/files/19197002/Default.vulkan.vert.txt)
Output of glslangValidator
```
C:\VulkanSDK\1.3.246.1\Bin\glslangValidator.exe F:\source\MaterialX\resources\Materials\Examples\StandardSurface\Default.vulkan.vert
F:\source\MaterialX\resources\Materials\Examples\StandardSurface\Default.vulkan.vert
ERROR: 0:26: 'vd' : undeclared identifier
ERROR: 0:26: 'normalWorld' : vector swizzle too long
ERROR: 0:26: 'normalWorld' : unknown swizzle selection
ERROR: 0:26: 'normalWorld' : unknown swizzle selection
ERROR: 0:26: 'normalWorld' : unknown swizzle selection
ERROR: 0:26: '' : compilation terminated
ERROR: 6 compilation errors.  No code generated.
```

ashwinbhat:
The expected shader should probably be as shown below where Vertexdata is placed in a layout 
```

#version 450

#pragma shader_stage(vertex)

// Uniform block: PrivateUniforms
layout (std140, binding=0) uniform PrivateUniforms_vertex
{
    mat4 u_worldMatrix;
    mat4 u_viewProjectionMatrix;
[...]

PR #2274, opened 2025-03-12
Emit layout blocks for stage interfaces in Vulkan

ashwinbhat:
The VkShaderGenerator was earlier flattening the VertexData into separate layout entries. While this is valid syntax, it is preferred to use the same struct as glsl to be consistent.

current emitted code:
```
layout (location = 0) out vec3 normalWorld;
layout (location = 1) out vec3 tangentWorld;
layout (location = 2) out vec3 positionWorld;
```
new proposed output:
```
layout (location = 0) out VertexData
{
    vec3 normalWorld;
    vec3 tangentWorld;
    vec3 positionWorld;
} vd;
```

Also fixes https://github.com/AcademySoftwareFoundation/MaterialX/issues/2271

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -210,8 +210,8 @@ jobs:
         C:/vcpkg/vcpkg install glslang[tools] --triplet=x64-windows-release
         C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe -v
         python python/Scripts/generateshader.py resources/Materials/Examples --target glsl --validator C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe
-        python python/Scripts/generateshader.py resources/Materials/Examples --target vulkan --validator C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe
         python python/Scripts/generateshader.py resources/Materials/Examples --target essl --validator C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe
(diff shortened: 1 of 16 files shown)


jstone-lucasfilm:
This change looks good to me, @ashwinbhat, and I'm CC'ing @kwokcb for his thoughts.

ashwinbhat:
@jstone-lucasfilm @bernardkwok 
I've updated the yml enable `vulkan` validation and generation.

.github/workflows/main.yml
```diff
@@ -208,8 +209,9 @@ jobs:
       run: |
         C:/vcpkg/vcpkg install glslang[tools] --triplet=x64-windows-release
         C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe -v
-        python python/Scripts/generateshader.py resources/Materials/Examples --target glsl --validator C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe
-        python python/Scripts/generateshader.py resources/Materials/Examples --target essl --validator C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe
+        python python/Scripts/generateshader.py ../resources/Materials/Examples --target glsl --validator C:/vcpkg/installed/x64-windows-release/tools/glslang/glslangValidator.exe
```
jstone-lucasfilm, line 212: Perhaps we don't need the `../` prefix to be added to these filepaths?  That might be a source of the new testing errors that we're seeing.

PR #2276, opened 2025-03-14
[partial] Extend the dielectric_bsdf node to carry a `priority` value

kprobinson:
This would be one straightforward approach to adding nested dielectric priority to MaterialX. I've made a very small extension to the dielectric_bsdf nodedef, just to include the priority parameter. The default is zero, and I've noted that higher values mean greater precedence, following the semantics of OSL (as well as that of Renderman and Arnold). Then the priority value is passed down to the underlying OSL shader so it can be used.

I believe this would be the most focused way to maintain compatibility with existing MaterialX workflows.


jstone-lucasfilm:
@kprobinson Let's merge this into your latest proposal PR at #2310, and we can focus our discussions there.

PR #2278, opened 2025-03-15
Move Standard Library Shader Nodes to StandardNodes doc 

dbsmythe:
Move the description of Standard Library Shader Nodes from the main Specification doc to the new StandardNodes doc.  Adjust other text and links as needed.


jstone-lucasfilm (approved):
This change looks good to me, thanks @dbsmythe!

Issue #2275, opened 2025-03-14
Used of uninitialized dirAlbedo in mx_sheen_bsdf in GLSL


tomas-davidovic:
When running the `renderglsl` tests, the `transparency_hints.mtlx` tests produce this code snippet:
```
        // Calculate the BSDF transmission for viewing direction
        ClosureData closureData = ClosureData(CLOSURE_TYPE_TRANSMISSION, L, V, N, P, occlusion);
        BSDF fuzz_bsdf_out = BSDF(vec3(0.0),vec3(1.0));
        mx_sheen_bsdf(closureData, fuzz_weight, fuzz_color, fuzz_roughness, geometry_normal, 1, fuzz_bsdf_out);
```

The `mx_sheen_bsdf` declares `float dirAlbedo;` without a default value, and then only handles the `CLOSURE_TYPE_REFLECTION` and `CLOSURE_TYPE_INDIRECT` cases, leaving `dirAlbedo` uninitialized when hitting the last line:
```
bsdf.throughput = vec3(1.0 - dirAlbedo * weight);
```

jstone-lucasfilm:
Great catch, @tomas-davidovic, and this seems valuable to fix!

jstone-lucasfilm:
Thanks for the report, @tomas-davidovic, and this has been addressed in #2280.

PR #2280, opened 2025-03-16
Fix edge case in mx_sheen_bsdf

jstone-lucasfilm:
This changelist fixes an edge case in the GLSL function `mx_sheen_bsdf`, where setting `closureData.closureType` to `CLOSURE_TYPE_TRANSMISSION` caused the `dirAlbedo` variable to be referenced without any assigned value.


jstone-lucasfilm:
I believe this should address issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/2275.

jstone-lucasfilm:
Here's a new GLSL/OSL render comparison with the codebase in this PR:

[MaterialXRenderTests_03_16_2025_GitHub.pdf](https://github.com/user-attachments/files/19275119/MaterialXRenderTests_03_16_2025_GitHub.pdf)

Issue #2277, opened 2025-03-14
MaterialX 1.39 upgrade path does not update NodeDef strings


JGamache-autodesk:
The upgrade path from 1.38 to 1.39 will not work if the document to upgrade has NodeDef strings.

To repro: upgrade this document to 1.39:

```xml
<?xml version="1.0"?>
<materialx version="1.38" xmlns:xi="http://www.w3.org/2001/XInclude">
  <standard_surface name="Surface" type="surfaceshader">
    <input name="base_color" type="color3" output="base_color_out" nodegraph="NG_Anonymized" />
  </standard_surface>
  <surfacematerial name="standard_surface1" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="Surface" />
  </surfacematerial>
  <nodegraph name="NG_Anonymized">
    <ramplr name="N1" type="color3" nodedef="ND_ramplr_color3">
      <input name="texcoord" type="vector2" nodename="N4" />
      <input name="valuel" type="color3" nodename="N3" />
      <input name="valuer" type="color3" nodename="N2" />
    </ramplr>
    <geompropvalue name="N4" type="vector2" nodedef="ND_geompropvalue_vector2">
      <input name="geomprop" type="string" value="st" />
    </geompropvalue>
    <dot name="N3" type="color3" nodedef="ND_dot_color3" />
    <swizzle name="N2" type="color3" nodedef="ND_swizzle_color3_color3">
      <input name="channels" type="string" value="bgr" />
      <input name="in" type="color3" nodename="N3" />
    </swizzle>
    <output name="base_color_out" type="color3" nodename="N1" />
  </nodegraph>
</materialx>
```

The swizzle node will get upgraded to a combine, but the node def string will still say `ND_swizzle_color3_color3`. That string should either be cleared or upgraded.

jstone-lucasfilm:
Good catch, @JGamache-autodesk, and we should fix this for 1.39.4.

jstone-lucasfilm:
Thanks for this original report, @JGamache-autodesk, and this has been addressed in #2282.

PR #2282, opened 2025-03-17
Improvements to `swizzle` upgrade logic

jstone-lucasfilm:
This changelist improves the upgrade logic for the `swizzle` node in v1.38 documents, extending support to nodes with explicit `nodedef` attributes, while maintaining support for nodes without them.

New examples of this usage have been added to the `TestSuite/stdlib/upgrade` folder for validation in unit testing.

### DIFF
--- a/resources/Materials/TestSuite/stdlib/upgrade/syntax_1_38.mtlx
+++ b/resources/Materials/TestSuite/stdlib/upgrade/syntax_1_38.mtlx
@@ -63,4 +63,15 @@
     <input name="surfaceshader" type="surfaceshader" nodename="N_surface_3" />
   </surfacematerial>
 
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:
This should address issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/2277 in MaterialX v1.39.4, and I'm CC'ing @JGamache-autodesk and @klucknav for their thoughts.

klucknav:
In the past we have hit code gen issues when the `nodeDefString` is not present. So I would prefer to update the nodeDef string, as we do with the `normalmap` node, instead of removing it.

jstone-lucasfilm:
That sounds reasonable to me, @klucknav, and I'll make that update to the PR from my side.

JGamache-autodesk (approved):
LGTM and would allow making the USD fix more palatable. Thanks!

Issue #2281, opened 2025-03-17
Richer metadata for OSL translation


boberfly:
Hi again, hopefully @jstone-lucasfilm is not sick of my issues yet. :)

I've come across something which I think I could make a trivial patch for, but I'll just briefly explain how I am currently using MaterialX.

I'm translating almost all OSL-compatible nodedefs separately by creating MaterialX documents with a single OSL node in them each and outputting an OSO. I am doing it this way as I'm finding that I need to mix and match renderer-native nodes with OSL MaterialX nodes once it hits the renderer's translation step, so I can use for instance Arnold's native OpenPBR surface but have all the inputs as MaterialX OSL nodes, or in Cycles' case, use its native `node_image_texture` for all of the ones that call into `texture()` to get around current limitations eg. `<UDIM>` tags only work with this node.

Using `geometry_tangent` from OpenPBR surface as an example here: https://github.com/AcademySoftwareFoundation/MaterialX/blob/7abbbdcca5e8dc7b3375146d2f3b57a52af99dbd/libraries/bxdf/open_pbr_surface.mtlx#L86C5-L87C44 I noticed there's quite a bit of rich metadata here, like `doc` or `defaultgeomprop`. I could use this data at the OSL-level for Cycles as I need a way of detecting if a given plug is for plugging in a tangent vector, as `dPdy` may not be calculated from the desired tangent implicitly (Cycles uses a hint to auto-plug the right tangent here internally for its default `PrincipledBSDF`, which I could also use if I can detect `defaultgeomprop==Tworld`).

In this function: https://github.com/AcademySoftwareFoundation/MaterialX/blob/7abbbdcca5e8dc7b3375146d2f3b57a52af99dbd/source/MaterialXGenOsl/OslShaderGenerator.cpp#L460 it looks like I could add `defaultgeomprop` but also  `doc` or `uiname`. My assumption as to why this hasn't been done yet is perhaps MaterialX is meant to take a full graph and translate this into a singular OSO that's not meant for use outside of the renderer's translation, so this stuff wouldn't be useful or exposed except for my odd use-case above.

If this seems like something that wouldn't be too problematic to add, I can make a patch that adds this metadata. I'm not sure if there's a default convention for OSL, but I think the `help` metadata tag might be a closer convention to `doc`. Or maybe some of these would need a `mtlx_` prefix like for `defaultgeomprop`.

Cheers

jstone-lucasfilm:
[...]

PR #2283, opened 2025-03-17
Added `defaultgeomprop` to OSL metadata if specified for the input

boberfly:
Hi, this adds the `defaultgeomprop` as metadata for GenOSL, referencing https://github.com/AcademySoftwareFoundation/MaterialX/issues/2281

I chose not to add `doc` (or `help`) metadata for this one, but that could be added also.

Cheers

### DIFF
--- a/source/MaterialXGenOsl/OslShaderGenerator.cpp
+++ b/source/MaterialXGenOsl/OslShaderGenerator.cpp
@@ -479,8 +479,9 @@ void OslShaderGenerator::emitMetadata(const ShaderPort* port, ShaderStage& stage
     auto widgetMetadataIt = UI_WIDGET_METADATA.find(port->getType());
     const ShaderMetadata* widgetMetadata = widgetMetadataIt != UI_WIDGET_METADATA.end() ? &widgetMetadataIt->second : nullptr;
     const ShaderMetadataVecPtr& metadata = port->getMetadata();
+    const string& geomprop = port->getGeomProp();
 
-    if (widgetMetadata || (metadata && metadata->size()))
+    if (widgetMetadata || (metadata && metadata->size()) || !geomprop.empty())
     {
         StringVec metadataLines;
         if (metadata)
@@ -499,9 +500,15 @@ void OslShaderGenerator::emitMetadata(const ShaderPort* port, ShaderStage& stage
         }
         if (widgetMetadata)
         {
+            const string& delim = geomprop.empty() ? EMPTY_STRING : Syntax::COMMA;
             const string& dataType = _syntax->getTypeName(widgetMetadata->type);
             const string dataValue = _syntax->getValue(widgetMetadata->type, *widgetMetadata->value, true);
-            metadataLines.push_back(dataType + " " + widgetMetadata->name + " = " + dataValue);
+            metadataLines.push_back(dataType + " " + widgetMetadata->name + " = " + dataValue + delim);
+        }
+        if (!geomprop.empty())
+        {
+            const string& dataType = _syntax->getTypeName(Type::STRING);
+            metadataLines.push_back(dataType + " defaultgeomprop = \"" + geomprop + "\"");
         }
         if (metadataLines.size())
         {


jstone-lucasfilm:
This proposal looks good to me, @boberfly, and I'm CC'ing @niklasharrysson for his thoughts on the handling of OSL metadata.

boberfly:
Just to add context this is really handy to set socket flags in Cycles that I'm upstreaming now:
https://projects.blender.org/blender/cycles/pulls/25/files#diff-2f6edf318189822c72afef23e9e400d4c3c4b647

source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -499,9 +500,15 @@ void OslShaderGenerator::emitMetadata(const ShaderPort* port, ShaderStage& stage
...
+            const string& delim = geomprop.empty() ? EMPTY_STRING : Syntax::COMMA;
             const string& dataType = _syntax->getTypeName(widgetMetadata->type);
             const string dataValue = _syntax->getValue(widgetMetadata->type, *widgetMetadata->value, true);
-            metadataLines.push_back(dataType + " " + widgetMetadata->name + " = " + dataValue);
+            metadataLines.push_back(dataType + " " + widgetMetadata->name + " = " + dataValue + delim);
+        }
+        if (!geomprop.empty())
+        {
+            const string& dataType = _syntax->getTypeName(Type::STRING);
+            metadataLines.push_back(dataType + " defaultgeomprop = \"" + geomprop + "\"");
```
ld-kerley, line 511: suggest using `mtlx_defaultgeomprop` instead.

ld-kerley (commented):
Oh just this PR now, I left a note on the issue, but I'll comment here too incase you don't see the issue.

I like this overall, but think it's a good idea for us to namespace any MaterialX specific OSL metadata conventions we're introducing.

source/MaterialXGenOsl/OslShaderGenerator.cpp
boberfly: Done

source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -479,8 +479,9 @@ void OslShaderGenerator::emitMetadata(const ShaderPort* port, ShaderStage& stage
     auto widgetMetadataIt = UI_WIDGET_METADATA.find(port->getType());
     const ShaderMetadata* widgetMetadata = widgetMetadataIt != UI_WIDGET_METADATA.end() ? &widgetMetadataIt->second : nullptr;
     const ShaderMetadataVecPtr& metadata = port->getMetadata();
+    const string& geomprop = port->getGeomProp();
```
niklasharrysson, line 482: @boberfly Sorry for the delay on this.. I noticed today that I was CC:ed on this PR.

I see that you are adding an explicit code path here for adding in the defaultgeomprop metadata on inputs. I think it might be better to handled this through the default code path for metadata registration instead. If you just register "defaultgeomprop" as an attribute that should be exported as metadata I think the system should pick this up, and it will be available among the other entries in the metadata vector for the port:

`const ShaderMetadataVecPtr& metadata = port->getMetadata();`

And hence it should be emitted below automatically.

The default metadata entries are specified and registered here:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/ShaderGenerator.cpp#L456

So you could add "defaultgeomprop" to this list. Or if you want it to only be registered for OSL you can do this in the override method for the OslShaderGenerator instead:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenOsl/OslShaderGenerator.cpp#L238
(Or of course if you have your own generator you could override and add it only there)

One important thing to note is that your application needs to call `registerShaderMetadata`, during creation and initialization of the GenContext, for this to work. See for example here how it's done in the render test suite:
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXTest/MaterialXRender/RenderUtil.cpp#L187

As you will see there are more metadata for OSL that is already supported. For example this is what some generated inputs from StandardSurface looks like, if you make sure to call `registerShaderMetadata` in your setup code.

```
    ...
    float diffuse_roughness = 0.000000
    [[
        float min = 0.000000,
        float max = 1.000000,
        string label = "Diffuse Roughness",
        string page = "Base",
        int uiadvanced = 1,
        string help = "Roughness of the diffuse reflection. Higher values cause the surface to appear flatter and darker.",
        string widget = "number"
    ]],
    float metalness = 0.000000
    [[
        float min = 0.000000,
        float max = 1.000000,
        string label = "Metalness",
        string page = "Base",
        string help = "Specifies how metallic the material appears. At its maximum, the surface behaves like a metal, using fully specular reflection and complex fresnel.",
        string widget = "number"
    ]],
    ...
```

boberfly: @niklasharrysson thanks for that insight, that makes a lot more sense! I did see a bit of scaffolding code there and was wondering why it wasn't emitting anything, but that call to register makes sense.

I'll try this soon, I guess the prefix wouldn't be applied doing it this way (as it looks like the others don't either, but they seem to be common OSL metadata names and `defaultgeomprop` would be more or less MaterialX-specific). And perhaps having this as a global could benefit other languages vs just doing it for GenOSL but I'll leave that one up to others to decide and I'll update this PR.

Cheers

niklasharrysson: Yeah, that's a good point that we should maybe add a "mtlx" prefix for all the standard metadata we generate. 

As you can see in the OSL override of `registerShaderMetadata` we're actually renaming the metadata there for the ones that have a corresponding OSL standard naming.

So we could add "mtlx" to all the MaterialX standard metadata, and any overriding generator could rename this if needed. But let's save this prefix stuff for another PR.

jstone-lucasfilm (approved):
Thanks for this improvement, @boberfly, and this looks good to me.

PR #2284, opened 2025-03-18
Turned at() calls to operator[] where bounds checks were not necessary.

tomas-davidovic:
In several places, where the variable was a pointer to `std::vector`, the code used `->at(x)` to fetch elements, most probably for the syntax convenience. However, `at()` is always bounds checked, unlike `operator[]` which is bounds checked only in debug mode. This makes `at()` slightly less efficient where performance matters, and if the bounds checking is not desired, using `(*vectorPtr)[x]` pattern makes the code slightly faster.

### DIFF
--- a/source/MaterialXGenOsl/OslShaderGenerator.cpp
+++ b/source/MaterialXGenOsl/OslShaderGenerator.cpp
@@ -112,7 +112,7 @@ ShaderPtr OslShaderGenerator::generate(const string& name, ElementPtr element, G
     {
         for (size_t j = 0; j < metadata->size(); ++j)
         {
-            const ShaderMetadata& data = metadata->at(j);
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm:
This change looks good to me, @tomas-davidovic, and assuming you've followed up with a support request ticket, the Linux Foundation should be able to resolve the CLA issues shortly.

I'm CC'ing @ld-kerley and @niklasharrysson for their thoughts on these specific changes to shader generation and rendering of materials leveraging structs.

tdavidovicNV:
I might have messed up the EasyCLA because I pushed from a branch under private username, with the commit tied to my NVIDIA account. I suggest we close this PR and I try again, with the proper logins everywhere.

PR #2288, opened 2025-03-19
Improved upgrade code that repairs missing node info

JGamache-autodesk:
Very useful when the submitted MaterialX document to upgrade was built without all the library info being available.


JGamache-autodesk:
Closing. Duplicate of #2290 which is already merged.

PR #2295, opened 2025-03-22
Add support for `closest` filtering in GLSL

jstone-lucasfilm:
This changelist adds support for the `closest` filter type in GLSL rendering, allowing it to correctly affect the magnification filtering of textures.


jstone-lucasfilm:
This PR addresses the remaining GLSL case for https://github.com/AcademySoftwareFoundation/MaterialX/issues/1525.

Issue #2286, opened 2025-03-18
Invalid guard in GlslProgram against setting Environment map in bindTextures


tomas-davidovic:
The guard in `GlslProgram.cpp` against setting `HW::ENV_RADIANCE` and `HW::ENV_IRRADIANCE` textures reads like this:
```
            // Always bind a texture unless it is a lighting texture.
            // Lighting textures are handled in the bindLighting() call.
            // If no texture can be loaded then the default color defined in
            // "samplingProperties" will be used to create a fallback texture.
            if (fileName != HW::ENV_RADIANCE &&
                fileName != HW::ENV_IRRADIANCE)
            {
                ImageSamplingProperties samplingProperties;
                samplingProperties.setProperties(uniform.first, publicUniforms);
                bindTexture(uniformType, uniformLocation, fileName, imageHandler, samplingProperties);
            }
```

However, it is comparing the filename of the texture (e.g., `image.png`) rather than the variable name (e.g., `u_envRadiance`) so it will never be false.

The relevant lines should read:
```
            if (uniform.first != HW::ENV_RADIANCE &&
                uniform.first != HW::ENV_IRRADIANCE)
```

jstone-lucasfilm:
Thanks for this report, @tomas-davidovic, as well as for the fix in #2296!

PR #2296, opened 2025-03-24
Fixes issue #2286, wrong string compared to not set env variables.

tdavidovicNV:
Fixes an issue where the actual `filename` (name of the image on disk) was compared against the variable name for radiance and irradiance.


jstone-lucasfilm (approved):

PR #2297, opened 2025-03-24
Remove unneeded uistep from UsdPreviewSurface

tdavidovicNV:
Fixes the incorrectly parsed uistep by simply removing the parameter, as it is not used.


jstone-lucasfilm (approved):

Issue #2089, opened 2024-10-29
Align GLSL matrix multiply and divide with OSL and C++


HardCoreCodin:
Shading languages use different matrix conventions: OSL is row major, GLSL is column major
This determines multiplication order: `M1 * M2` in OSL is `M2 * M1` in GLSL
MaterialX back-end implementations generate code with the same ordering, consistent with the node inputs.
If it is correct for OSL, it would be incorrect for GLSL (and vice versa)
This makes the same node (and hence any graph that uses it) non-portable across back-ends.

<img width="973" alt="MatMul_Blender_Vs_MGE_REVERSED" src="https://github.com/user-attachments/assets/64a4f1e7-212d-4b3f-b177-b779db72c0cf">
<img width="983" alt="MatMul_Blender_Vs_MGE" src="https://github.com/user-attachments/assets/72553ebf-43f5-4f06-aaf4-a1b3a63b20b8">
<img width="906" alt="MatMul_Blender_Vs_MGE_M1" src="https://github.com/user-attachments/assets/90dd43bc-1ee7-412f-ae41-d105f648eaa9">
<img width="979" alt="MatMul_Blender_Vs_MGE_M2" src="https://github.com/user-attachments/assets/e7657022-18db-4e3b-bb8b-365d9444b86d">

MaterialX specification mentions row major, but only with reference to matrix literal construction component ordering.
However, both OSL and GLSL use the same ordering anyway: ( X.x, X.y, X.z , Y.x, Y.y, Y.z ,  Z.x, Z.y, Z.z )
So this is not really saying anything meaningful.
The specification should state what the implied matrix convention is in MaterialX - explicitly.
All implementations should then be made to account for that convention.

If it is row major, then the ordering for the `multiply` node types that take a 3x3 or 4x4 matrix, should be reversed in GLSL.

<img width="917" alt="MatrialX_glsl_impl_incorrect_multiply_divide_matrix" src="https://github.com/user-attachments/assets/691ba1b2-45dd-480f-8b93-b37bfdd44793">

There should also be a special handling for `divide` node types, as just flipping the order will not suffice:
Matrix division is defined as the first matrix provided multiplied by the inverse of the second.
The correct matrix need to be inverted in column-major implementations.

<img width="931" alt="MGE_matrix_A_divided_by_B_flipped_order" src="https://github.com/user-attachments/assets/256d7ad9-9e60-40c6-8e5e-35ea57e487ac">
[...]

PR #2298, opened 2025-03-24
Fixed GLSL implementations of matrix operators

tdavidovicNV:
Fixed an issue with GLSL matrix multiplication and division, described in #2089 

The gist of the problem is this. Taking C++ code as the basis, MaterialX uses row major matrices, with translation in the last row, so it transforms points like so `p * M`. When composing transforms `M1 * M2`, the final result is `p * M1 * M2`.

However, GLSL has matrices in col-major storage, meaning that when you simply upload the CPU memory, the matrices on the GPU are transposed. That effectively means that translation is now in the last column, so transforming point works like `M * p`, and composing transforms works like `M2 * M1 * p`, to match the CPU `p * M1 * M2`.

That means that matrix multiplication has to have the inputs swapped. Alternatively, we could do `transpose(transpose({{in1}}) * transpose({{in2}}))` for the same effect.

The fix to division has two parts. In GLSL, only the `*` operator has special meaning (matrix multiplication), division operator `/` only does component wise division. So instead of `M1 / M2` you have to write `M1 * inverse(M2)`. And on top that, the matrix multiplication swap of arguments described above also applies, for the final form of `inverse(M2) * M1`. Again, alternatively we could transpose the arguments and then transpose the result.

### DIFF
--- a/libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
+++ b/libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
@@ -237,8 +237,8 @@
   <implementation name="IM_multiply_vector3FA_genglsl" nodedef="ND_multiply_vector3FA" target="genglsl" sourcecode="{{in1}} * {{in2}}" />
   <implementation name="IM_multiply_vector4_genglsl" nodedef="ND_multiply_vector4" target="genglsl" sourcecode="{{in1}} * {{in2}}" />
   <implementation name="IM_multiply_vector4FA_genglsl" nodedef="ND_multiply_vector4FA" target="genglsl" sourcecode="{{in1}} * {{in2}}" />
-  <implementation name="IM_multiply_matrix33_genglsl" nodedef="ND_multiply_matrix33" target="genglsl" sourcecode="{{in1}} * {{in2}}" />
-  <implementation name="IM_multiply_matrix44_genglsl" nodedef="ND_multiply_matrix44" target="genglsl" sourcecode="{{in1}} * {{in2}}" />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This fix looks good to me, @tdavidovicNV, though I'd like to run our GLSL/OSL render comparisons before moving forward.

jstone-lucasfilm:
@tdavidovicNV For now, I'd like to make sure our existing GLSL/OSL tests continue to succeed, but you're completely correct that we should add tests that validate matrix multiplication and division once these latest fixes have been integrated.  I'm not yet expecting those to succeed, and my local tests are just focused on visual validation of our existing render comparisons.

jstone-lucasfilm:
Attaching a new GLSL/OSL render comparison with the changes in this PR:

[MaterialXRenderTests_03_24_2025_GitHub.pdf](https://github.com/user-attachments/files/19438568/MaterialXRenderTests_03_24_2025_GitHub.pdf)

pablode:
> Again, alternatively we could transpose the arguments and then transpose the result.

That sounds much cheaper than the inversion that got merged.

tdavidovicNV:
@pablode the transpose would only resolve the swapping of arguments, which comes from the natural transpose when you upload row-major matrix to GLSL (with col-major matrices).

For the matrix division to actually work, you still need the matrix inverse, as the division operator in GLSL is just element-wise division.

It is just the difference between `transpose(transpose({{in1}}) * transpose(mx_inverse({{in2}})))` and `mx_inverse({{in2}}) * {{in1}})`. 

Unfortunately, there is no way to divide matrices without an inverse, and even if you have a built-in `/` operator that does matrix divide, like OSL, all it does underneath is multiply with a matrix inverse, give or take some precision and optimization tricks.

PR #2299, opened 2025-03-24
Fix division by singular matrices in test suite

tdavidovicNV:
The matrix division unit test uses a singular matrix as a divisor.
This change instead does matrix divison that will result is matrix with 1.5, 1.5, 1.5, 1.0 on the diagonal.

This in isolation will make the GLSL test produce NaNs, because of the issue #2089 , so merging PR #2298 first is highly recommended.


jstone-lucasfilm (approved):
Good catch, @tdavidovicNV, and this change looks good to me.

PR #2301, opened 2025-03-24
Dev/tdavidovic/remove texture baker ptr

tdavidovicNV:
The TextureBakerPtr defined in RenderPipeline.h isn't actually used in any place where a more concrete type of TextureBaker wouldn't be known. It seems like it was intended, or originally used, as the return type from `RenderPipeline::createTextureBaker` which is now using a full type erasure by returning a void pointer.

As such, a generic type whose actual type depends on compiler switches isn't really needed, and explicit types of `TextureBakerGlslPtr` and `TextureBakerMslPtr` seem clearer.

### DIFF
--- a/source/MaterialXRenderGlsl/TextureBaker.h
+++ b/source/MaterialXRenderGlsl/TextureBaker.h
@@ -23,7 +23,7 @@
 MATERIALX_NAMESPACE_BEGIN
 
 /// A shared pointer to a TextureBaker
-using TextureBakerPtr = shared_ptr<class TextureBakerGlsl>;
+using TextureBakerGlslPtr = shared_ptr<class TextureBakerGlsl>;
 
 /// A vector of baked documents with their associated names.
 using BakedDocumentVec = std::vector<std::pair<std::string, DocumentPtr>>;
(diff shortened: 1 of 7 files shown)


tdavidovicNV:
This also begs the question whether the `createTextureBaker` should even be part of the `RenderPipeline` interface, as it seems it is only ever called from within the specific `RenderPipeline`.

jstone-lucasfilm (approved):
This looks like a good change to me, @tdavidovicNV, and I'd agree it opens the door to additional simplifications in the future.

ld-kerley:
We may want to consider re-instating the `TextureBakerPtr` definition, as removing it breaks the current USD build.

We could probably provide a patch to USD, but I worry about increasing the friction between the projects, and the MaterialX integration complexity on the USD side.

jstone-lucasfilm:
That's a good point, @ld-kerley, and I defer to you and @tdavidovicNV on whether the `TextureBakerPtr` alias should be restored for USD compatibility.

tdavidovicNV:
If this creates a friction with USD, lets revert. The improvement is marginal at best and the end result is somewhat better, but not actually good, imo. Maybe if enough USD-compatibility-breaking changes accumulate, a cleanup like this could go along for the ride?

jstone-lucasfilm:
Since we're getting close to our testing and release candidate phase for 1.39.4, I thought I'd check in on this thread.

@ld-kerley @tdavidovicNV Would one of you like to make additional improvements here, either providing backwards compatibility for `TextureBakerPtr` or submitting a reversion of this change?

I don't have a strong opinion on this, and I'll defer to the judgement of you two.

tdavidovicNV:
I am in favor of clean revert since I do not know the USD implications and cannot investigate in reasonable time frame. Can it be done easily from this PR, or should I submit a separate reverting PR?

jstone-lucasfilm:
@tdavidovicNV That approach sounds fine to me, and the best approach is to submit a new PR that reverts the contents of this one.  We can then review and merge that change as usual.

PR #2302, opened 2025-03-24
Removed unnecesary getProgram()

tdavidovicNV:
The program has already been obtained and the following lines use it, so this is an unnecessary virtual call that makes reading the code a bit confusing.


jstone-lucasfilm (approved):

PR #2305, opened 2025-03-24
Replace `vector::at` with indexing when bounds checks are unneeded

tdavidovicNV:
In several places, where the variable was a pointer to `std::vector`, the code used `->at(x)` to fetch elements, most probably for the syntax convenience. However, `at()` is always bounds checked, unlike `operator[]` which is bounds checked only in debug mode. This makes `at()` slightly less efficient where performance matters, and if the bounds checking is not desired, using `(*vectorPtr)[x]` pattern makes the code slightly faster.

(This is a second attempt at the exact same PR, as the previous one was blocked by complicated CLA)


jstone-lucasfilm:
This change looks good to me, @tdavidovicNV, and I'm CC'ing @niklasharrysson and @ld-kerley for their expertise in this area of the code.

ld-kerley (approved):
Looks good to me - thanks for the PR

PR #2306, opened 2025-03-24
Update width and height consistently in MslRenderer::setSize

tdavidovicNV:
In the `MslRenderer::setSize`, if the `_framebuffer` already existed, the `_width` and `_height` haven't been updated.


jstone-lucasfilm (approved):

PR #2307, opened 2025-03-24
Remove redundant clearInputLists in MslProgram::setStages

tdavidovicNV:
`clearInputLists();` called explicitly in `MslProgram::setStages` is already part of `clearStages();` called at the start of the method.


jstone-lucasfilm (approved):

PR #2308, opened 2025-03-25
Refine matrix multiplication tests to enable validation

tdavidovicNV:
The `math_operators.mtlx` test for matrix multiplication was multiplying just identity matrices. If actually ran to produce an image, it would not show an error in matrices having swapped arguments.

In this PR I've changed the values so the result is `0.75, 0.25, 0.5, 1` when multiplies correctly, and `0.25, 0.75, 0.5, 1` when swapped.

I ran the tests with this `<input name="renderTestPaths" type="string" value="resources/Materials/TestSuite/stdlib/math" />`, before after the fix introduced in #2298.

Original values, before fix:
image
Original values, after fix:
image

Updated values, before fix:
image
Updated values, after fix:
image

The test now can be used to actually verify the calculated value (and shows that the fix in #2298 does indeed bring GLSL matrix multiplication in line with OSL)

### DIFF
--- a/resources/Materials/TestSuite/stdlib/math/math_operators.mtlx
+++ b/resources/Materials/TestSuite/stdlib/math/math_operators.mtlx
@@ -679,8 +679,8 @@
   </nodegraph>
   <nodegraph name="multiply_matrix33">
     <multiply name="multiply1" type="matrix33">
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm (approved):
Great work, thanks @tdavidovicNV!

PR #2310, opened 2025-03-25
Proposing a per-material priority attribute for handling nested diele…

kprobinson:
…ctrics

### DIFF
--- a/BRANCH.md
+++ b/BRANCH.md
@@ -1,0 +1,105 @@
+This file is temporary and will be removed before the associated PR is brought out of draft mode.
+
+---
+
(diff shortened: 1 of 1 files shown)


msuzuki-nvidia:
Some renderers use lower numbers for higher priority (similar to "1st priority"), while others use the reverse order. I assume this proposes that higher numbers indicate higher priority, as the default value of 0 is the lowest priority. It might be helpful to explicitly describe this to avoid any confusion.

kprobinson:
@msuzuki-nvidia Much appreciated. That's been part of the discussion. I've added it in a few places to make sure it's clear.

PR #2311, opened 2025-03-25
Add Graph Editor info to Developer Quick Start

michaelybecker:
Explicitly mention how to build the Graph Editor; it's included with the compiled binary, and I think it's fair to assume newcomers will not know to look for it but would appreciate knowing of its existence upfront.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,8 @@ MaterialX is an open standard for representing rich material and look-developmen
 - Download the latest version of the [CMake](https://cmake.org/) build system.
 - Point CMake to the root of the MaterialX library and generate C++ projects for your platform and compiler.
 - Select the `MATERIALX_BUILD_PYTHON` option to build Python bindings.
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks for catching that, @michaelybecker, and feel free to fill out a [support ticket](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2311#issuecomment-2752264637) if the EasyCLA approval process doesn't work as expected.

The very first contribution from a developer to an ASWF project often runs into this hurdle, but it usually works smoothly for all future contributions.

michaelybecker:
Thanks for the heads up, @jstone-lucasfilm  - I naively assumed the one I signed for USD contribution would cover it. Looks like it's sorted now!

jstone-lucasfilm (approved):
Thanks for the update, @michaelybecker!

PR #2314, opened 2025-03-26
Update values in example materials

tdavidovicNV:
Fixes:
* In the division matrix33 test, the divisor was still a singular matrix, now it is aligned with matrix44.
* In rotate3d, the rotation was 180 degrees which, if used as an actual test, would miss transposed rotation matrices. The rotation is now 135 degrees to clearly separate the left and right rotation.
* In rotate2d, the rotation was 90 degrees, and it was updated to 135 degrees (3/4 PI) to also avoid missing possible bugs in rotation matrix.

This set of changes makes running tests on all of `resources/Materials` more useful.


jstone-lucasfilm (approved):
Thanks for these improvements, @tdavidovicNV!

Issue #2304, opened 2025-03-24
Possible invalid memory access on Metal


tdavidovicNV:
In `MetalTextureHandler::createRenderResources`, there is a line that gets image data like so `void* imageData = image->getResourceBuffer();`.
Later it is checked before the data is copied into the created texture, but before that there is a piece of code:
```
if ((pixelFormat == MTLPixelFormatRGBA32Float || pixelFormat == MTLPixelFormatRGBA8Unorm) && channelCount == 3)
{
  ... rearrangedDataF[dstIdx] = ((float*) imageData)[srcIdx++]; 
}
```

that seems to read from `imageData` based solely on the image format, rather than `imageData` being valid or not. 
I guess it only comes up in `MetalRenderPipeline::updatePrefilteredMap` which creates an empty image of 3*HALF, which does not fall into this if statement so the problem is never hit.

Incidentally, 3xFLOAT in Metal gets converted to `MTLPixelFormatRGBA16Float`, so 4 channels, but the code that converts 3 channels to 4 channels (mentioned above) does not seem to act on Half floats, only on floats and 8Unorm.

jstone-lucasfilm:
CC'ing @ld-kerley and @dgovil for their thoughts.

ld-kerley:
It seems safe enough to just move the `if(imageData)` guard higher to guard access to that memory.  I'll put together a quick PR

PR #2315, opened 2025-03-27
Guard against null image in MetalTextureHandler::createRenderResources

ld-kerley:
addresses #2304


jstone-lucasfilm (approved):

Issue #2303, opened 2025-03-24
MetalTextureHandler::releaseRenderResources cannot release all resource.


tdavidovicNV:
The description of the `releaseRenderResources` says that if no image is provided, all cached images should have their resources release.

However, the Metal implementation simply early exits if the `image` parameter is `nil`, which is not consistent with the description, nor with the GLTextureHandler implementation.

jstone-lucasfilm:
This sounds worthwhile to investigate, thanks @tdavidovicNV, and I'm CC'ing @ld-kerley and @dgovil for their thoughts.

PR #2316, opened 2025-03-27
Release all images in `MetalTextureHandler::releaseRenderResources`

ld-kerley:
This PR addresses #2303.

This brings the MetalTextureHandler inline with the GLTextureHandler for releasing all the images when `nullptr` is passed.

Without this - both CPU and Metal texture memory is being leaked if the image caches are flushed.  

Recursively releasing the resources fixes the Metal texture memory leak, but that just highlighted the CPU side memory leak.  This is due to the MetalTextureHandler also retaining an `ImagePtr` in `_imageBindingInfo`.  Removing the `ImagePtr` from this storage fixes the memory leak, and it doesn't seem to be necessary to store the image, the only thing the image is used for is to access the resource ID - which is the same as the key to the `unordered_map`.


jstone-lucasfilm (approved):

Issue #2309, opened 2025-03-25
MslProgram::updateAttributesList does not cache result.


tdavidovicNV:
Whenever `MslProgram::updateAttributesList` is called, e.g., whenever `MslProgram::getAttributesList` is called it rechecks the variables in `_shader` vertex stage. Unlike GlslProgram, which caches the result on the first call and only updates when the code has changed.

I believe this comes from the fact that in Glsl, the whole `_attributeList` is created as part of the `updateAttributesList` call, but in Msl, the main part of it (everything except for `typeString`) is created in the `build` call.

A simple flag that indicates that the `updateAttributesList` has already been called will could be enough, but as I don't understand the way this code handles structs and currently do not have access to MacOS for testing, I cannot provide a PR.

PR #2317, opened 2025-03-27
Optimization to `MslProgram::updateAttributeList`

ld-kerley:
This PR addresses #2309

introduce guard to avoid repeatedly setting the value and type string for the `_attributeList`.


jstone-lucasfilm (approved):
Thanks for the optimization, @ld-kerley!

PR #2319, opened 2025-03-27
Fix OSL implementation of `frame` for failed read

ld-kerley:
```
  <nodedef name="ND_frame_float" node="frame" nodegroup="application">
    <output name="out" type="float" default="1.0" />
  </nodedef>
```

The node definition states the default is `1.0`. This PR ensures that if the `getattribute()` call fails then the node will return the specified default value.


libraries/stdlib/genosl/mx_frame_float.osl
```diff
@@ -1,4 +1,7 @@
 void mx_frame_float(output float result)
 {
+    // ND_frame_float has a default specified of 1.0, so in the event the getattribute()
+    // fails we should also return 1.0
+    result = 1.0;
```
jstone-lucasfilm, line 5: This looks like a good fix, @ld-kerley, though I wonder if we should create a new pattern for attributes with default values in MaterialX OSL, along the line of the following helper function:

```
float mx_get_attribute(string name, float default)
{
    float result = default;
    getAttribute(name, result);
    return result;
}
````

What are your thoughts?

dgovil:
(Test comment, please ignore)

libraries/stdlib/genosl/mx_frame_float.osl
ld-kerley: I'm not opposed to that change - at all - but that feels like it might be better staged as a separate improvement PR.  It feels like the sort of thing that might need a little more care to get right in all cases, and for all types.

jstone-lucasfilm: Good point, and I agree that this makes sense as a separate PR in the future.

jstone-lucasfilm (approved):

Issue #1841, opened 2024-05-28
Cloverleaf placement appears incorrect


kwokcb:
## Issue

If you create a `cloverleave` node and try and center it it will not be placed where expected. If you use the same placement inputs for `circle` or `hexagon` it does. 

I'm unsure if there is any upgrade issue as `channels` is used in the implementation nodegraph or just is pre-existing ?

### Test Data

Output on a 2d plane. BTW a center of 1,1 places it in the center for some reason ?

<table>
<tr>
<td>
<img src="https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/ccdbba79-928f-44d2-9afc-c0c61be152a5">

<td>
<img src="https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/913d6b2a-d852-4c52-80b7-41e3e9ee250e">

<td>
<img src="https://github.com/AcademySoftwareFoundation/MaterialX/assets/49369885/d63dc810-04e0-4cc8-ad02-81935ef9ae82">

</tr>
</table>

File used:

```xml
<?xml version="1.0"?>
<materialx version="1.39" colorspace="lin_rec709">
  <surfacematerial name="surfacematerial_material1" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="convert_surfaceshader1" />
    <input name="displacementshader" type="displacementshader" value="" />
  </surfacematerial>
  <tiledcloverleafs name="tiledcloverleafs_color4" type="color3">
    <input name="texcoord" type="vector2" value="0,0" />
    <input name="uvtiling" type="vector2" value="1,1" />
    <input name="uvoffset" type="vector2" value="0,0" />
    <input name="size" type="float" value="0.5" />
    <input name="staggered" type="boolean" value="false" />
  </tiledcloverleafs>
  <convert name="convert_surfaceshader1" type="surfaceshader">
    <input name="in" type="color3" nodename="tiledcloverleafs_color4" />
  </convert>
  <surfacematerial name="surfacematerial_material2" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="convert_surfaceshader2" />
    <input name="displacementshader" type="displacementshader" value="" />
  </surfacematerial>
  <cloverleaf name="cloverleaf_float1" type="float">
    <input name="texcoord" type="vector2" value="0,0" />
    <input name="center" type="vector2" value="0.5,0.5" />
    <input name="radius" type="float" value="0.25" />
[...]

PR #2327, opened 2025-04-02
Fix Cloverleaf offset

zicher3d:
Cloverleaf Center was not doubled as other dimensions within the node, resulting in a different offset than other similar shape nodes (circle, hexagon).

The Tiledcloverleafs node, which uses Cloverleaf, is not affected by the fix as the scale and offsets of the tiled pattern is achieved through UVs and Center is always 0,0.

This fixes the following issue:
https://github.com/AcademySoftwareFoundation/MaterialX/issues/1841

@bernardkwok Please let me know if it works on your previous tests


jstone-lucasfilm (approved):
Thanks for the fix, @zicher3d!

Issue #2324, opened 2025-04-01
Inconsistent results with `dielectric_bsdf` tinting


ld-kerley:
I was stumbling around in the graph editor and I noticed something that felt, at least visually, inconsistent to me. So I thought it was worth raising here so we can clarify what the intended result is.

[matx_layer_bug.mtlx.zip](https://github.com/user-attachments/files/19543423/matx_layer_bug.mtlx.zip)

In this MaterialX file I have four separate `surface` nodes. 

* `Surface_A` - A `dielectric_bsdf` node with a `tint` specified.
* `Surface_B` - A `dielectric_bsdf` node without a `tint`, but instead multiplied by the same tint value using `multiply`.
* `Surface_C` - The dielectric node from A layered over an `oren_nayar_diffuse_bsdf`.
* `Surface_D` - The dielectric node from B layered over the same `oren_nayar_diffuse_bsdf`.

I've rendered all of these in both GLSL via the graph editor, and in OSL's testrender (note I wrote the OSL code by hand, but replicated the same idea in the MaterialX graph. 

| Surface | GLSL | OSL |
|---------|------|-----|
| A | <img width="300" alt="Image" src="https://github.com/user-attachments/assets/c319e75c-1574-44d1-95f2-49d937bd245f" />  |  <img width="300" alt="Image" src="https://github.com/user-attachments/assets/69e67b63-bdec-466e-a210-e22baa29756b" /> |
| B | <img width="300" alt="Image" src="https://github.com/user-attachments/assets/e3570760-3ca2-411f-8068-3d4c7b36a868" />  |  <img width="300" alt="Image" src="https://github.com/user-attachments/assets/4e86da91-05f9-4ce9-8e41-abda6508deb3" />  |
| C | <img width="300" alt="Image" src="https://github.com/user-attachments/assets/103cb597-ed87-4529-b8a0-c23d6a664e6c" />  |  <img width="300" alt="Image" src="https://github.com/user-attachments/assets/fb230e65-e8eb-4090-877d-8dcc8b098e43" />  |
| D | <img width="300" alt="Image" src="https://github.com/user-attachments/assets/87124d1e-02e7-4951-bbe7-f33246bc4357" />  |  <img width="300" alt="Image" src="https://github.com/user-attachments/assets/04baebde-9f0b-4383-81ee-1d8a23b47c3f" />  |

[...]

PR #2328, opened 2025-04-02
Fix bug in multiply_bsdf for GLSL

niklasharrysson:
This change list fixes a bug in the GLSL implementation of node multiply_bsdf, as logged in #2324.

### DIFF
--- a/libraries/pbrlib/genglsl/mx_multiply_bsdf_color3.glsl
+++ b/libraries/pbrlib/genglsl/mx_multiply_bsdf_color3.glsl
@@ -4,5 +4,5 @@ void mx_multiply_bsdf_color3(ClosureData closureData, BSDF in1, vec3 in2, out BS
 {
     vec3 tint = clamp(in2, 0.0, 1.0);
     result.response = in1.response * tint;
-    result.throughput = in1.throughput * tint;
+    result.throughput = in1.throughput;
 }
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:

@ld-kerley It makes sense that this should resolve the issue you reported in #2324, but perhaps you can confirm this using your original test data?

niklasharrysson:
Sorry for not adding image results. Here's a comparison between before and after using Lee's test data:

before

after

ld-kerley:
@niklasharrysson Thanks for this - looks like it brings things back inline with the OSL implementation.

I do still think it's important that we make sure we get consensus across a number of renderer implementations that this is the correct approach to take, and ensure we correctly document this.

ld-kerley (approved):

ld-kerley:
I guess reading this again raises another question if the same idea should be applied to the `mx_multiply_bsdf_float()` case too.

With this patch in place we have a discontinuity between `float` and `color` handling here - which feels confusing to me, but perhaps this implies more the need to think more deeply about what the 100% correct answer is here.

jstone-lucasfilm:
@ld-kerley Doesn't this change already handle the BSDF*float case in addition to the BSDF*color case?  I see that in Niklas's code, but maybe I'm missing the edge case that you're thinking of.

ld-kerley:
@jstone-lucasfilm sorry - you're completely right - not enough coffee yet today.

jstone-lucasfilm:
@ld-kerley Although I see your point about gathering additional examples from production renderers, it's my sense that Niklas's solution is very likely the best that we can do in a _deterministic rasterizer_, while Monte Carlo renderers will do a slightly better job using probability-based selection of lobes.  Given that context, I'd be inclined to move forward with Niklas's fix, since it's such a dramatic improvement over the current logic, and we can consider how to borrow ideas from the Monte Carlo world in the future.

ld-kerley:
Yeah - I agree that this is at least a great step in the right direction and might be the final answer - I just want to make sure we're doing the due diligence to get the spec writing part written correctly.  Maybe we can add this to an upcoming TAC conversation

PR #2329, opened 2025-04-02
Add animated clock example

zicher3d:
Adding an animated procedural sample for a simple clock.

This was created in Maya/LookdevX and exported through the LookdevX Publishing option. Manual changes were then made to remove xpos/ypos attributes, and change the texcoord input adding defaultgeomprop="UV0".

Currently there is no way to test the animation in Viewer or Graph Editor. In Graph Editor, when clock.mtlx is loaded, it appers as a node in the "animated" nodegroup. Connect the color output to a material base color. A Time node should be plugged into the Time input to properly see the animation once the ability to play a timeline is available.


jstone-lucasfilm (approved):
Thanks for the great contribution, @zicher3d!

In the MaterialX viewer and editor, we still have work to do in handling animated examples materials, but this should provide a great first example that we can work with.

Issue #2335, opened 2025-04-06
Missing libraries when building MaterialX 1.39.3


luyatshimbalanga:
Hello team,

the build of MaterialX 1.3.9  failed due to missing some libraries in the respective headers as highlighted below from a patch as suggested by the error message:
```
diff -ru MaterialX-1.39.3.original/source/MaterialXGenShader/ShaderGenerator.h MaterialX-1.39.3/source/MaterialXGenShader/ShaderGenerator.h
--- MaterialX-1.39.3.original/source/MaterialXGenShader/ShaderGenerator.h	2025-04-05 22:48:18.949032619 -0700
+++ MaterialX-1.39.3/source/MaterialXGenShader/ShaderGenerator.h	2025-04-05 22:47:59.165943936 -0700
@@ -14,6 +14,7 @@
 #include <MaterialXGenShader/ColorManagementSystem.h>
 #include <MaterialXGenShader/Factory.h>
 #include <MaterialXGenShader/ShaderStage.h>
+#include <cstdint>
 #include <MaterialXGenShader/Syntax.h>
 
 #include <MaterialXFormat/File.h>
diff -ru MaterialX-1.39.3.original/source/MaterialXGenShader/ShaderNode.h MaterialX-1.39.3/source/MaterialXGenShader/ShaderNode.h
--- MaterialX-1.39.3.original/source/MaterialXGenShader/ShaderNode.h	2025-04-05 22:47:01.851687043 -0700
+++ MaterialX-1.39.3/source/MaterialXGenShader/ShaderNode.h	2025-04-05 22:44:41.944059886 -0700
@@ -14,6 +14,7 @@
 #include <MaterialXGenShader/ShaderNodeImpl.h>
 #include <MaterialXGenShader/TypeDesc.h>
 #include <MaterialXGenShader/GenUserData.h>
+#include <cstdint>
 
 #include <MaterialXCore/Node.h>
 
diff -ru MaterialX-1.39.3.original/source/MaterialXGenShader/TypeDesc.h MaterialX-1.39.3/source/MaterialXGenShader/TypeDesc.h
--- MaterialX-1.39.3.original/source/MaterialXGenShader/TypeDesc.h	2025-03-07 13:13:25.000000000 -0800
+++ MaterialX-1.39.3/source/MaterialXGenShader/TypeDesc.h	2025-04-05 22:54:14.141624815 -0700
@@ -13,6 +13,8 @@
[...]

PR #2338, opened 2025-04-08
Add missing `cstdint` include for GCC 15

luyatshimbalanga:
This commit includes the missing libraries detected by the build system #2335

### DIFF
--- a/source/MaterialXGenShader/TypeDesc.h
+++ b/source/MaterialXGenShader/TypeDesc.h
@@ -13,7 +13,6 @@
 #include <MaterialXCore/Value.h>
 #include <MaterialXCore/Document.h>
 
-#include <cstdint>
 #include <string_view>
 
 MATERIALX_NAMESPACE_BEGIN
--- a/source/MaterialXGenShader/TypeDesc.h
+++ b/source/MaterialXGenShader/TypeDesc.h
@@ -13,6 +13,8 @@
 #include <MaterialXCore/Value.h>
 #include <MaterialXCore/Document.h>
 
+#include <cstdint>
+#include <stdint.h>
 #include <string_view>
 
 MATERIALX_NAMESPACE_BEGIN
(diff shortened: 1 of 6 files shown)


source/MaterialXGenShader/TypeDesc.h
```diff
@@ -13,6 +13,8 @@
 #include <MaterialXCore/Value.h>
 #include <MaterialXCore/Document.h>
 
+#include <cstdint>
+#include <stdint.h>
```
jstone-lucasfilm, line 17: @luyatshimbalanga I can see the need for `cstdint` in new versions of GCC, but what is the motivation for additionally including `stdint.h` in just this one source file?

luyatshimbalanga: Thar was a suggestion from the compiler when it encounter the failure on that source. If `sdtint.h` is redundant, feel free to remove it.

jstone-lucasfilm: Thanks, @luyatshimbalanga, and I can make that change over here.  Structurally, we may also want to add a single include of `cstdint` in Library.h within MaterialXCore, rather than including it in each source file that references a fixed-width type.

jstone-lucasfilm: I've gone ahead and made that simplification to the current PR, @luyatshimbalanga, and let me know if this continues to build successfully in your GCC 15 environment.

luyatshimbalanga: The simplified PR modification runs successfully in the GCC build environment as seen on 
https://download.copr.fedorainfracloud.org/results/@designsuite/blender/fedora-rawhide-x86_64/08882889-materialx/builder-live.log

jstone-lucasfilm (approved):
Thanks for catching this, @luyatshimbalanga, and I believe we can move forward with this fix for MaterialX v1.39.4.

PR #2339, opened 2025-04-08
Fixed incorrect number of faces and vertices in createQuadMesh

tdavidovicNV:
createQuadMesh claimed the created mesh had 6 faces, and 0 vertices, while it has 2 faces and 4 vertices.


jstone-lucasfilm (approved):

PR #2341, opened 2025-04-10
Move cache clear outside material loop in viewer

tdavidovicNV:
The call to clear implementations does not need to be called for every newly added material individually.


jstone-lucasfilm (approved):

Issue #2117, opened 2024-11-17
Flattening produces invalid document


pablode:
I'm running into an issue where when I flatten following document

<details>
<summary>Show code</summary>

```xml
<?xml version="1.0"?>
<materialx version="1.38" xmlns:xi="http://www.w3.org/2001/XInclude">
  <UsdPreviewSurface name="Surface" type="surfaceshader">
    <input name="diffuseColor" type="color3" output="diffuseColor_rgb" nodegraph="prismarine2" colorspace="srgb_texture" />
    <input name="normal" type="vector3" output="normal_out" nodegraph="prismarine2" />
    <input name="roughness" type="float" output="roughness_r" nodegraph="prismarine2" colorspace="lin_rec709" />
    <input name="metallic" type="float" value="0" />
    <input name="opacity" type="float" value="1" />
    <input name="useSpecularWorkflow" type="integer" value="0" />
  </UsdPreviewSurface>
  <surfacematerial name="prismarine" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="Surface" />
  </surfacematerial>
  <nodegraph name="prismarine2">
    <UsdUVTexture name="DiffuseTexture" type="multioutput" nodedef="ND_UsdUVTexture">
      <input name="file" type="filename" value="/Users/pablode/Work/usd-wg-assets/full_assets/McUsd/McUsd_materials/tex/prismarine.png" />
      <input name="wrapS" type="string" value="periodic" />
      <input name="wrapT" type="string" value="periodic" />
      <input name="st" type="vector2" nodename="UVReader" />
    </UsdUVTexture>
    <UsdPrimvarReader name="UVReader" type="vector2" nodedef="ND_UsdPrimvarReader_vector2">
      <input name="fallback" type="vector2" value="0, 0" />
      <input name="varname" type="string" value="st" />
    </UsdPrimvarReader>
    <output name="diffuseColor_rgb" type="color3" output="rgb" nodename="DiffuseTexture" />
    <convert name="convert0" type="vector3" nodedef="ND_convert_color3_vector3">
      <input name="in" type="color3" output="rgb" nodename="NormalTexture" colorspace="lin_rec709" />
    </convert>
    <UsdUVTexture name="NormalTexture" type="multioutput" nodedef="ND_UsdUVTexture">
      <input name="bias" type="color4" value="-1, -1, -1, 0" />
[...]

PR #2348, opened 2025-04-14
Fix invalid connections in `flattenSubgraphs`

bowald:
As described in this issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2117

Sometimes when flattening a document, we can produce documents which connects nodes with other nodes not existing in the document anymore (they have been replaced during flattensubgraph).

The issue happens if a node downstream is flatten before the upstream node. The new instance will be connected to the upstream node when created, but when flattening the upstream node, it's downstream connections will be updated using the downstreamPortMap, which don't refers to the new created instances.

This PR makes sure the downstreamPortMap is in sync.

### DIFF
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -322,6 +322,12 @@ void GraphElement::flattenSubgraphs(const string& target, NodePredicate filter)
                         if (sourceInput)
                         {
                             destInput->copyContentFrom(sourceInput);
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Good catch, @bowald, and this looks like a promising fix!

jstone-lucasfilm (approved):
Thanks for this fix, @bowald, and this looks good in local tests!

PR #2350, opened 2025-04-15
Clarity improvements to translation

jstone-lucasfilm:
This changelist implements clarity improvements to the translation graph from Standard Surface to UsdPreviewSurface, making no functional changes to its behavior.


jstone-lucasfilm:
I'm CC'ing @JGamache-autodesk to make sure I'm not misunderstanding the intent of these nodes with my proposed renames!

PR #2352, opened 2025-04-18
Fix default input on `atan2` nodes

anderslanglands:
default input was set to 'in1', changed to 'iny'


ld-kerley (approved):
Nice catch thanks Anders. Looks good to me - sorry I missed this when I updated the inputs.

Issue #2149, opened 2024-12-16
Proposal : Named Values


ld-kerley:
(Related to [Generic Type System proposal](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2148).)

# Problem

Current XML based system for describing node definitions and nodegraphs requires each specialization of the node to be spelled out explicitly (e.g. `ND_add_float`, `ND_add_color3`, `ND_add_vector2`, ...). This leads to a verbose data library, and encourages copy/paste authoring which can be prone to errors.

As well as the variation of types, each different type has different value strings that need to be authored.

## Solutions

The possible solutions suggested in the [Generic Type System proposal](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2148) also needs a way to define values that correspond to the type. Most of the default values in the MaterialX Data Library are really just _zero_ or _one_. So in this proposal we suggest additional syntax to describe the concrete named values indirectly at their usage site.

The `<typedef>` element should be extended with additional attributes to hold the concrete values for the named values.

The `value` attribute then uses a special token to identify the new named value syntax, including a label identifying which of the available named values to use.

The exact selection for these tokens and the list of provided named values is a subject for further debate, but for the sake of example, we're proposing the `value` attribute use the prefix `TypeValue:` to identify this new named value behavior. This proposal also recommends as a minimum `zero` and `one` always be provided as concrete named values.

## Example
Note here we intentionally don't take any opinion on any generic type mechanism, and instead make the named value proposal in isolation. This could be implemented even if none of the generic type suggestions above are actioned.

```xml
<!-- type definitions are extended with their own concrete values for each of the named values -->
<typedef name="boolean" zero="False" one="True"/>
<typedef name="float" zero="0.0" one="1.0"/>
<typedef name="color3" zero="0.0, 0.0, 0.0" one="1.0, 1.0, 1.0"/>
<typedef name="vector3" zero="0.0, 0.0, 0.0" one="1.0, 1.0, 1.0"/>

<!-- original node definition -->
<nodedef name="ND_multiply_float" node="multiply" nodegroup="math">
  <input name="in1" type="float" value="0.0" />
  <input name="in2" type="float" value="1.0" />
  <output name="out" type="float" defaultinput="in1" />
</nodedef>

<!-- new node definition -->
<nodedef name="ND_multiply_float" node="multiply" nodegroup="math">
  <input name="in1" type="float" value="TypeValue:zero" />
[...]

Issue #2148, opened 2024-12-16
Proposal : Generic Type System


ld-kerley:
# Problem
Current  XML based system for describing node definitions and node graphs requires each specialization of the node to be spelled out explicitly, e.g. `ND_add_float`, `ND_add_color3`, `ND_add_vector2`, etc. This leads to a verbose data library, and encourages copy/paste authoring which can be prone to errors.

The size of both the source and the installed artifacts is also a consideration for the MaterialX project, and if we can reduce the footprint of the project, that would benefit several downstream MaterialX integrations.

Finally, because of the exhaustive nature of the current system, adding a new type to the system becomes incrementally more difficult each time, due to having to add new node definitions for each of the general purpose nodes. `ND_add_myType`, `ND_multiply_myType`, `ND_image_myType`, etc. This problem becomes combinatorially worse if there multiple templated types involved.

# Solutions
The general theme of the possible solutions below is to develop a system that describes the different specializations of a given node with a single, or small number, of descriptions, referencing a set of types that are valid for a given description.

## Build time template system
The simplest solution proposed here would be a build time template based system.  Here we propose augmenting the Data Library source files with additional information describing families of types, that could then be used to concretely instantiate the node definitions currently required by the system during the building of the MaterialX project.

A build time template system could be built upon:
* An existing template based systems, such as Jinja (https://jinja.palletsprojects.com/en/stable/).
* A bespoke metadata based system using custom syntax embedded in the Data Library files, or perhaps side-car files or pure python.
* An extension of the existing XML syntax, that is processed and then removed during the build stage.

Extending the XML syntax would feel the most natural option to use to progress forward in to the next solution, though perhaps would be the most cumbersome to implement as an initial step.  XML is not the easiest syntax to describe loops or other control flow within, and other systems perhaps more python based might lead to a simpler, cleaner source syntax for the Data Library.

A build time solution has the significant benefit of being incremental. It would not require any existing downstream MaterialX integrations to adopt any changes.

## Runtime polymorphic system
Another possible solution would be a runtime polymorphic system.  We could extend the XML syntax to describe the necessary the relationships between the types and/or explicitly define groups of types that could be used inplace of concrete types in node definitions.

### Example
```xml
<template name="numeric" options="float, vector2, vector3, vector4, color3, color4" />

<nodedef name="ND_range_numeric" node="range" nodegroup="adjustment">
  <input name="in" type="numeric" value="zero" />
[...]

PR #2362, opened 2025-05-02
Standard Library Templating and Named Value implementation.

ld-kerley:
This PR proposes a concrete implementation inspired by proposals presented in #2148 and #2149.  We will recap these briefly.

## Problem Statement
The standard data library is repetitive due to the type strictness of MaterialX.  A lot of node definitions need to be concretely instantiated for each type they support. For instance `<add>` ends up with 16 different node definitions, spanning about 80 lines of XML. Due to the repetitive nature of these lines of XML, it can be hard to identify typos, and encourages copy/pasting which can be error prone. 

## Proposed Solution
Instead of authoring each node definition explicitly, this PR introduces the concept of a `<template>` for specifying repetitive elements.  We also introduce a new syntax to abstract the specification of common values that have a different expression for different types, but are semantically the same, ie. `zero` and `one`. 

```
<template name="TP_ND_add" varName="typeName" options="float, integer, color3, color4, vector2, vector3, vector4">
  <nodedef name="ND_add_@typeName@" node="add" nodegroup="math">
    <input name="in1" type="@typeName@" value="Value:zero" />
    <input name="in2" type="@typeName@" value="Value:zero" />
    <output name="out" type="@typeName@" defaultinput="in1" />
  </nodedef>
</template>
```

Here we wrap a variation of the node definition for `<add>` inside a container `<template>` element. The `<template>` element has the following three attributes:
* `name` - unique identifier 
* `varName` - the name of a variable to be identified and replace within the contents of the template
* `options` - a list of strings that will used to instantiate copies of the template.

The length of the `options` list determines how many copies of the template are instantiated. Each copy will substitute a value from the `options` list for each attribute that contains the value of the `varName` attribute wrapped in `@` characters. 

In the example above `@typeName@` is replaced with each of the elements in the `options` list. 

```
<nodedef name="ND_add_float" node="add" nodegroup="math">
  <input name="in1" type="float" value="Value:zero" />
  <input name="in2" type="float" value="Value:zero" />
  <output name="out" type="float" defaultinput="in1" />
</nodedef>
<nodedef name="ND_add_integer" node="add" nodegroup="math">
  <input name="in1" type="integer" value="Value:zero" />
  <input name="in2" type="integer" value="Value:zero" />
  <output name="out" type="integer" defaultinput="in1" />
</nodedef>
<nodedef name="ND_add_color3" node="add" nodegroup="math">
  <input name="in1" type="color3" value="Value:zero" />
  <input name="in2" type="color3" value="Value:zero" />
  <output name="out" type="color3" defaultinput="in1" />
</nodedef>
<nodedef name="ND_add_color4" node="add" nodegroup="math">
  <input name="in1" type="color4" value="Value:zero" />
  <input name="in2" type="color4" value="Value:zero" />
  <output name="out" type="color4" defaultinput="in1" />
</nodedef>
<nodedef name="ND_add_vector2" node="add" nodegroup="math">
  <input name="in1" type="vector2" value="Value:zero" />
  <input name="in2" type="vector2" value="Value:zero" />
  <output name="out" type="vector2" defaultinput="in1" />
</nodedef>
<nodedef name="ND_add_vector3" node="add" nodegroup="math">
  <input name="in1" type="vector3" value="Value:zero" />
  <input name="in2" type="vector3" value="Value:zero" />
  <output name="out" type="vector3" defaultinput="in1" />
</nodedef>
<nodedef name="ND_add_vector4" node="add" nodegroup="math">
  <input name="in1" type="vector4" value="Value:zero" />
  <input name="in2" type="vector4" value="Value:zero" />
  <output name="out" type="vector4" defaultinput="in1" />
</nodedef>
```

The second expansion that happens is the resolution of the named values.  In the expanded template above we still have inputs that have `value` attributes that are not currently legal values, we see `Value:zero` and `Value:one`.

The expansion of these named values relies upon additional attributes specified in the type definitions.

```
  <typedef name="integer" zero="0" one="1"/>
  <typedef name="float" zero="0.0" one="1.0"/>
  <typedef name="color3" semantic="color" zero="0.0,0.0,0.0" one="1.0,1.0,1.0"/>
  <typedef name="color4" semantic="color" zero="0.0,0.0,0.0,0.0" one="1.0,1.0,1.0,1.0"/>
  <typedef name="vector2" zero="0.0,0.0" one="1.0,1.0"/>
  <typedef name="vector3" zero="0.0,0.0,0.0" one="1.0,1.0,1.0"/>
  <typedef name="vector4" zero="0.0,0.0,0.0,0.0" one="1.0,1.0,1.0,1.0"/>
```

Again the named value expansion is just another text replacement, we located any `value` attributes that starts with the string `Value:`, we then interrogate the type of that value, and look up the concrete value value in the corresponding type definition.

## Build vs Runtime evaluation
In order to maintain compatibility with the existing specification, the build has been updated to perform these template expansions and string replacements during the build.  This ensures that the generated data library remains compatible with the current specification.  These build time expansions are performed by C++ based tools built during the project, and using the MaterialX library itself.

The MaterialX C++ library has also been updated to provide implementation to evaluate these new concepts at runtime.  Concretely, the template expansion can be performed during file read if the read option `XmlReadOptions.expandTemplateElems` is enabled.  If this is used during the runtime, the list of node definitions returned by `Document::GetNodeDefs()` would remain unchanged, and the complete list would be returned. 

The runtime evaluation of the named value replacement takes place in `ValueElement::GetValueString()`, where we optionally apply the same string replacement logic, based on the type and the type definition.

## CMake changes

CMake arguments have been added to control these build vs runtime choices.
* `MATERIALX_BUILD_EXPAND_TEMPLATE_ELEMS` - if enabled, then the templates will be expanded during the build, and the standard data library generated by the build will contain the concrete instances of the template.
* `MATERIALX_BUILD_BAKE_NAMED_VALUES` - if enabled, the named values will be baked out to their concrete type based values. Note, if this option is enabled, then `MATERIALX_BUILD_EXPAND_TEMPLATE_ELEMS` is forced to be true, as any named values inside a template may not be able to resolve to a specific type.

In order to support the cross compilation used to generate MaterialX library for uses on other platforms, such as iOS, we have to ensure that the tools used to generate the data library are built for the build machines architecture.  The easiest way to support this cross compilation approach in CMake is to create a small subproject that allows the tools to be built with a separate toolchain. (source references [here](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html) and [here](https://stackoverflow.com/questions/36173840/how-to-instruct-cmake-to-use-the-build-architecture-compiler))

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,12 +31,21 @@ jobs:
           compiler_version: "14"
           python: 3.9
 
-        - name: Linux_GCC_14_Python311
+        - name: Linux_GCC_14_Python311_expand_lib
           os: ubuntu-24.04
           compiler: gcc
           compiler_version: "14"
           python: 3.11
           test_render: ON
+          cmake_config: -DMATERIALX_BUILD_EXPAND_TEMPLATE_ELEMS=ON -DMATERIALX_BUILD_BAKE_NAMED_VALUES=ON
+
+        - name: Linux_GCC_14_Python311_unexpanded_lib
+          os: ubuntu-24.04
+          compiler: gcc
+          compiler_version: "14"
+          python: 3.11
+          test_render: ON
+          cmake_config: -DMATERIALX_BUILD_EXPAND_TEMPLATE_ELEMS=OFF -DMATERIALX_BUILD_BAKE_NAMED_VALUES=OFF
 
         - name: Linux_GCC_CoverageAnalysis
           os: ubuntu-24.04
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,8 @@ option(MATERIALX_BUILD_BENCHMARK_TESTS "Build benchmark tests." OFF)
 
 option(MATERIALX_BUILD_SHARED_LIBS "Build MaterialX libraries as shared rather than static." OFF)
 option(MATERIALX_BUILD_DATA_LIBRARY "Build generated products from the MaterialX data library." OFF)
+option(MATERIALX_BUILD_EXPAND_TEMPLATE_ELEMS "Process the data library files at build time to expand any template elements." ON)
+option(MATERIALX_BUILD_BAKE_NAMED_VALUES "Process the data library files at build time to bake out the named values." ON)
 option(MATERIALX_BUILD_MONOLITHIC "Build a single monolithic MaterialX library." OFF)
 option(MATERIALX_BUILD_USE_CCACHE "Enable the use of ccache to speed up build time, if present." ON)
 option(MATERIALX_PYTHON_LTO "Enable link-time optimizations for MaterialX Python." ON)
@@ -67,6 +69,10 @@ if (MATERIALX_BUILD_IOS)
     message(DEPRECATION "The MATERIALX_BUILD_IOS option is deprecated. Set CMAKE_SYSTEM_NAME to iOS instead.")
     set(CMAKE_SYSTEM_NAME iOS)
 endif()
+
+list(APPEND CMAKE_MODULE_PATH
+        ${PROJECT_SOURCE_DIR}/cmake/macros)
+include(Public)
 
 # Apple ecosystem cross-compilation
 # https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-visionos-or-watchos
@@ -306,129 +312,21 @@ else()
     endif()
 endif()
 
-# Shared functions
-function(assign_source_group prefix)
-    foreach(_source IN ITEMS ${ARGN})
-        if(IS_ABSOLUTE "${_source}")
-            file(RELATIVE_PATH _source_rel "${CMAKE_CURRENT_SOURCE_DIR}" "${_source}")
-        else()
-            set(_source_rel "${_source}")
-        endif()
-        get_filename_component(_source_path "${_source_rel}" PATH)
-        string(REPLACE "/" "\\" _source_path_msvc "${_source_path}")
-        source_group("${prefix}\\${_source_path_msvc}" FILES "${_source}")
-    endforeach()
-endfunction(assign_source_group)
-
-function(mx_add_library MATERIALX_MODULE_NAME)
-    set(options ADD_OBJECTIVE_C_CODE)
-    set(oneValueArgs EXPORT_DEFINE)
-    set(multiValueArgs
-        SOURCE_FILES
-        HEADER_FILES
-        INLINED_FILES
-        MTLX_MODULES)
-    cmake_parse_arguments(args
-        "${options}"
-        "${oneValueArgs}"
-        "${multiValueArgs}"
-        ${ARGN})
-
-    if (APPLE AND args_ADD_OBJECTIVE_C_CODE)
-        file(GLOB_RECURSE materialx_source_oc "${CMAKE_CURRENT_SOURCE_DIR}/*.m*")
-        set_source_files_properties(${materialx_source_oc} PROPERTIES
-            COMPILE_FLAGS "-x objective-c++")
-        set(args_SOURCE_FILES ${args_SOURCE_FILES} ${materialx_source_oc})
-    endif()
-
-    assign_source_group("Source Files" ${args_SOURCE_FILES})
-    assign_source_group("Source Files" ${args_INLINED_FILES})
-    assign_source_group("Header Files" ${args_HEADER_FILES})
-
-    if (NOT MATERIALX_BUILD_MONOLITHIC)
-        set(TARGET_NAME ${MATERIALX_MODULE_NAME})
-        add_library(${TARGET_NAME})
-
-        # Create version resource
-        if(MATERIALX_BUILD_SHARED_LIBS AND MSVC)
-            configure_file(${PROJECT_SOURCE_DIR}/cmake/modules/MaterialXVersion.rc.in ${CMAKE_CURRENT_BINARY_DIR}/version.rc)
-            target_sources(${TARGET_NAME} PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/version.rc)
-        endif()
-
-        target_link_libraries(${TARGET_NAME}
-            PUBLIC
-            ${args_MTLX_MODULES}
-            ${CMAKE_DL_LIBS})
-
-        target_include_directories(${TARGET_NAME}
-            PUBLIC
-            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../>
-            $<INSTALL_INTERFACE:${MATERIALX_INSTALL_INCLUDE_PATH}>
-            PRIVATE
-            ${EXTERNAL_INCLUDE_DIRS})
-
-        set_target_properties(
-            ${TARGET_NAME} PROPERTIES
-            OUTPUT_NAME ${MATERIALX_MODULE_NAME}${MATERIALX_LIBNAME_SUFFIX}
-            COMPILE_FLAGS "${EXTERNAL_COMPILE_FLAGS}"
-            LINK_FLAGS "${EXTERNAL_LINK_FLAGS}"
-            INSTALL_RPATH "${MATERIALX_SAME_DIR_RPATH}"
-            VERSION "${MATERIALX_LIBRARY_VERSION}"
-            SOVERSION "${MATERIALX_MAJOR_VERSION}")
-    else()
-        set(TARGET_NAME ${MATERIALX_MONOLITHIC_TARGET})
-        add_library(${MATERIALX_MODULE_NAME} ALIAS ${MATERIALX_MONOLITHIC_TARGET})
-
-        # Store the aliased MaterialX modules name to create CMake export aliases later.
-        set_property(GLOBAL APPEND PROPERTY MATERIALX_MODULES ${MATERIALX_MODULE_NAME})
-    endif()
-
-    set_target_properties(${TARGET_NAME} PROPERTIES CXX_VISIBILITY_PRESET hidden)
-    set_target_properties(${TARGET_NAME} PROPERTIES CMAKE_VISIBILITY_INLINES_HIDDEN 1)
-
-    target_sources(${TARGET_NAME}
-        PRIVATE
-            ${args_SOURCE_FILES}
-        PUBLIC
-            FILE_SET
-                mxHeaders
-            TYPE
-                HEADERS
-            BASE_DIRS
-                ${CMAKE_CURRENT_SOURCE_DIR}/..
-                ${CMAKE_CURRENT_BINARY_DIR}/..
-            FILES
-                ${args_HEADER_FILES}
-                ${args_INLINED_FILES})
-
-    target_include_directories(${TARGET_NAME} PUBLIC
-        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../>)
-
-    target_compile_definitions(${TARGET_NAME} PRIVATE "-D${args_EXPORT_DEFINE}")
-
-    if(NOT SKBUILD)
-        if(NOT MATERIALX_BUILD_MONOLITHIC)
-            install(TARGETS ${MATERIALX_MODULE_NAME}
-                    EXPORT MaterialX
-                    ARCHIVE DESTINATION ${MATERIALX_INSTALL_LIB_PATH}
-                    LIBRARY DESTINATION ${MATERIALX_INSTALL_LIB_PATH}
-                    RUNTIME DESTINATION bin
-                    FILE_SET mxHeaders DESTINATION ${MATERIALX_INSTALL_INCLUDE_PATH})
-        endif()
-
-        install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/${MATERIALX_MODULE_NAME}.pdb"
-                DESTINATION "${MATERIALX_INSTALL_LIB_PATH}/" OPTIONAL)
-    endif()
-
-    # Pass TARGET_NAME back to call site, so the caller can modify the build target.
-    set(TARGET_NAME ${TARGET_NAME} PARENT_SCOPE)
-endfunction()
-
 # Propagate shared library setting to NanoGUI
 if(MATERIALX_BUILD_SHARED_LIBS)
     set(BUILD_SHARED_LIBS "ON")
 else()
     set(BUILD_SHARED_LIBS "OFF")
+endif()
+
+# If we're baking the named "Value:" entries - then we have to also expand any
+# template elements
+if (MATERIALX_BUILD_BAKE_NAMED_VALUES)
+    set(MATERIALX_BUILD_EXPAND_TEMPLATE_ELEMS ON)
+endif()
+
+if (MATERIALX_BUILD_EXPAND_TEMPLATE_ELEMS OR MATERIALX_BUILD_BAKE_NAMED_VALUES)
+    set(MATERIALX_BUILD_DATA_LIBRARY ON)
 endif()
 
 # Build a monolithic target - needs to be added before the other build targets that may be included.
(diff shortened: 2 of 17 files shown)


ashwinbhat:
I like this <template> node concept and the agree with the motivation. It does help reduce the duplication. I'm curious if we could just extend `nodedef` to support multiple types so that will also help with publishing/authoring workflows? In the current proposal, a template node is expanded based on types but authoring tools don't have any knowledge about this. I understand that is intentional to maintain compatibility but won't it still lead to authoring issues with repetitive edits?

jstone-lucasfilm:
Although I'm a huge fan of templated node definitions, and this feature has been a goal for the MaterialX project since before it was open-sourced, I think it's worthwhile to consider leading with a run-time implementation rather than a build-time implementation, as suggested in @ld-kerley's original [GitHub Issue](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2148#issuecomment-2622207873):

> There's a lot to consider here, and just to round out the discussion, I'll note that the Runtime Polymorphic approach has the potential to dramatically reduce the size of our runtime data libraries, potentially improving runtime performance for all MaterialX and USD applications.

> If we pursued the Runtime Polymorphic approach, then this would become a proposed feature for MaterialX v1.40, and we would use our version system to upgrade legacy documents to the latest syntax.

In our MaterialX TSC discussion today, we had the chance to refine this idea a bit further, as expressed in the following proposed steps:

- Write up a [proposal](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Proposals.md) for a `template` element, supporting run-time templating of both `nodedef` and `nodegraph` elements.
- Work through the details of this proposal with the community.
- Implement the new `template` element in MaterialXCore, with unit tests to validate all of its use cases.
- Refine the new implementation with the community.
- When we're ready for MaterialX v1.40 or v2.0, switch our standard data libraries over to leverage the new `template` element.

I think it's still worthwhile to consider the pros and cons of both the run-time and build-time solutions, and I'm very interested in additional thoughts from other teams.

ld-kerley:
I would like to highlight some of the other aspects of the discussion at today's TSC, as they relate to this PR.

* Numerous stakeholders and TSC participants expressed interest in pursuing this build time approach.
* Build time template expansion has a number of benefits.
  * Low risk.
  * Zero adoption cost downstream. No changes to the data library delivered downstream to MaterialX consumers. 
  * A practical working example will make it a lot easier to explore some of the real world complexities of this new language feature, without committing it to the spec.
  * Incremental changes that are much easier to roll back if necessary, as the change is completely self contained inside MaterialX. 
  * Does not precludes future work towards a complete runtime polymorphic system.
  * Makes building future runtime polymorphic system less work.

For those interested, a reminder of the robust discussions on the original proposals can be found [here](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2149) and [here](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2148).

jstone-lucasfilm:
This is still in-progress thinking on my part, but I wanted to share some of the concrete cases that concern me about a build-time template system:

- Because the build-time template system would be outside of the MaterialX specification, with no guarantees of data versioning, developers could easily create custom nodes that would be stranded as the build-time template system evolves.  The run-time template system, on the other hand, would be fully integrated with MaterialX data versioning, with all of the usual guarantees that legacy custom nodes remain valid indefinitely.
- The build-time template system could create a disconnect between the specified rules of MaterialX and the actual syntax that developers need to author when developing new custom nodes or implementing fixes to standard nodes.  This seems like a cognitive burden that would potentially be significant, especially for new MaterialX developers that we're encouraging to work with MaterialX for the first time or make substantial contributions to our project.  The run-time template system, on the other hand, would always maintain our traditional integration between the specified rules of MaterialX and the syntax that we expect developers to work with.

If we ended up stuck with a build-time template system for a long period of time, e.g. for months or years, I can imagine we might end up regretting our choice to follow this approach, where the costs of the build-time template system would greatly outweigh its advantages.

None of this is to say that a build-time template system is off the table, but I wanted to share some of my own thinking about the cases that concern me about this approach.

ld-kerley:
The scope of this PR is intended to be constrained solely to the standard data library files that live inside the MaterialX project.

I'm not sure I 100% understand what you mean by "custom nodes", but if that means nodes specified outside the MaterialX project codebase, i.e. nodes not in the current standard data library, then I would provide a point of clarification. The build time template expansion tools would not be installed as part of the MaterialX installation package, and thus not be available to custom node authors at all.  Those authors would and should be constrained to the syntax as defined in the specification, as they always have. This allows us the safety of iterating on the data library template definition, without impacting anyone else.

I do concede that we are considering introducing syntax to the `libraries` folder that is not 'yet' defined in the specification, but I do not believe this introduces a cognitive burden at all.  I believe its easier to read, understand and check for bugs in this

```
  <template name="TP_ND_add" varName="typeName" options="float, integer, color3, color4, vector2, vector3, vector4">
    <nodedef name="ND_add_@typeName@" node="add" nodegroup="math">
      <input name="in1" type="@typeName@" value="Value:zero" />
      <input name="in2" type="@typeName@" value="Value:zero" />
      <output name="out" type="@typeName@" defaultinput="in1" />
    </nodedef>
  </template>
```
instead of this
```
  <nodedef name="ND_add_float" node="add" nodegroup="math">
    <input name="in1" type="float" value="0.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="float" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_integer" node="add" nodegroup="math">
    <input name="in1" type="integer" value="0" />
    <input name="in2" type="integer" value="0" />
    <output name="out" type="integer" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_color3" node="add" nodegroup="math">
    <input name="in1" type="color3" value="0.0, 0.0, 0.0" />
    <input name="in2" type="color3" value="0.0, 0.0, 0.0" />
    <output name="out" type="color3" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_color4" node="add" nodegroup="math">
    <input name="in1" type="color4" value="0.0, 0.0, 0.0, 0.0" />
    <input name="in2" type="color4" value="0.0, 0.0, 0.0, 0.0" />
    <output name="out" type="color4" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_vector2" node="add" nodegroup="math">
    <input name="in1" type="vector2" value="0.0, 0.0" />
    <input name="in2" type="vector2" value="0.0, 0.0" />
    <output name="out" type="vector2" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_vector3" node="add" nodegroup="math">
    <input name="in1" type="vector3" value="0.0, 0.0, 0.0" />
    <input name="in2" type="vector3" value="0.0, 0.0, 0.0" />
    <output name="out" type="vector3" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_vector4" node="add" nodegroup="math">
    <input name="in1" type="vector4" value="0.0, 0.0, 0.0, 0.0" />
    <input name="in2" type="vector4" value="0.0, 0.0, 0.0, 0.0" />
    <output name="out" type="vector4" defaultinput="in2" />
  </nodedef>
  <nodedef name="ND_add_matrix33" node="add" nodegroup="math">
    <input name="in1" type="matrix33" value="1.0,0.0,0.0, 0.0,1.0,0.0, 0.0,0.0,1.0" />
    <input name="in2" type="matrix33" value="0.0,0.0,0.0, 0.0,0.0,0.0, 0.0,0.0,0.0" />
    <output name="out" type="matrix33" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_matrix44" node="add" nodegroup="math">
    <input name="in1" type="matrix44" value="1.0,0.0,0.0,0.0, 0.0,1.0,0.0,0.0, 0.0,0.0,1.0,0.0, 0.0,0.0,0.0,1.0" />
    <input name="in2" type="matrix44" value="0.0,0.0,0.0,0.0, 0.0,0.0,0.0,0.0, 0.0,0.0,0.0,0.0, 0.0,0.0,0.0,0.0" />
    <output name="out" type="matrix44" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_color3FA" node="add" nodegroup="math">
    <input name="in1" type="color3" value="0.0, 0.0, 0.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="color3" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_color4FA" node="add" nodegroup="math">
    <input name="in1" type="color4" value="0.0, 0.0, 0.0, 0.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="color4" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_vector2FA" node="add" nodegroup="math">
    <input name="in1" type="vector2" value="0.0, 0.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="vector2" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_vector3FA" node="add" nodegroup="math">
    <input name="in1" type="vector3" value="0.0, 0.0, 0.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="vector3" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_vector4FA" node="add" nodegroup="math">
    <input name="in1" type="vector4" value="0.0, 0.0, 0.0, 0.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="vector4" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_matrix33FA" node="add" nodegroup="math">
    <input name="in1" type="matrix33" value="1.0,0.0,0.0, 0.0,1.0,0.0, 0.0,0.0,1.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="matrix33" defaultinput="in1" />
  </nodedef>
  <nodedef name="ND_add_matrix44FA" node="add" nodegroup="math">
    <input name="in1" type="matrix44" value="1.0,0.0,0.0,0.0, 0.0,1.0,0.0,0.0, 0.0,0.0,1.0,0.0, 0.0,0.0,0.0,1.0" />
    <input name="in2" type="float" value="0.0" />
    <output name="out" type="matrix44" defaultinput="in1" />
  </nodedef>
```

Finally I'll also note, because this is completely self contained inside the MaterialX project - if we all agree it's been a terrible idea, and we want to revert back to a non-template world. It will be trivial to bake out the templates and commit them to the project.  Leaving us exactly where we are today.

rafalSFX:
It seems to me that the build-time solution is a stepping stone towards the run-time solution. Ie, the work on real-time will require code that bakes out the concrete `nodedefs` from the `templates`. It's just that the build-time solution exposes it sooner.

Personally I prefer incremental and evolutionary approach to development, so I would lean towards the build-time approach, because it allows assessment of the direction earlier in the process. As long as Jonathan's reservations are addressed.

So, for example, regarding custom nodes, I would imagine developers would still keep using the formal XML syntax. I see the build-time step as an unofficial prototype to do some ground work for run-time, for testing, and to get early benefits of the templates.

I understand Jonathan's concern about unsanctioned syntax, though. I'm not sure if there is a good way of isolating it (and template development) from the official .mtlx sources. Maybe one way is to use  templates for some small subset of test (ie, non-standard) nodedefs only, until the proposed specification stabilizes?

It's also a valid concern about using the build-time solution for too long before run-time solution arrives. But than again, run-time solution will not be stable unless we test it, and therefore even the run-time-only solution may experience evolution and yield undesirable, stranded documents.

I also wonder how much is involved in going from build-time to run-time support.

mapetikea:
I read the PR as "make it easier to author standard library nodes without changing how you use MaterialX"? If that assumption is correct; I think it makes a lot of sense.

I'm not sure if I fully understand the reasoning/goals behind the runtime proposal. I think we need to be very clear with our goals when we say that it will become more efficient. For example, is the goal to have an effective solution for desktop use, or do we want to simplify the use of MaterialX in a web environment? Are we targeting the complete MaterialX API or do we want to create minimal evaluation engine without authoring parts?

Maybe the solution for all of the above is (partly) a generic type system, but I think we need to clear on what we are trying to achieve?

Jan-Jordan:
Some first remarks:
especially with the originally proposed syntax it feels more like function overloading than function templates. 

For overloading it has a very restrictive set of allowed overloads (can be good and bad)
1. all overloads  have to have the same number of parameters
2. inputs/outputs are either fixed type or unfixed (typed based on a defined list of types). 
3. all unfixed types are from that list or????  (whats the rule that allows ND_add_color3FA to be generated from TP_ND_add + types? what is altogether legal in this set of overloads?)

con: having both a list of overloads and a list of types, like in the original proposal, seems both superfluous and error prone.  
pro: starting with a small set of allowed overloads and simple rules for resolving the overloads makes it easier to expand later

I don't think putting this into "build" is not at all more flexible than putting it into "runtime" when it comes to changing/reverting the feature. This needs to be reasonably sound in any case.

jesschimein:
My primary concern would be to keep USD and MaterialX aligned and working together smoothly, which I think is doable with either option.  

My understanding  is that the build-time version wouldn't change the data library and a run-time option would be far enough in the future that we could start  discussions early on how to keep USD in alignment.

ld-kerley:
Thanks to everyone for chiming in here. Having a wider set of community input for important projects within MaterialX is important, and really helps bring balance to the conversation. It's important that the project considers input from as many of the stakeholders as possible.

Firstly I'd like to highlight the great work added here by @anderslanglands. He's taken the time to template-ize a large section of the standard library here. This will not only hugely help with the ongoing documentation effort, but also helped highlight a few potential errors in the existing standard library.  Anders nicely documents all the "diffs" in the PR he posted to my fork [here](https://github.com/ld-kerley/MaterialX/pull/1).  

Notably the `atan2` input default values are inconsistent across different types.  I believe this to be a bug in the standard data library, one that has likely been there for a while now. This example highlights the value of templates in making sure we don't have drift between implementations, and how adopting a template system will help make discrepancies a thing of the past.  The templated `stdlib_defs.mtx` file is now less than half the size of the untemplated version, shrinking to under 2500 lines from just over 5000. I expect other files to reduce even further, especially those with many repeated nodegraph implementations.

This nicely feathers in to responding to @mapetikea.  Yes, your understanding is correct, one the major benefits to this approach is exactly that we "make it easier to author standard library nodes without changing how you use MaterialX". I also think you raise very valid points regarding making sure we scope any possible runtime implementation of this system, particularly as we have different downstream implementors with very different use-cases.

As @jesschimein points out, this build-time approach would ensure the data library would remain compatible with USD today, and allow a much longer period of time for development and adoption of any possible run-time implementation.

@Jan-Jordan - I certainly agree with your observations, and somewhat intentionally opted to start with the simplest for of this templating system I could imagine, while still being useful.  To my mind of on the benefits of this build-time approach is that it allows us to iterate internally to the project source, without commit to a system.  I think its very possible the final system we end up with will be richer in terms of its expressiveness and capabilities, but as @rafalSFX points out that incremental journey is probably an easier one for the community to walk along together.

@rafalSFX - I'd like to try and provide an alternative view to your comment about unsanctioned syntax.  This "work-in-progress" template syntax would only exist in the MaterialX source repository, and never in any build/deployment of the project. So in some senses I would consider that "private", until we are shipping `mtlx` files with the syntax in (in a potential runtime solution in the future).  If there is a concern regarding people using the GitHub repo as a reference repository, which is certainly valid, then I'd point out first that no tooling will create files with this syntax, and if anyone does create it by hand, we could ensure that it fails MaterialX validation, and it wouldn't function in any of the existing MaterialX tooling correctly either.  We could also bake out a version of the expanded data library to some location in the repository for reference, if that was deemed a helpful source of information for people wanting to understand the formal MaterialX syntax, without building it themselves.

Finally, @jstone-lucasfilm, I respectfully disagree that the syntax is "an entirely different syntax". I feel that anyone who is knowledgeable enough to be reading the MaterialX data library files directly, is likely to be able to infer the intention of the `<template>` element, and the significant reduction in the size of the data library, coupled with it being less error prone, seems like it would actually make the task easier.  I'd also point out that the idea has been through a proposal process of sorts, in the form of the previously posted GitHub issues.

jstone-lucasfilm:
@ld-kerley I'd still really like to see this idea go through our specification proposal process before committing any code or changes to our data libraries, but I'm certainly open to further discussion on this!

rafalSFX:
Anders' [PR](https://github.com/ld-kerley/MaterialX/pull/1) looks really good! Thank you! 

Seems like the spec proposal could be based on that.

rafalSFX:
One thing that caught my eye is how the `uniform="true"` attribute of an input element forces a separate template. It would be nice to have just one template, but I suspect it's gonna be hard to fold the `uniform` attribute into the template. Especially in the first version of the system. Anyway, I just wanted to point it out.

```
  <template name="TP_ND_constant" varName="typeName" options="float, color3, color4, vector2, vector3, vector4, boolean, integer">
    <nodedef name="ND_constant_@typeName@" node="constant" nodegroup="procedural">
      <input name="value" type="@typeName@" value="Value:zero" />
      <output name="out" type="@typeName@" default="Value:zero" />
    </nodedef>
  </template>

  <template name="TP_ND_constant_uniform" varName="typeName" options="string, filename">
    <nodedef name="ND_constant_@typeName@" node="constant" nodegroup="procedural">
      <input name="value" type="@typeName@" value="Value:zero" uniform="true" />
      <output name="out" type="@typeName@" default="Value:zero" />
    </nodedef>
  </template>
```

jstone-lucasfilm:
@ld-kerley This is more of a question for a future specification proposal, but would it be possible to restrict the names of templated elements to the allowable character set for [MaterialX element names](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#materialx-names), so that they won't be marked as invalid when these elements appear in runtime documents?

Some design discussion is needed here, and I'd want @dbsmythe to provide his review of any syntax proposal for our specification, but one possibility would simply be to omit the bracketing `@` characters in your templated names, applying the rule that any appearance of this name will be replaced by the appropriate type.

ld-kerley:
@rafalSFX - one of the really nice things about going through this process (well Anders doing it) is that we found a number of other small inconsistencies in the data library.  Some that I believe are bugs (see the atan2 diff in [Anders PR](https://github.com/ld-kerley/MaterialX/pull/1)), and others that are more of a grey area (does `<add>` for a matrix really need its default values to be identity?), but would be good to re-examine.  I think this is a really strong argument for having templates at least at the build step - as it makes these things so much easier to see. 

@jstone-lucasfilm - I'm very open to changes here - my choice of `@` was somewhat arbitrary.  I just wanted to select a character that I didn't think would clash with anything else.  When doing string substitution like this, it's important to ensure there is no accidental substitution, I've been bitten by that one too many times before.  So if we wanted to select another character thats fine, but I think we should say that whatever character (or series of characters) we select is not valid in other attribute values, to try and avoid incorrect substitutions.

ld-kerley:
Closing this PR in favor of #2451 which includes some improvements based on feedback received, and also includes the document building tools - which have been really helpful to experiment with the template system.

PR #2368, opened 2025-05-04
Fix GLSL implementation of layer node

jstone-lucasfilm:
This changelist fixes a typo in the GLSL implementation of the `layer` node, where the throughput term was computed via addition rather than multiplication.

For more straightforward validation, the `layer_bsdf.mtlx` example has been updated so that the throughput computation in `layer` influences its rendered result.


jstone-lucasfilm:
Here's a render of the layering example before the fix, where energy is unintentionally created between layers:

LayerExample_BeforeFix

And here's a render of the layering example after the fix:

LayerExample_AfterFix

ld-kerley (approved):
Looks good to me - thanks for catching my typo - sorry it slipped thru.

I'm glad we have a test case that exercises this correctly now.

PR #2369, opened 2025-05-05
ShaderGen : Fix interface input value for 

kwokcb:
## Issue

When a node is created in shader generation the incorrect values may be stored due to:
1. A node definition's input interface resolved values are not used.
2. A node instance's input interface  resolved  values are not used.

## Notes

There were past fixes for proper meta-data passing such as colorspace / units and geometry stream bindings on interface inputs. This addresses the missing evaluation of resolved values.

## Fix

1. On node creation query the resolved value on nodedef interface inputs instead of unresolved values
2. On node creation check for interface input values and stored resolved ones if found.

Note that the logic for token resolution has not changed, it was just not being called for properly.

## Test

Add a new variant for token resolution by adding downstream materials to the previous token test.
The previous test resolves properly for some cases if the interface inputs are at the nodegraph level of an unconnected output.

In order, parent, nodedef, top and sibling level token override tests.

image
image
image
image

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -576,6 +576,14 @@ ShaderGraphPtr ShaderGraph::create(const ShaderGraph* parent, const string& name
             if (nodeInput)
             {
                 ValuePtr value = nodeInput->getResolvedValue();
+                if (!value)
+                {
+                    InputPtr interfaceInput = nodeInput->getInterfaceInput();
+                    if (interfaceInput)
+                    {
+                        value= interfaceInput->getResolvedValue();
+                    }
+                }
                 if (value)
                 {
                     const string& valueString = value->getValueString();
--- a/source/MaterialXGenShader/ShaderNode.cpp
+++ b/source/MaterialXGenShader/ShaderNode.cpp
@@ -349,7 +349,7 @@ void ShaderNode::initialize(const Node& node, const NodeDef& nodeDef, GenContext
                 InputPtr interfaceInput = nodeInput->getInterfaceInput();
                 if (interfaceInput)
                 {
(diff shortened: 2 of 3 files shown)


kwokcb:
@niklasharrysson, @jstone-lucasfilm I think both of you were involved in previous interface input resolve issues so pinging you first.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -576,6 +576,14 @@ ShaderGraphPtr ShaderGraph::create(const ShaderGraph* parent, const string& name
             if (nodeInput)
             {
                 ValuePtr value = nodeInput->getResolvedValue();
+                if (!value)
```
kwokcb, line 579: Seems interface input values were not being checked at all on node instances.

source/MaterialXGenShader/ShaderNode.cpp
```diff
@@ -349,7 +349,7 @@ void ShaderNode::initialize(const Node& node, const NodeDef& nodeDef, GenContext
                 InputPtr interfaceInput = nodeInput->getInterfaceInput();
                 if (interfaceInput)
                 {
-                    portValue = interfaceInput->getValue();
+                    portValue = interfaceInput->getResolvedValue();
```
kwokcb, line 352: Interface inputs on nodedefs should get the resolved value.

niklasharrysson:
> @niklasharrysson, @jstone-lucasfilm I think both of you were involved in previous interface input resolve issues so pinging you first.

Thanks Bernard, this looks like a good fix!

kwokcb:
Thanks @niklasharrysson , @jstone-lucasfilm let me know what you think. I would be nice to get this in as I think it
may also resolve some of the other existing input interface data passing issues but would need to check.

PR #2373, opened 2025-05-07
Remove C++ unlit surface implementation 

ld-kerley:
This PR removes the C++ implementation of `ND_surface_unlit` and replaces it with source code for both GLSL and MSL shader language targets.


kwokcb:
I'm curious if this should be a nodegraph implementation ?

ld-kerley:
I'm not sure if this can be a nodegraph implementation - because the node outputs a `surfaceshader` type and some target shader language use different data representations for that type.  Open to ideas around that if you have any....

libraries/stdlib/genglsl/mx_surface_unlit.glsl
```diff
@@ -0,0 +1,8 @@
+
+void mx_surface_unlit(float emission, vec3 emission_color, float transmission, vec3 transmission_color, float opacity, out surfaceshader result)
```
jstone-lucasfilm, line 2: What would you think about combining these terms as follows?

```
    result.color = emission * emission_color * opacity;
    result.transparency = mix(vec3(1.0), transmission * transmission_color, opacity);
```

To my eye, that seems more straightforward for the reader to understand, with one assignment each to `result.color` and `result.transparency`.

jstone-lucasfilm (commented):
This looks like a really promising change, @ld-kerley, and I had one suggestion for an improvement.

kwokcb:
Perhaps I'll leave it as a suggestion as I tried to create this graph and `surfaceshader` cannot really handle the alpha component properly from my attempt.

jstone-lucasfilm (approved):
This looks great to me, @ld-kerley, and thanks for implementing this improvement!

PR #2374, opened 2025-05-08
Fix failing hextile unit testing

ld-kerley:
The recent addition of an example hextile unit test is currently failing the OSL render tests.  The do not run in CI because of the time it takes to build OSL - so this failure was not caught during the merge checks.

This PR adds a filtering mechanism to the render tests, similar to the shader generation tests, to allow for specific files to be excluded for render testing.  The hextile tests are then excluded from OSL render testing, as that node has not yet been implemented.

Also removing the hextile exclusion for shader generation for MSL.

### DIFF
--- a/libraries/README.md
+++ b/libraries/README.md
@@ -75,4 +75,4 @@ This folder contains the standard data libraries for MaterialX, providing declar
     - point, directional, spot
 - Shader generation does not currently support:
     - `displacementshader` and `volumeshader` nodes for hardware shading targets (GLSL, MSL).
-    - `hextiledimage` and `hextilednormalmap` for MSL, OSL, and MDL.
+    - `hextiledimage` and `hextilednormalmap` for OSL and MDL.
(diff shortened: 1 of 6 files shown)


jstone-lucasfilm (commented):
This looks good to me, @ld-kerley!

In order to maintain the accuracy of our new target support documentation, should we remove MSL from the list of unsupported platforms for `hextiledimage` and `hextilednormalmap` here?

https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/libraries#target-support

ld-kerley:
Sure thing - I can update that document - I always forget it even exists - as we were discussing in the TSC I think we need to find a way to consolidate this information with the spec. 

Also I'm going to revert the title of the PR - as I am not changing the language support available in the codebase here at all - I'm just modifying the test setup so it doesn't fail.

msuzuki-nvidia:
Yes thank you for doing this @ld-kerley, it looks good to me too!

jstone-lucasfilm:
@ld-kerley That original title doesn't follow our conventions on the MaterialX project, especially with that final period that goes against common GitHub practice, but it's too late now!  :D

PR #2378, opened 2025-05-12
Ensure determinism of ShaderGen with color and unit transforms

ld-kerley:
While working on some shader code generation caching systems, I noticed that I was getting non-deterministic code generation if colorspace transformation was enabled.  Upon further investigation the cause appears to be the use of `std::unordered_map` with a pointer as the key.  

```
 std::unordered_map<ShaderInput*, ColorSpaceTransform> _inputColorTransformMap;
```

I'm not super familiar with the history of the code here to know if there's a good reason this is a map, but it appears safe to change this to a vector of pairs, to ensure we process the colorspace transformations in a consistent order.

The same logic appears to apply to the unit transformations, so I changed the storage there too.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -1116,11 +1116,11 @@ void ShaderGraph::populateColorTransformMap(ColorManagementSystemPtr colorManage
             {
                 if (asInput)
                 {
-                    _inputColorTransformMap.emplace(static_cast<ShaderInput*>(shaderPort), transform);
+                    _inputColorTransformMap.emplace_back(static_cast<ShaderInput*>(shaderPort), transform);
                 }
                 else
                 {
-                    _outputColorTransformMap.emplace(static_cast<ShaderOutput*>(shaderPort), transform);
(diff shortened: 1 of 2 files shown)


kwokcb:
Thanks for fixing this up. LGTM.

jstone-lucasfilm (commented):
Great work, @ld-kerley, and I don't see any issues with these code changes.

Would it be acceptable for us to standardize the title of this pull request before merging?  The usual GitHub practice is for the first word to be an imperative verb, with 50 or fewer total characters in the title, and no period at the end.

One suggested replacement for the title would be `Improve determinism of ShaderGen with color or unit transforms`, but let me know how that sounds to you.

ppenenko:
Thanks for fixing this @ld-kerley . I believe I've run into this too, coincidentally, while working on shader caching code. Thought I'd link to it in case there's any overlap: https://github.com/PixarAnimationStudios/OpenUSD/pull/3661

At the risk of stating the obvious, codegen non-determinism is a performance issue, because video drivers cache final shader binaries by hashing the input shader sources, so the always-changing sources cause constant recompilation. We've observed the performance impact of this on Windows+OpenGL and MacOS+Metal.

PR #2383, opened 2025-05-15
Improve Add Node Workflow in Graph Editor

kiyoyukinakagaki-fn:
For #1542 

Continuing on from @beersandrew 's early work on adding the ability to add a node by drawing a link from a pin.

For reference, this is @beersandrew 's original PR: https://github.com/AcademySoftwareFoundation/MaterialX/pull/2157

What I have added/modified so far:
* Improvement on @beersandrew 's filtering logic for simpler if statements and also for categorized filtered candidate nodes in the tree menu
* Filter out nodes having no inputs of matching type rather than filtering by node types only as per @beersandrew 's original plan
* Added the logic to automatically link the newly created node back to the original pin, from which the link was drawn from

Also done the reverse direction too.

* Distinguish if the currently drawn link is forward or backward
* Candidate nodes are filtered based on output types if backward
* Connect the original input pin back to the output pin of the newly created node

Ready for review now.

https://github.com/user-attachments/assets/f7c7fd5e-34d0-4aa9-ad14-acf5f4aaa36f

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3677,7 +3677,7 @@ void Graph::showHelp() const
 
 void Graph::addNodePopup(bool cursor)
 {
-    bool open_AddPopup = ImGui::IsWindowFocused(ImGuiFocusedFlags_RootWindow) && ImGui::IsKeyReleased(ImGuiKey_Tab) ||
+    bool open_AddPopup = (ImGui::IsWindowFocused(ImGuiFocusedFlags_RootWindow) && ImGui::IsKeyReleased(ImGuiKey_Tab)) ||
         (_pinFilterType != mx::EMPTY_STRING && ImGui::IsMouseReleased(0));
     static char input[32]{ "" };
     if (open_AddPopup)
@@ -3720,7 +3720,8 @@ void Graph::addNodePopup(bool cursor)
                 // Allow spaces to be used to search for node names
                 std::replace(subs.begin(), subs.end(), ' ', '_');
 
-                if ((subs.size() == 0 || str.find(subs) != std::string::npos) && (_menuFilterType == mx::EMPTY_STRING || node.getType() == _menuFilterType))
+                if ((subs.size() == 0 || str.find(subs) != std::string::npos) &&
+                    (_menuFilterType == mx::EMPTY_STRING || node.getType() == _menuFilterType))
                 {
                     if (ImGui::MenuItem(getUserNodeDefName(nodeName).c_str()) || (ImGui::IsItemFocused() && ImGui::IsKeyPressedMap(ImGuiKey_Enter)))
                     {
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -2007,12 +2007,18 @@ void Graph::addNode(const std::string& category, const std::string& name, const 
         newNode->_showAllInputs = true;
         node->setType(type);
         ++_graphTotalSize;
+        _pinIdToLinkToForAddedNode = ed::PinId();
         for (mx::InputPtr input : defInputs)
         {
             UiPinPtr inPin = std::make_shared<UiPin>(_graphTotalSize, &*input->getName().begin(), input->getType(), newNode, ax::NodeEditor::PinKind::Input, input, nullptr);
             newNode->inputPins.push_back(inPin);
             _currPins.push_back(inPin);
             ++_graphTotalSize;
+
+            if (_pinIdToLinkFromForAddedNode != ed::PinId() && _pinIdToLinkToForAddedNode == ed::PinId() && _menuFilterType == input->getType())
+            {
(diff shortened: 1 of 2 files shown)


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2001,12 +2007,18 @@ void Graph::addNode(const std::string& category, const std::string& name, const
...
         ++_graphTotalSize;
+        _pinIdToLinkToForAddedNode = ed::PinId();
         for (mx::InputPtr input : defInputs)
         {
             UiPinPtr inPin = std::make_shared<UiPin>(_graphTotalSize, &*input->getName().begin(), input->getType(), newNode, ax::NodeEditor::PinKind::Input, input, nullptr);
             newNode->inputPins.push_back(inPin);
             _currPins.push_back(inPin);
             ++_graphTotalSize;
+
+            if (_pinIdToLinkFromForAddedNode != ed::PinId() && _pinIdToLinkToForAddedNode == ed::PinId() && _menuFilterType == input->getType())
```
lfl-eholthouser, line 2018: Hi @kiyoyukinakagaki-fn ! Thanks for picking up this issue. I think things are looking great, I just have a few comments. The variable names `_pinIdToLinkToForAddedNode` and `_pinIdToLinkFromForAddedNode` are a bit hard to distinguish from each other since they are both so long and similar. Maybe they could be shorted to just, `_pinIdToLinkTo` and `_pinIdToLinkFrom`.

lfl-eholthouser, line 2018: Since  _`pinIdToLinkToForAddedNode = ed::PinId()` is always being set on line 2010, do we need the check in this if statement?

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2562,179 +2570,202 @@ void Graph::addLink(ed::PinId startPinId, ed::PinId endPinId)
...
         }
     }
+    return true;
+}
 
-    if (ed::AcceptNewItem())
+void Graph::addLink(ed::PinId startPinId, ed::PinId endPinId)
+{
+    // Prefer to assume left to right - start is an output, end is an input; swap if inaccurate
+    if (UiPinPtr inputPin = getPin(endPinId); inputPin && inputPin->_kind != ed::PinKind::Input)
```
lfl-eholthouser, line 2598: I think I may be missing something, I see this line is also being called on 2523, but I'm a bit confused about calling the pin inputPin but then also making sure the kind isn't Input and that it's not an input pin.

source/MaterialXGraphEditor/Graph.cpp
kiyoyukinakagaki-fn: Hi @lfl-eholthouser 
Thank you for your valuable feedback!
I agree they are hardly distinguishable. 'll rename those variables.

kiyoyukinakagaki-fn: Yes, we do. We initialize `_pinIdToLinkToForAddedNode` before the loop. Then, during the loop, we want to set it to the pin of matching type only if it has not been set previously during the loop. That way, this if block can pick up the **first** pin of the matching type. If we don't do the check, that variable will keep overwritten by subsequent inputs of matching types, so the pin to automatically link to will be the **last** pin rather than the **first** pin, which we prefer to automatically connect to.

source/MaterialXGraphEditor/Graph.cpp
kiyoyukinakagaki-fn: That logic was originally in addLink() before my change. My change splits addLink() into checkCanAddLink() and addLink(), so they both inherited the same logic. I just copied&pasted what was there into both.

 By the look of it, the original intention was:
* To start with, this function assumes startPin is an output while endPin is an input. If that way all good.
* If they come in reversed (i.e startPin being an input while endPin being an output), this if block "corrects" the order as a safe-guard convenience mechanism.
* To that end, I'm guessing the variable name is named inputPin, because the endPin is assumed to be an input. The if block is correct the order to meet the assumption.

Having observed above and taken your concern in, I also agree that calling it inputPin is a bit confusing. Since its the pin counterpart of endPinId, I'll rename it to endPin.

source/MaterialXGraphEditor/Graph.cpp
kiyoyukinakagaki-fn: I've got rid of the line in a subsequent commit
 `_pinIdToLinkToForAddedNode = ed::PinId()`

lfl-eholthouser: Ah, I missed the check from the previous loop. Thanks for clarifying!

source/MaterialXGraphEditor/Graph.cpp
lfl-eholthouser: Got it, thanks @kiyoyukinakagaki-fn !

source/MaterialXGraphEditor/Graph.cpp
lfl-eholthouser: Thanks for all of your work on this issue. I think the changes look great!

Issue #1847, opened 2024-05-30
Support updating references when renaming a node or nodegraph


kwokcb:
## Issue

This was raised by @dgovil in Slack that there is no such API in existence. `Element::setName()` will just the name but not update references resulting in breaking connections.

## Proposal
- The suggestion as to add a optional argument to `setName(name, updateReferences=false)` which defaults to not updating references.
  This might work but only `Node` and `NodeGraph` classes support the require call to find downstream pots.
- Another suggestion would be to add this on the derived classes of `Node` and `NodeGraph`.
- Another possibility is to add in a `rename()` method.

## Implementation
This is what I cam up with as the logic required. 
```python
 def renameNode(node, newName : str, updateReferences : bool = True):

    if not node or not newName:
        return
    if not (node.isA(mx.Node) or node.isA(mx.NodeGraph)):
        print('A non-node or non-nodegraph was passed to renameNode()')
        return 
    if node.getName() == newName:
        return

    parent = node.getParent()
    if not parent:
        return

    newName = parent.createValidChildName(newName)

    if updateReferences:
        downStreamPorts = node.getDownstreamPorts()
        if downStreamPorts:
            for port in downStreamPorts:
                #if (port.getNodeName() == node.getName()): This is assumed from getDownstreamPorts()
                oldName = port.getNodeName()
                if (port.getAttribute('nodename')):
                    port.setNodeName(newName)
                    print('  > Update downstream port: "' + port.getNamePath() + '" from:"' + oldName + '" to "' + port.getAttribute('nodename') + '"')
                elif (port.getAttribute('nodegraph')):
                    port.setAttribute('nodegraph', newName)
                    print('  > Update downstream port: "' + port.getNamePath() + '" from:"' + oldName + '" to "' + port.getAttribute('nodegraph') + '"')
                elif (port.getAttribute('interfacename')):
                    port.setAttribute('interfacename', newName)
                    print('  > Update downstream port: "' + port.getNamePath() + '" from:"' + oldName + '" to "' + port.getAttribute('interfacename') + '"')

    node.setName(newName)
```
It might be "nice" to add in a `renameConnection()` which would check the usage of `interfacename` or `nodename` or `nodegraph`. Long term it would better to just use 1 connection string identifier.

dgovil:
Thanks, @kwokcb!

rasmustilljander:
[...]

PR #2386, opened 2025-05-15
Add setName with updateReferences

rasmustilljander:
This solves (or partially solves): https://github.com/AcademySoftwareFoundation/MaterialX/issues/1847

This commit adds the `setName` method on `Node` and `NodeGraph`.
- When `updateReferences == false`, behave exactly like the original `setName()`.  
- When `updateReferences == true`, also walk downstream ports or inputs and updates corresponding node or nodegraph name.

 **API added**
  ```cpp
  void Node::setName(const std::string& newName, bool updateReferences);
  void NodeGraph::setName(const std::string& newName, bool updateReferences);

### DIFF
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -20,6 +20,26 @@ const string Backdrop::HEIGHT_ATTRIBUTE = "height";
 // Node methods
 //
 
+void Node::setName(const std::string& name, bool updateReferences)
+{
+    if (!updateReferences)
+    {
+        Element::setName(name);
+        return;
+    }
+    if (name.empty() || getName() == name)
+    {
+        return;
+    }
+    std::string validName = getParent()->createValidChildName(name);
+    vector<PortElementPtr> downStreamPorts = getDownstreamPorts();
+    for (PortElementPtr& port : downStreamPorts)
+    {
+        port->setNodeName(validName);
+    }
+    setName(validName);
+}
+
 void Node::setConnectedNode(const string& inputName, ConstNodePtr node)
 {
     InputPtr input = getInput(inputName);
@@ -559,6 +579,26 @@ string GraphElement::asStringDot() const
 //
 // NodeGraph methods
 //
+
+void NodeGraph::setName(const std::string& name, bool updateReferences)
+{
+    if (!updateReferences)
+    {
+        Element::setName(name);
+        return;
+    }
+    if (name.empty() || getName() == name)
+    {
+        return;
+    }
+    std::string validName = getParent()->createValidChildName(name);
+    vector<PortElementPtr> downStreamPorts = getDownstreamPorts();
+    for (PortElementPtr& port : downStreamPorts)
+    {
+        port->setNodeGraphString(validName);
+    }
+    setName(validName);
+}
 
 vector<OutputPtr> NodeGraph::getMaterialOutputs() const
 {
--- a/source/MaterialXCore/Node.h
+++ b/source/MaterialXCore/Node.h
@@ -58,6 +58,16 @@ class MX_CORE_API Node : public InterfaceElement
     }
     virtual ~Node() { }
 
+    /// @name Name
+    /// @{
+
+    /// Set the element's name string.  The name of a MaterialX element must be
+    /// unique among all elements at the same scope. Optionally updates downstream ports.
+    /// @throws Exception if an element at the same scope already possesses the
+    ///    given name.
+    void setName(const std::string& name, bool updateReferences = false);
+
+    /// @}
     /// @name Connections
     /// @{
 
@@ -327,6 +337,13 @@ class MX_CORE_API NodeGraph : public GraphElement
     }
     virtual ~NodeGraph() { }
 
+    /// Set the element's name string.  The name of a MaterialX element must be
+    /// unique among all elements at the same scope. Optionally updates downstream ports.
+    /// @throws Exception if an element at the same scope already possesses the
+    ///    given name.
+    void setName(const std::string& name, bool updateReferences = false);
+
+    /// @}
     /// @name Material References
     /// @{
 
--- a/source/MaterialXTest/MaterialXCore/Node.cpp
+++ b/source/MaterialXTest/MaterialXCore/Node.cpp
@@ -774,6 +774,123 @@ TEST_CASE("Node Definition Creation", "[nodedef]")
 
         doc->removeChild(graph->getName());
     }
-    
+
     REQUIRE(doc->validate());
+}
+
+TEST_CASE("renameElement with connections", "[node, nodegraph]")
+{
+    mx::DocumentPtr doc = mx::createDocument();
+
+    const std::string type = "float";
+    const std::string new_name = "new_name";
+    SECTION("node")
+    {
+        // Upstream -> Downstream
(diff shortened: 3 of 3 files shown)


source/MaterialXCore/Node.cpp
```diff
@@ -20,6 +20,26 @@ const string Backdrop::HEIGHT_ATTRIBUTE = "height";
 // Node methods
 //
 
+void Node::setName(const std::string& name, bool updateReferences)
+{
+    if (!updateReferences)
```
ld-kerley, line 25: I know the logic here was lifted directly from Bernards python prototype - but I'm wondering if it's better here to just call `Element::setName(name)` first all the time. To ensure that we're using consistent naming functionality across all the elements, and then perform the reference updating afterwards. 

The use of `createValidChildName()` means we'll have different behavior for some elements than others.  I think we could (and should) look at rolling that in to `Element::setName()` itself - but maybe as a separately considered piece of work.

Here I think it would be safe to do something along the lines of 

```
{
  Element::set(name);
  if (!updateReferences)
    return;
  string newName = getName();
  <reference updating logic here>
}
```

(The same comment applies to the NodeGraph case too)

source/MaterialXCore/Node.cpp
```diff
@@ -20,6 +20,26 @@ const string Backdrop::HEIGHT_ATTRIBUTE = "height";
...
+    {
+        Element::setName(name);
+        return;
+    }
+    if (name.empty() || getName() == name)
+    {
+        return;
+    }
+    std::string validName = getParent()->createValidChildName(name);
```
kwokcb, line 34: The name logic is a bit inconsistent here. 
- If not updating references it will go to the base class and throw an exception on duplicate names.
- Here it will:
  - Return if there is a duplicate so will never reach trying to create a unique name. 
  - Also the description says it will throw an exception on duplicates unconditionally.

If it's always creating a unique name we can remove lines 30-32 and update the description. 
Otherwise we should mimic the exception logic of the base class.

Similar applies to the NodeGraph method.

Thanks.

ld-kerley (commented):
Thanks @rasmustilljander - this looks like it's gonna make editing documents a lot more straight forwards!

source/MaterialXCore/Node.cpp
rasmustilljander: Thanks for the excellent feedback! I made some changes, and the call to `getDownstreamPorts` needs to be done before the call to `Element::setName(name);` but otherwise I do believe that my changes aligns well with your suggestion.

source/MaterialXCore/Node.cpp
rasmustilljander: Thanks for the excellent feedback, I tried to tackle that and I think I got it more consistent.

source/MaterialXCore/Node.h
```diff
@@ -58,6 +58,16 @@ class MX_CORE_API Node : public InterfaceElement
...
     virtual ~Node() { }
 
+    /// @name Name
+    /// @{
+
+    /// Set the element's name string.  The name of a MaterialX element must be
+    /// unique among all elements at the same scope. Optionally updates downstream ports.
+    /// @throws Exception if an element at the same scope already possesses the
+    ///    given name.
+    void setName(const std::string& name, bool updateReferences = false);
```
jstone-lucasfilm, line 68: One concern that I have with this approach is that a non-virtual override of the base `Element::setName` method would lead to ambiguous code when the `updateReferences` argument is omitted.

As an alternative, what would you think about assigning this new method a unique name such as `setNameGlobal`?  This new method would no longer require an `updateReferences` argument at all, and it would only be invoked when the global rename behavior was explicitly intended by the caller.

jstone-lucasfilm (commented):
This looks very promising, @rasmustilljander, and I've written up a suggested improvement.

source/MaterialXCore/Node.h
ld-kerley: I almost think that we would want this new behavior to be the default - and thus the `setName()` functionality.  If we need a separate function, then maybe we call it `setNameLocalOnly()`?  

I'm sort of curious if we have any existing use-cases where it would be important to rename a node without following the references and also renaming those.  I can see we might rename a node in isolation, but I think that once you've connected it, then wouldn't you always the connected reference to be similarly updated?

kwokcb: This is where we get into file-I/O vs run-time APIs :). 

Historically the main usage case is I/O and thus these methods have on-purpose been kept as simple data setters. I don't think we want to adversely affect performance or cause any side-effects here. (*) 

Thus I'd concur that we use a new name for the "run-time" / "connectivity" interface. Maybe `rename()`  or `setNameGlobal()` as Jonathan suggests ?.

---
(@ld-kerley -- I think as part of API surface "exposure" these should be hidden and accessible only for file-I/O over time, but a different topic).

ld-kerley: Thanks @kwokcb that makes a lot of sense - helpful to understand the backstory here for sure. 

I quite like the idea of `rename()` actually.

jstone-lucasfilm: If we choose a new name for the global version of `setName`, I'd like to establish a pattern that can apply to other global variants of methods in the future, in the same spirit as the distinction between `getFilePrefix` and `getActiveFilePrefix`:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Element.h#L186

Along those lines, I'd probably advocate for adding some new word to the global version of the method, where that same word can be added to other global methods in the future.

It's just an initial suggestion that this new word be `Global`, and I'd love to hear other suggestions as well.

jstone-lucasfilm: In the spirit of giving @rasmustilljander a clear path forward, I'll recommend the name `setNameGlobal` for the new global version of the `setName` method, but feel free to make a case for other options in this thread.

rasmustilljander: Thanks for the great discussion, really interesting!

In my latest commit I’ve renamed it to `setNameGlobal` while still using the original `setName` under the hood, which I think follows our last suggestion. Do you have any feedback on the docstrings?

source/MaterialXCore/Node.h
```diff
@@ -58,6 +58,18 @@ class MX_CORE_API Node : public InterfaceElement
     }
     virtual ~Node() { }
 
+    /// @name Name
+    /// @{
+
+    /// Set the element's name string. The name of a MaterialX element must be
```
jstone-lucasfilm, line 64: For these new docstrings, my recommendation would be to be specific about the element subclass (e.g. Node, NodeGraph), and to omit the repeated language of the general rules for MaterialX element names.

Here's a proposed update to the doc string for the Node method, and the same approach should work for NodeGraph as well:

```
/// Set the name string of this Node, propagating the updated name to all
/// downstream ports.
/// @throws Exception if an element at the same scope already possesses the
///     given name.
```

jstone-lucasfilm (commented):
Great work, @rasmustilljander, and I've written up a proposal for the language of the doc strings.

source/MaterialXCore/Node.h
rasmustilljander: Looks great, I updated the docs to align with your suggestion!

PR #2387, opened 2025-05-15
Validate that node output matches node definitions

rasmustilljander:
Enforce that a node’s type aligns with its NodeDef outputs:
- **Single-output** NodeDefs require the node’s type to match that output’s type.  
- **Multi-output** NodeDefs require the node’s type to be `"multioutput"`.

Changes
- In `Node::validate()`, inspect `nodeDef->getOutputs()`:  
  - **1 output** → `node.getType() == output.getType()`  
  - **>1 output** → `node.getType() == "multioutput"`  

No existing API has been altered; this strictly adds validation logic.

### DIFF
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -200,14 +200,15 @@ bool Node::validate(string* message) const
         bool exactMatch = hasExactInputMatch(nodeDef, &matchMessage);
         validateRequire(exactMatch, res, message, "Node interface error: " + matchMessage);
 
-        vector<OutputPtr> nodeDefOutputs = nodeDef->getOutputs();
-        if(nodeDefOutputs.size() == 1)
+        const vector<OutputPtr>& activeOutputs = nodeDef->getActiveOutputs();
+        const size_t numActiveOutputs = activeOutputs.size();
+        if (numActiveOutputs > 1)
         {
-            validateRequire(getType() == nodeDefOutputs[0]->getType(), res, message, "The attribute type must match the type of the output of the node definition");
+            validateRequire(getType() == MULTI_OUTPUT_TYPE_STRING, res, message, "Node type is not 'multioutput' for node with multiple outputs");
         }
-        else if(nodeDefOutputs.size() > 1)
+        else if (numActiveOutputs == 1)
         {
-            validateRequire(getType() == "multioutput", res, message, "The attribute type must be multioutput to match the implementation");
+            validateRequire(getType() == activeOutputs[0]->getType(), res, message, "Node type does not match output port type");
         }
     }
     else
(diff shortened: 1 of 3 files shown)


ld-kerley (commented):
This looks like a great addition @rasmustilljander.  I left some small style notes below regarding the error message reported.

Additional bonus points awarded for adding the unit test as well! thank you :)

source/MaterialXCore/Node.cpp
```diff
@@ -199,6 +199,16 @@ bool Node::validate(string* message) const
         string matchMessage;
         bool exactMatch = hasExactInputMatch(nodeDef, &matchMessage);
         validateRequire(exactMatch, res, message, "Node interface error: " + matchMessage);
+
+        vector<OutputPtr> nodeDefOutputs = nodeDef->getOutputs();
```
kwokcb, line 203: Small notes: 

- `getActiveOutputs()` should be used instead of `getOutputs()` as the former handles definition inheritance. 
- Please use MULTI_OUTPUT_TYPE_STRING constant instead of "multioutput".

Note: this is very similar to `NodeDef::getType()` where you can see usage, though that function should probably throw an exception if there are no outputs and not a string.

source/MaterialXCore/Node.cpp
```diff
@@ -199,6 +199,16 @@ bool Node::validate(string* message) const
...
+        if(nodeDefOutputs.size() == 1)
+        {
+            validateRequire(getType() == nodeDefOutputs[0]->getType(), res, message, "The attribute type must match the type of the output of the node definition");
+        }
+        else if(nodeDefOutputs.size() > 1)
+        {
+            validateRequire(getType() == "multioutput", res, message, "The attribute type must be multioutput to match the implementation");
+        }
```
kwokcb, line 211: More a pathological case but still possible to create is a `nodedef` with 0 outputs. Might want to check that as well. Not sure what it means for type of a node vs type of nodedef.

ld-kerley: I wonder if zero outputs from a nodedef element should itself be a validation check we add on the nodedef - is there a reasonable use-case for a nodedef to have zero outputs?

rasmustilljander: I agree that it may be something to check for, but I see it an validation notice for the NodeDef and not an error for the node. But I made an attempt and added a validation check in NodeDef::validate but it caused issues elsewhere- I ended up removing it in my last commit again. See image for the testcases that failed due to it.

image

source/MaterialXCore/Node.cpp
rasmustilljander: Great feedback! I'm now using `MULTI_OUTPUT_TYPE_STRING` and I've tried to align the style with could be found in `NodeDef::getType`.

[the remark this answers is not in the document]
rasmustilljander: Great feedback! I've updated my messages to align with the style!

jstone-lucasfilm (approved):
Great work, thanks @rasmustilljander!

Issue #1952, opened 2024-07-26
GLSL ShaderGen: `time_float` and `frame_float` do not have any effect


hybridherbst:
Related to #1951 

I believe the attached file should have some animation based on application time, however it just stays still in MaterialXWebViewer.
[marbleTime.mtlx.zip](https://github.com/user-attachments/files/16389605/marbleTime.mtlx.zip)

<img width="973" alt="image" src="https://github.com/user-attachments/assets/271e3298-b3f4-4eaa-b327-e3f0c1dc078b">

hybridherbst:
Edit: in some applications the time nodes show up as not connected, this one here should work:
[myFile2.mtlx.zip](https://github.com/user-attachments/files/16393512/myFile2.mtlx.zip)

kwokcb:
There is no "play" functionality here nor in the desktop viewer. For the hardware case the `u_frame` shader uniform needs to be updated and bound as mentioned [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/ShaderGeneration.md)

hybridherbst:
Please consider it a feature request :)

jstone-lucasfilm:
Thanks for this feature request, @hybridherbst, and I've just updated the title to clarify that this feature is missing in our GLSL shader generation, which both the Viewer and Graph Editor leverage for rendering

jstone-lucasfilm:
Good point, @hybridherbst, and indeed those updates should be made to the Viewer and Graph Editor when this feature is added.

fnRaihanKibria:
Hello, I'd like to work on this for ASWF Dev Days 2025.

jstone-lucasfilm:
Excellent, thanks @fnRaihanKibria!

jstone-lucasfilm:
Thanks for posting this original issue, @hybridherbst, and thanks to @fnRaihanKibria for the fix in #2389!

If there's still an outstanding compatibility issue with this feature on MacOS, feel free to write up a new GitHub Issue for this, so that we don't lose track.

fnRaihanKibria:
> Thanks for all the work here! However, the file I originally attached to the issue still is not animated – I think this issue here should be reopened.

Are you on a Mac? There is an issue on Macs unfortunately: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2391

hybridherbst:
[...]

PR #2389, opened 2025-05-15
Bind frame uniform in graph editor and viewer

fnRaihanKibria:
Addressing issue #1952

Bind the frame uniform in the render calls of the graph editor and viewer.

### DIFF
--- a/source/MaterialXGraphEditor/Main.cpp
+++ b/source/MaterialXGraphEditor/Main.cpp
@@ -15,7 +15,6 @@
 
 #include <iostream>
 #include <limits>
-#include <type_traits>
 
 namespace
 {
@@ -256,9 +255,8 @@ int main(int argc, char* const argv[])
         ImGui::NewFrame();
 
         auto renderer = graph->getRenderer();
-        using FrameType = std::invoke_result_t<decltype(&RenderView::getFrame), RenderView>;
-        constexpr auto frameMaxValue = std::numeric_limits<FrameType>::max();
-        renderer->setFrame((renderer->getFrame() + 1) % frameMaxValue);
+        constexpr auto FRAME_MAX_VALUE = std::numeric_limits<unsigned int>::max();
+        renderer->setFrame((renderer->getFrame() + 1) % FRAME_MAX_VALUE);
         renderer->drawContents();
         if (!captureFilename.empty())
         {
--- a/source/MaterialXGraphEditor/Main.cpp
+++ b/source/MaterialXGraphEditor/Main.cpp
@@ -14,6 +14,8 @@
 #include <GLFW/glfw3.h>
 
 #include <iostream>
+#include <limits>
+#include <type_traits>
 
 namespace
 {
@@ -254,7 +256,9 @@ int main(int argc, char* const argv[])
         ImGui::NewFrame();
 
         auto renderer = graph->getRenderer();
-        renderer->setFrame((renderer->getFrame() + 1) % 32768);
+        using FrameType = std::invoke_result_t<decltype(&RenderView::getFrame), RenderView>;
+        constexpr auto frameMaxValue = std::numeric_limits<FrameType>::max();
+        renderer->setFrame((renderer->getFrame() + 1) % frameMaxValue);
         renderer->drawContents();
         if (!captureFilename.empty())
         {
--- a/source/MaterialXGraphEditor/Main.cpp
+++ b/source/MaterialXGraphEditor/Main.cpp
@@ -253,7 +253,9 @@ int main(int argc, char* const argv[])
         ImGui_ImplGlfw_NewFrame();
         ImGui::NewFrame();
 
-        graph->getRenderer()->drawContents();
+        auto renderer = graph->getRenderer();
+        renderer->setFrame((renderer->getFrame() + 1) % 32768);
+        renderer->drawContents();
         if (!captureFilename.empty())
         {
             break;
--- a/source/MaterialXGraphEditor/RenderView.cpp
+++ b/source/MaterialXGraphEditor/RenderView.cpp
@@ -144,7 +144,8 @@ RenderView::RenderView(mx::DocumentPtr doc,
     _renderTransparency(true),
     _renderDoubleSided(true),
     _captureRequested(false),
-    _exitRequested(false)
+    _exitRequested(false),
+    _frame(0)
 {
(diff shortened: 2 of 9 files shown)


fnRaihanKibria:
Note:
-  As far as I understand, the time uniform is computed from the frame in GLSL, hence it does not need to be bound.
- I do not have a Mac available right now and wrote the Metal pipeline code ad hoc without testing.

ld-kerley (approved):
This looks like a great change to me - I can confirm all appears to work in the GLSL render path, but the Metal render path doesn't appear to work yet, but I think that might be due to the some differences in how the uniforms are handled in Metal.  So likely unrelated to this change. 

I think we should move forward with this patch as is and file further issues to track down where things are not working in Metal. 

I think we might also want to add some form of playback controls to the viewer, but I'll file that as a separate issue as well.

source/MaterialXGraphEditor/RenderView.cpp
```diff
@@ -806,6 +807,10 @@ void RenderView::renderFrame()
         {
             material->getProgram()->bindUniform(mx::HW::ALPHA_THRESHOLD, mx::Value::createValue(0.99f));
         }
+        if (material->getProgram()->hasUniform(mx::HW::FRAME))
+        {
+            material->getProgram()->bindUniform(mx::HW::FRAME, mx::Value::createValue(static_cast<float>(_frame++)));
```
kwokcb, line 812: This will update the frame count for every material drawn instead of once per refresh. Same in the other code updating frame rate. Suggest to modify / update frame outside of the refresh logic. This will allow for a proper frame setting (vs refresh count being used as  frame count)
even if not handled with this change.

kwokcb (commented):
This feels a bit strange way to set "frame" as this is counting the number of refreshes that occur but as noted this can be added in later. I would remove frame setting from inside the refresh as noted though. If / when you have proper frame setting then it can be  set outside without changing the binding code again.

As the number increases continuously while the app is running, Unsure what this means when hitting integer limits on in some shading languages. e.g. medium precision OpenGLES has a max of 32767 (*). Might be okay for desktop / high precision.

(*) This does not handle the Web Viewer (which uses ES)  as noted in the issue.

source/MaterialXGraphEditor/RenderView.cpp
fnRaihanKibria: I've moved the increment from during the binding to where the draw call for the frame is made, and added a limit to 32767.

source/MaterialXGraphEditor/Main.cpp
```diff
@@ -253,7 +253,9 @@ int main(int argc, char* const argv[])
         ImGui_ImplGlfw_NewFrame();
         ImGui::NewFrame();
 
-        graph->getRenderer()->drawContents();
+        auto renderer = graph->getRenderer();
+        renderer->setFrame((renderer->getFrame() + 1) % 32768);
```
jstone-lucasfilm, line 257: Instead of this hardcoded maximum, perhaps we should use `std::numeric_limits<unsigned int>::max() - 1`?  That would help to clarify the motivation for the maximum, as well as providing more headroom than the existing `32768`.

source/MaterialXView/Viewer.cpp
```diff
@@ -2204,6 +2204,7 @@ void Viewer::draw_contents()
     }
 
     // Render the current frame.
+    _renderPipeline->_frame = (_renderPipeline->_frame + 1) % 32768;
```
jstone-lucasfilm, line 2207: Same as above, where `std::numeric_limits<unsigned int>::max() - 1` may be a better match than `32768`.

jstone-lucasfilm (commented):
This looks very promising, @fnRaihanKibria, and I had one suggestion for an improvement.

source/MaterialXGraphEditor/Main.cpp
fnRaihanKibria: Using modulo with the counter type's maximum.

source/MaterialXView/Viewer.cpp
fnRaihanKibria: Using modulo with the counter type's maximum.

source/MaterialXGraphEditor/Main.cpp
```diff
@@ -253,7 +255,11 @@ int main(int argc, char* const argv[])
         ImGui_ImplGlfw_NewFrame();
         ImGui::NewFrame();
 
-        graph->getRenderer()->drawContents();
+        auto renderer = graph->getRenderer();
+        using FrameType = std::invoke_result_t<decltype(&RenderView::getFrame), RenderView>;
+        constexpr auto frameMaxValue = std::numeric_limits<FrameType>::max();
```
jstone-lucasfilm, line 260: This is a really interesting approach, and I especially like the one-line version of this logic that you use in Viewer.cpp below:

```
constexpr auto frameMaxValue = std::numeric_limits<decltype(_renderPipeline->_frame)>::max();
```

Is there a way to maintain the same, simpler approach in Main.cpp as well, rather than bringing in the more complex machinery of `std::invoke_result_t` from `type_traits`?

fnRaihanKibria: In Viewer.cpp the _frame variable is part of the class using it, whereas here we don't have direct access to it as it is a private variable of another class. I didn't want to make the variable public just for this. Unless there is a solution I am missing?
I don't think the longer code here is a performance issue at least, as it should all be resolved at compile time. If it's too confusing I suppose we could simply use the type explicitly?

jstone-lucasfilm: Yes, it's more the additional complexity for future readers that I'm concerned with, as `std::invoke_result_t` is such an unusual technique, and `type_traits` is not otherwise used in this source file.

I like your suggestion to use the type explicitly, which would give the following one-line implementation:

```
constexpr auto FRAME_MAX_VALUE = std::numeric_limits<unsigned int>::max();
```

Note that I've also updated the name of the constant from `frameMaxValue` to `FRAME_MAX_VALUE`, in order to better match the coding conventions in MaterialX.

Does that approach seem reasonable to you as well?

fnRaihanKibria: Sure, made the change.

jstone-lucasfilm (approved):
Great work, thanks @fnRaihanKibria!

PR #2394, opened 2025-05-15
Handle UI display color space in MaterialX Viewer

mialana:
_ASWF Dev Days May 2025
**Original Issue:** #1209_ 

This PR fixes discrepancies between the rendered preview material in MaterialXView and the NanoGUI color picker. In both the `Color3` and `Color4` cases, it completes the following conversions:

1. Uses `Color3::linearToSrgb()` to convert the stored value in "lin_rec709" color space to sRGB space for use in the color picker.
2. After user selects a color in the GUI and presses the "Pick" button, converts the selected RGB values to linear space using the `Color3::srgbToLinear()` conversion function. The converted linear color value is what is ultimately written to the material.

In the `Color4` case, the alpha channel is left untouched, both during reading and writing.

Note: Similar changes may have to be implemented for MaterialXGraphEditor. However, while attempting to implement those changes, I observed that ImGUI performs automatic gamma correction on its color picker widgets (color space in ImGUI seems to be being discussed in-depth in this [issue](https://github.com/ocornut/imgui/issues/578)). All-in-all, after discussion with @ld-kerley, it seems to be unclear what desired behavior in the ImGUI / MaterialXGraphEditor case is, so it is not handled yet in this PR.

### DIFF
--- a/source/MaterialXView/Editor.cpp
+++ b/source/MaterialXView/Editor.cpp
@@ -11,8 +11,6 @@
 #include <nanogui/slider.h>
 #include <nanogui/vscrollpanel.h>
 
-#include <iostream>
-
 namespace
 {
 
@@ -334,10 +332,6 @@ void PropertyEditor::addItemToForm(const mx::UIPropertyItem& item, const std::st
             mx::Color3 displayCol = v.linearToSrgb();
 
             ng::Color c(displayCol[0], displayCol[1], displayCol[2], 1.0);
-            ng::Color linCol(v[0], v[1], v[2], 1.0);
-
-            std::cout << "[" << label << " Color3 READ] linear: " << linCol << std::endl;
-            std::cout << "display: " << c << std::endl;
 
             new ng::Label(twoColumns, label);
             auto colorVar = new EditorColorPicker(twoColumns, c);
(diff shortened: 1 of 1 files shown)


mialana:
I will also include some "before-and-after" images to illustrate the effects of my changes.

_**Case 1:** Upon load, color picker does not correctly convert stored linear color values to RGB, thus creating a discrepancy between the color picker preview and the rendered material preview. In this case, **the rendered material preview is correct**, and _the color picker is incorrect.__

**BEFORE:**
viewer-initial_load_incorrect

**AFTER:**
viewer-initial_load_corrected

_**Case 2:** When user picks a desired material color using the NanoGUI color picker, the raw RGB values are not converted to linear color space before being written to the material, thus creating a discrepancy between the color picker preview and the rendered material preview. In this case, **the color picker is correct**, and **the rendered material preview is incorrect**._

**BEFORE:**
viewer-midtone_red_incorrect

**AFTER:**
viewer-midtone_red_corrected


Please let me know if there are any flaws in my logic. Thank you!

jstone-lucasfilm (approved):
Great work, @mialana, and this looks good to me!

Issue #2220, opened 2025-02-07
Graph Editor: Crash when multiple node outputs are present


zicher3d:
When a nodedef+nodegraph defines multiple outputs, connecting some might result in crashes. I have a simple case that crashes on a float output.
I didn't investigate other output types in details. So far it seems to happen on float only.

Steps:

1. Unzip the attached zip file. There are 2 mtlx nodegraphs in it.
2. Copy the files in the libraries folder of the MaterialX install
3. Start the Graph Editor (I tried 1.38.10 and 1.39.2)
4. (**See video for the rest of the steps or**...)
5. Delete the NG_marble1 default nodegraph
6. Create a **test_output** (singular) node
7. Connect the float output to the standard surface SR_marble1 **Base** input
8. Verify that editing the value works and all is well
9. Delete the **test_output** node
10. Create a **test_outputs** (plural) node
11. Connect the color output to the standard surface **Base_color** input
12. Verify all is good and color is working
13. Connect the float output to **Base**
14. ISSUE1: Preview object will disappear
15. ISSUE2: Editing the float value will crash Graph Editor

https://github.com/user-attachments/assets/bac7dca4-3982-465f-9252-d23934f6e762

[test_nodes.zip](https://github.com/user-attachments/files/18713531/test_nodes.zip)

pedrosanchez-fn:
I'd be interested on picking this up for ASWF Dev days

jstone-lucasfilm:
Excellent, thanks @pedrosanchez-fn!

pedrosanchez-fn:
Hello, I've been working on these today. Unfortunately, I have not been able to get a solution done yet, but I'm hoping to spend a bit more time tomorrow if that's okay?
I believe I've managed to find the source of the problem, this being that the `test_outputs` outputs are not been added to the `Node` instance on creation. As a consequence, during connection, `Node->getNodeDefOutput` returns an empty `OutputPtr`, and the `ShaderNode` defaults to the first output:

``` C++
// ShaderGraph.cpp:127
OutputPtr nodeDefOutput = connectingElement ? upstreamNode->getNodeDefOutput(connectingElement) : nullptr; // nodeDefOutput is empty
ShaderOutput* output = nodeDefOutput ? newNode->getOutput(nodeDefOutput->getName()) : newNode->getOutput(); // newNode->getOutput() is run, returning the output at index 0
```

[...]

PR #2397, opened 2025-05-16
Fix Graph Editor crash with multioutput nodes

pedrosanchez-fn:
When connecting multiple output nodes to surface shaders, `Graph::addLink` was not specifying the output being connected. As a consequence, the shader would end up using the first output on the upstream node which could cause shader compilation errors and exceptions if the input and output types did not match.

This change ensures that the output is specified on the input pin.


jstone-lucasfilm:
Thanks @pedrosanchez-fn, and I'll link this to GitHub Issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/2220.

jstone-lucasfilm (approved):
This looks good to me, thanks @pedrosanchez-fn!

PR #2398, opened 2025-05-17
Patch GLFW to improve macOS monitor name detection

### LINKED ISSUE #2384: Graph Editor (Mac) : glfw port error
## Issue

This error appears to only show up on Mac and at least seems to occur with M1 but not sure if
it shows up on other hardware.

Glfw Error 65544: Cocoa: Failed to find service port for display

This does not seem to cause any side-effects so logging as lower priority issue.

Notes:
- This error seems to have been there for a very long time.
- Unsure if this started to occur when switch to Metal was done.
- This could be an outdated glfw library issue (used by NanoGUI).

This does no occur with MaterialxView (which does not use NanoGUI).

warrenm:
This patch fixes #2384 by porting some of the Cocoa monitor name handling changes from GLFW release 3.4 to the copy of GLFW in the Graph Editor source tree. Specifically, it uses `NSScreen` APIs where possible, instead of older IOKit APIs that do not work with AirPlay connected displays or on Apple Silicon.

I chose to only bring over the essential changes from GLFW 3.4, since a full update to 3.4 is much more extensive, harder to review, and probably not merited for such small benefit. Nonetheless, if and when a full update to 3.4 or later is necessary, this patch should merge pretty cleanly.

Tested on a MacBook Pro M2 Max running macOS Sequoia 15.4.1 and a MacBook Pro (2019) running macOS Monterey 12.7.5. Gracefully handles dynamic connection/disconnection of external displays and AirPlay connected displays, and removes the spurious log message noted in #2384.

### DIFF
--- a/source/MaterialXGraphEditor/External/Glfw/src/cocoa_monitor.m
+++ b/source/MaterialXGraphEditor/External/Glfw/src/cocoa_monitor.m
@@ -39,18 +39,31 @@
 
 // Get the name of the specified display, or NULL
 //
-static char* getDisplayName(CGDirectDisplayID displayID)
+static char* getMonitorName(CGDirectDisplayID displayID, NSScreen* screen)
 {
+    // IOKit doesn't work on Apple Silicon anymore
+    // Luckily, 10.15 introduced -[NSScreen localizedName].
+    // Use it if available, and fall back to IOKit otherwise.
+    if (screen)
+    {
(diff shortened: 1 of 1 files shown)


meshula (commented):
Thanks Warren! This looks good to me.

jstone-lucasfilm (approved):
Thanks for the fix, @warrenm!

PR #2395, opened 2025-05-16
Graph Editor - UI support for enum

inter-sakot4:
**Description**
In this PR, UI support for enumerator types are added to GraphEditor.

Some assumptions are made to execute this task:
- Fix the issue for only GraphEditor , though similar issue exist also in MaterialXView ( NanoGUI )
- Generalize the solution to int, bool, string, float, double, Vector2, Vector3, Vector4, Color3, Color4, Matrix types

With the above assumptions, scope of this task is limited only to GraphEditor ( IMGUI )

**Solution**
- `renderEnumInput<input_type>(input, uiProperties, temp)` is a template function that does the rendering of enumerator input types.

Following edge cases are handled:
- ui.enumerationValues and ui.ui.enumeration are not always populated ( See /MaterialXRender/Util.cpp ). In such cases, the enum and enumvalues are parsed directly from the attribute values to render the enumerators.
- Its common to have string input with enum fields  only , but no enumvalues

**How to test?**

Slightly modified `standard_surface_jade.mtlx` can be used like below.

```
<?xml version="1.0"?>
<materialx version="1.39" colorspace="lin_rec709">
  <standard_surface name="SR_jade" type="surfaceshader">
    <input name="base" type="float" value="0.5" />
    <input name="base_color" type="color3" value="0.0603,0.4398,0.1916" />
    <input name="specular_roughness" type="float" value="0.25" enum="0.0, 0.25, 0.5, 0.75" enumvalues="0.0,0.25,0.5,0.75"/>
    <input name="specular_IOR" type="float" value="2.418" />
    <input name="specular_anisotropy" type="float" value="0.5" />
    <input name="subsurface" type="float" value="0.4" enum="0.0,0.4,0.5,0.6,0.7" enumvalues="0.0,0.4,0.5,0.6,0.7" />
    <input name="subsurface_color" type="color3" value="0.0603, 0.4398, 0.1916" enum="DEFAULT_GREEN, RED" enumvalues="0.0603,0.4398,0.1916,0.8,0.04398,0.01916" />
  </standard_surface>
  <surfacematerial name="Jade" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_jade" />
  </surfacematerial>
</materialx>

```
**Expected Result**

image

PR #2402, opened 2025-05-21
Fix color space issues with sRGB framebuffer in Graph Editor

mialana:
_Related [issue](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1209) and [Dev Days PR for Viewer](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2394)_

This PR addresses color space discrepancies / logic in MaterialX Graph Editor.

---
#### Summary of Problem: 
Before, the graph editor used an sRGB framebuffer (i.e. `glEnable(GL_FRAMEBUFFER_SRGB)`) for the entire GUI, meaning both the rendered material (with its color values given in linear space) _and_ the ImGUI components had an automatic "to-sRGB" transformation applied to them.

Although this is desired for the material preview, which is a rendered, PBR result and needs to be color corrected to display space, it has adverse results on ImGUI components that handle color. This is because currently, ImGUI is not completely compatible with sRGB-enabled framebuffers (ref: [578](https://github.com/ocornut/imgui/issues/578), [1724](https://github.com/ocornut/imgui/issues/1724), [2943](https://github.com/ocornut/imgui/pull/2943) [4890](https://github.com/ocornut/imgui/issues/4890), [6583](https://github.com/ocornut/imgui/issues/6583)). To exemplify a problem that arises, alpha-blending is calculated incorrectly when using an sRGB framebuffer.

Compare a "correct" color picker rendered w/o sRGB-enabled framebuffer with a color picker rendered w/ sRGB-enabled framebuffer, which is how it was done before this PR. I gave it the same color value, but in linear space:

Screenshot 2025-05-21 at 16 55 20
Screenshot 2025-05-21 at 16 57 52

There is clearly a non-negligible difference in both the color ramp and the overall canvas (which is the result of a completely different blending equation being applied, as commented on [here](https://github.com/ocornut/imgui/issues/578#issuecomment-1585432977)). This calculation error results in the color picker UI component being in a nonlinear display space and should be avoided.

---
#### Resolution Implemented:

1. To avoid these calculation errors in ImGUI, the sRGB framebuffer should be disabled for the UI. However, the rendered material view needs to still have a sRGB transformation applied to it. To solve this, I used the `ImGui::GetWindowDrawList()->AddCallback()` function to to inject a call to`glEnable(GL_FRAMEBUFFER_SRGB)` into ImGUI's iterative draw list right before the materialview texture is drawn, and disable it immediately after.

This effectively maintains the original material render result, but solves the color problems in the UI.

2. Then, regarding the original [issue](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1209), `Color3::linearToSrgb()` and `Color3::srgbToLinear()` are used to transform between color spaces. The result is below. 

_**Before:** User previews rendered material and color picker swatch in display space, but the actual color value numbers are actually in the linear range. These values will not match up in other applications, such as Adobe products, which is misleading._
Screenshot 2025-05-21 at 13 29 01

_**After:** User always interacts with color in display space (pick + view), but behind-the-scenes, the color is written / stored in linear space._
Screenshot 2025-05-21 at 16 26 17

_Rendered material result is successfully maintained:_

Screenshot 2025-05-21 at 17 19 02
Screenshot 2025-05-21 at 17 18 24

3. Lastly, a simple tooltip is added to the color picker component, clarifying the difference between front-facing UI values and stored values.
Screenshot 2025-05-21 at 17 44 40

---
Please let me know if I can clarify on any of the points I mentioned above. I tried to balance digital color principles / correctness with overall UI experience in these changes. Thanks!

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -104,10 +104,12 @@ std::string getUserNodeDefName(const std::string& val)
     return result;
 }
 
-static void EnableSRGBCallback(const ImDrawList*, const ImDrawCmd*)   {
+static void EnableSRGBCallback(const ImDrawList*, const ImDrawCmd*)
+{
     glEnable(GL_FRAMEBUFFER_SRGB);
 }
-static void DisableSRGBCallback(const ImDrawList*, const ImDrawCmd*)  {
+static void DisableSRGBCallback(const ImDrawList*, const ImDrawCmd*)
+{
     glDisable(GL_FRAMEBUFFER_SRGB);
 }
 
@@ -979,7 +981,7 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
         {
             mx::Color3 prev, temp;
 
-            // read material value in converted display space
+            // Read material value in converted display space
             prev = temp = val->asA<mx::Color3>().linearToSrgb();
 
             // Use ImGuiColorEditFlags_Uint8 flag for built-in Uint8 input fields
@@ -992,7 +994,7 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
             // Set input value and update materials if different from previous value
             if (prev != temp)
             {
-                // convert back to linear color space for writing to material and node input
+                // Convert back to linear color space for writing to material and node input
                 mx::Color3 linearCol = temp.srgbToLinear();
 
                 addNodeInput(_currUiNode, input);
@@ -1008,12 +1010,12 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
         mx::ValuePtr val = input->getValue();
         if (val && val->isA<mx::Color4>())
         {
-            // read material value and convert RGB components to display space
+            // Read material value and convert RGB components to display space
             mx::Color4 linearCol = val->asA<mx::Color4>();
             mx::Color3 displayCol3 = mx::Color3(linearCol[0], linearCol[1], linearCol[2]).linearToSrgb();
 
             mx::Color4 prev, temp;
-            // create 4D vector with converted RGB and non-converted, stored Alpha value
+            // Create 4D vector with converted RGB and non-converted, stored Alpha value
             prev = temp = mx::Color4(displayCol3[0], displayCol3[1], displayCol3[2], linearCol[3]);
 
             // Use ImGuiColorEditFlags_Uint8 flag for built-in Uint8 input fields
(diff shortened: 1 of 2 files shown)


lfl-eholthouser (approved):
Thanks for fixing this issue @mialana. I think the changes look great!

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -971,21 +978,27 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
...
-            float min = minVal ? minVal->asA<mx::Color3>()[0] : 0.f;
-            float max = maxVal ? maxVal->asA<mx::Color3>()[0] : 100.f;
-            float speed = (max - min) / 1000.0f;
-            ImGui::PushItemWidth(-100);
-            ImGui::DragFloat3("##hidelabel", &temp[0], speed, min, max);
-            ImGui::PopItemWidth();
-            ImGui::SameLine();
-            ImGui::ColorEdit3("##color", &temp[0], ImGuiColorEditFlags_NoInputs);
+
+            // read material value in converted display space
```
lfl-eholthouser, line 982: I want to flag that this pr #2395 is also modifying the color3 and color4 section of setConstant()

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -104,6 +104,13 @@ std::string getUserNodeDefName(const std::string& val)
     return result;
 }
 
+static void EnableSRGBCallback(const ImDrawList*, const ImDrawCmd*)   {
```
jstone-lucasfilm, line 107: This is a minor note, but we use Allman-style braces in MaterialX, where the opening and closing brackets are placed on their own line, as in:

```
static void EnableSRGBCallback(const ImDrawList*, const ImDrawCmd*)
{
    glEnable(GL_FRAMEBUFFER_SRGB);
}
```

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -971,21 +978,27 @@ void Graph::setConstant(UiNodePtr node, mx::InputPtr& input, const mx::UIPropert
...
+            ImGui::ColorEdit3("##color", &temp[0], ImGuiColorEditFlags_Uint8);
+            if (ImGui::IsItemHovered(ImGuiHoveredFlags_AllowWhenDisabled))
+            {
+                ImGui::SetTooltip("Color is selected and rendered to Viewer in sRGB display space, \nbut written to .mtlx file in linear format.");
+            }
 
             // Set input value and update materials if different from previous value
             if (prev != temp)
             {
+                // convert back to linear color space for writing to material and node input
```
jstone-lucasfilm, line 995: Another minor note, our convention for code comments is to capitalize the first word of the comment text.

jstone-lucasfilm (commented):
This looks great to me, @mialana, and I've added two minor suggestions to match the coding style of the MaterialX project.

mialana:
I have just made the style fixes. Thank you @lfl-eholthouser and @jstone-lucasfilm!

jstone-lucasfilm (approved):

PR #2404, opened 2025-05-21
Fix overwriting of alpha values in Viewer colorpicker

mialana:
Currently, if an user uses the MaterialXView colorpicker for a `Color4` field, the alpha component is automatically overwritten and transformed to a value of 1, regardless of original value. Although it is shown unmodified in the UI, if written out to file, the alpha component is always "1". e.g.
```
<input name="bias" type="color4" value="0.0908109, 0.0908109, 0.0908109, 1" />
```
This is because the Nanogui `ColorWheel` class does not handle color values with 4 components, yet the Nanogui `ColorPicker` class refers to its color wheel member to set the new picked color ([code ref](https://github.com/mitsuba-renderer/nanogui/blob/6452dd6944d2ba5c0c9bc0042a1894f703ce1ace/src/colorpicker.cpp#L58-L65)).

To prevent this from happening, I modified the current `class EditorColorPicker : public ng::ColorPicker`, and created custom callbacks for `m_color_wheel` and `m_pick_button`. The logic essentially is -- since we have additional color widget members in our color picker, we instead refer to their values when setting the color value. Then, the alpha component can only be changed via the color widgets, so interacting with the color wheel does not change / overwrite the original alpha component.

The following GIF is a simple visual demonstration of the alpha being correctly preserved and previewed in the `PickButton`. I've also tested saving out to file to validate the written value.

### DIFF
--- a/source/MaterialXView/Editor.cpp
+++ b/source/MaterialXView/Editor.cpp
@@ -374,7 +374,7 @@ void PropertyEditor::addItemToForm(const mx::UIPropertyItem& item, const std::st
             {
                 // Transform sRGBA color picker value to linear space for writing to material
                 mx::Color3 linearCol = mx::Color3(c.r(), c.g(), c.b()).srgbToLinear();
-                mx::Vector3 v(linearCol[0], linearCol[1], linearCol[2]);
+                mx::Vector4 v(linearCol[0], linearCol[1], linearCol[2], c.a());
 
                 material->modifyUniform(path, mx::Value::createValue(v));
             }
--- a/source/MaterialXView/Editor.cpp
+++ b/source/MaterialXView/Editor.cpp
@@ -48,13 +48,37 @@ class EditorColorPicker : public ng::ColorPicker
             });
         }
 
-        // The color wheel does not handle alpha properly, so only
-        // overwrite RGB in the callback.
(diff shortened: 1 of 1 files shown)


source/MaterialXView/Editor.cpp
```diff
@@ -374,7 +398,7 @@ void PropertyEditor::addItemToForm(const mx::UIPropertyItem& item, const std::st
             {
                 // Transform sRGBA color picker value to linear space for writing to material
                 mx::Color3 linearCol = mx::Color3(c.r(), c.g(), c.b()).srgbToLinear();
-                mx::Vector3 v(linearCol[0], linearCol[1], linearCol[2]);
+                mx::Vector4 v(linearCol[0], linearCol[1], linearCol[2], c.a());
```
mialana, line 401: Additionally, I realized that in my [previous PR](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2394) to handle color space transformations, I erroneously concatenated all 4D color vectors to 3D when writing to material. This is my mistake, and is fixed in this PR.

jstone-lucasfilm (approved):
Thanks for these fixes, @mialana!

There's one minor coding style issue in the change, but I'll just make the adjustment on my side before merging.

Issue #2379, opened 2025-05-12
WebGPU/WGSL: Option for GLSL Vulkan shader functions to pass separate texture2D and sampler parameters, instead of only sampler2D.


scotbrew:
**Request:** 
To support WebGPU/WGSL (translated from GLSL), GLSL Vulkan code option for needs separate texture2D and sampler binding handles passed in as function parameters instead of only sampler2D. 

**Description:**
WebGPU/WGSL is stricter and can only bind separate texture2D and sampler, contrary to GLSL Vulkan which has options to bind a combined sampler2D (texture2D+sampler) as well as bind separate texture2D and sampler.  There also looks to be a limitation in both GLSL and WGSL that prevents one from constructing a sampler2D struct (with texture2D and sampler) when one binds the texture2D and sampler separately.

To allow MaterialX GLSL Vulkan shader code to work with USD WebGPU, the generated MaterialX GLSL Vulkan code needs to be able to pass separate texture2D and sampler variables.

**Steps to Reproduce:**
With shader code using separate binding of texture2D and sampler, try to compile compile with `glslang`.
`./glslang --target-env vulkan1.3 `

**Original, using sampler2D: (works with GLSL Vulkan, but not WebGPU/WGSL)**
```
layout(binding = 13) uniform sampler2D domeLightFallback;

void mx_image_color3(sampler2D tex_sampler, int layer, vec3 defaultval, vec2 texcoord, int uaddressmode, int vaddressmode, int filtertype, int framerange, int frameoffset, int frameendaction, vec2 uv_scale, vec2 uv_offset, out vec3 result)
{
    ...
}

void NG_tiledimage_vector3(sampler2D file, vec3 default1, vec2 texcoord1, vec2 uvtiling, vec2 uvoffset, vec2 realworldimagesize, vec2 realworldtilesize, int filtertype, int framerange, int frameoffset, int frameendaction, out vec3 out1)
{
    ...
    mx_image_vector3(file, 0, default1, N_multtilesize_vector3_out, 2, 2, filtertype, framerange, frameoffset, frameendaction, vec2(1.000000, 1.000000), vec2(0.000000, 0.000000), N_img_vector3_out);
    ...
}
```

**Proposed Option, using texture2D and sampler:**
```
layout(binding = 0, set = 1) uniform texture2D textureBind_domeLightFallback;
layout(binding = 0, set = 2) uniform sampler samplerBind_domeLightFallback;

[...]

Issue #2377, opened 2025-05-12
WebGPU/WGSL: Compiler Error, LightData struct member variable "type" is a reserved word.


scotbrew:
When integrating MaterialX into USD WebGPU, we have come across a shader code issue in MaterialX.
In short, in the WebGPU/WGSL compiler,  "type" is a reserved word, so cannot be used as a member variable name for the structure LightData GLSL code that WebGPU code is translated from. 

As a test fix, "LightData.type" can be changed to "LightData.light_type" (or something similar).  From sample shaders reviewed so far, this appears to be used internal to the shader source and is not set externally, so should minimize risk. 

This is a blocker for USD WebGPU to use MaterialX materials as-is.
```
struct LightData
{
    int type;
    vec3 position;
    vec3 color;
    float intensity;
    float decay_rate;
    vec3 direction;
    float shadowOcclusion;
};
```

For the WGSL spec, see: https://www.w3.org/TR/WGSL/

**Steps to Reproduce (standalone):**  
1) In Chrome browser, Launch https://jsfiddle.net/cx20/vmkaqw2b/
2) Switch bottom right shader code to "fragment"
3) Change the GLSL code to be:
```
#version 450
layout(location = 0) out vec4 outColor;

struct LightData
{
    int type;
    vec3 position;
    vec3 color;
    float intensity;
    float decay_rate;
    vec3 direction;
    float shadowOcclusion;
};

void main() {
    LightData s;
    s.type = 2;
    s.color = vec3(1.0, 1.0, 1.0);
    outColor = vec4(0.0, 0.0, 1.0, 1.0);
}
```

Resulting WGSL / WebGPU code:
```
struct LightData {
  type : i32,
  position : vec3<f32>,
  color : vec3<f32>,
  intensity : f32,
  decay_rate : f32,
  direction : vec3<f32>,
  shadowOcclusion : f32,
}

var<private> outColor : vec4<f32>;

fn main_1() {
  var s : LightData;
  s.type = 2i;
  s.color = vec3<f32>(1.0f, 1.0f, 1.0f);
  outColor = vec4<f32>(0.0f, 0.0f, 1.0f, 1.0f);
  return;
}

struct main_out {
  @location(0)
  outColor_1 : vec4<f32>,
}

@fragment
fn main() -> main_out {
  main_1();
  return main_out(outColor);
}
```

[...]

PR #2407, opened 2025-05-22
WGSL/WebGPU GLSL ShaderGenerator

scotbrew:
Adding WgslShaderGenerator and its support classes to generate WGSL-compliant GLSL code.  This is targeted to allow USD to render MaterialX shaders with the WebGPU implementation.

**Main changes:**
1. LightData.type renamed to LightData.light_type -- "type" is a reserved word in WGSL.  The change only affects the WGSL flavor of GLSL.
2. Changing the function signature and calls for all functions that use (sampler2D) to (texture2D, sampler), as WGSL GLSL is more restrictive and does not support sampler2D.

**Pull Request Sections:**
1. WgslShaderGenerator and its support classes
3. WGSL-compliant GLSL shaders ~~(filenames: *_wgsl.glsl)~~ (using `#define HW_SEPARATE_SAMPLERS` in glsl shader code)
4. Small modifications to base classes ~~or refactoring a few .glsl files.~~ and adding 2 baseclass virtual functions.

**Issues Resolved:**
This PR addresses the following logged issues.
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/2377
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/2379


scotbrew:
This Pull Request has 3 main sections:
1. WgslShaderGenerator and its support classes
2. WGSL-compliant GLSL shaders (filenames: *_wgsl.glsl)
3. Small modifications to base classes or refactoring a few .glsl files.

libraries/pbrlib/genglsl/lib/mx_latlong.glsl
```diff
@@ -0,0 +1,16 @@
+#define M_PI 3.1415926535897932
```
scotbrew, line 1: Extracted from libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl

libraries/pbrlib/genglsl/lib/mx_microfacet.glsl
```diff
@@ -1,5 +1,10 @@
+#ifndef M_PI
```
scotbrew, line 1: M_PI and M_PI_INV used in this function snippet though may have already been defined above in the resulting shader.  Scoping behind #ifndef so not redefined if defined above in resulting shader.

libraries/targets/genglsl_wgsl.mtlx
```diff
@@ -0,0 +1,14 @@
...
+
+  <!--
+    Copyright Contributors to the MaterialX Project
+    SPDX-License-Identifier: Apache-2.0
+  -->
+
+  <!-- ======================================================================== -->
+  <!-- Target definition for WGSL target. Derives from base GLSL target         -->
+  <!-- ======================================================================== -->
+  <targetdef name="genglsl_wgsl" inherit="genglsl" />
```
scotbrew, line 12: Assuming this is needed.  Note that Vulkan that WgslShaderGenerator derives from still uses "genglsl" (from its base class) and not have a "genvulkan".  If it does, then the "inherit" should be changed here.

resources/Materials/TestSuite/_options.mtlx
```diff
@@ -21,7 +21,7 @@
 
     <!-- Comma separated list of target" specifiers to indicate which
          code generators to use. -->
-    <input name="targets" type="string" value="genglsl,genosl,genmdl,genessl,genmsl" />
+    <input name="targets" type="string" value="genglsl,genglsl_wgsl,genosl,genmdl,genessl,genmsl" />
```
scotbrew, line 24: Added "genglsl_wgsl" to this list, though no other explicit tests added.  Anything more that should be done in the TestSuite?

scotbrew:
No testing done with Js currently.  Likely something would need to change around the LightData struct so it can use LightData.light_type instead of LightData.type to have Js work with WebGPU/WGSL.

source/MaterialXGenGlsl/WgslResourceBindingContext.cpp
```diff
@@ -0,0 +1,82 @@
...
+void WgslResourceBindingContext::initialize()
+{
+    // Reset bind location counter.
+    _hwUniformBindLocation = _hwInitUniformBindLocation;
+}
+
+
+// Copied from VkResourceBindingContext::emitResourceBindings().  
+// Modified the Type::FILENAME uniform codegen.
+void WgslResourceBindingContext::emitResourceBindings(GenContext& context, const VariableBlock& uniforms, ShaderStage& stage)
```
scotbrew, line 29: Copied from VkResourceBindingContext::emitResourceBindings() and then modified if uniform is Type::FILENAME to bind texture2D and sampler, instead of sampler2D.

source/MaterialXGenGlsl/WgslResourceBindingContext.cpp
```diff
@@ -0,0 +1,82 @@
...
+    {
+        if (uniform->getType() != Type::FILENAME)
+        {
+            hasValueUniforms = true;
+            break;
+        }
+    }
+    if (hasValueUniforms)
+    {
+        generator.emitLine("layout (std140, binding=" + std::to_string(_hwUniformBindLocation++) + ") " +
```
scotbrew, line 46: This makes it so the .glsl shader code compiles.  However, the bindLocation likely needs to be tweaked.  The main focus for this Pull Request is to generate proper WGSL code, so that is achieved.  If wanting to have MaterialX render WebGPU/WGSL, then handling texture binding and more likely has to be done.

source/MaterialXGenGlsl/WgslResourceBindingContext.cpp
```diff
@@ -0,0 +1,82 @@
...
+        generator.emitScopeEnd(stage, true);
+    }
+
+    // Second, emit all sampler uniforms as separate uniforms with separate layout bindings
+    for (auto uniform : uniforms.getVariableOrder())
+    {
+        if (uniform->getType() == Type::FILENAME)
+        {
+            // Bind separately as texture2D + sampler
+            generator.emitString("layout (binding=" + std::to_string(_hwUniformBindLocation++) + ") " + syntax.getUniformQualifier() + " ", stage);
```
scotbrew, line 69: This makes it so the .glsl shader code compiles. However, the bindLocation likely needs to be tweaked. The main focus for this Pull Request is to generate proper WGSL code, so that is achieved. If wanting to have MaterialX render WebGPU/WGSL, then handling texture binding and more likely has to be done.

source/MaterialXGenGlsl/WgslShaderGenerator.cpp
```diff
@@ -0,0 +1,90 @@
...
+    // For functions described in ::emitSpecularEnvironment()
+    _tokenSubstitutions[HW::T_ENV_RADIANCE+"_texture"] = HW::ENV_RADIANCE+"_texture";
+    _tokenSubstitutions[HW::T_ENV_RADIANCE+"_sampler"] = HW::ENV_RADIANCE+"_sampler";
+    _tokenSubstitutions[HW::T_ENV_IRRADIANCE+"_texture"] = HW::ENV_IRRADIANCE+"_texture";
+    _tokenSubstitutions[HW::T_ENV_IRRADIANCE+"_sampler"] = HW::ENV_IRRADIANCE+"_sampler";
+
+    // image overrides
+    // Registering these Implementations as done with Implementations in GlslShaderGenerator
+    // See also stdlib_genglsl_impl.mtlx
+    StringVec elementNames = {
```
scotbrew, line 33: This follows the same recipe done in the glslShaderGenerator.  It looks like it needs to be done so that the additional 2 texture function parameters are auto-added like the others.  See glslShaderGenerator for the other IM_* being registered.

source/MaterialXGenMsl/Nodes/LightSamplerNodeMsl.cpp
```diff
@@ -42,7 +42,7 @@ void LightSamplerNodeMsl::emitFunctionDefinition(const ShaderNode& node, GenCont
             string ifstatement = "if ";
             for (const auto& it : lightShaders->get())
             {
-                shadergen.emitLine(ifstatement + "(light.type == " + std::to_string(it.first) + ")", stage, false);
+                shadergen.emitLine(ifstatement + "(light." + shadergen.get_lightdata_typevar_string() + " == " + std::to_string(it.first) + ")", stage, false);
```
scotbrew, line 45: Using new ShaderGenerator virtual function instead of hardcoded value, for consistency.

source/MaterialXGenShader/HwShaderGenerator.cpp
```diff
@@ -308,7 +308,7 @@ ShaderPtr HwShaderGenerator::createShader(const string& name, ElementPtr element
     VariableBlockPtr psPrivateUniforms = ps->createUniformBlock(HW::PRIVATE_UNIFORMS, "u_prv");
     VariableBlockPtr psPublicUniforms = ps->createUniformBlock(HW::PUBLIC_UNIFORMS, "u_pub");
     VariableBlockPtr lightData = ps->createUniformBlock(HW::LIGHT_DATA, HW::T_LIGHT_DATA_INSTANCE);
-    lightData->add(Type::INTEGER, "type");
+    lightData->add(Type::INTEGER, get_lightdata_typevar_string() );
```
scotbrew, line 311: Using new ShaderGenerator virtual function instead of hardcoded value, so will work with both WGSL (that uses LightData.light_type) and the other ShaderGenerators (that use LightData.type).

source/MaterialXGenShader/Nodes/CompoundNode.cpp
```diff
@@ -73,7 +73,18 @@ void CompoundNode::emitFunctionDefinition(const ShaderNode&, GenContext& context
         // Add all inputs
         for (ShaderGraphInputSocket* inputSocket : _rootGraph->getInputSockets())
         {
-            shadergen.emitString(delim + syntax.getTypeName(inputSocket->getType()) + " " + inputSocket->getVariable(), stage);
+            // Conditional behavior for WgslShaderGenerator
```
scotbrew, line 76: There are a few different ways of handling this, where changes need to be done for WGSL.  We can have this exception code in the CompoundNode or make a new virtual function ShaderGenerator::emitFunctionDefinitionParameter() that WgslShaderGenerator can then override.

What is the preferred way to determine what ShaderGenerator is being used?
-  Idea1: use dynamic_cast or other methods to test for WgslShaderGeneratorPtr.
-  Idea2: (context.getShaderGenerator().getTarget() == "genglsl_wgsl")

ashwinbhat:
Hi @scotbrew greatly appreciate you getting this work started 👍🏽  I briefly looked at the changes and think you are addressing all the key areas wrt the sampler split. However, I see that we will have to duplicate major blocks of shader code for minor differences in function call. 
While having separate `_wgsl.glsl` and `.glsl` allows us to tweak the wgsl version later, I'm wondering for the initial release if we could consolidate `mx_environment_*` and `hex_tile` by using a compile time macro. So something like this:
      #define VULKAN 100
        #if VULKAN > 100
            mx_latlong_map_lookup(Lw, $envMatrix, lod, $envRadiance_texture, $envRadiance_sampler)
        #else
            mx_latlong_map_lookup(Lw, $envMatrix, lod, $envRadiance)
        #endif

Could we introduce a `hwSeparateSamplers` option in https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/GenOptions.h that allows this branching and emits the required `#define`. This will also reduce the need to create a separate WgslResourceBindingContext since Vulkan ResourceBindingContext could simply choose to sperate or combine based on hwSeparateSamplers option setting. 
        
Based on suggestion from Niklas slack I think we want to use `genwgsl` instead of `genglsl_wgsl` so this will allow underlying implementation to change and remove the `glsl` notion in future.

kwokcb:
In terms of design / implementation choices, I am assuming this as the context:
1. The output just needs to sufficiently compatible Vulkan based GLSL to produce WGSL via the suitable tool-chain which uses `tint`.
2.  That we also want to ensure what is done here does not prevent us from migrating to SLANG longer term to get a direct SLANG ->WGSL output.
3. We change as little as possible to existing generators inserting tweaks only if absolutely needed. This allow easier increment integration.

source/MaterialXGenShader/Nodes/CompoundNode.cpp
scotbrew: I agree and that makes sense.  I will then add a new virtual function to ShaderGenerator that will be used by this function.
`ShaderGenerator::emitFunctionDefinitionParameter()` that WgslShaderGenerator can then override.

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
```diff
@@ -484,20 +484,6 @@ vec3 mx_refraction_solid_sphere(vec3 R, vec3 N, float ior)
     return refract(R, N1, ior);
 }
 
-vec2 mx_latlong_projection(vec3 dir)
```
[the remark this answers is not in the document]
scotbrew: No longer implemented this way.  Closing out comment thread.

libraries/pbrlib/genglsl/lib/mx_latlong.glsl
scotbrew: No longer implemented this way.  Closing out comment thread.

libraries/pbrlib/genglsl/lib/mx_microfacet.glsl
scotbrew: No longer implemented this way.  Closing out comment thread.

resources/Materials/TestSuite/_options.mtlx
scotbrew: Removed use of "genglsl_wgsl" as it was not needed.  Keeping testing to the broader genglsl category and specializing from there.  Closing as no longer implementing this.

source/MaterialXGenGlsl/GlslShaderGenerator.cpp
```diff
@@ -211,10 +211,14 @@ void GlslShaderGenerator::emitSpecularEnvironment(GenContext& context, ShaderSta
     int specularMethod = context.getOptions().hwSpecularEnvironmentMethod;
     if (specularMethod == SPECULAR_ENVIRONMENT_FIS)
     {
+        emitLibraryInclude("pbrlib/genglsl/lib/mx_latlong.glsl", context, stage);
```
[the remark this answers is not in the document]
scotbrew: No longer implemented this way.  Closing out comment thread.

source/MaterialXGenGlsl/WgslResourceBindingContext.h
```diff
@@ -0,0 +1,49 @@
...
+
+    // Initialize the context before generation starts.
+    void initialize() override;
+
+    // Emit uniforms with binding information
+    void emitResourceBindings(GenContext& context, const VariableBlock& uniforms, ShaderStage& stage) override;
+
+  protected:
+    // Binding location for uniform blocks
+    size_t _hwUniformBindLocation = 0;
```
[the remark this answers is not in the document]
scotbrew: Streamlined away these variables as they did not need to be unique to the Wgsl implementation.  Closing out comment thread.

source/MaterialXGenGlsl/WgslShaderGenerator.cpp
scotbrew: No longer implemented this way.  Is not needed as not adding a "genglsl_wgsl" type.  Closing out comment thread.

source/MaterialXGenShader/Nodes/CompoundNode.cpp
scotbrew: Closing out this discussion as the pathway has been discussed, decided, and implemented.

source/MaterialXGenGlsl/WgslShaderGenerator.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+
+void WgslShaderGenerator::emitSpecularEnvironment(GenContext& context, ShaderStage& stage) const
+{
+    int specularMethod = context.getOptions().hwSpecularEnvironmentMethod;
+    if (specularMethod == SPECULAR_ENVIRONMENT_FIS)
+    {
+        emitLibraryInclude("pbrlib/genglsl/lib/mx_latlong_wgsl.glsl", context, stage);
```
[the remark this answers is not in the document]
scotbrew: No longer implemented this way.  Closing out this thread.

source/MaterialXGenShader/ShaderGenerator.h
```diff
@@ -211,6 +214,9 @@ class MX_GENSHADER_API ShaderGenerator
     /// export of metadata.
     virtual void registerShaderMetadata(const DocumentPtr& doc, GenContext& context) const;
 
+    // Retrieve the string used for the LightData.type member variable
+    virtual const string getLightDataTypevarString() const { return string("type"); }
```
scotbrew, line 218: Is this fine to return a constructed string, or would we prefer to create a static string in the class and return that?

source/MaterialXGenMsl/Nodes/LightSamplerNodeMsl.cpp
scotbrew: Changed virtual function name to be in line with naming conventions.

source/MaterialXGenShader/HwShaderGenerator.cpp
scotbrew: Renaming virtual function to align with naming conventions.  Resulting generated shader code is the same.

source/MaterialXGenShader/ShaderGenerator.h
scotbrew: I am answering my own question.  Changing to a string& and returning a static string.  Cleaner solution.  This aligns with how it is done in other segments of the MaterialX code.

scotbrew:
I've combed over the changes to streamline a bit further. PR should be ready for final review.

kwokcb (commented):
This looks great with some minor items: 

1. It would be good to expose the environment texture / sampler uniform names as a courtesy to implementors so they know what uniforms to use for binding.

I can't think of any other way to make the source code cleaner than what's done here with the preprocessor directive. 

2. I think the sampler gradient call syntax is incorrect ?

Thanks.

libraries/stdlib/genglsl/mx_hextilednormalmap.glsl
```diff
@@ -29,9 +34,16 @@ void mx_hextilednormalmap_vector3(
 
     HextileData tile_data = mx_hextile_coord(coord, rotation, rotation_range, scale, scale_range, offset, offset_range);
 
+#ifdef HW_SEPARATE_SAMPLERS
+    vec3 nm1 = textureGrad(sampler2D(tex_texture, tex_sampler), tile_data.coord1, tile_data.ddx1, tile_data.ddy1).xyz;
```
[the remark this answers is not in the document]
scotbrew: Good question.  Though pure WGSL uses `textureSampleGrad`, the  code that MaterialX is generating here is GLSL Vulkan shader code that is WGSL-compliant (that the Tint library can then later convert from GLSL Vulkan to pure WGSL).  So here we want to keep generating the GLSL function `textureGrad()`.

libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
```diff
@@ -64,6 +68,10 @@ vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 alpha, int distributio
 
 vec3 mx_environment_irradiance(vec3 N)
 {
+#ifdef HW_SEPARATE_SAMPLERS
+    vec3 Li = mx_latlong_map_lookup(N, $envMatrix, 0.0, $envIrradiance_texture, $envIrradiance_sampler);
```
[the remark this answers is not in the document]
kwokcb: Since it's only for WGSL this is a good place to leave this.

JGamache-autodesk (approved):
Apart from missing WGSL forbidden keywords in the Syntax, everything else looks fine to me.

JGamache-autodesk (approved):

source/MaterialXGenGlsl/WgslSyntax.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include <MaterialXGenGlsl/WgslSyntax.h>
+
+MATERIALX_NAMESPACE_BEGIN
+
+WgslSyntax::WgslSyntax(TypeSystemPtr typeSystem) : VkSyntax(typeSystem)
+{
+    // Add in WGSL specific keywords
+    registerReservedWords( { 
+    	"type"
```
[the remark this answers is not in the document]
scotbrew: Done -- added the full list of WGSL keywords and reserved words.

kwokcb (approved):

libraries/stdlib/genglsl/mx_image_color3.glsl
```diff
@@ -1,7 +1,15 @@
 #include "lib/$fileTransformUv"
 
+#ifdef HW_SEPARATE_SAMPLERS
+void mx_image_color3(texture2D tex_texture, sampler tex_sampler, int layer, vec3 defaultval, vec2 texcoord, int uaddressmode, int vaddressmode, int filtertype, int framerange, int frameoffset, int frameendaction, vec2 uv_scale, vec2 uv_offset, out vec3 result)
+{
+    vec2 uv = mx_transform_uv(texcoord, uv_scale, uv_offset);
+    result = texture(sampler2D(tex_texture, tex_sampler), uv).rgb;
```
jstone-lucasfilm, line 7: @scotbrew This is probably a naïve question, but I wanted to ask:

In the current pull request, I'm often seeing a pattern where the separated `texture2d` and `sampler` are recombined as a `sampler2d` before they are passed to texture calls.

What would stop us from simply passing this combined `sampler2d` into the majority of our MaterialX GLSL functions, e.g. the `mx_image_color3` function seen here, so that they retain a unified interface across GLSL and WGSL?

Would that potentially simplify this proposal, reducing the number of locations where the new `HW_SEPARATE_SAMPLERS` macro is needed?

scotbrew: That is a very reasonable question, and the answer underscores the reason for most of this work.

WGSL only supports separate texture2D and sampler, and not sampler2D directly.  If one tries to construct sampler2D(my_tex, my_sampler) from texture2D and sampler bindings and then pass that sampler2D in as a function parameter, the glslang compiler throws the error below. I could not find a way around that.
`ERROR: unknown:2917: 'call argument' : sampler constructor must appear at point of use`.

See Issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/2379 for more information.

To try this out, with shader code using separate binding of texture2D and sampler, compile with glslang.
`./glslang --target-env vulkan1.3`

jstone-lucasfilm: Got it, @scotbrew, and thanks for the explanation.

You may have already considered this alternative, but just in case, I'll propose it here:

What if we were to define a `HwTextureSampler` struct, using your existing `HW_SEPARATE_SAMPLERS` macro used to handle the platform-specific nature of its contents?

```
struct HwTextureSampler
{
#ifdef HW_SEPARATE_SAMPLERS
    texture2D tex_texture;
    sampler   tex_sampler;
#else
    sampler2D tex_sampler;
#endif
};
```

This would allow the interfaces of functions such as `mx_image_color3` to remain unified, and we could include an additional `mx_get_sampler_2d` function to handle the platform-specific conversion of an `HwTextureSampler` to the final `sampler2d` object:

```
void mx_image_color3(HwTextureSampler tex_sampler, int layer, vec3 defaultval, vec2 texcoord, int uaddressmode, int vaddressmode, int filtertype, int framerange, int frameoffset, int frameendaction, vec2 uv_scale, vec2 uv_offset, out vec3 result)
{
    vec2 uv = mx_transform_uv(texcoord, uv_scale, uv_offset);
    result = texture(mx_get_sampler_2d(tex_sampler), uv).rgb;
}
```

Assuming the syntaxes of GLSL and WGSL allow this inclusion of texture/sampler objects in structures (a detail that I don't actually know), perhaps this would make the new feature slightly simpler and easier to maintain?

scotbrew: Yes, that is a good idea, though have tried that `struct` alternative and was unsuccessful.  The shader compiler throws an error that one cannot add a binding variable to a structure for GLSL (though it can be done with Metal).  I agree if someone can get it to work, then that would simplify and unify the shaders.  

To test possible shader scenarios, use:
`./glslang --target-env vulkan1.3`

jstone-lucasfilm: That makes sense, @scotbrew, and I'm OK proceeding with the high-level approach that you've proposed here.

scotbrew: Sounds good. If the WGSL tech, spec, or compiler limitations later evolve, then the HW_SEPARATE_SAMPLERS define in the code will make it relatively straightforward to refactor.

jstone-lucasfilm:
@scotbrew Oh, I just noticed that our render validation tests are returning all-black renders with this PR!

I've confirmed the same results locally on my Windows machine, and here are the current renders from the Viewer and Graph Editor with this PR:

Viewer:

Viewer

Graph Editor:

GraphEditor

jstone-lucasfilm:
Looking back through recent GitHub Actions runs, here's the first one I see where the render validation tests are returning black renders, in case that helps to track down the cause:

https://github.com/AcademySoftwareFoundation/MaterialX/actions/runs/15331741009

scotbrew:
That is definitely one to catch and fix before fully integrating.  

Updates:
1. I'll take a look to repro it on my end and compare shader code to identify the fix.
2. Verified locally that the latest MaterialXView with this PR has black renders.  Tracking down the introduced issue.
3. The issue has been isolated to a refactoring change for the PR in the commit "WGSL: Refactor shader code and CompountNode shader gen" (SHA: 131d322aad76b1a676b65c4813dc6b8299393814 ).
4. Fixed in the latest commit.  The refactoring of ShaderGenerator::emitFunctionDefinitionParameter() needed to treat input and output ShaderPorts slightly differently.

scotbrew:
The identified render issue regression has been addressed.  Here are the latest images from the Linux test runners.
GraphEditor_MarbleSolid Viewer_BrassAverage Viewer_CarpaintTranslated

scotbrew:
For code spot-checking, a comparison was done between the standard_surface_look_brass_tiled shader code generated by the latest PR2407 code and latest main branch.  Everything looks as expected.
- VS: Identical
- PS: Diff adding only code scoped by #ifdef HW_SEPARATE_SAMPLERS. (see attached *-DIFF.txt)
- MDL: Identical

[standard_surface_look_brass_tiled_ps-DIFF.txt](https://github.com/user-attachments/files/20657436/standard_surface_look_brass_tiled_ps-DIFF.txt)

jstone-lucasfilm (approved):
Thanks so much for this contribution, @scotbrew, and I believe this looks ready to go!

Further in the future, I'd be interested in trying additional strategies to reduce the interface duplication in the hardware implementations of our image nodes, but the benefits of this initial implementation greatly outweigh those minor issues.

PR #2409, opened 2025-05-23
Remove ShaderNodeImpl::getTarget() function

ld-kerley:
This PR removes `ShaderNodeImpl::getTarget()` and all its downstream overloaded functions.

It appears that this function isn't used anywhere, and removing it allows us to simplify the class hierarchy, and possibly opens the door to more consolidation between the GLSL/MSL C++ implementation nodes. 

If removing this function is going to cause problems downstream anywhere, it would be great if someone could give some context - and we should probably add some tests to capture that behavior.


niklasharrysson (approved):
Nice cleanup! This looks good to me

PR #2400, opened 2025-05-21
Consolidate ClosureSourceCode and SourceCode C++ nodes

ld-kerley:
There was really very little difference left after the recent closure data refactor, so I think it makes sense to merge the two together and further reduce the complexity of the C++ implementation nodes.

PR #2410, opened 2025-05-23
Consolidate ClosureCompound with Compound C++ node

ld-kerley:
Similar to the recent work in #2400.

This PR consolidates the few minor differences left between the ClosureCompound and Compound C++ nodes., further reducing the complexity of the C++ implementation nodes.

Also consolidates the logic used to determine if the output is a closure, to be shared here and with the SourceCode node.


source/MaterialXGenShader/ShaderNodeImpl.cpp
```diff
@@ -76,6 +76,19 @@ ShaderGraph* ShaderNodeImpl::getGraph() const
...
+    if (outputs.empty())
+    {
+        // This should never happen as we auto populate the default 'out' output based on the
+        // node type if no output is present.
+        throw ExceptionShaderGenError("Node has no outputs defined'");
+    }
+    return outputs[0]->getType().isClosure();
+}
+
+
```
niklasharrysson, line 91: Nitpick: Extra newline

niklasharrysson (approved):
Looks good to me! Just added two minor comments above.

Issue #1551, opened 2023-10-04
Blur node size inconsistent when zooming viewport


zicher3d:
In the attached sample, a texture (or a procedural) that is blurred by a Blur node seem to blur based on the screen pixels. I would expect the same amount of blur to be consistent in UV space (the documentation seem to say so), but in the video you can clearly see the texture being blurred more when zooming out, and less when zooming in.

The video uses a texture, the MTLX sample uses a procedural.

https://github.com/AcademySoftwareFoundation/MaterialX/assets/7254958/33753259-a903-48df-a5fe-8b821de9dd0d

[blur_zoom_issue.zip](https://github.com/AcademySoftwareFoundation/MaterialX/files/12806554/blur_zoom_issue.zip)

jstone-lucasfilm:
Let's close out this issue, as we've removed the existing implementations of `blur` in change #2411, with our goal being to reimplement them from scratch in a future release of MaterialX.

When time permits, we should write up a new plan for the `blur` nodes in a GitHub Issue, taking into account all of our learnings from earlier implementations.

zicher3d:
Good for me

PR #2411, opened 2025-05-24
Remove Blur node implementation

ld-kerley:
I'm submitting this PR - because I think it's the best we can do for now.  I looked in to trying to find a portable robust implementation for this, and couldn't really come up with anything.

I think it's better to be explicitly do nothing, than to do something that is confusing. 

MDL currently just passes through the `in` input as the output, so I'm using that as the template for the others as it seems like a reasonable fallback - implemented as a nodegraph that is now shared across all languages. 

I'll happily take this PR down, if we can find a robust blur implementation.  Ultimately if we move forward with this, and we don't find an implementation that can work, I would propose we move the `blur` node to the proposals specification in our next breaking change release.


jstone-lucasfilm:
This is definitely an important issue to resolve, and I'll link your proposal to @zicher3d's earlier report on visual issues with the `blur` node in real-time shading:

https://github.com/AcademySoftwareFoundation/MaterialX/issues/1551

jstone-lucasfilm:
Following our approach with `heighttonormal`, I wonder if we can use our knowledge of the relationship between screen-space and texture-space derivatives to propose a better implementation of `mx_compute_sample_size_uv` for the `blur` node:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/genglsl/lib/mx_sampling.glsl#L9

It seems very possible that this function is the source of the unintended screen-space dependency in the current `blur` node implementation, and we may be able to apply our learnings here as well.

jstone-lucasfilm:
Before we move forward with such a major change, I'd like to loop in our Autodesk colleagues who contributed this feature to MaterialX in their original work on ShaderGen, and who may have ideas for taking it forward in the future.

I believe @kwokcb originally authored this code in his work on ShaderGen at Autodesk, and @ashwinbhat has been part of recent conversations about evolving our support for `blur` and other convolution nodes in the future, potentially using a stochastic approach that could scale between real-time and film-quality results.

If our Autodesk colleagues would prefer to delete the current code and start fresh sometime in the future, I would support that approach, but it should definitely be their call.

ld-kerley:
To be clear about what's being proposed here.  I'm not suggesting we delete the node, just that we revert its implementation to be a passthru of the `in` input.  Which is what MDL stdlib files implementations, is doing already.

 I feel this would be less confusing to users than the current implementation that highly view dependent.

libraries/README.md
```diff
@@ -76,3 +76,4 @@ This folder contains the standard data libraries for MaterialX, providing declar
 - Shader generation does not currently support:
     - `displacementshader` and `volumeshader` nodes for hardware shading targets (GLSL, MSL).
     - `hextiledimage` and `hextilednormalmap` for OSL and MDL.
+    - `blur` the implementation passes through`in` unmodified.
```
jstone-lucasfilm: Very minor, but there should be a space between "through" and "`in`".

ashwinbhat:
After discussing with @zicher3d and @JGamache-autodesk we are ok with removing the current implementation for `blur`

jstone-lucasfilm (approved):
This looks good to me as well, thanks @ld-kerley!

PR #2413, opened 2025-05-24
Align screen colors between Editor and Viewer

jstone-lucasfilm:
Following up on recent work from @mialana, this changelist updates the viewport screen color in the MaterialX Graph Editor to match that of the MaterialX Viewer.

Now that the UI elements in the Graph Editor have their intended dark color scheme, the darker screen color from the Viewer makes a good choice for the Graph Editor as well.


jstone-lucasfilm:
Here's a screen capture of the Graph Editor with the updated viewport screen color, aligning with that of the Viewer:

MaterialXGraphEditor_AlignedScreenColor

PR #2415, opened 2025-05-27
Adjust binding of frame uniform to ensure Metal renderers animate

warrenm:


jstone-lucasfilm (approved):
Thanks for this fix, @warrenm!

PR #2418, opened 2025-05-28
Add `color4` input support to web viewer UI

hybridherbst:
Seems this was missing.

It looks like this – for the color4 property "dot_Color", two inputs are added to the UI, that modify .rgb and .a:
image


jstone-lucasfilm (approved):
This change looks good to me, thanks @hybridherbst!

Issue #2417, opened 2025-05-28
Restore support for frame-independent `time` in HW shading


jstone-lucasfilm:
Following up on issue #2391, we should restore support for a frame-independent implementation of the `time` node in GLSL and MSL.  In the linked issue, @warrenm described this need as follows:

> With the [death](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2192) of the `HwTimeNode` class, it seems that time essentially became defined as _frame index divided by FPS_. This is, I assume, a common definition in offline and production work, but it doesn't make sense in a real-time context where an application could easily have a variable frame rate (e.g. on an iPhone that throttles conditionally from 120FPS to 60 FPS, beyond the user's or application's control). For animation that's as smooth as possible in a real-time context, it's desirable to use the host's best attempt at a monotonic clock that runs independently of frame rate.

> The lack of frame rate independence can be seen (at least on macOS) by loading a material with a `frame` node in the viewer: in the steady state it animates a low frame rate, but when the orbit controls are active, the frame rate spikes upward, causing drastic differences in animation speed. This cannot be alleviated with a time node that is implemented as the quotient of the frame index and FPS, even if the FPS is sampled instantaneously.

Our own MaterialX Viewer is an example of an application that requires a frame-independent time for correct rendering of animated materials, and it's likely that UsdView and LookdevX will need this functionality in the future as well.

jstone-lucasfilm:
I've written up an initial PR for this change in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2442

PR #2419, opened 2025-05-28
Add support for `time` and `frame` in web viewer

hybridherbst:
Time is currently calculated based on the time at the start of the frame, in seconds – so not framecount / FPS as per the spec, but the more useful calculation in real-time settings (related: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2417)

This should be the last piece missing to fix
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/1952.


jstone-lucasfilm (approved):
This looks like a good improvement to me, thanks @hybridherbst!

Issue #2422, opened 2025-05-29
Shader Generation problem with `<postition>` and `interfacename`


ld-kerley:
I suspect I have found a shader generation bug. It appears that using a `<position>` inside a nodegraph, and exposing the `space` parameter via the `interface name` mechanism appears to not generate valid shader code, either in MSL or OSL. 

Using the `<position>` node alone appears to work.

```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">

  <position name="myPos" type="vector3">
    <input name="space" type="string" value="object"/>
  </position>

  <convert name="conv" type="color3">
    <input name="in" type="vector3" nodename="myPos"/>
  </convert>
  <surface_unlit name="Srf" type="surfaceshader">
    <input name="emission_color" type="color3" nodename="conv" />
  </surface_unlit>
  <surfacematerial name="Mtl" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="Srf" />
  </surfacematerial>

</materialx>
```

<img width="1348" alt="Image" src="https://github.com/user-attachments/assets/d75640b0-7f2f-46bc-b9b5-c57c9882eb1a" />

But the following doesn't work

```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709">

  <nodedef name="ND_positionwrapper" node="positionwrapper">
    <input name="wrap_space" type="string" value="object" enum="model, object, world" uniform="true"/>
    <output name="out" type="vector3"/>
  </nodedef>

  <nodegraph name="NG_positionwrapper" nodedef="ND_positionwrapper">
    <position name="myPos" type="vector3">
      <input name="space" type="string" interfacename="wrap_space"/>
    </position>
    <output name="out" type="vector3" nodename="myPos"/>
  </nodegraph>


  <positionwrapper name="myPosWrapper" type="vector3">
    <input name="wrap_space" type="string" value="object"/>
  </positionwrapper>

  <convert name="conv" type="color3">
    <input name="in" type="vector3" nodename="myPosWrapper"/>
  </convert>
  <surface_unlit name="Srf" type="surfaceshader">
    <input name="emission_color" type="color3" nodename="conv" />
  </surface_unlit>
  <surfacematerial name="Mtl" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="Srf" />
  </surfacematerial>

</materialx>
```

[...]

PR #2423, opened 2025-05-29
Fix enumeration interfacename bug

ld-kerley:
Related to #2422.

When creating the `Input`, during construction of the graph for code generation, currently if an enumeration is exposed via `interfacename`, then the `value` member ends up getting set to `nullptr`, which means later the code generation cannot determine the correct type.

This PR guards the setting of the value, to avoid setting this to `nullptr`.

### DIFF
--- a/source/MaterialXGenShader/ShaderNode.cpp
+++ b/source/MaterialXGenShader/ShaderNode.cpp
@@ -353,16 +353,21 @@ void ShaderNode::initialize(const Node& node, const NodeDef& nodeDef, GenContext
                 }
             }
             const string& valueString = portValue ? portValue->getValueString() : EMPTY_STRING;
-            std::pair<TypeDesc, ValuePtr> enumResult;
-            const string& enumNames = nodeDefInput->getAttribute(ValueElement::ENUM_ATTRIBUTE);
-            const TypeDesc type = context.getTypeDesc(nodeDefInput->getType());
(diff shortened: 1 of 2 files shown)


source/MaterialXGenShader/ShaderNode.cpp
```diff
@@ -353,16 +353,21 @@ void ShaderNode::initialize(const Node& node, const NodeDef& nodeDef, GenContext
                 }
             }
             const string& valueString = portValue ? portValue->getValueString() : EMPTY_STRING;
-            std::pair<TypeDesc, ValuePtr> enumResult;
-            const string& enumNames = nodeDefInput->getAttribute(ValueElement::ENUM_ATTRIBUTE);
-            const TypeDesc type = context.getTypeDesc(nodeDefInput->getType());
-            if (context.getShaderGenerator().getSyntax().remapEnumeration(valueString, type, enumNames, enumResult))
```
kwokcb, line 359: The "bug" appears to be in GLSL and MSL `remapEnumeration` method. It is returning true when there is no value mapped. MDL returns the correct value of `false` in this case.

ld-kerley: Chatted offline with Bernard - we still think the original fix is valid - but I'm gonna put up a PR to fix the return value from `remapEnumeration` as well.

PR #2426, opened 2025-05-30
Allow keyboard deletion of selected links in Graph Editor

hybridherbst:
This PR adds the ability to delete selected links using the keyboard, same as it works for nodes.
The existing functionality to Option-Click on a link is still the same, but now deletion of links should be more discoverable.

Personally I only found Option-click by looking at the code while trying to understand how I can delete links; the obvious select-and-press-delete hadn't worked, with this PR it does.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -4280,6 +4280,27 @@ void Graph::drawGraph(ImVec2 mousePos)
                     }
                     linkGraph();
                 }
+                else if (selectedLinks.size() > 0)
+                {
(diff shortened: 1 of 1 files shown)


hybridherbst:
Thanks! After submitting this PR I ran into a few cases where seemingly the wrong thing got deleted when pressing "delete" while having an edge selected, so now I'm not super sure anymore about the Link ID logic being correct here. Although from the code I don't know how it can be wrong.

In other words: would be great if someone else could also take a look to verify the code.

jstone-lucasfilm:
If you know of specific cases where deletion behaves incorrectly with links, I'd recommend posting a GitHub Issue for them, and we can focus developers on those issues next.

PR #2435, opened 2025-06-03
Improve remapEnumeration() for GLSL and MSL (and align MDL source)

ld-kerley:
Related to #2423 

After the PR above was merged Bernard Kwok rightly pointed out that `remapEnumeration()` itself should also have guard against empty input values.


jstone-lucasfilm:
These code changes look fine to me, thanks @ld-kerley.

The one detail that I would change is the title, where we don't usually include punctuation marks such as parentheses, and we aim to keep the character count below 50 when possible.

PR #2442, opened 2025-06-06
Restore support for frame-independent `time`

jstone-lucasfilm:
This changelist restores support for frame-independent `time` in our hardware shader generators, allowing each renderer to choose whether its implementation should be based on wall-clock time or a frame multiplier.

The MaterialX Viewer and Graph Editor, which have frame rates that vary with user interactions and system resources, have been updated to use wall-clock time in their `time` implementations.

The animated clock example has been assigned a more visible starting speed, maintaining its usefulness in validating `time` implementations.


jstone-lucasfilm:
I'm interested in feedback on this proposal, and I'll initially CC @ld-kerley and @hybridherbst for their thoughts.

jstone-lucasfilm:
Additionally CC'ing @ashwinbhat and @zicher3d, in case they would like to try this with their latest animated materials.

hybridherbst:
I'm in favor of this PR – making a local build from your branch now to test it out.

hybridherbst:
Looks good – one issue I found is that MaterialXView will render at ~10fps when not interacting, but better-looking 60fps while interacting. MaterialXGraphEditor always renders at a smooth framerate. Can the viewer be updated to do that as well? The samples look much better with it.

Attached one more test file that ticks a number up each second, making it easier to see "time works" than the clock asset: [SimpleTimeGraph.zip](https://github.com/user-attachments/files/20659036/SimpleTimeGraph.zip)

ScreenRecording2025-06-09at19 57 35-ezgif com-optimize

jstone-lucasfilm:
That's a good note about the impact of user interactions on the viewer, and although it was originally an intentional choice to minimize the GPU overhead of the application, it's worthwhile to reconsider today.

If we ultimately wanted to switch to rendering at the monitor refresh rate in the viewer, it should be straightforward to make that change, and I believe we already use that rendering mode when the user enables the frame rate display.

jstone-lucasfilm:
Thanks for the testing and feedback, @hybridherbst.  Are there any additional thoughts from the community on this change before we move forward?

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -1488,11 +1488,9 @@
   <!--
     Node: <time>
     The current time in seconds as defined by the host environment.
-    (Default values is 1/24.0)
   -->
   <nodedef name="ND_time_float" node="time" nodegroup="application">
-    <input name="fps" type="float" value="24.0" />
```
ld-kerley, line 1494: I don't think that we can just remove inputs to a node without handling this correctly in a breaking change in the data library? 

This is going to change peoples materials.

jstone-lucasfilm: This is an interesting case, @ld-kerley, as the MaterialX specification never actually included this `fps` input, but it was present in our shader generators.

What are your thoughts on how best to handle this?

ld-kerley: Removing appears to me to break the promise of not changing any existing materialX documents. I did notice the input is not mentioned in the specification, but also we know that there are implementations out there that depend on the data library as the definition of the standard, and others that reference the specification document.  Thats why we've been working towards aligning to two.

I certainly think a change like this should be discussed a lot more widely than a single PR, as it has the possibility of changing existing data.

Maybe the best thing we can do here is to add an additional node definition for `time`, instead of modifying the existing one.  I believe that might protect any existing data, but that would need testing, and is probably worth discussing more widely too.

jstone-lucasfilm: I like the suggestion above, and here's an initial GitHub Issue to add a second signature for frame-based time:

https://github.com/AcademySoftwareFoundation/MaterialX/issues/2449

jstone-lucasfilm: Based on a bit of further study of MaterialX `time` nodes across the industry, I've put together a slight refinement of this idea as https://github.com/AcademySoftwareFoundation/MaterialX/pull/2450

Issue #2444, opened 2025-06-08
Documentation: Questions regarding usage and implementation of Geometry Name Expressions


hybridherbst:
Hi, I have some understanding and implementation questions for [Geometry Name Expressions](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.GeomExts.md#geometry-name-expressions). I'm numbering them so that it's hopefully easier to answer specific ones, and ideally some clarifying language can be added to the spec.

1. The spec mentions that `a/b/c` matches matches `a/b/c` and `a/b/c/d`, but fails to mention if it should also match `a/b/cd` and if not, why not. (I would assume it does not match the latter, but I can't derive that from the spec.)

3. Is it correct to assume that `//My Node Name` is supposed to match `My Node Name` at any depth?
4. How would a runtime differentiate between a string that's supposed to be a geom expression, and a string that just happens to be a fancy node name? E.g. I can easily have nodes named `Hello * World [4]` in a glTF file (and have seen that in the wild), but it looks to me like that would be interpreted as an expression, not a node name. Is there a way missing in the spec to escape characters?
5. Same as the last question, it's not clear to me if forward slashes in node names `/` can be escaped somehow. I understand some formats like OpenUSD don't allow slashes in node names, but others (glTF, FBX, ...) allow it.
6. A comment in Geom.h mentions `// @todo Geometry name expressions are not yet supported.` – so which parts are supposed to work, e.g. can I assume prepending  `//` or using `*` should work, or should I assume nothing but exact name matches works?
7. From an implementation perspective, since materialassign "falls through" and later ones override earlier ones, it seems that a naive implementation will easily be of quadratic complexity (each node needs to be checked against each materialassign). Thus, it's probably a reasonable strategy to match in reverse (starting with the last item) and removing matches from the set of nodes to check. Is that a valid assumption?
8. For the `[a-z]`syntax, is the assumption that "character preceding" and "character following" mean: in ASCII? Or is it meant more strict? I assume that's the intention but it could probably be spelled out... e.g. right now it's not clear to me if [A..z] or [Y..b] (which includes Y,Z,a,b) are allowed, and if [7..A] also matches the characters ? and @, besides others, because they're in-between 7 and A in ASCII.
9. These two sections in the spec are confusing to me:
[...]

PR #2445, opened 2025-06-09
Keep original object and material names in Web Viewer

hybridherbst:
Previously, object names were taken directly from the three.js scene. However, three.js enforces specific constraints for object names (e.g. spaces, braces and so on are replaced with `_`).

This renaming breaks DAG matching for geom expressions, as the compared name might not be right. three.js keeps the original name around though, so this PR makes sure the original name is used for path matching instead of the in-scene name.

The PR also sets the material name properly from the `materialassign` instead of keeping it empty.

### DIFF
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -889,7 +889,7 @@ export class Material
                         assigned += viewer.getScene().updateMaterial(matassign);
                         matassign.setGeometry(temp);
                         assignedSolo = true;
-                        break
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm (approved):
Thanks for the fixes, @hybridherbst!

hybridherbst:
Thanks! To do further improvements it would be great to get some feedback on the spec questions in https://github.com/AcademySoftwareFoundation/MaterialX/issues/2444#issuecomment-2954305575.

PR #2446, opened 2025-06-09
Move uniform updates to object scope in Web Viewer

hybridherbst:
The current handling of uniforms was per-material, which means that when a material is used on multiple objects, they all share uniforms – thus, they were all rendered with the uniforms of the last user of that material.

Any multi-object files were affected. It just happened to work for the default shader ball because there all objects have the same pivot.

As a side effect of this, the previous on-demand updating of uniforms (using `#_updateTransforms`) isn't possible anymore, so I removed the related code.

| File | Before PR | Now
| - | - | - |
| Primitives.glb<br>[Primitives.glb.zip](https://github.com/user-attachments/files/20656983/Primitives.glb.zip) | image | image | 
| ComplexScene.glb<br>[ComplexScene.glb.zip](https://github.com/user-attachments/files/20656988/ComplexScene.glb.zip) | image | image |

### DIFF
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -106,6 +106,15 @@ export class Scene
         }
         scene.add(model);
 
+        // Set up onBeforeRender callbacks so that we can update uniforms per object right before rendering.
+        model.traverse((child) =>
+        {
+            child.onBeforeRender = (_renderer, _scene, camera, _geometry, material, _group) =>
+            {
+                this.updateObjectUniforms(child, material, camera);
+            };
+        })
+
         console.log("- Scene load time: ", performance.now() - geomLoadTime, "ms");
 
         // Always reset controls based on camera for each load. 
@@ -116,7 +125,6 @@ export class Scene
 
         viewer.getMaterial().clearSoloMaterialUI();
         viewer.getMaterial().updateMaterialAssignments(viewer, "");
-        this.setUpdateTransforms();
     }
 
     //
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:
This looks like a good improvement to me, @hybridherbst, and I'm CC'ing @kwokcb for his expertise on this code.

javascript/MaterialXView/source/viewer.js
```diff
@@ -106,6 +106,15 @@ export class Scene
         }
         scene.add(model);
 
+        // Set up onBeforeRender callbacks so that we can update uniforms per object right before rendering.
+        model.traverse((child) =>
+        {
+            child.onBeforeRender = (_renderer, _scene, camera, _geometry, material, _group) =>
```
kwokcb, line 112: The frequency appears to differ no where it was things that affected view related uniforms would be updated on "dirty".

javascript/MaterialXView/source/viewer.js
```diff
@@ -222,55 +230,30 @@ export class Scene
         orbitControls.update();
     }
 
-    setUpdateTransforms(val=true)
+    updateObjectUniforms(child, material, camera)
```
kwokcb, line 233: Sorry but it's a bit hard to tell based on the code diff what the key change is. 

How does this differ from traversing the scene and updating per object vs adding a per object callback to update ? 

Can you point it out. Thanks.

kwokcb (commented):
Hi @hybridherbst .

This looks like a really good fix to have. 

1) I'm just having trouble see what the key difference that makes the transforms work properly.  

2) Do we need to update very render as opposed to when a view update is required ? (-- I can see you want to naturally update on load).
 
Thanks.

javascript/MaterialXView/source/viewer.js
hybridherbst: Sorry for missing your question here.

When the same material is shared between different objects, the uniforms need to be correct _right before_ rendering a specific object with a specific material. Essentially, right before doing the draw call, the cbuffer setup needs to be correct for the object to be rendered – otherwise model matrices are "reused" and effectively objects render in the wrong place.

So updating object-specific uniforms needs to happen in a callback right before render, not just in the update loop.

javascript/MaterialXView/source/viewer.js
hybridherbst: I'm not sure I understand your question, could you please rephrase it? Thanks

hybridherbst:
The GLB+MTLX in https://github.com/AcademySoftwareFoundation/MaterialX/issues/2508 with and without this PR:

With this PR:
image

Without this PR:
image

I think any kind of mapping that's not 1:1 between objects and materials has this issue.

javascript/MaterialXView/source/viewer.js
kwokcb: Thanks for the explanation. Makes sense now. I thought that it was caching a transform per object to apply.

jstone-lucasfilm:
Good point, @kwokcb, and I agree that this looks safe and worthwhile to include in 1.39.4.

PR #2450, opened 2025-06-11
Add deprecated `fps` input to `time` node

jstone-lucasfilm:
This changelist adds a deprecated `fps` input to the specification and implementations of the `time` node, providing continuity of node signatures for applications using earlier versions of the MaterialX data libraries.

In a future minor version upgrade, we can then formally remove this input, providing version upgrade logic to handle earlier assets.

In a quick survey of `time` node implementations, the `fps` input appears to be ignored, making this proposed approach a reasonable match with industry expectations:

Houdini: https://www.sidefx.com/docs/houdini/nodes/vop/mtlxtime.html
RealityKit: https://developer.apple.com/documentation/shadergraph/application/time-(float)
Three.js: https://github.com/mrdoob/three.js/pull/29149

### DIFF
--- a/documents/Specification/MaterialX.StandardNodes.md
+++ b/documents/Specification/MaterialX.StandardNodes.md
@@ -515,6 +515,7 @@ Standard Application nodes:
 <a id="node-time"> </a>
 
 * **`time`**: the current time in seconds, as defined by the host environment.  This node must be of type float.  Applications may use whatever method is appropriate to communicate the current time to the &lt;time> node's implementation, whether via an internal state variable, a custom input, dividing the current frame number by a local "frames per second" value, or other method; real-time applications may return some variation of wall-clock time.
+    * `fps` (float): an unused input, to be removed in a future specification version.
 
 <br>
 
--- a/libraries/stdlib/genosl/mx_time_float.osl
+++ b/libraries/stdlib/genosl/mx_time_float.osl
@@ -1,4 +1,4 @@
-void mx_time_float(output float result)
+void mx_time_float(float fps, output float result)
 {
     // Use the standard default value if the attribute is not present.
     result = 0.0;
--- a/libraries/stdlib/stdlib_defs.mtlx
+++ b/libraries/stdlib/stdlib_defs.mtlx
@@ -1490,6 +1490,7 @@
     The current time in seconds as defined by the host environment.
   -->
   <nodedef name="ND_time_float" node="time" nodegroup="application">
+    <input name="fps" type="float" value="24.0" /> <!-- An unused input, to be removed in a future specification version -->
     <output name="out" type="float" default="0.0" />
   </nodedef>
 
--- a/source/MaterialXGenMdl/mdl/materialx/stdlib_1_6.mdl
+++ b/source/MaterialXGenMdl/mdl/materialx/stdlib_1_6.mdl
@@ -1584,6 +1584,7 @@ export float mx_frame_float(
 }
 
 export float mx_time_float(
+	uniform float mxp_fps = float(24.0) [[ anno::unused() ]]
 )
 	[[
 		anno::description("Node Group: application")


jstone-lucasfilm:
I'm very interested in discussion of this proposal, and I'll initially CC @ld-kerley, @ashwinbhat, and @hybridherbst for their thoughts.

ld-kerley:
Reintroducing the port - but having it not be functional will still change existing peoples materials - if they were previously using the nodegraph implementation of `time` that respected the `fps` input. Won't it?  That breaks the MaterialX promise.

I also think there are possible offline rendering use-cases where frame is a consistent entity, where it might still be useful to have time defined as a function of frames. 

I think we should support both flavors of `time`, and use the presence of the `fps` input to select between the two, as we do for other nodes.

At Imageworks we had a time based node where the `fps` input could be driven by shot configuration information, so that materials could be portable across shots at different fps.

jstone-lucasfilm:
@ld-kerley That's an interesting point, though my thinking is that each renderer should be free to _choose_ its interpretation of the `time` node based on context.  As seen in Houdini, RealityKit, Omniverse, and Three.js, our suggested `fps` input in MaterialX is ignored, and it's not clear to me that this input is providing any real value.

I wouldn't see any conflict between our parameter-free `time` node and your proposed Imageworks use case, as the renderer has access to shot configuration information at render time, and requiring the authoring artist to bake FPS information into every material seems counterproductive.

jstone-lucasfilm:
I'm additionally CC'ing @ashwinbhat and @dbsmythe for their perspectives, and feel free to include other interested folks in the thread.

dbsmythe:
The original idea of the fps input would be that its value would be **provided by the application**, and would not be baked into an .mtlx file. Otherwise, shaders may not work as expected when imported into a setup running at a different frame rate.

jstone-lucasfilm:
Agreed, @dbsmythe, and I believe the inclusion of the `fps` input in the nodedef for `time` was simply an error in the original ShaderGen implementation for MaterialX.

Fortunately, external applications have omitted this input from their custom implementations of `time` in Houdini, RealityKit, Omniverse, and Three.js, and no support for `time` has ever been available in the MaterialX Viewer, MaterialX Web Viewer, MaterialX Graph Editor, or UsdView in any release that I'm aware of.

This gives us the opportunity to mark the `fps` input as deprecated in 1.39.4, and we can formally remove it through our upgrade system in 1.40.0.

ashwinbhat:
I discussed with @JGamache-autodesk and @zicher3d. We are ok this change and the proposed plan.

ashwinbhat (approved):

jstone-lucasfilm:
Thanks for the review, @ashwinbhat, @JGamache-autodesk, and @zicher3d.

Has the discussion above addressed your earlier concerns, @ld-kerley?  Do folks in the community have any additional concerns that we should discuss before moving forward?

ld-kerley:
If I'm understanding @dbsmythe above, that was the same idea we had at Imageworks.  The renderer (or shader source consumer) isn't always the application that is controlling the dispatch of the scene generation, where things like shot context are present.  For instance Katana knows about shot environment, frame number etc, but the renderer need to know about any of those things, somewhat intentionally.  Obviously you could insist the renderer handle this and feed things back in to the context the shader is running, in my single Imageworks data point, that would be additional engineering.

I think there is a legitimate production use case for both flavors of the time node, and that we should keep support for both.

Though if preferred I'd be open to them existing as two separate nodes, if thats clearer to a user. There is a fundamental different between wall-clock time in a realtime application and the time a frame is rendered within a shot for an offline rendered production.

jstone-lucasfilm:
@ld-kerley From the original design conversations with @dbsmythe, our intent was never for `fps` to be an artist-facing input to the `time` node, and its addition was just an error in the original ShaderGen implementation for MaterialX.

Having two distinct interfaces for `time` would imply that we actually wanted artists to bake their `fps` choices into MaterialX files, which I believe would go against both the original design and current practice for `time` nodes in the industry.

ld-kerley:
Then I'm misunderstanding @dbsmythe's comment above - how does an application provide the fps, if not via an input on the time node?

jstone-lucasfilm:
That's a common misunderstanding, @ld-kerley, and I believe it's the same one that led to the unintended introduction of the `fps` input in the original ShaderGen implementation.

The implementation of the `time` node is fully under the control of the application itself, and this is what allows an offline renderer to provide a frame-based interpretation of time, while a real-time renderer provides its true wall-clock time.

A material that is passed between off-line and real-time renderers should never have `fps` baked into its nodes, allowing the application to provide its best interpretation of time to the material, and providing a consistent visual experience in both cases.

jstone-lucasfilm:
Here's how we originally expressed these ideas in the specification, and we're fully open to ideas for making it clearer:

**time**: the current time in seconds, as defined by the host environment. This node must be of type float. Applications may use whatever method is appropriate to communicate the current time to the <time> node's implementation, whether via an internal state variable, a custom input, dividing the current frame number by a local "frames per second" value, or other method; real-time applications may return some variation of wall-clock time.

jstone-lucasfilm:
Let's move forward with this PR for now, so that we don't leave our codebase in a limbo state between an unintended breakage and its fix.

Feel free to propose additional ideas and improvements, though, and there's still plenty of time for discussion before we mark MaterialX 1.39.4 for release.

PR #2452, opened 2025-06-13
Add `MATERIALX_INSTALL_BIN_PATH` option in CMake

mattyjams:
Hello!

These changes add a `MATERIALX_INSTALL_BIN_PATH` to complement `MATERIALX_INSTALL_LIB_PATH` that allows re-pathing files that would typically be installed into `bin`. For the most part, this really only affects Windows when building shared libraries, but you could also use `MATERIALX_INSTALL_BIN_PATH` to re-path executables like `MaterialXGraphEditor` and `MaterialXView` on other platforms.

For some additional context, we provide pre-compiled binaries for multiple platforms as part of Unreal Engine installations, so we end up wanting a somewhat unusual file layout underneath the installation "root" where libraries are installed into platform-specific directories. Using `MATERIALX_INSTALL_LIB_PATH` supports this already, since both shared and static libraries for Mac and Linux are typically installed in a `lib` directory, but when trying to redirect shared libraries on Windows (`DLL`s), they end up always going into a `bin` directory in the root of the installation. This extra control over where the `bin` directory is located in the MaterialX installation lets us keep all of the platform-specific binaries together straight out of the build without having to post-process it and move things around.

Let me know what you think! Thanks!


ld-kerley:
hey @mattyjams - your PR looks good, but I did have a question regarding how you're using this.  Basically I'm wondering if you also need control the install locations of the data library and resource files, both folders also hold files that I think are required for the two MaterialX binary tools to function correctly.

I was wondering if we should look at finding a clean way to represent that dependency in cmake, so if someone does relocate the binaries the associated library/resource files also get installed to the relocated location so things work transparently, but it would be helpful to understand more about your context here first I think.

mattyjams:
Thanks @ld-kerley!

In our case, we _only_ need to re-path the platform-specific binaries, so for us that's just the shared/static libraries since we don't provide executables. The headers in the `include` directory, and the `libraries` and `resources` directories are shared between platforms and still go at the root of the installation (as specified with `CMAKE_INSTALL_PREFIX`).

So instead of this:
```
<installation root>/
    bin/
    include/
    lib/
    libraries/
    resources/
```

Where the binaries for all of the platforms would inter-mingle and some would likely collide, we're aiming for a layout like this:
```
<installation root>/
    include/
    libraries/
    resources/
    Mac/
        lib/ # universal binaries
    Unix/
        aarch64-unknown-linux-gnueabi/
            lib/
        x86_64-unknown-linux-gnu/
            lib/
    VS2015/ # a weird legacy thing, but implies Windows
        ARM64/
            bin/ # if building shared libraries
            lib/ # either static libraries, or import libraries when building shared
        x64/
            bin/ # similar to ARM64
            lib/
```



If you're setup with access to the Unreal Engine source on GitHub, you can see that layout at least partially here:
https://github.com/EpicGames/UnrealEngine/tree/ue5-main/Engine/Source/ThirdParty/MaterialX/Deploy/MaterialX-1.39.3

And the build scripts we use to produce those build products per-platform here:
https://github.com/EpicGames/UnrealEngine/tree/ue5-main/Engine/Source/ThirdParty/MaterialX


Finding the data libraries at runtime is a bit of a different story. :)

For engine build/distribution reasons we duplicate the `libraries` directory from the installation into a different location, but otherwise we direct MaterialX and OpenUSD to find them in code, which is unfortunately in two places at the moment. Our "direct" MaterialX import code gives the path directly to `loadLibraries()`:
https://github.com/EpicGames/UnrealEngine/blob/ue5-main/Engine/Plugins/Interchange/Runtime/Source/Import/Private/MaterialX/MaterialXUtils/MaterialXManager.cpp#L473

While the OpenUSD side uses the `PXR_MTLX_STDLIB_SEARCH_PATHS` environment variable
https://github.com/EpicGames/UnrealEngine/blob/ue5-main/Engine/Plugins/Runtime/USDCore/Source/UnrealUSDWrapper/Private/UnrealUSDWrapper.cpp#L1116

In case it's interesting, there was a little bit of chatter about this on the ASWF Slack a while ago:
https://academysoftwarefdn.slack.com/archives/C02HJH53RN3/p1739503106645029

I'm not sure that there's anything more we'd need of MaterialX itself here, although I'd be interested to hear any ideas you might have around keeping the data libraries better "connected" to the compiled libraries.

I also saw https://github.com/PixarAnimationStudios/OpenUSD/commit/983e2d9601323bbe6634610266af28c056617cae go by recently which we might be able to take advantage of, but I haven't looked into it yet to see whether we could make the relative pathing work such that we can link up `usdMtlx` to where we put MaterialX's data libraries. That also wouldn't change anything for direct use of MaterialX either, of course.

PR #2454, opened 2025-06-16
Fix race in format settings for multithreaded codegen

ppenenko:
When testing [parallel MaterialX codegen in Storm](https://github.com/PixarAnimationStudios/OpenUSD/pull/3567), and comparing the generated code between test runs using a [persistent MaterialX cache](https://github.com/PixarAnimationStudios/OpenUSD/pull/3661), I noticed that the results were varying around `float` literal formatting. It turned out to be due to a data race on `static` variables controlling the formatting settings.

This is similar to #2378 but limited to multithreaded codegen.

Waiting on the CLA approval before I promote this draft to a full PR.

### DIFF
--- a/source/MaterialXCore/Value.cpp
+++ b/source/MaterialXCore/Value.cpp
@@ -14,11 +14,11 @@
 MATERIALX_NAMESPACE_BEGIN
 
 Value::CreatorMap Value::_creatorMap;
-Value::FloatFormat Value::_floatFormat = Value::FloatFormatDefault;
-int Value::_floatPrecision = 6;
 
 namespace
 {
+thread_local Value::FloatFormat _floatFormat = Value::FloatFormatDefault;
+thread_local int _floatPrecision = 6;
 
 template <class T> using enable_if_mx_vector_t =
     typename std::enable_if<std::is_base_of<VectorBase, T>::value, T>::type;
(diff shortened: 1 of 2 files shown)


source/MaterialXCore/Value.cpp
```diff
@@ -14,11 +14,11 @@
 MATERIALX_NAMESPACE_BEGIN
 
 Value::CreatorMap Value::_creatorMap;
-Value::FloatFormat Value::_floatFormat = Value::FloatFormatDefault;
-int Value::_floatPrecision = 6;
 
 namespace
 {
+thread_local Value::FloatFormat _floatFormat = Value::FloatFormatDefault;
+thread_local int _floatPrecision = 6;
```
ppenenko, line 22: Previously, these were `static` data members of the `Value` class, mutated during codegen by `ScopedFloatFormatting`.

Making them `thread_local` fixed the data race, but they couldn't remain members because the class is marked as `__declspec(dllexport)`  and TLS variables can't be exported. So, moving them to file scope.

I'm not quite sure what the naming convention for file-scoped variables is.

jstone-lucasfilm:
Good catch, @ppenenko, and I'm fine with your newly proposed naming convention for thread-local variables.

Once you've worked through the EasyCLA process, I'd support the idea of moving forward with this change.

ppenenko:
Passed EasyCLA.

jstone-lucasfilm (approved):
Thanks for the fix, @ppenenko!

PR #2455, opened 2025-06-18
Static analysis fixes

jstone-lucasfilm:
This changelist addresses a handful of static analysis warnings flagged by PVS-Studio:

- Resolve two conflicting declarations of `_resourceBindingCtx` in `VkShaderGenerator` and `WgslShaderGenerator`.  Only a single declaration of this variable is needed, as `WgslShaderGenerator` inherits from `VkShaderGenerator`.
- Remove two unneeded null checks in `CgltfLoader::load`.  The variables checked in this logic are non-null by definition, so the null checks can never be hit.
- Replace `insert` with `emplace` in a handful of Graph methods, in situations where clarity and efficiency are improved by this usage.

Additionally, this changelist removes an unused variable name in the local `emitInputAnnotations` function within MdlShaderGenerator.cpp, as naming unused variables triggers a warning in the very latest version of MSVC, and GitHub Actions is beginning to bring this new MSVC version online.


source/MaterialXGenMdl/MdlShaderGenerator.cpp
```diff
@@ -700,7 +700,7 @@ ShaderPtr MdlShaderGenerator::createShader(const string& name, ElementPtr elemen
 namespace
 {
 
-void emitInputAnnotations(const MdlShaderGenerator& _this, ConstDocumentPtr doc, const ShaderPort* variable, ShaderStage& stage)
+void emitInputAnnotations(const MdlShaderGenerator& _this, ConstDocumentPtr, const ShaderPort* variable, ShaderStage& stage)
```
lfl-eholthouser, line 703: These changes looks good to me @jstone-lucasfilm though I'm a bit confused by the removal of `doc` was that due to warning as well?

lfl-eholthouser (approved):

source/MaterialXGenMdl/MdlShaderGenerator.cpp
jstone-lucasfilm: Good catch, @lfl-eholthouser, and I've added a sentence to describe the motivation for that update.  The latest version of MSVC emits warnings for unused variables in local functions, and GitHub Actions is just starting to bring that version of MSVC online.

PR #2458, opened 2025-06-21
Optimization to mx_ggx_smith_G2

jstone-lucasfilm:
This changelist implements an optimization to the `mx_ggx_smith_G2` function in GLSL, replacing two divides with four multiplies.

Validation of this proposed optimization was conducted by rendering the `standard_surface_default.mtlx` example in the MaterialX Viewer on an NVIDIA RTX A6000 at 4K resolution, with the directional albedo mode set to Monte Carlo, and this change reduces the total frame time from 72ms to 70ms.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -78,7 +78,7 @@ float mx_ggx_smith_G2(float NdotL, float NdotV, float alpha)
     float alpha2 = mx_square(alpha);
     float lambdaL = sqrt(alpha2 + (1.0 - alpha2) * mx_square(NdotL));
     float lambdaV = sqrt(alpha2 + (1.0 - alpha2) * mx_square(NdotV));
-    return 2.0 / (lambdaL / NdotL + lambdaV / NdotV);
+    return 2.0 * NdotL * NdotV / (lambdaL * NdotV + lambdaV * NdotL);
 }
 
 // Rational quadratic fit to Monte Carlo data for GGX directional albedo.


jstone-lucasfilm:
I'm attaching a new set of GLSL/OSL render comparisons with the code in this PR, which look identical to render comparisons before the change:

[MaterialXRenderTests_06_21_2025_GitHub.pdf](https://github.com/user-attachments/files/20847882/MaterialXRenderTests_06_21_2025_GitHub.pdf)

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
```diff
@@ -78,7 +78,7 @@ float mx_ggx_smith_G2(float NdotL, float NdotV, float alpha)
     float alpha2 = mx_square(alpha);
     float lambdaL = sqrt(alpha2 + (1.0 - alpha2) * mx_square(NdotL));
     float lambdaV = sqrt(alpha2 + (1.0 - alpha2) * mx_square(NdotV));
-    return 2.0 / (lambdaL / NdotL + lambdaV / NdotV);
+    return 2.0 * NdotL * NdotV / (lambdaL * NdotV + lambdaV * NdotL);
```
kwokcb, line 81: Looks good to me. 

Just a quick question -- do we need to handle divide by 0 ?
`(2.0 * NdotL * NdotV) / max(epsilon, lambdaL * NdotV + lambdaV * NdotL)`

kwokcb (approved):

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
jstone-lucasfilm: Good question, and our current approach is to individually clamp the `NdotV` and `NdotL` terms to epsilon in higher-level functions, so that lower-level functions don't need to repeatedly do this work.

PR #2381, opened 2025-05-13
Add Hextile implementation to MDL backend

msuzuki-nvidia:
This PR adds support for Hextile nodes to the MDL backend.

Nodes to add:
- `<hextiledimage_color3>`
- `<hextiledimage_color4>`
- `<hextilednormalmap_vector3>`

PR #2460, opened 2025-06-25
Revive mx_constant_float() in MDL backend

msuzuki-nvidia:
Revive `mx_constant_float()` in MDL backend that had been accidentally deleted by #2381.


jstone-lucasfilm (approved):
Thanks for the fix, @msuzuki-nvidia!

PR #2462, opened 2025-06-25
Order variable blocks by name to fix a diff between platforms

ppenenko:


source/MaterialXGenShader/ShaderStage.h
```diff
@@ -50,8 +51,9 @@ extern MX_GENSHADER_API const string PIXEL;
 class VariableBlock;
 /// Shared pointer to a VariableBlock
 using VariableBlockPtr = std::shared_ptr<VariableBlock>;
-/// Shared pointer to a map between string identifiers and VariableBlocks
-using VariableBlockMap = std::unordered_map<string, VariableBlockPtr>;
+/// Shared pointer to a map between string identifiers and VariableBlocks.
+/// The order may affect the order of the respective definitions in the generated code.
+using VariableBlockMap = std::map<string, VariableBlockPtr>;
```
ppenenko, line 56: This `std::unordered_map` is iterated over in [GlslShaderGenerator::emitUniforms()](https://github.com/AcademySoftwareFoundation/MaterialX/blob/84984f0abeacacebdb8da58c7426bced3d005521/source/MaterialXGenGlsl/GlslShaderGenerator.cpp#L264) during codegen, so the iteration order affects the order of private and public uniform blocks in the generated code.

The iteration order of a `std::unordered_map` is not guaranteed by the C++ standard, but in practice happens to be deterministic for each given platform. It is does however vary between the platforms that MaterialX and USD are built for.

In particular, [the `testHdStMaterialXShaderGen` USD tests](https://github.com/PixarAnimationStudios/OpenUSD/blob/ceab7e6541c02c9376a03de84bb1174e95e7264e/pxr/imaging/hdSt/CMakeLists.txt#L2884) can't be enabled across all platforms, because they share baseline shader files between platforms. Comparing local test results with the baselines is also complicated because of these diffs.

Overall, these diffs come up as false positives when diagnosing run-time codegen nondeterminism issues such as #2454 and #2378.

Issue #2437, opened 2025-06-04
MaterialXView: Toggling boolean input on and then back off does not revert the value


hybridherbst:
![Image](https://github.com/user-attachments/assets/2da57333-0d66-441f-bba2-3bb3c9f53dee)

# Reproduction steps

1. Download [MyKeywordGraph.mtlx.zip](https://github.com/user-attachments/files/20596813/MyKeywordGraph.mtlx.zip)
2. Open in MaterialXView
3. Open "Property Editor"
4. Flip the toggle – note visual output changes, as expected
5. Flip the toggle back – note visual output does not change
6. Continue flipping the toggle – output will not change ever again until reloading the file

**Edit:** This works fine in MaterialXGraphEditor and MaterialXWebViewer.

jstone-lucasfilm:
Whoa, good catch -- that's unexpected!

warrenm:
After doing some spelunking, the cause of this issue seems to be a type mismatch between the value created in the [editor code](https://github.com/AcademySoftwareFoundation/MaterialX/blob/84984f0abeacacebdb8da58c7426bced3d005521/source/MaterialXView/Editor.cpp#L306) and the value type emitted by the Metal shader generator: the editor boxes up the checkbox value as a `float`, which ultimately gets copied into the uniform buffer down in `MslProgram::bindUniformBuffers` around [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/84984f0abeacacebdb8da58c7426bced3d005521/source/MaterialXRenderMsl/MslPipelineStateObject.mm#L1289-L1305). The type of the property in the shader is `bool` rather than `float`. Copying the 4 bytes of float value `1.0f` results in a `0x00` byte at the offset where a boolean is expected, so the value seen by the shader can never become true again after the box is toggled off.

I'm not sure what a comprehensive fix looks like, but boxing the checkbox value into a bool value instead of a float value (on line 306 referenced above) fixes this in MaterialXView on macOS when using the Metal rendering pipeline.

warrenm:
The above fix also seems to work (i.e., not break) on the OpenGL path on macOS: there, the uniform binding code bottoms out in a call to `glUniform1i` rather than `glUniform1f` with the fix applied, with both having the same (expected) effect. I'm not sure how many other configurations need to be tested, but I'll draft a PR with this suggested fix.

PR #2463, opened 2025-06-26
MaterialXView: Treat checkbox-driven uniforms as bool rather than float

warrenm:
This PR seeks to address #2437, wherein toggling a checkbox off and on in the Property Editor causes the corresponding uniform to stop updating, remaining in the "off" state indefinitely thereafter. The proposed fix is to pass the value of the checkbox to shaders as a bool rather than a float.

Tested in MaterialXView on macOS 15.5 (OpenGL and Metal rendering paths).


jstone-lucasfilm:
This fix looks right to me, thanks @warrenm, and I'm CC'ing @kwokcb in case he has additional thoughts.

PR #2464, opened 2025-06-26
Fixes for installing PDB files in debug builds

mattyjams:
Hello!

For our debug builds on Windows, I was noticing that the PDB files that carry the debug information did not appear to be getting installed along with the binaries.

The farthest back I could trace the addition of the PDB handling was from this commit about six years ago, which looks to have been included in [v1.36.3](https://github.com/AcademySoftwareFoundation/MaterialX/releases/tag/v1.36.3) as its first release:
https://github.com/AcademySoftwareFoundation/MaterialX/commit/749bb42bf46fe42238e2c15dcf3e1b71bee6c797#diff-3b1f19b41d696ced243017e7d2eaffe91809d77dbce2e91aa6d4f112dc809d53

The pattern used to assemble the path to the PDB files in the build directory back then looked like this:
```
"${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/MaterialXCore.pdb"
```

And that's effectively still what they look like today. That does not seem to produce the path where they are currently being generated though.

In my builds (using CMake version `3.31.6-msvc6`, currently bundled with the latest Visual Studio), that expression evaluates to this:
```
<build dir>/source/MaterialXCore//MaterialXCore.pdb
```

While a debug build of static libraries actually generates the file here:
```
<build dir>/lib/Debug/MaterialXCore_d.pdb
```

So there are a few problems with the expression currently being used:
- Static builds now appear to create the PDB files under `CMAKE_LIBRARY_OUTPUT_DIRECTORY` (`<build dir>/lib`) as opposed to `CMAKE_CURRENT_BINARY_DIR` (`<build dir>/source/MaterialXCore`).
- `CMAKE_BUILD_TYPE` is not necessarily set in all cases, particularly for "multi-config" generators that can support multiple configs in a single generation pass (as is the case for Visual Studio). The `CONFIG` configuration expression looks to be what we want instead.
  - More detail on this here:
    - https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html
    - https://www.kitware.com/cmake-and-the-default-build-type/
    - https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:CONFIG
- Use of additional options that affect the output library name (in my case, `CMAKE_DEBUG_POSTFIX=_d`) are not accounted for. We can use the `TARGET_FILE_BASE_NAME` generator expression to get at the fully decorated output name.

There's also different behavior between static and shared library builds. With PDB files on Windows, there seems to be a distinction between those that are generated by the linker for use with shared libraries and executables, and those that are generated by the compiler for use with static libraries. CMake provides the `TARGET_PDB_FILE` generator expression to access the path to the former flavor, but there is no equivalent `TARGET_COMPILE_PDB_FILE` for the latter. There is some additional detail and discussion on this here:
https://cmake.org/cmake/help/latest/prop_tgt/PDB_NAME.html
https://gitlab.kitware.com/cmake/cmake/-/issues/16935

So for static builds (`MATERIALX_BUILD_SHARED_LIBS=OFF`), we still need to produce a path ourselves, but the changes mentioned above should now correctly identify where that file lives in the build directory. For shared library builds (`MATERIALX_BUILD_SHARED_LIBS=ON`) and executables, we can just use `TARGET_PDB_FILE`.

Finally, since compiler-generated PDB files are associated with static libraries, we direct those to install into `MATERIALX_INSTALL_LIB_PATH` (usually just `lib`), while linker-generated PDB files are instead directed into `MATERIALX_INSTALL_BIN_PATH` (usually just `bin`).

For reference, my CMake incantation looks roughly like this:
```
# generate
cmake -G "Visual Studio 17 2022" <source location> <options ...>

#build
cmake --build . --config Debug

#install
cmake --install . --config Debug
```


mattyjams:
One additional update in 9ee9dfbe63adc849cd3d963bd5b2fd8236516039, where I wrapped the PDB file handling in `if(MSVC)`, since the generator expressions being used are only available on Windows and their usage emits errors with other toolchains.

jstone-lucasfilm (approved):
This looks like a good set of improvements to me, thanks @mattyjams!

PR #2465, opened 2025-06-26
Fix MDL code gen after removing ClosureSourceCodeNode

krohmerNV:
In #2400 the ClosureSourceCodeNode has been removed and replaced by the regular SourceCodeNode.
For that to work, the SourceCodeNode needs to emit dependent function calls to generate code for the inputs.

### DIFF
--- a/source/MaterialXGenMdl/Nodes/SourceCodeNodeMdl.cpp
+++ b/source/MaterialXGenMdl/Nodes/SourceCodeNodeMdl.cpp
@@ -43,14 +43,6 @@ void SourceCodeNodeMdl::initialize(const InterfaceElement& element, GenContext& 
         throw ExceptionShaderGenError("Can't find nodedef for implementation element " + element.getName());
     }
 
-    // store the output type for special handling during function call emission
-    vector<OutputPtr> outputs = nodeDef->getActiveOutputs();
-    if (outputs.empty())
-    {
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:
This looks good to me, @krohmerNV, and I've added @ld-kerley as a reviewer.

source/MaterialXGenMdl/Nodes/SourceCodeNodeMdl.cpp
```diff
@@ -77,6 +85,13 @@ void SourceCodeNodeMdl::emitFunctionCall(const ShaderNode& node, GenContext& con
     {
         const ShaderGenerator& shadergen = context.getShaderGenerator();
         const MdlShaderGenerator& shadergenMdl = static_cast<const MdlShaderGenerator&>(shadergen);
+
+        if (_outputType.isClosure())
```
ld-kerley: You might want to use the new `nodeOutputIsClosure(node)` function I added when I made the refactor that broke this - instead of storing the output explicitly.  Thats what we're doing in the main sourceCodeNode. 

see [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/Nodes/SourceCodeNode.cpp#L112)

krohmerNV: okay, yes. That's better.

krohmerNV:
Thanks!

PR #2334, opened 2025-04-05
Fix GLSL implementation of BSDF addition

jstone-lucasfilm:
This changelist fixes the computation of throughput in the GLSL implementation of BSDF addition, with a comment describing the motivation for the provided math.

PR #2467, opened 2025-06-27
Optimizations to glTF PBR graph

jstone-lucasfilm:
This changelist implements optimizations to the graph for glTF PBR, replacing `mix` operations on leaf BSDFs with pre-multiplied `add` operations, and removing duplicate computations.

glTF PBR Carpaint: 25ms -> 11ms
glTF PBR Glass: 23ms -> 9ms
glTF PBR Gold: 23ms -> 10ms
glTF PBR Boombox: 11ms -> 4ms


jstone-lucasfilm:
glTF PBR Carpaint: 25ms -> 11ms

GltfPbrCarpaint_Before
GltfPbrCarpaint_After

glTF PBR Glass: 23ms -> 9ms

GltfPbrGlass_Before
GltfPbrGlass_After

glTF PBR Gold: 23ms -> 10ms

GltfPbrGold_Before
GltfPbrGold_After

glTF PBR Boombox: 11ms -> 4ms

GltfPbrBoombox_Before
GltfPbrBoombox_After

pablode:
Just tried this new version of the glTF PBR with guc's test suite (that uses USD Storm), and I'm getting some shading differences in some tests (the commit before works fine). I tried to reproduce the issue with MaterialXView, but was not successful. Not sure what to make of it, just wanted to bring it up.

|   | Before | After |
| ---  | --- | --- |
| Material_00 | Material_00_mtlx_ref |  Material_00_mtlx |

USD+Mtlx file: [Archive.zip](https://github.com/user-attachments/files/20959570/Archive.zip)

Simplified .mtlx file that does not crash viewer: [Material_00.mtlx.zip](https://github.com/user-attachments/files/20959578/Material_00.mtlx.zip)

jstone-lucasfilm:
@pablode I'm sure you've already run this test, but here are the results that I get for the example above in the MaterialX Viewer, before and after the optimization in this PR.

Before:

PabloExample_Before

After:

PabloExample_After

As far as I can tell, these images are a pixel-perfect match, with the exception of the FPS display.

I'm definitely interested in the visual difference that you're seeing in Hydra/Storm, but so far I'm at a loss to explain the source of it.

jstone-lucasfilm:
@pablode One thought on the difference that you're seeing in Hydra/Storm but not in the MaterialX Viewer:

In MaterialX v1.39.4, we've fixed three incorrect computations for the throughput of the BSDF [multiply](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2328), [add](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2334), and [layer](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2368) nodes, and it seems likely that recent optimizations to the OpenPBR Surface and glTF PBR graphs will require that this math be correct.

Could it be that you'll see matching visuals if you fully integrate MaterialX v1.39.4 into your Hydra/Storm build, rather than selectively updating just the glTF PBR graph?

proog128 (approved):
The changes look good to me, it's a nice improvement.

Depending on how a renderer implements layering, I could imagine there might be differences in edge cases, for example if thin film and transmission intensity is between 0 and 1. Just to be on the safe side, did you check the MDL output in Iray or the MDL SDK renderer?

jstone-lucasfilm:
@proog128 I haven't yet had an opportunity to test MDL rendering with this optimization, and I'm CC'ing @krohmerNV in case he's in a good position to give this a try.

I seem to recall that there are edge cases where BSDF operators in MDL are still in progress, and Kai would have the best sense as to whether this category of optimization would be problematic in that context.

jstone-lucasfilm:
To provide additional support for the continuity of GLSL and OSL renders, here are GLSL/OSL render comparisons for the glTF Boombox material, shown before and after the current optimization.

glTF Boombox Before:

GltfPbr_Boombox_Before

glTF Boombox After:

GltfPbr_Boombox_After

jstone-lucasfilm:
Any thoughts from @niklasharrysson, @kwokcb, or @krohmerNV before we proceed with this optimization?

niklasharrysson (approved):
The performance improvements are very impressive, so I understand the wish to merge these changes in.

However, I share the concerns others have raised in the OpenPBR PR for the same change — that this makes the shading model design more difficult to understand.

I know you've put significant effort into finding a better solution with the same performance improvements, but so far, no straightforward alternative has emerged. So perhaps this is the best solution for now, to get the performance gains in for the next release.

In that context, it might be worthwhile to document and explain this choice of design in the graph? As well as maybe open a new issue to log this as a future improvement, to review this design later and see if a better solution can be found. Perhaps by having the code generator, or some other transformation, take care of this. And only for the targets where it makes a difference.

jstone-lucasfilm:
Thanks @niklasharrysson, and I'll write up a linked GitHub Issue so that our discussion of automating this class of optimizations doesn't get lost to history.

It's worth noting that an automated version of the optimization in this PR would need to recognize two important facts about BSDF graphs:

- Because `mix` and `layer` nodes are linear operations, they can be reordered as long as the author is careful to preserve the exact visual meaning of the graph.
- A `mix` of two leaf BSDF nodes can be refactored as an `add` of two pre-multiplied BSDF nodes, as long as the author is careful to preserve the exact visual meaning of the graph.

Writing automatic refactoring logic that understands and implements these two ideas is definitely possible, but my sense is that it's a significant research project for a knowledgeable MaterialX developer.

Perhaps writing this up as a GitHub Issue will inspire someone to take this project on!

jstone-lucasfilm:
The call to action has been posted!

https://github.com/AcademySoftwareFoundation/MaterialX/issues/2480

Let's continue the discussion of this category of optimization there, and feel free to add any details from related discussions that I may have missed.

PR #2475, opened 2025-07-07
Fixed a typo in GLSL rotation matrix used in hextiling

tdavidovicNV:
There is a typo in the GLSL rotation matrix used in hextiling. The 1st row, 2nd column should be `-a.z` instead of `-a.x`.
See `mx_rotate_vector3.glsl` for the other axis rotation matrix implementation (or [`Rotation matrix from axis and angle` ](https://en.wikipedia.org/wiki/Rotation_matrix#Rotation_matrix_from_axis_and_angle) on wikipedia)

### DIFF
--- a/libraries/stdlib/genglsl/lib/mx_geometry.glsl
+++ b/libraries/stdlib/genglsl/lib/mx_geometry.glsl
@@ -34,7 +34,7 @@ mat3 mx_axis_rotation_matrix(vec3 a, float r)
     float c = cos(r);
     float omc = 1.0 - c;
     return mat3(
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks for the proposed fix, @tdavidovicNV, and I'm CC'ing @msuzuki-nvidia to validate that this matches his intent.  If this change is included, then we should make sure to align the MDL implementation of this node.

msuzuki-nvidia:
yes this was a typo, thanks @tdavidovicNV!

[MDL implementation](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenMdl/mdl/materialx/stdlib_1_6.mdl#L2370) seems correct.

PR #2476, opened 2025-07-07
Use shader code token replacement instead of #ifdef HW_SEPARATE_SAMPLERS 

scotbrew:
Produce cleaner generated shader code by removing use of #ifdef HW_SEPARATE_SAMPLERS to handle both combined and separate texture samplers.  This restores the generated shader code to look like it did before adding the ability to handle separate texture + sampler (for WGSL).

Methodology:
* Swaps out use of #ifdef HW_SEPARATE_SAMPLERS with token replacement.

Impact: 
* Improves readability of shader code
* Reduces number of lines of shader code
* Makes standard path (combined sampler2D) almost a pass-through
* Adds new tokens to the shader template code.

Shader code now uses these tokens instead of direct values:
    $envIrradiance
    $envIrradianceSampler2D
    $envRadiance
    $envRadianceSampler2D
    $texSamplerSampler2D
    $texSamplerSignature

The tokens values are specified in HwShaderGenerator.cpp for the general case and then overridden in WgslShaderGenerator.cpp for separating out the texture and sampler.

### DIFF
--- a/source/MaterialXGenShader/HwShaderGenerator.cpp
+++ b/source/MaterialXGenShader/HwShaderGenerator.cpp
@@ -74,14 +74,14 @@ const string T_ALPHA_THRESHOLD                = "$alphaThreshold";
 const string T_NUM_ACTIVE_LIGHT_SOURCES       = "$numActiveLightSources";
 const string T_ENV_MATRIX                     = "$envMatrix";
 const string T_ENV_RADIANCE                   = "$envRadiance";
-const string T_ENV_RADIANCE_TEXTURE           = "$envRadiance_texture";
-const string T_ENV_RADIANCE_SAMPLER           = "$envRadiance_sampler";
+const string T_ENV_RADIANCE_SAMPLER2D         = "$envRadianceSampler2D";
 const string T_ENV_RADIANCE_MIPS              = "$envRadianceMips";
 const string T_ENV_RADIANCE_SAMPLES           = "$envRadianceSamples";
-
 const string T_ENV_IRRADIANCE                 = "$envIrradiance";
-const string T_ENV_IRRADIANCE_TEXTURE         = "$envIrradiance_texture";
-const string T_ENV_IRRADIANCE_SAMPLER         = "$envIrradiance_sampler";
+const string T_ENV_IRRADIANCE_SAMPLER2D       = "$envIrradianceSampler2D";
+const string T_TEX_SAMPLER_SAMPLER2D          = "$texSamplerSampler2D";
+const string T_TEX_SAMPLER_SIGNATURE          = "$texSamplerSignature";
+
 const string T_ENV_LIGHT_INTENSITY            = "$envLightIntensity";
 const string T_ENV_PREFILTER_MIP              = "$envPrefilterMip";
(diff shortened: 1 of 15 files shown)


source/MaterialXGenShader/HwShaderGenerator.cpp
```diff
@@ -74,14 +74,14 @@ const string T_ALPHA_THRESHOLD                = "$alphaThreshold";
 const string T_NUM_ACTIVE_LIGHT_SOURCES       = "$numActiveLightSources";
 const string T_ENV_MATRIX                     = "$envMatrix";
 const string T_ENV_RADIANCE                   = "$envRadiance";
-const string T_ENV_RADIANCE_TEXTURE           = "$envRadiance_texture";
```
scotbrew, line 77: Removing now unneeded tokens that were introduced in the previous PR for separate texture + sampler using #ifdef.

source/MaterialXGenShader/HwShaderGenerator.cpp
```diff
@@ -74,14 +74,14 @@ const string T_ALPHA_THRESHOLD                = "$alphaThreshold";
...
-const string T_ENV_RADIANCE_SAMPLER           = "$envRadiance_sampler";
+const string T_ENV_RADIANCE_SAMPLER2D         = "$envRadianceSampler2D";
 const string T_ENV_RADIANCE_MIPS              = "$envRadianceMips";
 const string T_ENV_RADIANCE_SAMPLES           = "$envRadianceSamples";
-
 const string T_ENV_IRRADIANCE                 = "$envIrradiance";
-const string T_ENV_IRRADIANCE_TEXTURE         = "$envIrradiance_texture";
-const string T_ENV_IRRADIANCE_SAMPLER         = "$envIrradiance_sampler";
+const string T_ENV_IRRADIANCE_SAMPLER2D       = "$envIrradianceSampler2D";
```
scotbrew, line 81: Verify token names seem reasonable.

scotbrew:
FIXED: Testing identified an issue specific to WGSL code generation regarding binding.  Can reproduce it locally.  Putting together a fix.

Test (local):
`python3 /Users/sbrew/Github/MaterialX/build/installed/python/Scripts/generateshader.py ~/Github/MaterialX/build/bin/resources/Materials/Examples/StandardSurface/standard_surface_brass_tiled.mtlx --target wgsl --validator "/Users/sbrew/bin/glslang --quiet"`

Code areas to review:
* source/MaterialXGenShader/HwShaderGenerator.cpp:366:
* source/MaterialXGenGlsl/WgslResourceBindingContext.cpp:36

jstone-lucasfilm (approved):
I really like this approach to merging the GLSL and WGSL shader code paths, thanks @scotbrew!

I'm CC'ing @niklasharrysson and @kwokcb in case they catch a subtle edge case that I'm missing, but overall I'd say we should aim to include this for 1.39.4.

scotbrew:
For those more familiar with the wider MaterialX codebase and conventions, please review the token names used and also provide any suggestions to improve the inline token documentation in HwShaderGenerator.h.

niklasharrysson (approved):
This looks like an excellent improvement. Thanks @scotbrew

PR #2477, opened 2025-07-07
Replace OSL C++ Material node with data driven implementation

ld-kerley:
Continuing the work of trying to simplify the shader generation and make MaterialX a more data driven system.

Simplification of code export ignores the node classification, which the OSL code generator doesn't need to emit the correct code.  We now simply walk backwards from each of the outputs and emit the code along the DAG, skipping any nodes that have previously been emitted.

This brings us closer to being able to experiment with emitting an OSL command string of precompiled iso nodes.


jstone-lucasfilm:
This looks great to me, thanks @ld-kerley!

I'll defer to @niklasharrysson on the formal review, as he has much greater expertise with this code than I, but I like the idea of including this improvement in 1.39.4.

source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -180,22 +174,16 @@ ShaderPtr OslShaderGenerator::generate(const string& name, ElementPtr element, G
         }
     }
 
-    // Emit all texturing nodes. These are inputs to any
-    // closure/shader nodes and need to be emitted first.
-    emitFunctionCalls(graph, context, stage, ShaderNode::Classification::TEXTURE);
-
     // Emit function calls for "root" closure/shader nodes.
     // These will internally emit function calls for any dependent closure nodes upstream.
```
niklasharrysson: The code comment here is outdated now and could be updated or removed.

niklasharrysson (approved):
This looks like a great improvement an cleanup. Thanks @ld-kerley 

I only added a small note above.

PR #2478, opened 2025-07-08
_options renderTestPaths can accept files, not just directories

tdavidovicNV:
The `renderTestPaths` could only accept directories, and would run all tests within that directory. This modification allows adding `.mtlx` files as well, useful for debugging individual tests.


jstone-lucasfilm (approved):
I've confirmed that I get the same output HTML in local tests, and this looks like an improvement to me.  Thanks @tdavidovicNV!

PR #2483, opened 2025-07-09
Optimizations to Standard Surface graph

jstone-lucasfilm:
This changelist implements optimizations to the graph for Standard Surface, replacing `mix` operations on leaf BSDFs with pre-multiplied `add` operations.  Pre-multiplied `add` operations take better advantage of dynamic branching in hardware shading languages, and should have a neutral or positive impact on software shading languages.

Standard Surface Marble: 3ms -> 1ms
Standard Surface Carpaint: 3ms -> 1ms
Standard Surface Glass: 5ms -> 3ms
Standard Surface Plastic: 2ms -> 1ms
Standard Surface Brushed Metal: 2ms -> 1ms
Standard Surface Chess Set: 9ms -> 4ms

### DIFF
--- a/libraries/bxdf/standard_surface.mtlx
+++ b/libraries/bxdf/standard_surface.mtlx
@@ -226,7 +226,7 @@
       <input name="normal" type="vector3" interfacename="normal" />
     </oren_nayar_diffuse_bsdf>
     <translucent_bsdf name="translucent_bsdf" type="BSDF">
-      <input name="weight" type="float" value="1.0" />
+      <input name="weight" type="float" interfacename="subsurface" />
       <input name="color" type="color3" nodename="coat_affected_subsurface_color" />
       <input name="normal" type="vector3" interfacename="normal" />
     </translucent_bsdf>
@@ -235,7 +235,7 @@
       <input name="in2" type="float" interfacename="subsurface_scale" />
     </multiply>
     <subsurface_bsdf name="subsurface_bsdf" type="BSDF">
-      <input name="weight" type="float" value="1.0" />
+      <input name="weight" type="float" interfacename="subsurface" />
       <input name="color" type="color3" nodename="coat_affected_subsurface_color" />
       <input name="radius" type="color3" nodename="subsurface_radius_scaled" />
       <input name="anisotropy" type="float" interfacename="subsurface_anisotropy" />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Standard Surface Marble: 3ms -> 1ms

StandardSurface_MarbleProcedural_Before
StandardSurface_MarbleProcedural_After

Standard Surface Carpaint: 3ms -> 1ms

StandardSurface_Carpaint_Before
StandardSurface_Carpaint_After

Standard Surface Glass: 5ms -> 3ms

StandardSurface_Glass_Before
StandardSurface_Glass_After

Standard Surface Plastic: 2ms -> 1ms

StandardSurface_Plastic_Before
StandardSurface_Plastic_After

Standard Surface Brushed Metal: 2ms -> 1ms

StandardSurface_MetalBrushed_Before
StandardSurface_MetalBrushed_After

jstone-lucasfilm:
Here's a new set of GLSL/OSL render comparisons with the optimization in this PR:

[MaterialXRenderTests_07_09_2025_GitHub.pdf](https://github.com/user-attachments/files/21146837/MaterialXRenderTests_07_09_2025_GitHub.pdf)

jstone-lucasfilm:
Feedback from the community on this proposal is encouraged, and I'm specifically CC'ing @niklasharrysson, @kwokcb, @krohmerNV, @ld-kerley, and @portsmouth for their thoughts.

jstone-lucasfilm:
Just to give a sense of the impact on more complex production scenes, here's a comparison of the Open Chess Set before and after the optimization in this changelist, using the same GPU and resolution as in the comparisons above.

Open Chess Set: 9ms -> 4ms

OpenChessSet_Before
OpenChessSet_After

libraries/bxdf/standard_surface.mtlx
```diff
@@ -226,7 +226,7 @@
       <input name="normal" type="vector3" interfacename="normal" />
     </oren_nayar_diffuse_bsdf>
     <translucent_bsdf name="translucent_bsdf" type="BSDF">
-      <input name="weight" type="float" value="1.0" />
+      <input name="weight" type="float" interfacename="subsurface" />
```
kwokcb, line 229: Is this an intended part of this change to use the interface value vs 1 ?

jstone-lucasfilm: This is a great question, and indeed it's related to a subtle aspect of the original graph logic that is not immediately obvious.

In the original Standard Surface graph, the `subsurface` input controls the intensity of _both_ the `subsurface_bsdf` and `translucent_bsdf` lobes, because these two BSDFs are mixed together using the thin-walled flag as a selector, and then the resulting blend is mixed with the diffuse term using the `subsurface` input.

In order to accurately capture this behavior in the optimized graph, the `subsurface` input needs to be used as the `weight` of both the `subsurface_bsdf` and `translucent_bsdf` nodes, and the inverse of of the `subsurface` input needs to be used as the `weight` for the diffuse term.

In some ways, the optimized graph clarifies this dual role of the `subsurface` input better than the original graph does, in addition to providing a significant performance win in renderers that have the ability to skip BSDF nodes with a zero weight.

kwokcb: Thanks for the clarification. This all looks good then.

portsmouth:
Looks fine to me, with the same caveat as the OpenPBR change, I.e. this optimization would ideally be part of the code generation eventually.

Issue #2473, opened 2025-07-02
Crash in MaterialXView when loading specific file


hybridherbst:
Follow-up to https://github.com/AcademySoftwareFoundation/MaterialX/issues/2470#issuecomment-3028606504.

Loading the mtlx file from this ZIP https://github.com/user-attachments/files/21022477/MeshAttributes.zip in MaterialXView leads to a crash with the following error message in the latest MaterialX version:
```sh
herbst@Felixs-MacBook-Pro build % ./bin/MaterialXView
-[_MTLCommandEncoder dealloc]:134: failed assertion `Command encoder released without endEncoding'
zsh: abort      ./bin/MaterialXView
```

Curiously, MaterialXGraphEditor does not crash, but also does not display anything (I believe it should display UV0 as color, as it does in MaterialXWebViewer).

warrenm:
This is a consequence of an [exception thrown by `MslProgram` during attribute binding](https://github.com/AcademySoftwareFoundation/MaterialX/blob/87c4a2bcf411a4ae601586220a3f1995ae5344ae/source/MaterialXRenderMsl/MslPipelineStateObject.mm#L374-L375) when it fails to find a required attribute. Since the `MetalRenderPipeline::renderFrame` method is exited abnormally, it doesn't get a chance to call `MTL(endCommandBuffer())` or `MTL(endEncoder());`, which leaves the encoder in an invalid state when it's subsequently cleaned up after the exception handler.

It's not immediately obvious how to fix this elegantly, but one possible approach is to abuse `std::unique_ptr` deleters as a scope guard and "defuse" the guard when it looks like the frame is going to complete successfully:

```diff
diff --git a/source/MaterialXView/RenderPipelineMetal.mm b/source/MaterialXView/RenderPipelineMetal.mm
index 733d92d1..c39bdfae 100644
--- a/source/MaterialXView/RenderPipelineMetal.mm
+++ b/source/MaterialXView/RenderPipelineMetal.mm
@@ -22,6 +22,12 @@
 
 const float PI = std::acos(-1.0f);
 
+template<class F>
+auto finally(F f) noexcept(noexcept(F(std::move(f)))) {
+    auto x = [f = std::move(f)](void*){ f(); };
+    return std::unique_ptr<void, decltype(x)>((void*)1, std::move(x));
+}
+
 }
 
 MetalRenderPipeline::MetalRenderPipeline(Viewer* viewerPtr) :
[...]

PR #2484, opened 2025-07-09
Add guard to clean up Metal objects on exception

warrenm:
This PR addresses the crash reported in #2473. It introduces a scope guard object that automatically ends the current command encoder and command buffer if the `renderFrame` method exits abnormally due to an exception. Without such a precaution, the viewer crashes shortly after the exception is handled, because the encoder is released by its autorelease pool before `endEncoding` is called on it, causing a hard assert in Metal.

### DIFF
--- a/source/MaterialXView/RenderPipelineMetal.mm
+++ b/source/MaterialXView/RenderPipelineMetal.mm
@@ -453,6 +453,11 @@ void MetalRenderPipeline::renderFrame(void* color_texture, int shadowMapSize, co
     }
     
     MTL(beginCommandBuffer());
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This looks good to me, thanks @warrenm.  I'm CC'ing @hybridherbst to make sure this addresses the issue that he encountered, as well as @ld-kerley and @dgovil for visibility.

dgovil:
Good catch. This looks sensible to me!

jstone-lucasfilm (approved):
This looks good to me, thanks @warrenm, and let's plan to include this in MaterialX 1.39.4.

PR #2487, opened 2025-07-11
Update NanoGUI version to support CMake 4+

tdavidovicNV:
The bin2c.cmake requirement for cmake 2.8 caused the NanoGUI compilation to fail to configure with any CMake 4.x.y
The newest NanoGUI fixes this, and the rest of the updates seems rather innocuous.

Without this, MaterialXView does not compile in the latest vscode on Windows (and possibly elsewhere).


jstone-lucasfilm (approved):

PR #2495, opened 2025-07-19
Modernize CMake handling of OIIO dependency

JGamache-autodesk:
By asking CMake to write MATERIALX_BUILD_OIIO in the generated CMake files if MaterialXRender was built with OpenImageIO support.

### DIFF
--- a/source/MaterialXRender/CMakeLists.txt
+++ b/source/MaterialXRender/CMakeLists.txt
@@ -26,7 +26,7 @@ if(UNIX)
 endif()
 
 if(MATERIALX_BUILD_OIIO)
-    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/External/OpenImageIO")
     find_package(OpenImageIO CONFIG REQUIRED)
     target_link_libraries(${TARGET_NAME} PRIVATE OpenImageIO::OpenImageIO OpenImageIO::OpenImageIO_Util)
+    target_compile_definitions(MaterialXGraphEditor PUBLIC MATERIALX_BUILD_OIIO)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,9 +229,6 @@ if(MATERIALX_BUILD_RENDER AND MATERIALX_BUILD_GEN_OSL AND MATERIALX_BUILD_TESTS)
 endif()
 
 # Add global definitions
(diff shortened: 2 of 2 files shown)


source/MaterialXRender/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if(UNIX)
 endif()
 
 if(MATERIALX_BUILD_OIIO)
-    list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/External/OpenImageIO")
```
JGamache-autodesk, line 29: No longer necessary. That directory was removed for 1.39.

source/MaterialXRender/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if(UNIX)
...
     find_package(OpenImageIO CONFIG REQUIRED)
     target_link_libraries(${TARGET_NAME} PRIVATE OpenImageIO::OpenImageIO OpenImageIO::OpenImageIO_Util)
+    target_compile_definitions(MaterialXGraphEditor PUBLIC MATERIALX_BUILD_OIIO)
```
JGamache-autodesk, line 31: Makes sure that any project linking with MaterialXRender, both locally and in external projects, are aware that this MaterialX was built with OIIO support. The info will be written in the generated CMake for this library.

Please note that the `target_link_library` can stay `PRIVATE` since `OiioImageLoader.h` hides the underlying OIIO implementation. It still needs the compile definition to make the class visible.

jstone-lucasfilm (approved):
This looks good to me, thanks @JGamache-autodesk!

source/MaterialXRender/CMakeLists.txt
marcin-serwin, line 31: This breaks building without `MaterialXGraphEditor`:
```
CMake Error at source/MaterialXRender/CMakeLists.txt:31 (target_compile_definitions):
  Cannot specify compile definitions for target "MaterialXGraphEditor" which
  is not built by this project.
```
I think it was meant to say
```suggestion
    target_compile_definitions(MaterialXRender PUBLIC MATERIALX_BUILD_OIIO)
```

jstone-lucasfilm: That looks like a good catch to me, @marcin-serwin.  Would one of you two like to propose that fix in a PR?

PR #2497, opened 2025-07-21
Fix MaterialXRender private dependency to OIIO

JGamache-autodesk:
Also declare downstream dependencies when building static libraries.


jstone-lucasfilm (approved):
Thanks so much, @JGamache-autodesk and @marcin-serwin!

PR #2498, opened 2025-07-22
Remove blur helper functions

jstone-lucasfilm:
Following up on #2411, this changelist removes the blur helper functions in GLSL, MSL, and OSL that are no longer used by the codebase.


jstone-lucasfilm:
For validation, here's a new set of GLSL/OSL renders using the codebase in this PR:

[MaterialXRenderTests_07_22_2025_GitHub.pdf](https://github.com/user-attachments/files/21372589/MaterialXRenderTests_07_22_2025_GitHub.pdf)

ld-kerley:
Thanks for catching some of this clean up I missed.

I think it's safe to remove the constant variables from the C++ shader generator code.

I do wonder If we should leave the shader code in the data library for 1.39.4.  Technically the shader code in the data library is part of the API surface of MaterialX, as downstream custom shader generators could be referencing the files, or other custom defined nodes could be using the includes.  I do think the includes here have methods that could potentially be useful in custom code.

I did check and none of this is directly referenced in OpenUSD, so at least one major downstream consumer shouldn't be affected.  I think, especially as we're so close to the release being cut, that we should leave the shader code and broadcast to the community that we're planning on removing this.  Giving downstream customers more time to refactor any uses they may have.

If others feel strongly that we need to remove this now, then thats ok, I'm just trying to highlight possible places where it might cause some issues.

jstone-lucasfilm:
Good points, @ld-kerley, though I would lean towards the idea of removing shading code that's never referenced, so that we don't end up maintaining legacy code that no longer receives any testing or validation.

If teams need to retrieve this code, it's always available in our GitHub history, and it seems more consistent to remove unused shading code at the same time that we remove all of the C++ code that referenced it.

ashwinbhat:
I'm fine with the cleanup, we should call it out in release notes as there might be integrations that have direct references to `mx_sampling` shader files.

ashwinbhat (approved):

Issue #2480, opened 2025-07-08
Automated optimization of shading model graphs


jstone-lucasfilm:
Two recent PRs to MaterialX demonstrated that shading model graphs can be significantly optimized, especially in the context of real-time and interactive rendering, by applying a set of visually invariant refactor steps:

- https://github.com/AcademySoftwareFoundation/MaterialX/pull/2459 (OpenPBR Surface)
- https://github.com/AcademySoftwareFoundation/MaterialX/pull/2467 (glTF PBR)

It would be ideal if we could implement this category of graph optimization _automatically_, e.g. at the level of our shader code generation, allowing all shading models to benefit without manual analysis or modification of their graphs being required.

So far, our most successful shading model graph optimizations have been based on the following two observations:

- Because BSDF `mix` and `layer` nodes are linear operations, they can be reordered as long as the author is careful to preserve the exact visual meaning of the graph.
- A `mix` of two leaf BSDF nodes can be refactored as an `add` of two pre-multiplied BSDF nodes, as long as the author is careful to preserve the exact visual meaning of the graph.

In combination, these two observations can be used to make a shading model graph more amenable to common renderer optimizations that skip the computations associated with a BSDF node when its `weight` is less than epsilon.

This GitHub Issue is a call for interested developers to propose and implement an automated solution that applies this category of optimizations to shading model graphs.

This project would likely require a non-trivial research and development effort, but the benefits to the community could be enormous, speeding up virtually all shading models that the industry renders in the context of MaterialX or USD.

jstone-lucasfilm:
This has been addressed in #2832.

PR #2499, opened 2025-07-22
Shader Generator ND_mix_bsdf optimization

ld-kerley:
Taking inspiration from #2459, which is a significant win in terms of HW shader performance. Instead of optimizing the node graph directly, here we apply the same idea, but programmatically at shader generation time.

If we meet the following condition.

"A `mix` node that has `BSDF` type inputs, and if those two nodes have `weight` inputs themselves."

Then we can replace the `mix` node with an `add` and modify the `weight` inputs, by multiplying by the `mix` input, and the inverse of the `mix` input.

### Old nodegraph
```mermaid
---
config:
      theme: redux
---
flowchart TD
    weightA{{"weight_A"}}
    weightB{{"weight_B"}}
    mix{{"mix"}}
    BsdfA["BSDF_A"]
    BsdfB["BSDF_B"]
    Mix["Mix"]
    weightA --> BsdfA
    weightB --> BsdfB
    BsdfA --> Mix
    BsdfB --> Mix
    mix --> Mix
```

### New nodegraph
```mermaid
flowchart TD
    weightA{{"weight_A"}}
    weightB{{"weight_B"}}
    mix{{"mix"}}
    Invert["Invert"]
    MultA["Multiply"]
    MultB["Multiply"]
    BsdfA["BSDF_A"]
    BsdfB["BSDF_B"]
    Add["Add"]
    mix --> Invert
    weightA --> MultA
    mix --> MultA
    weightB --> MultB
    Invert --> MultB

    MultA --> BsdfA
    MultB --> BsdfB

    BsdfA --> Add
    BsdfB --> Add
    
```

This allows the BSDF nodes to early out if their contribution is not necessary in the mix.

Taking this approach has a couple of advantages:
* We programmatically match all instances of this possible optimization, not just the few that are hand crafted.
* We can leave the uber shader nodegraphs, in their original state.  The use of the `mix` node better conveys the idea behind the nodegraph.
* Its possible that some backends might be able to optimize the original graph more effectively than the current hand-crafted representation.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -873,6 +873,10 @@ bool ShaderGraph::removeNode(ShaderNode* node)
     {
         input->breakConnection();
     }
+    for (ShaderOutput* output : node->getOutputs())
+    {
+        output->breakConnections();
+    }
     _nodeMap.erase(mapIt);
     _nodeOrder.erase(vecIt);
     return true;


jstone-lucasfilm:
I'm very excited about the future of this work, @ld-kerley, and my main recommendation would be that we consider it a new investment of research and engineering effort for MaterialX 1.39.5, so that we have the time to validate this approach for all MaterialX shading models, from both a performance and visual parity perspective.

Once we're confident that this gives us the same or better performance as https://github.com/AcademySoftwareFoundation/MaterialX/pull/2483, https://github.com/AcademySoftwareFoundation/MaterialX/pull/2459, and https://github.com/AcademySoftwareFoundation/MaterialX/pull/2467, we can replace all of those manual optimizations with this automated approach, with the expectation that other shading models (even custom studio models) will benefit in the same way.

kwokcb:
Hi @ld-kerley ,
What are your thoughts on integrations that don't use shader generation and hence won't ever see / use this optimization.
It feels like this can be a pre-processing step or even a stand-alone optimization utility. 
It can still be integrated here to automatically perform these (or other optimizations) but is not directly tied to code generation.

fpsunflower:
I'm not super familiar with the details of codegen here, but can you explain what is different between the two implementations?

What can the add/mult nodes do that the mix node cannot ?

ld-kerley:
Related to #2480

ld-kerley:
@fpsunflower Its not so much a feature of codegen, as much as function of the HW shader language backends here.  The implementation of the BSDF nodes that accept a weight input are written to early out if the weight is zero.  In the `mix` case the weight is often not zero, but unfolding the mix math and moving the weight to be combined with the weight input to the BSDF node does often end up with zero weights in the BSDF nodes.   In a language like OSL I would expect the OSL optimizer to be able to fold away the unused branches if the mix value is set to 0 or 1. 

@bernardkwok today anyone not using codegen wouldn't have access to the optimization in this form.  We could pre-process nodegraphs during build with this logic if we wanted.  Personally I'm not a fan of optimizing the nodegraphs, when we've only validated it a performance improvement in subset of the backend languages.  

I actually wonder if we should put this optimization behind a code gen option switch to make it easier for downstream consumers to profile either way and make their own decision if they want it.

If we grow the number of these sorts of optimizations, which I hope we do, there are a number of other really valuable ones we used at Imageworks, then I can imagine a world where perhaps downstream consumers could run "code gen" to generate an optimized version of the graph - but not go all the way to shader code.  I think this might actually be an important part of any future overhaul of the code gen system, to centralize the graph based construction and optimization, and allow an "exit-point" here.  But this is all me just shooting from the hip.

fpsunflower:
Ah I see now -- the issue is the mix happens _after_ the BSDFs are already evaluated, so there's no chance to short-circuit them. In the new layout they can be. This makes sense.

ld-kerley:
Just for some more context - Slack conversations with Kai Rohmer [here](https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1756473253774029), suggest that at least in MDL this optimization might not be ideal. So we should refactor things here to be constrained to the HW shading languages for now.

ld-kerley:
I've added a GenOption based option to toggle this behavior on and off.  It's also enabled for the GLSL/MSL code generators used in MaterialXView and MaterialXGraphEditor.

Before this is merged the specialized versions of the graph should be removed from (libraries/bxdf/genglsl) - but I'm leaving them here for now to allow easier A/B testing between the modified nodegraph approach and the programmatic approach in this PR.

jstone-lucasfilm:
@ld-kerley Following up on our last MaterialX TSC meeting, I'd love to bring this PR back into focus again, so that we can make it available as a platform for #2680 and further generator optimizations in the future.

When you have a moment, would you mind resolving the handful of merge conflicts in `MaterialXGenShader`?

I can then take the next step and compare the performance of these latest generator optimizations against the manually optimized shading model graphs, and we can work through any minor differences that may remain.

ld-kerley:
I rebased and pushed - and re-ran the test suite locally - but didn't have time to validate the optimizations again - I think they should still be valid, but it would be really interesting to find a way to add a unit test to verify they are applied correctly.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -1049,8 +1065,382 @@ void ShaderGraph::optimize(GenContext& context)
...
+    // replace those with the outgoing connection from the new add node.
+    auto mixNodeOutput = mix2Node->getOutput("out");
+    auto mixNodeOutputConns = mixNodeOutput->getConnections();
+    for (auto conn : mixNodeOutputConns)
+    {
+        addNode->getOutput("out")->makeConnection(conn);
+    }
+
+    removeNode(mix1Node);
+    removeNode(mix2Node);
```
ppenenko, line 1442: This crashes for me (VS2022 Debug, when starting MaterialXView) n `ShaderInput::breakConnection()` and I think https://github.com/ld-kerley/MaterialX/pull/3/files is the fix.

ld-kerley: Not quite sure why this doesn't crash for me - but I agree it's important to ensure all the connections are severed. I merged your fix in to this PR

source/MaterialXGraphEditor/RenderView.cpp
```diff
@@ -162,6 +162,7 @@ RenderView::RenderView(mx::DocumentPtr doc,
     // Make sure all uniforms are added so value updates can
     // find the corresponding uniform.
     _genContext.getOptions().shaderInterfaceType = mx::SHADER_INTERFACE_COMPLETE;
+    _genContext.getOptions().optReplaceBsdfMixWithLinearCombination = true;
```
jstone-lucasfilm, line 165: In order to match the existing behavior in MaterialX exactly, I'd recommend that we enable this flag _by default_ in our hardware shader generators, rather than only enabling it in the Viewer and Graph Editor.

source/MaterialXGenShader/GenOptions.h
```diff
@@ -204,6 +205,10 @@ class MX_GENSHADER_API GenOptions
     /// inside the bitangent node.
     bool hwImplicitBitangents;
 
+    /// Analyse the graph of ShaderNodes and replace any ND_mix_bsdf nodes
+    /// with a linear combination of their weighted inputs
+    bool optReplaceBsdfMixWithLinearCombination;
```
jstone-lucasfilm, line 210: Since many of our existing generator options also influence performance, I'd recommend omitting the `opt` prefix here, and just naming the flag `replaceBsdfMixWithLinearCombination`.

ld-kerley: I think we're going to be adding a number of additional "optimization" steps - so I think it would be helpful to catagorize them as such.  Especially as it hints as where the code lives in the shader generator as well.

Are there other existing options that are really steps that optimize the generated material in some way? or are they more options where performance level can be tuned to user need?

source/MaterialXGraphEditor/RenderView.cpp
ld-kerley: I'm not sure we have a clean way in the API to set specific option defaults for specific shader generators. We could hard code it for certain generators all the time, but that would then disallow users from disabling the optimization. 

I'm open to suggestions on where to put this - or perhaps we should just merge this and log an issue to revisit how to enable this by default in certain shader generators as a separate task. 

I was looking at the GenOptions class the other day - and it struck me that it would be nice to have generator specific sub-classes. There are a lot of options that don't apply to the OSL generator for instance, and some new ones that only apply to the OSL generator. It might be nice to refactor things to make the API surface clearer about which options apply to which generators.

niklasharrysson: +1 for sub-classing the GenOptions per generator. Most of the options are HW specific and irrelevant for path tracers, and there are now OSL specific ones as well. Would be nice to refactor this for a better separation of concerns.

source/MaterialXGenShader/GenOptions.h
niklasharrysson: One that fits this category is `elideConstantNodes`. An optimization to remove any nodes classified as constants, which is also done in ShaderGraph::optimize().

ld-kerley: we should make a note to prefix this with `opt` as well at some point when we can break the interface then.

source/MaterialXGraphEditor/RenderView.cpp
ld-kerley: When I looked really quick I couldn't see a nice way to do this without breaking the API - we might need to wait until we refactor the shader gen interface - unless someone else sees a nice path.

source/MaterialXGenShader/GenOptions.h
jstone-lucasfilm: I'm not sure an `opt` prefix for all optimization-related flags is a good naming convention, since so many of our shader generation flags influence performance, but also have subtle impacts on visual fidelity and other properties of rendering.

For now, I'd recommend following the pattern in @niklasharrysson's example of `elideConstantNodes`, where the name of the flag compactly describes what it does, without aiming to group all of our flags into optimization, visual fidelity, or other categories.

source/MaterialXGraphEditor/RenderView.cpp
jstone-lucasfilm: I like the idea of subclassing `GenOptions` in the future, but we'll need a short-term solution as well, so that we're not reducing real-time rendering performance for external clients with our next release of MaterialX.

As one initial suggestion, what about adding a virtual `updateGenOptions` method to the `ShaderGenerator` class, which would then be invoked in the constructor for `GenContext`?

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/GenContext.cpp#L15

The base implementation of the virtual method would be empty, and the implementation for `HwShaderGenerator::updateGenOptions` would set the new flag that you've added in this PR.

ld-kerley: maybe this is best added as a separate PR?

source/MaterialXGenShader/GenOptions.h
ld-kerley: Thats exactly why I feel like we should have the `opt` prefix - as it is not a performance related selection that affects visual fidelity - but purely an optimization that should lead to an identical image (modulo some floating point precision due to changing in operation order). 

This prefix is intended to indicate options that should be safe to enable/disable without any change to the image - and thus encourage integrators to experiment freely to find the choices that are most optimal for their platform.

source/MaterialXGraphEditor/RenderView.cpp
jstone-lucasfilm: That sounds fine to me!  We'll just need to make that change before we fully remove the target-specific graphs for OpenPBR Surface, Standard Surface, and glTF PBR, but both of those can be future PRs.

source/MaterialXGenShader/GenOptions.h
jstone-lucasfilm: I see your perspective here, but that categorization also applies to existing flags such as `elideConstantNodes` and `shaderInterfaceType`, where there's a tradeoff between performance and some non-visual aspect of rendering.

In all of these cases, the interplay between performance, shading language feature set, and shader reusability is a complex one, and a singular categorization such as 'opt` seems too broad of a brush to paint them with.

My recommendation would be to follow the current naming convention with this new feature, and we can take a step back once the landscape of optimizations has evolved, seeing whether meaningful patterns across the set of shader generation options become apparent.  With that clearer perspective, we may even choose to break `GenOptions` into multiple classes, based on the categorizations that we find to be most useful in real-world workflows.

source/MaterialXGraphEditor/RenderView.cpp
niklasharrysson: (not for this PR but for the next change to update GenOptions)

> As one initial suggestion, what about adding a virtual updateGenOptions method to the ShaderGenerator class, which would then be invoked in the constructor for GenContext?

I like that idea. A possible variation of that is to instead have a virtual `createOptions()` method, called at the same place in GenContext constructor, to both create the GenOptions instance and set it's default values for a particular shader generator. This would then support creating subclasses of GenOptions in the future.

source/MaterialXGenShader/GenOptions.h
ppenenko: FWIW, I would also vote for introducing the `opt` prefix, especially because more optimizations are coming. E.g. #2680 would require something like `optLobePruning`, and if we implement an optimization pass manager, we'll need something like `optMaxPassIterations` to limit the iterations if it doesn't converge. So it would be helpful to group these options.

jstone-lucasfilm: @ppenenko That's a completely valid perspective, but since we already have similar optimization flags in our GenOptions structure (e.g. `elideConstants`, `shaderInterfaceType`), it may make more sense to continue with the current naming convention, and then apply a _global_ change once we've agreed upon a new pattern.  That's the usual approach that we take in MaterialX, where naming conventions are not applied piecemeal as new features are added, but are applied universally based on a new design decision.  By the time we reach that future design decision, we may actually decide to move subsets of our `GenOptions` structure into multiple _structures_ for hardware options, OSL options, MDL options, and language-independent generation controls, making the existing prefixes unnecessary.

Since @niklasharrysson is one of the original architects of this system, I'd be interested in his thoughts on how to approach the organization of our `GenOptions` controls, both in the short term of this PR and in the longer term, where more significant design decisions might  be made.

ld-kerley: FWIW I don't really think `shaderInterfaceType` is an optimization that would be given the `opt` prefix.  To me the `opt` prefix designates an optimization that only restructures things internal to the shader, and doesn't change the rendered artifact (ignoring floating point operation order differences), or change the exposed shader interface. 

I agree with @ppenenko here, and think we may even need a system to manage these optimizations further, so I think having the `opt` prefix would make it easy to communicate to users that their settings may even override by a more complex optimization manager at some point, and its possible we get to that point before we're ready for the large (breaking change) refactor.

jstone-lucasfilm: This is a really worthwhile discussion, and I'm CC'ing @niklasharrysson for his expertise in this domain, as I feel that his perspective will be valuable here.

My own thinking is that the proposed `opt` prefix takes too narrow of a view of what _optimization_ means in the context of our `GenOptions` structure, since nearly all of our ShaderGen flags are making some form of _tradeoff_ that influences performance, exchanging it for other qualities such as visual accuracy, shader instruction count, flexibility when shader inputs are modified, or the set of platforms on which the generated shader is appropriate.

Because each flag involves a subtle tradeoff involving multiple competing goals, it seems more appropriate to document the tradeoff for each flag clearly, rather than broadly grouping our flags into "optimization" and "non-optimization".  An example of this pattern can be seen for the recently-added [hwAiryFresnelIterations](https://github.com/AcademySoftwareFoundation/MaterialX/blob/834edcf474ba0fc2e263c4ec3d943cfa1eaea5f2/source/MaterialXGenShader/GenOptions.h#L166C18-L166C41), which we document in the following way:

```
    /// Sets the number of iterations for Airy Fresnel reflection calculations.
    /// Higher values provide more accurate thin-film interference patterns
    /// but increase computational cost. Defaults to 2.
```

To my mind, that's the template that I'm hoping we can follow for future `GenOptions` flags that influence performance, where we document the precise tradeoff that the client is making when they modify it.

niklasharrysson: I don’t have a strong preference on this. The main thing is that the options are clearly documented and accurately describe what they do.

It's a great intent to try to group and make the options more clear by prefix. But it might be hard to keep the grouping clean over time when options span multiple categories. For example, a future optimization that’s specific to HW rendering, or OSL rendering, where we already have established groupings.

Also, a quick note: we’ve already mentioned refactoring GenOptions down the line. Given that, it may not be worth spending too much time and energy on tightening naming conventions in the current GenOptions.

ppenenko: I guess my take on this is that partial/incomplete grouping would be more helpful than no grouping at all. Plus, going back and renaming options after the fact is guaranteed to increase confusion. Deferring till another refactoring can take a while, making the temporary names de facto permanent.

As for what is and what isn't an optimization, I think there's a  very well defined class of flags that control ShaderGraph optimization/rewriting specifically, and they are rather analogous to optimization flags traditionally passed to C++ and shader compilers. On a related note, ShaderGraph is becoming an API surface with the introduction of the visitor pattern, so this class of flags is becoming even better-defined.  We have an opportunity to group these flags here and now.

niklasharrysson: > partial/incomplete grouping would be more helpful than no grouping at all

FWIW, I think an incomplete grouping may cause more confusion than no grouping at all. If there is a group it will direct the user focus (that’s the whole point of having a group). So related options placed inconsistently outside the group will loose visibility. 

> As for what is and what isn't an optimization, I think there's a very well defined class of flags that control ShaderGraph optimization/rewriting specifically

This is a great point. A good next step might be to clarify what the proposed “opt” grouping is meant to cover. If the intent is "anything that affects performance" (including things like hwAiryFresnelIterations), that feels too broad to group consistently over time.

But if it’s specifically limited to ShaderGraph edits, and is analogous to the optimization flags we pass to C++ and shader compilers (great analogy), then it feels much more focused, and much easier to keep consistent.

Just a note that if we decide on such a grouping we should rename the existing elideConstants in a future PR, to make this fully consistent.

ld-kerley: I interpret `opt` as the later definition here.  ShaderGraph changes that may or may not affect performance, but are explicitly intended not to affect the final image.  Something that very well may need to be tuned, depending on render destination. 

I think that `elideConstants` is the only existing option that would need to be reclassified.  Because there is only one, perhaps we can just add an `optElideContants` option that takes precedence if it's enabled.  This would allow for a graceful deprecation of the old name, and allow us to move forward with the proposed classification definition Niklas outlines above.

niklasharrysson: This sounds good to me 👍

jstone-lucasfilm: @ld-kerley @ppenenko @niklasharrysson I like this direction forward, and the idea of a prefix that purely represents graph refactoring steps seems like a natural evolution of language-specific prefixes such as `hw` and `osl`.  Graph refactoring steps can be enabled in any generated shading language, though naturally their value may be greater in some environments and languages than others.

In order to create an intuitive connection between the new option prefix and its expected behavior, what would you think about the new prefix string being `graph`?  To my mind, this seems like a self-documenting convention that users would quickly understand, and it allows no ambiguity over whether any future added option belongs in this category or not.

ppenenko: @jstone-lucasfilm TBH `graph` sounds ambiguous to me, mostly because there are concepts of Node Graph and Shader Graph. Could it be `shaderGraph` or `sg`?

jstone-lucasfilm: Sure, I would be open to any variation on this idea, as long as we're communicating the intended meaning of a "shader graph refactoring step".

I'll leave the specific choice between `graph`, `shaderGraph`, and `sg` to the group, though in the world of VFX and real-time rendering, I usually think of `sg` as referring to "scene graph" rather than shader graph.

source/MaterialXGenShader/ShaderGraph.cpp
ppenenko, line 1028: By the way, this elision seems essential not just for optimization, but also for correctness: filenames or samplers can't be local variables in GLSL and HLSL. So there can't be a Dot node whose result is cached in a local.

ppenenko:
Should a test run be added with the optimization enabled?

PR #2507, opened 2025-07-28
Improve UDIM support in viewer

jstone-lucasfilm:
This changelist improves the support for multi-UDIM materials in the MaterialX Viewer, allowing materials with explicit UDIM-associated names to be bound to implicit geometric UDIMs.

Although explicit UDIM-associated names are not a recommended convention for multi-UDIM materials, they are common enough in studio pipelines that it's valuable to support them as an alternative.

### DIFF
--- a/source/MaterialXView/Viewer.cpp
+++ b/source/MaterialXView/Viewer.cpp
@@ -1496,15 +1496,17 @@ void Viewer::loadDocument(const mx::FilePath& filename, mx::DocumentPtr librarie
                 }
             }
 
-            // Apply implicit udim assignments, if any.
+            // Apply UDIM assignments, if any.
             for (mx::MaterialPtr mat : newMaterials)
             {
                 mx::NodePtr materialNode = mat->getMaterialNode();
                 if (materialNode)
                 {
+                    // First check for implicit UDIM assignments.
                     std::string udim = mat->getUdim();
                     if (udim.empty())
                     {
+                        // Fall back to name-based UDIM assignments.
                         for (const std::string& token : mx::splitString(materialNode->getName(), UDIM_SEPARATORS))
                         {
                             if (token.size() == 4 && std::all_of(token.begin(), token.end(), isdigit))


source/MaterialXView/Viewer.cpp
```diff
@@ -1501,6 +1503,16 @@ void Viewer::loadDocument(const mx::FilePath& filename, mx::DocumentPtr librarie
                 if (materialNode)
                 {
                     std::string udim = mat->getUdim();
+                    if (udim.empty())
```
kwokcb, line 1507: I guess this is okay. It's a bit hidden what it's doing and any material with 4-digit numbers in the name will be treated as a udim mapping. Should it be an advanced option to allow turning this on / off ?

jstone-lucasfilm: Good points, @kwokcb, and I agree more refinement would be valuable here.

Would improving the code documentation be enough, or do you believe there's a real-world situation where the user would want to disable this behavior?

It's worth noting that implicit UDIM assignments in materials will still take precedence over these name-based assignments, and this mapping only occurs when the geometry itself has multiple UDIMs.

jstone-lucasfilm: I've added comments to help clarify the new proposed behavior, and let me know how this looks to you.

kwokcb: As far as I could see, UDIM syntax at the Document and at the Material level doesn't seem to be documented at all in the specification, so I'm wondering if this is just a good opportunity to improve this vs code comments here ?

jstone-lucasfilm: @kwokcb It's beyond the scope of this PR, but I'd agree documentation of the expected behavior for UDIMs is an important goal for MaterialX.

In particular, the AOUSD Materials Working Group is expecting to contribute a normative description of UDIMs to the MaterialX specification, as historically the behavior of UDIMs has only been defined by industry conventions and product-specific documentation.

lfl-eholthouser (commented):
This looks great to me!

PR #2509, opened 2025-07-31
Simplify OSL shader generator for `surfaceshader` output

ld-kerley:
Currently if the OSL shader generator is passed an element that outputs a single `surfaceshader` type, then it injects the shader code from the OSL implementation of `surfacematerial` node.  Because the `surfaceshader` type is a struct and it needs to be resolved down to a `closure color` for output.

This PR modifies the shader generator to insert an instance of the `surfacematerial` node, instead of hardcoding the shader source in the shader generator.  This decouples the library, and means we just have a single source of truth for the shader code in the `surfacematerial` node for OSL.

Adds a new API call on the `ShaderGraph` node to insert a node upstream of an output of the graph.

### DIFF
--- a/source/MaterialXGenOsl/OslShaderGenerator.cpp
+++ b/source/MaterialXGenOsl/OslShaderGenerator.cpp
@@ -166,8 +166,7 @@ ShaderPtr OslShaderGenerator::generate(const string& name, ElementPtr element, G
     for (size_t i = 0; i < outputs.size(); ++i)
     {
         const ShaderGraphOutputSocket* outputSocket = graph.getOutputSocket(i);
-        const string result = getUpstreamResult(outputSocket, context);
-        emitLine(outputSocket->getVariable() + " = " + result, stage);
+        emitLine(outputSocket->getVariable() + " = " + getUpstreamResult(outputSocket, context), stage);
     }
 
     // End shader body
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -789,7 +789,7 @@ ShaderNode* ShaderGraph::inlineNodeBeforeOutput(ShaderGraphOutputSocket* output,
     InputPtr nodeDefInput = nullptr;
     if (!inputName.empty())
     {
-        // only look for the input if we are given an inputName - its valid to insert the node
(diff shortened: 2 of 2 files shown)


niklasharrysson (approved):
This looks good to me, thank @ld-kerley

source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -178,29 +162,12 @@ ShaderPtr OslShaderGenerator::generate(const string& name, ElementPtr element, G
...
     {
-        // Assign results to final outputs.
-        for (size_t i = 0; i < outputs.size(); ++i)
-        {
-            const ShaderGraphOutputSocket* outputSocket = graph.getOutputSocket(i);
-            const string result = getUpstreamResult(outputSocket, context);
-            emitLine(outputSocket->getVariable() + " = " + result, stage);
-        }
+        const ShaderGraphOutputSocket* outputSocket = graph.getOutputSocket(i);
+        const string result = getUpstreamResult(outputSocket, context);
```
jstone-lucasfilm: Minor note: Was this intended to be a `const` reference to a string?  If so, then we should add the reference symbol; otherwise, this would be clearer as a non-const string value.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -757,7 +766,78 @@ ShaderNode* ShaderGraph::createNode(ConstNodePtr node, GenContext& context)
...
+    OutputPtr nodeDefOutput = nodeDef->getOutput(outputName);
+    if (!nodeDefOutput)
+    {
+        throw ExceptionShaderGenError("Output '"+outputName+"' not found on NodeDef '"+nodeDefName+"'");
+    }
+
+    InputPtr nodeDefInput = nullptr;
+    if (!inputName.empty())
+    {
+        // only look for the input if we are given an inputName - its valid to insert the node
```
jstone-lucasfilm: Minor note: These new comments should begin with capitalized words for consistency.

jstone-lucasfilm (approved):
Other than the minor notes I've added inline, this change looks good to me.

I'd probably recommend that we hold this change in reserve for 1.39.5, since it's a substantial change in logic.

PR #2513, opened 2025-08-01
Introduce NanoColor names to default color management

JGamache-autodesk:
For 1.39 we only remap to existing color spaces.

Todo: For 1.40:
  - rename node defs/graphs to make nanocolor names the default
  - implement an upgrade path for legacy color spaces names
  - provide new nodegraphs for lin_ap0_scene, lin_rec2020_scene, and srgb_ap1_scene

### DIFF
--- a/source/MaterialXGenShader/DefaultColorManagementSystem.cpp
+++ b/source/MaterialXGenShader/DefaultColorManagementSystem.cpp
@@ -20,7 +20,21 @@ const StringMap COLOR_SPACE_REMAP =
     { "gamma18", "g18_rec709" },
     { "gamma22", "g22_rec709" },
     { "gamma24", "rec709_display" },
-    { "lin_ap1", "acescg" }
+    { "lin_ap1", "acescg" },
+
+    // In 1.39 we remap namespaces from ASWF recommended color interop spaces
+    //  See https://github.com/AcademySoftwareFoundation/ColorInterop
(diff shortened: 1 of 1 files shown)


JGamache-autodesk:
@meshula, this should help get correct MaterialX default color management for color spaces recognized by UsdColorSpaceAPI.

meshula:
Thanks! What would it take to add the missing lin_rec2020? I think it would involve adding another matrix here & there?

This is a huge step forward, even without the missing cs.

jstone-lucasfilm:
@meshula Assuming our goal is to include this in the upcoming MaterialX 1.39.4, we should keep the change as simple and minimal as possible, but I like the idea of including `lin_rec2020` in a subsequent improvement for 1.39.5.

jstone-lucasfilm (approved):
@JGamache-autodesk This looks like the right approach for 1.39.4, and I'd support our including this in the release, though I'll give folks a few days to see if any concerns arise.

ld-kerley:
When we have TODO list items - as in the PRs description. I think it would be a good idea to create a GitHub issue. @JGamache-autodesk - do you think you could add one - so we don't lose track of these things?

JGamache-autodesk:
Opened issue #2535 with the 1.40 items. Please note that the 3 missing color spaces will be introduced in 1.39.5 via pull request #2514.

PR #2516, opened 2025-08-05
Improve MaterialXCore API for Shader Generation

ld-kerley:
Introduce more concrete API calls to MaterialXCore, to help make the interface to Shader Generation more direct.

Functions added:
-  Access for the "bsdf" attribute on a `nodedef` element
- Access for the "sourcecode" attribute on an `implementation` element.

Also use these MaterialXCode API calls, and others in the Shader Generation code,  in favor of more generic `getAttribute(attrName)` calls.

### DIFF
--- a/source/MaterialXCore/Definition.cpp
+++ b/source/MaterialXCore/Definition.cpp
@@ -23,11 +23,13 @@ const string NodeDef::TRANSLATION_NODE_GROUP = "translation";
 
 const string NodeDef::NODE_ATTRIBUTE = "node";
 const string NodeDef::NODE_GROUP_ATTRIBUTE = "nodegroup";
+const string NodeDef::BSDF_ATTRIBUTE = "bsdf";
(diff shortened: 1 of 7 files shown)


niklasharrysson (approved):
This looks good to me.

I just want to point out that the "bsdf" attribute on NodeDefs isn’t documented in the specification, as far as I know. It was originally added as a hint for codegen, to indicate whether a BSDF is reflective, transmissive, or both. One could therefore consider it an implementation detail that should remain internal, in which case it might be better not to publish an API for it.

That said, I think it’s a useful hint that could benefit any backend. So exposing it in the API makes sense to me. Perhaps we should then consider adding it to the specification as well.

ld-kerley:
I've actually been bumping in to some other things with the node classification stuff in the shadergen - so I might actually want to hold off on this change for now. 

I'll try and write up a different issue to discuss the classification things I've been thinking about.

PR #2517, opened 2025-08-07
Fix zenity invocation for KDE

jreichel-nvidia:
The KDE file open dialog hangs because of the trailing space.


jreichel-nvidia:
Can someone take care of NanoGUI which has the same problem (affects MaterialXView)?

jstone-lucasfilm (approved):

jreichel-nvidia:
For reference: In newer versions of NanoGUI the problem has been fixed here:

commit 1cbff2e8d79507e77ab000ee5282c35694689ade
Author: Wenzel Jakob <wenzel.jakob@epfl.ch>
Date:   Fri Aug 29 18:55:10 2025 +0200

    Switch file open/save dialog provider to ``nativefiledialog-extended``

    This commit replaces the hand-written file dialog code in common.cpp
    with the library https://github.com/btzy/nativefiledialog-extended.

    The main consequence of this change is that the Linux implementation of
    these operations switches from zenity to xdg-desktop-portal, causing the
    file dialogs to also work in containers (e.g., flatpak) and non-GTK
    based distributions like KDE. The library also has explicit support for
    Wayland.

jstone-lucasfilm:
Ah, that's good to know, @jreichel-nvidia.  I'm glad Wenzel is still making improvements to NanoGUI, and we should update our submodule reference to the latest commit when time permits.

PR #2518, opened 2025-08-08
Remove unneeded DocumentPtr from MDL ShaderGen

ld-kerley:
Remove unneeded ConstDocumentPtr being passed around inside the MDL shader generator.

While poking around with some Shader Generator refactor experiments I noticed this argument not being used anywhere. 

Tagging @krohmerNV for visibility.


krohmerNV:
looks good to me. I believe this was introduced by while adding annotations. I never managed to get parameter description and range annotations from the mtlx definitions to pass them to MDL. It's fine to remove the pointer. If needed, I can add it back,
Thanks, Lee!

jstone-lucasfilm:
Let's go ahead and include this change in the upcoming 1.39.4, since it's localized to MDL generation and should not impact any functionality.

PR #2519, opened 2025-08-08
add an enhanced validation to catch addition problems.

bhouston:
In creating my blender materialx exporter, I ran into a lot of issues with invalid mtlx files being generator.  In order to address that I created my own validator for mtlx file issues.  I've ported that back to the mxvalidate.py file and made this PR to contribute those additional validates back.

Basically just run "mxvalidate --enhanced glass.mtlx" and it will check now for the following:

- proper node connections.
- surface shader setup, if there is one.
- material structure, if there is one.

I've validated it against all example *.mtlx materials and they all pass, but I did find one warning:

open_pbr_default.mtlx: It has an empty displacementshader input, which I consider to be incorrect:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/Examples/OpenPbr/open_pbr_default.mtlx#L5

### DIFF
--- a/python/Scripts/mxvalidate.py
+++ b/python/Scripts/mxvalidate.py
@@ -13,6 +13,7 @@ def main():
     parser.add_argument("--resolve", dest="resolve", action="store_true", help="Resolve inheritance and string substitutions.")
     parser.add_argument("--verbose", dest="verbose", action="store_true", help="Print summary of elements found in the document.")
     parser.add_argument("--stdlib", dest="stdlib", action="store_true", help="Import standard MaterialX libraries into the document.")
+    parser.add_argument("--enhanced", dest="enhanced", action="store_true", help="Perform enhanced validation including surface shader checks, node connections, and material structure.")
     parser.add_argument(dest="inputFilename", help="Filename of the input document.")
     opts = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


kwokcb:
Hi @bhouston ,
It's good to see you've look at validation more closely. I just want to leave a couple of comments:

1. There is currently an issue that you don't really get "proper" validation if you don't load in the standard libraries -- and this is an option. I'd be curious if you get better results if this flag is enabled. I logged an issue to remove the flag and load the libraries always but maybe you can do it with this PR.

2. I'm wondering if we have gaps in the material->shader connection setup. It is possible but in general upstream node connection existence should be checked.

3. As for constans or nodes without explicit values, it's possible to be valid as there should be a default value specified by the definition. So may be resolved by 1 -- but worth testing out.

4. We can't really assume only certain types of nodes for upstream surfaceshader's as an arbitrary PBR network could be rotted in a surfaceshader output node.

For anything amiss we should move this into the C++ code which performs validation so all code paths perform the same validation check, not just this Python utility.

For your last point It currently  valid to not have a displacement shader (value="") as not all materials require this. We could make this be an invalid case but we'd have to perform add  upgrade logic to remove the displacment input instead. Probably something to discuss some more.

bhouston:
Yeah, I will close this.

PR #2520, opened 2025-08-08
Initial AOUSD formatting proposals

jstone-lucasfilm:
This changelist contains an initial set of formatting proposals for the Standard Nodes specification, organized by the AOUSD Materials Working Group.

### DIFF
--- a/documents/Specification/MaterialX.StandardNodes.md
+++ b/documents/Specification/MaterialX.StandardNodes.md
@@ -72,51 +72,88 @@ Standard Texture nodes:
 
 <a id="node-image"> </a>
 
-* **`image`**: samples data from a single image, or from a layer within a multi-layer image.  When used in the context of rendering a geometry, the image is mapped onto the geometry based on geometry UV coordinates, with the lower-left corner of an image mapping to the (0,0) UV coordinate (or to the fractional (0,0) UV coordinate for tiled images).
+### `image`
+
+Samples data from a single image, or from a layer within a multi-layer image.  When used in the context of rendering a geometry, the image is mapped onto the geometry based on geometry UV coordinates, with the lower-left corner of an image mapping to the (0,0) UV coordinate (or to the fractional (0,0) UV coordinate for tiled images).
+
 The type of the &lt;image> node determines the number of channels output, which may be less than the number of channels in the image file, outputting the first N channels from the image file.  So a `float` &lt;image> would return the Red channel of an RGB image, and a `color3` &lt;image> would return the RGB channels of an RGBA image.  If the type of the &lt;image> node has more channels than the referenced image file, then the output will contain zero values in all channels beyond the N channels of the image file.
-    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.
-    * `layer` (uniform string): the name of the layer to extract from a multi-layer input file.  If no value for `layer` is provided and the input file has multiple layers, then the "default" layer will be used, or "rgba" if there is no "default" layer.  Note: the number of channels defined by the `type` of the `<image>` must match the number of channels in the named layer.
-    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;_geometry token_>, [_interface token_] or {_hostattr_} is included in the filename but no substitution value or default is defined, or if the resolved `file` URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
-    * `texcoord` (vector2): the name of a vector2-type node specifying the 2D texture coordinate at which the image data is read.  Default is to use the current u,v coordinate.
-    * `uaddressmode` (uniform string): determines how U coordinates outside the 0-1 range are processed before sampling the image; see below.  Default is "periodic".
-    * `vaddressmode` (uniform string): determines how V coordinates outside the 0-1 range are processed before sampling the image; see below.  Default is "periodic".
-    * `filtertype` (uniform string): the type of texture filtering to use; standard values include "closest" (nearest-neighbor single-sample), "linear", and "cubic".  If not specified, an application may use its own default texture filtering method.
+
+The `file` input can include one or more substitutions to change the file name that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.  The `filtertype` input supports options `closest` (nearest-neighbor single-sample), `linear`, and `cubic`.
+
+|Port            |Description                                                                                                      |Type                  |Default  |Accepted Values                  |
+|----------------|-----------------------------------------------------------------------------------------------------------------|----------------------|---------|---------------------------------|
+|`file`          |The URI of an image file                                                                                         |filename              |__empty__|                                 |
+|`layer`         |The name of the layer to extract from a multi-layer input file                                                   |string                |__empty__|                                 |
+|`default`       |A default value to use if the file reference can not be resolved                                                 |float, colorN, vectorN|__zero__ |                                 |
+|`texcoord`      |The 2D texture coordinate at which the image data is read                                                        |vector2               |_UV0_    |                                 |
+|`uaddressmode`  |Determines how U coordinates outside the 0-1 range are processed before sampling the image                       |string                |periodic |constant, clamp, periodic, mirror|
+|`vaddressmode`  |Determines how V coordinates outside the 0-1 range are processed before sampling the image                       |string                |periodic |constant, clamp, periodic, mirror|
+|`filtertype`    |The type of texture filtering to use                                                                             |string                |linear   |closest,linear,cubic             |
+|`framerange`    |A string 'minframe-maxframe', e.g. '10-99', to specify the range of frames that the image file is allowed to have|string                |__empty__|                                 |
+|`frameoffset`   |A number that is added to the current frame number to get the image file frame number                            |integer               |0        |                                 |
+|`frameendaction`|What to do when the resolved image frame number is outside the `framerange` range                                |string                |constant |constant, clamp, periodic, mirror|
+|`out`           |Output: the sampled texture value                                                                                |Same as `default`     |__zero__ |                                 |
 
 <a id="node-tiledimage"> </a>
 
-* **`tiledimage`** (NG): samples data from a single image, with provisions for tiling and offsetting the image across uv space.
-    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.
-    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;geomtoken>, [interfacetoken] or {hostattr} is included in the filename but no substitution value or default is defined, or if the resolved file URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
-    * `texcoord` (vector2): the name of a vector2-type node specifying the 2D texture coordinate at which the image data is read.  Default is to use the current u,v coordinate.
-    * `uvtiling` (vector2): the tiling rate for the given image along the U and V axes. Mathematically equivalent to multiplying the incoming texture coordinates by the given vector value. Default value is (1.0, 1.0).
-    * `uvoffset` (vector2): the offset for the given image along the U and V axes. Mathematically equivalent to subtracting the given vector value from the incoming texture coordinates. Default value is (0.0, 0.0).
-    * `realworldimagesize` (vector2): the real-world size represented by the `file` image, with unittype "distance".  A `unit` attribute may be provided to indicate the units that `realworldimagesize` is expressed in.
-    * `realworldtilesize` (vector2): the real-world size of a single square 0-1 UV tile, with unittype "distance".  A `unit` attribute may be provided to indicate the units that `realworldtilesize` is expressed in.
-    * `filtertype` (uniform string): the type of texture filtering to use; standard values include "closest" (nearest-neighbor single-sample), "linear", and "cubic".  If not specified, an application may use its own default texture filtering method.
+### `tiledimage`
+Samples data from a single image, with provisions for tiling and offsetting the image across uv space.
+
+The `file` input can include one or more substitutions to change the file name that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.
+
+|Port                |Description                                                                                                      |Type                  |Default  |Accepted Values                  |
+|--------------------|-----------------------------------------------------------------------------------------------------------------|----------------------|---------|---------------------------------|
+|`file`              |The URI of an image file                                                                                         |filename              |__empty__|                                 |
+|`default`           |A default value to use if the file reference can not be resolved                                                 |float, colorN, vectorN|__zero__ |                                 |
+|`texcoord`          |The 2D texture coordinate at which the image data is read                                                        |vector2               |_UV0_    |                                 |
+|`uvtiling`          |The tiling rate for the given image along the U and V axes                                                       |vector2               |1.0, 1.0 |                                 |
+|`uvoffset`          |The offset for the given image along the U and V axes                                                            |vector2               |0.0, 0.0 |                                 |
+|`realworldimagesize`|The real-world size represented by the file image                                                                |vector2               |1.0, 1.0 |                                 |
+|`realworldtilesize` |The real-world size of a single square 0-1 UV tile                                                               |vector2               |1.0, 1.0 |                                 |
+|`filtertype`        |The type of texture filtering to use                                                                             |string                |linear   |closest,linear,cubic             |
+|`framerange`        |A string 'minframe-maxframe', e.g. '10-99', to specify the range of frames that the image file is allowed to have|string                |__empty__|                                 |
+|`frameoffset`       |A number that is added to the current frame number to get the image file frame number                            |integer               |0        |                                 |
+|`frameendaction`    |What to do when the resolved image frame number is outside the `framerange` range                                |string                |constant |constant, clamp, periodic, mirror|
+|`out`               |Output: the sampled texture value                                                                                |Same as `default`     |__zero__ |                                 |
 
 <a id="node-latlongimage"> </a>
 
-* **`latlongimage`** (NG): samples an equiangular map along a view direction with adjustable latitudinal offset.
-    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](#filename-substitutions) section above.
-    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;_geometry token_>, [_interface token_] or {_hostattr_} is included in the filename but no substitution value or default is defined, or if the resolved `file` URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
-    * `viewdir` (vector3): the view direction determining the value sampled from the projected equiangular map.
-    * `rotation` (float): the longitudinal sampling offset, in degrees.
+### `latlongimage`
+Samples an equiangular map along a view direction with adjustable latitudinal offset.
+
+The `file` input can include one or more substitutions to change the file name that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.
+
+|Port      |Description                                                                        |Type    |Default      |
+|----------|-----------------------------------------------------------------------------------|--------|-------------|
+|`file`    |The URI of an image file                                                           |filename|__empty__    |
+|`default` |A default value to use if the file reference can not be resolved                   |color3  |0.0, 0.0, 0.0|
+|`viewdir` |The view direction determining the value sampled from the projected equiangular map|vector3 |0.0, 0.0, 1.0|
+|`rotation`|The longitudinal sampling offset, in degrees                                       |float   |0.0          |
+|`out`     |Output: the sampled texture value                                                  |color3  |0.0, 0.0, 0.0|
 
 <a id="node-triplanarprojection"> </a>
 
-* **`triplanarprojection`** (NG): samples data from three images (or layers within multi-layer images), and projects a tiled representation of the images along each of the three respective coordinate axes, computing a weighted blend of the three samples using the geometric normal.
-    * `filex` (uniform filename): the URI of an image file to be projected in the direction from the +X axis back toward the origin.
-    * `filey` (uniform filename): the URI of an image file to be projected in the direction from the +Y axis back toward the origin with the +X axis to the right.
-    * `filez` (uniform filename): the URI of an image file to be projected in the direction from the +Z axis back toward the origin.
-    * `layerx` (uniform string): the name of the layer to extract from a multi-layer input file for the x-axis projection.  If no value for `layerx` is provided and the input file has multiple layers, then the "default" layer will be used, or "rgba" if there is no "default" layer.  Note: the number of channels defined by the `type` of the `<image>` must match the number of channels in the named layer.
-    * `layery` (uniform string): the name of the layer to extract from a multi-layer input file for the y-axis projection.
-    * `layerz` (uniform string): the name of the layer to extract from a multi-layer input file for the z-axis projection.
-    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if any `file<em>X</em>` reference can not be resolved (e.g. if a &lt;geomtoken>, [interfacetoken] or {hostattr} is included in the filename but no substitution value or default is defined, or if the resolved file URI cannot be read)  The `default` value must be the same type as the `<triplanarprojection>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
-    * `position` (vector3): a spatially-varying input specifying the 3D position at which the projection is evaluated.  Default is to use the current 3D object-space coordinate.
-    * `normal` (vector3): a spatially-varying input specifying the 3D normal vector used for blending.  Default is to use the current object-space surface normal.
-    * `upaxis` (integer enum): which axis is considered to be "up", either 0 for X, 1 for Y, or 2 for Z.  Default is Y (1).
-    * `blend` (float): a 0-1 weighting factor for blending the three axis samples using the geometric normal, with higher values giving softer blending.  Default is 1.0.
-    * `filtertype` (uniform string): the type of texture filtering to use; standard values include "closest" (nearest-neighbor single-sample), "linear", and "cubic".  If not specified, an application may use its own default texture filtering method.
+### `triplanarprojection`
+Samples data from three images (or layers within multi-layer images), and projects a tiled representation of the images along each of the three respective coordinate axes, computing a weighted blend of the three samples using the geometric normal.
+
+|Port            |Description                                                                                                      |Type                  |Default  |Accepted Values                  |
+|----------------|-----------------------------------------------------------------------------------------------------------------|----------------------|---------|---------------------------------|
+|`filex`         |The URI of an image file to be projected in the direction from the +X axis back toward the origin                |filename              |__empty__|                                 |
+|`filey`         |The URI of an image file to be projected in the direction from the +Y axis back toward the origin                |filename              |__empty__|                                 |
+|`filez`         |The URI of an image file to be projected in the direction from the +Z axis back toward the origin                |filename              |__empty__|                                 |
+|`layerx`        |The name of the layer to extract from a multi-layer input file for the X-axis projection                         |string                |__empty__|                                 |
+|`layery`        |The name of the layer to extract from a multi-layer input file for the Y-axis projection                         |string                |__empty__|                                 |
+|`layerz`        |The name of the layer to extract from a multi-layer input file for the Z-axis projection                         |string                |__empty__|                                 |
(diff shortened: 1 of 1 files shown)


dbsmythe:
Generally speaking, I really like this new proposal: it's quite a bit easier to read, and the increased consistency of formatting and the use of the tables for inputs improves the ease of finding specific information.

A few notes:

1. The first column in each table should be labeled "Input" or "Input name" rather than "Port", to be consistent with the terminology used elsewhere in the Specification.
2. We need to ensure that all descriptive and clarifying text that appeared in the legacy-format document is replicated in the new-format document somewhere: for example, for the `input` node, the additional text for the "layer" and "default" input descriptions is omitted, while for the "luminance" node, there is no notation of what the given default value for "lumacoeffs" actually represents (they are the ACEScg/ap1 luma coefficients).  I like the conciseness that the current abbreviated Descriptions provide, so I would suggest adding that extra text as separate notes above or below the table, similar to how the extra descriptive notes are presented for the "file" and "filtertype" inputs (although in the case of "filtertype", the same information about acceptable values is currently being presented twice: once in the table, and again in the separate leading paragraph).
3. We should clean up the capitalization and punctuation of the node and input descriptions, e.g. always capitalize the first word in a node or input description (the descriptions for "atan2" and all the logical operator nodes are currently not) and end with a period ("ramplr" is missing the ending ".").  We should probably also be more consistent about whether the description starts with "Outputs ..." or "Returns..." or not.
4. If a default value is of type `string`, I think the value should be shown within double-quotes.
5. The font size currently used for node name headers is a bit large (at least on my display), appearing larger than sub-section headers, e.g. in the "Blend Nodes" subsection, the "plus" and "minus" node name text is larger than the "Blend Nodes" header font above them; this makes it more difficult to see the sectional organization of the document. 

Some things we can discuss:

- Should there always be an "Accepted Values" column in the tables, or (as is the case now) only when one of the inputs has a requirement about allowable values?
- In the legacy-format document, I tried to consistently identify with "(NG)" which nodes were implemented as nodegraphs in the standard library, to distinguish them from nodes that needed native implementations in each target.  Do others find value in that, or does adding "(NG)" just add noise to the descriptions?  There are a few nodes like "tiledimage" that were marked with "(NG)" in the legacy-format document that have lost their "(NG)" designation in the new-format document.
- Having the widths of the tables change from one node to the next is a bit visually distracting: is there a Markdown syntax that could allow more consistent table presentation and column widths?
- If we are writing up a normative description, should there be a standard way to express in this document what the exact math or algorithm (pseudo code) is expected to be?  For the simple math operators this may be overkill, but for nodes like "range", "rgbtohsv" or "colorcorrect", this could be very handy.

I would also suggest that while we're in this transition period, we should keep the original-format version around e.g. "MaterialX.StandardNodes.legacyformat.md" or something like that; once we're happy with the new format and are certain that no information from the legacy document is lost, the legacy-formatted version could be removed in a separate PR.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -243,17 +321,27 @@ Standard Noise nodes:
 
 <a id="node-noise2d"> </a>
 
-* **`noise2d`**: 2D Perlin noise in 1, 2, 3 or 4 channels.
-    * `amplitude` (float or vector<em>N</em>): the center-to-peak amplitude of the noise (peak-to-peak amplitude is 2x this value).  Default is 1.0.
-    * `pivot` (float): the center value of the output noise; effectively, this value is added to the result after the Perlin noise is multiplied by `amplitude`.  Default is 0.0.
-    * `texcoord` (vector2): the 2D texture coordinate at which the noise is evaluated.  Default is to use the first set of texture coordinates.
+### `noise2d`
```
ld-kerley, line 354: This table is missing the following nod definitions 
* ND_noise2d_color3
* ND_noise2d_color4
* ND_noise2d_color3FA
* ND_noise2d_color4FA
* ND_noise2d_vector2FA
* ND_noise2d_vector3FA
* ND_noise2d_vector4FA

The corresponding node definitions are also missing from noise3d and fractal3d

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -266,66 +354,103 @@ Standard Noise nodes:
...
+|`position`|The 3D position at which the noise is evaluated|vector3|_Pobject_|
+|`out`     |Output: the computed noise value               |float  |0.0      |
 
 <a id="node-worleynoise2d"> </a>
 
-* **`worleynoise2d`**: 2D Worley noise using centered jitter, outputting float (distance metric to closest feature), vector2 (distance metrics to closest 2 features) or vector3 (distance metrics to closest 3 features).
-    * `jitter` (float): amount to jitter the cell center position, with smaller values creating a more regular pattern.  Default is 1.0.
-    * `style` (integer): the output style, one of "distance" (distance to the cell center), or "solid" (constant value for each cell).
-    * `texcoord` (vector2): the 2D position at which the noise is evaluated.  Default is to use the first set of texture coordinates.
+### `worleynoise2d`
```
ld-kerley, line 421: Missing "Accepted Values" derived from the enum values for style.

Also for worleynoise3d.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -266,66 +354,103 @@ Standard Noise nodes:
...
+|`outmin`     |The lowest output value                                                                    |float  |0      |
+|`outmax`     |The highest output value                                                                   |float  |1      |
+|`clampoutput`|If enabled the output is clamped between the min and max output values                     |boolean|true   |
+|`octaves`    |The number of octaves of noise to be summed                                                |integer|3      |
+|`lacunarity` |The exponential scale between successive octaves of noise                                  |float  |2      |
+|`diminish`   |The rate at which noise amplitude is diminished for each octave                            |float  |0.5    |
+|`type`       |The type of noise function to use.  One of 0 (Perlin), 1 (Cell), 2 (Worley), or 3 (Fractal)|integer|0      |
+|`style`      |The output style                                                                           |integer|0      |
+|`out`        |Output: the computed noise value                                                           |float  |0.0    |
```
ld-kerley: The default value specified here is not actual defined in the data library - this may be in a bug in the data library that we should fix separately, but currently this specification statement does not accurately reflect the data library as it stands. 

Similarly for unifiednoise3d

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -556,307 +785,635 @@ Math nodes have one or two spatially-varying inputs, and are used to perform a m
...
+|Port |Description                       |Type                  |Default |
+|-----|----------------------------------|----------------------|--------|
+|`in1`|The primary input stream          |float, colorN, vectorN|__zero__|
+|`in2`|The stream to multiply `in1` by   |Same as `in1` or float|__one__ |
+|`out`|Output: product of `in1` and `in2`|Same as `in1`         |`in1`   |
+
+|Port |Description                       |Type                  |Default|
+|-----|----------------------------------|----------------------|-------|
+|`in1`|The primary input stream          |matrixNN              |__one__|
+|`in2`|The stream to multiply `in1` by   |Same as `in1` or float|__one__|
```
ld-kerley: There is no ND_multiply_matrix33FA or ND_multiply_matrix44FA.

This looks like a copy/paste bug to include the "or float" here.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -556,307 +785,635 @@ Math nodes have one or two spatially-varying inputs, and are used to perform a m
...
+|Port |Description                      |Type    |Default |
+|-----|---------------------------------|--------|--------|
+|`in` |Vector to be transformed         |vector2 |__zero__|
+|`mat`|Matrix to transform `in` by      |matrix33|__one__ |
+|`out`|Output: `in` transformed by `mat`|vector2 |`in`    |
+
+|Port |Description                      |Type                |Default |
+|-----|---------------------------------|--------------------|--------|
+|`in` |Vector to be transformed         |vector3             |__zero__|
+|`mat`|Matrix to transform `in` by      |matrix33 or matrix44|__one__ |
```
ld-kerley: should be "matrixNN"

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -556,307 +785,635 @@ Math nodes have one or two spatially-varying inputs, and are used to perform a m
...
 
-* **`invertmatrix`**: output the inverse of the incoming matrix; if the input matrix is not invertible, the output matrix will consist of all floating-point "NaN" values.
-    * `in` (matrix<em>NN</em>): the input value or nodename
+### `invertmatrix`
+Invert the incoming matrix.
+
+|Port |Description            |Type    |Default|
+|-----|-----------------------|--------|-------|
+|`in` |The input matrix       |matrixNN|__one__|
+|`out`|Output: inverse of `in`|matrixNN|`in`   |
```
ld-kerley: Could be "Same as `in`.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -865,80 +1422,131 @@ Adjustment nodes have one input named "in", and apply a specified function to va
...
+|`low` |Low value for the input range             |Same as `in` or float |__zero__|
+|`high`|Low value for the output range            |Same as `low`         |__one__ |
+|`out` |Output: the adjusted value                |Same as `in`          |`in`    |
 
 <a id="node-luminance"> </a>
 
-* **`luminance`**: (color3 or color4 only) output a grayscale value containing the luminance of the incoming RGB color in all color channels, computed using the dot product of the incoming color with the luma coefficients of the working colorspace; the alpha channel is left unchanged if present.
-    * `in` (color3/color4): the input value or nodename
-    * `lumacoeffs` (uniform color3): the luma coefficients of the current working color space; if no specific color space can be determined, the ACEScg (ap1) luma coefficients [0.2722287, 0.6740818, 0.0536895] will be used.  Applications which support color management systems may choose to retrieve the luma coefficients of the working colorspace from the CMS to pass to the &lt;luminance> node's implementation directly, rather than exposing it to the user.
+### `luminance`
```
ld-kerley, line 1547: Missing "Accepted Values" defined in the enum values.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -865,80 +1422,131 @@ Adjustment nodes have one input named "in", and apply a specified function to va
...
-    * `in` (float or color<em>N</em> or vector<em>N</em>): the input value or nodename
-    * `amount` (float): a multiplier for saturation; the saturate operator performs a linear interpolation between the luminance of the incoming color value (copied to all three color channels) and the incoming color value itself.  Note that setting amount to 0 will result in an R=G=B gray value equal to the value that the `luminance` node (below) returns.  Default is 1.0.
-    * `lumacoeffs` (uniform color3): the luma coefficients of the current working color space; if no specific color space can be determined, the ACEScg (ap1) luma coefficients [0.272287, 0.6740818, 0.0536895] will be used.  Applications which support color management systems may choose to retrieve this value from the CMS to pass to the &lt;saturate> node's implementation directly, rather than exposing it to the user.
+### `saturate`
```
ld-kerley, line 1591: Missing "Accepted Values" defined in the enum values.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -952,13 +1560,30 @@ Premult nodes operate on 4-channel (color4) inputs/outputs, have one input named
...
+|Port |Description                         |Type  |Default           |
+|-----|------------------------------------|------|------------------|
+|`in` |The input stream to be premultiplied|color4|0.0, 0.0, 0.0, 1.0|
+|`out`|Output: premultiplied `in`          |color4|`in`              |
 
 <a id="node-unpremult"> </a>
 
-* **`unpremult`**: divide the RGB channels of the input by the Alpha channel of the input.  If the Alpha value is zero, the original color4 input value is passed through unchanged.
-    * `in` (color4): the input value or nodename; default is (0,0,0,1).
+### `unpremult`
```
ld-kerley, line 1643: This seems to infer a lot more unpremult nodes than actually exist. There is only 
* ND_unpremult_color4

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -1114,33 +1850,78 @@ Conditional nodes are used to compare values of two streams, or to select a valu
 
 <a id="node-ifgreater"> </a>
 
-* **`ifgreater`**: output the value of the `in1` or `in2` stream depending on whether the value of one test input is greater than the value of another.  Ifgreater nodes can be of output type float, integer, color<em>N</em>, vector<em>N</em> or matrix<em>NN</em>.  There is also a "boolean" output-type **`ifgreater`** node, with `value1` and `value2` inputs but no `in1` or `in2`: output is "true" if `value1` > `value2`.
-    * `value1` (integer or float): the first value or nodename to compare.  Default is 1.0.
-    * `value2` (integer or float): the second value or nodename to compare must be the same type as `value1`.  Default is 0.0.
-    * `in1` (float or integer or color<em>N</em> or vector<em>N</em> or matrix<em>NN</em>): the value or nodename to output if `value1` > `value2`; must be the same type as the `ifgreater` node's output.  Default is 0.0 in all channels.
-    * `in2` (float or integer or color<em>N</em> or vector<em>N</em> or matrix<em>NN</em>): the value or nodename to output if `value1` &lt;= `value2`; must be the same type as the `ifgreater` node's output.  Default is 0.0 in all channels.
+### `ifgreater`
```
ld-kerley, line 1916: This is missing the boolean cases.
* ND_ifgreater_boolean
* ND_ifgreater_booleanI

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -1114,33 +1850,78 @@ Conditional nodes are used to compare values of two streams, or to select a valu
...
 <a id="node-ifgreatereq"> </a>
 
-* **`ifgreatereq`**: output the value of the `in1` or `in2` stream depending on whether the value of one test input is greater than or equal to the value of another.  Ifgreatereq nodes can be of output type float, integer, color<em>N</em>, vector<em>N</em> or matrix<em>NN</em>. There is also a "boolean" output-type **`ifgreatereq`** node, with `value1` and `value2` inputs but no `in1` or `in2`: output is "true" if `value1` >= `value2`.
-    * `value1` (integer or float): the first value or nodename to compare.  Default is 1.0.
-    * `value2` (integer or float): the second value or nodename to compare; must be the same type as `value1`.  Default is 0.0.
-    * `in1` (float or integer or color<em>N </em>or vector<em>N</em> or matrix<em>NN</em>): the value or nodename to output if `value1` >= `value2`; must be the same type as the `ifgreatereq` node's output.  Default is 0.0 in all channels.
-    * `in2` (float or integer or color<em>N </em>or vector<em>N</em> or matrix<em>NN</em>): the value or nodename to output if `value1` &lt; `value2`; must be the same type as the `ifgreatereq` node's output.  Default is 0.0 in all channels.
+### `ifgreatereq`
```
ld-kerley, line 1936: This is missing the boolean cases
* ND_ifgreatereq_boolean
* ND_ifgreatereq_booleanI

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -1114,33 +1850,78 @@ Conditional nodes are used to compare values of two streams, or to select a valu
...
 
 <a id="node-ifequal"> </a>
 
-* **`ifequal`**: output the value of the `in1` or `in2` stream depending on whether the value of two test inputs are equal or not.  Ifequal nodes can be of output type float, integer, color<em>N</em>, vector<em>N</em> or matrix<em>NN</em>. There is also a "boolean" output-type **`ifequal`** node, with `value1` and `value2` inputs but no `in1` or `in2`: output is "true" if `value1` == `value2`.
-    * `value1` (boolean or integer or float): the first value or nodename to compare.  Default is 0 or "false".
-    * `value2` (boolean or integer or float): the second value or nodename to compare; must be the same type as `value1`.  Default is 0 or "false".
-    * `in1` (float or integer or color<em>N </em>or vector<em>N</em> or matrix<em>NN</em>): the value or nodename to output if `value1` == `value2`; must be the same type as the `ifequal` node's output.  Default is 0.0 in all channels.
-    * `in2` (float or integer or color<em>N </em>or vector<em>N</em> or matrix<em>NN</em>): the value or nodename to output if `value1` != `value2`; must be the same type as the `ifequal` node's output.  Default is 0.0 in all channels.
+### `ifequal`
```
ld-kerley, line 1956: This is missing 
* ND_ifequal_boolean
* ND_ifequal_booleanI

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -1114,33 +1850,78 @@ Conditional nodes are used to compare values of two streams, or to select a valu
...
+|`value2`|The second value to be compared                   |boolean                                  |false   |
+|`in1`   |The value stream to output if `value1` = `value2` |float, colorN, vectorN, matrixNN, integer|__zero__|
+|`in2`   |The value stream to output if `value1` != `value2`|Same as `in1`                            |__zero__|
+|`out`   |Output: the result of the comparison              |Same as `in1`                            |`in1`   |
+
+|Port    |Description                       |Type   |Default|
+|--------|----------------------------------|-------|-------|
+|`value1`|The first value to be compared    |boolean|false  |
+|`value2`|The first value to be compared    |boolean|false  |
+|`out`   |Output: true if `value1` = `value2|boolean|       |
```
ld-kerley: Missing default output value - data library states `false`.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -72,51 +72,88 @@ Standard Texture nodes:
...
+|`default`       |A default value to use if the file reference can not be resolved                                                 |float, colorN, vectorN|__zero__ |                                 |
+|`texcoord`      |The 2D texture coordinate at which the image data is read                                                        |vector2               |_UV0_    |                                 |
+|`uaddressmode`  |Determines how U coordinates outside the 0-1 range are processed before sampling the image                       |string                |periodic |constant, clamp, periodic, mirror|
+|`vaddressmode`  |Determines how V coordinates outside the 0-1 range are processed before sampling the image                       |string                |periodic |constant, clamp, periodic, mirror|
+|`filtertype`    |The type of texture filtering to use                                                                             |string                |linear   |closest,linear,cubic             |
```
ld-kerley: suggest "closest, linear, cubic" instead of "closest,linear,cubic"

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -72,51 +72,88 @@ Standard Texture nodes:
...
+|`uvtiling`          |The tiling rate for the given image along the U and V axes                                                       |vector2               |1.0, 1.0 |                                 |
+|`uvoffset`          |The offset for the given image along the U and V axes                                                            |vector2               |0.0, 0.0 |                                 |
+|`realworldimagesize`|The real-world size represented by the file image                                                                |vector2               |1.0, 1.0 |                                 |
+|`realworldtilesize` |The real-world size of a single square 0-1 UV tile                                                               |vector2               |1.0, 1.0 |                                 |
+|`filtertype`        |The type of texture filtering to use                                                                             |string                |linear   |closest,linear,cubic             |
```
ld-kerley: suggest "closest, linear, cubic" instead of "closest,linear,cubic"

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -72,51 +72,88 @@ Standard Texture nodes:
...
+|`position`      |A spatially-varying input specifying the 3D position at which the projection is evaluated                        |vector3               |_Pobject_|                                 |
+|`normal`        |A spatially-varying input specifying the 3D normal vector used for blending                                      |vector3               |_Nobject_|                                 |
+|`upaxis`        |Which axis is considered to be 'up', either 0 for X, 1 for Y, or 2 for Z                                         |integer               |2        |0, 1, 2                          |
+|`blend`         |Weighting factor for blending samples using the geometric normal, with higher values giving softer blending      |float                 |1.0      |                                 |
+|`filtertype`    |The type of texture filtering to use                                                                             |string                |linear   |closest,linear,cubic             |
```
ld-kerley: suggest "closest, linear, cubic" instead of "closest,linear,cubic"

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -556,307 +785,635 @@ Math nodes have one or two spatially-varying inputs, and are used to perform a m
...
 
 <a id="node-divide"> </a>
 
-* **`divide`**: divide an incoming float/color/vector/matrix by a value; dividing a channel value by 0 results in floating-point "NaN".  Division of two vectors is interpreted as a component-wise division of the first vector by the second, while division of two matrices is interpreted as a standard matrix product of the `in1` matrix and the inverse of the `in2` matrix.
-    * `in1` (float or color<em>N</em> or vector<em>N</em> or matrix<em>NN</em>): the value or nodename for the primary input
-    * `in2` (same type as `in1` or float): the value or nodename to divide by; default is 1.0 in all channels for float/color/vector types, or the identity matrix for matrix types.
+### `divide`
+Divide one value by another. Scalar and vector types divide component-wise, while for matrices `in1` is multiplied with the inverse of `in2`.
+
+|Port |Description                   |Type                           |Default |
```
ld-kerley: suggest removing additional spaces in table.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -556,307 +785,635 @@ Math nodes have one or two spatially-varying inputs, and are used to perform a m
...
 
 <a id="node-power"> </a>
 
-* **`power`**: raise incoming float/color values to the specified exponent, commonly used for "gamma" adjustment.
-    * `in1` (float or color<em>N</em> or vector<em>N</em>): the value or nodename for the primary input
-    * `in2` (same type as `in1` or float): exponent value or nodename; output = pow(in1, in2); default is 1.0 in all channels
+### `power`
+Raise incoming float/color values to the specified exponent, commonly used for "gamma" adjustment.
+
+|Port |Description                   |Type                           |Default |
```
ld-kerley: suggest removing additional spaces in table.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -556,307 +785,635 @@ Math nodes have one or two spatially-varying inputs, and are used to perform a m
...
 
 <a id="node-safepower"> </a>
 
-* **`safepower`** (NG): raise incoming float/color values to the specified exponent.  Unlike the standard [&lt;power>](#node-power) node, negative `in1` values for &lt;safepower> will result in negative output values, e.g. `out = sign(in1)*pow(abs(in1),in2)`.
-    * `in1` (float or color<em>N</em> or vector<em>N</em>): the value or nodename for the primary input
-    * `in2` (same type as `in1` or float): exponent value or nodename; default is 1.0 in all channels
+### `safepower`
+Raise incoming float/color values to the specified exponent. Negative "in1" values will result in negative output values.
+
+|Port |Description                   |Type                           |Default |
```
ld-kerley: suggest removing additional spaces in table.

ld-kerley (changes requested):
I've made some good progress on the automatic generation of this doc.

The comments below are the outstanding differences.  Mainly either minor formatting changes, or places that look like small errors introduced in the human refactoring.

documents/Specification/MaterialX.StandardNodes.md
ld-kerley: Infact I think it has to be "Same as `in`" here - because otherwise we suggesting that you could have a `matrix33` as input and a `matrix44` as output - which isn't the case for `invertmatrix`.

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Data library fix - will take a while to merge

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Should be there

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Data library fix

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Include

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Do this

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Do this

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Go with numbers as it's explicit. Need to remove coefficients in the future and rely on color management system instead. Make a note of this in the spec

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Go with numbers as it's explicit. Need to remove coefficients in the future and rely on color management system instead. Make a note of this in the spec

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Fix this

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Add these

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Add these

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Add these

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Add this

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Go with spaces everywhere

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Go with spaces everywhere

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Go with spaces everywhere

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Do this

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Do this

documents/Specification/MaterialX.StandardNodes.md
anderslanglands: Do this

anderslanglands:
@dbsmythe regarding your questions:

1. We use "Port" here as it covers both inputs and outputs so felt like it needed a new word and "Port" is somewhat common elsewhere. Would you be OK with this is we added it to the front of the spec explaining clearly what is meant here?

For Accepted Values, we tried with it always there and it was pretty ugly, so we're inclined to keep it optional. We can try again if we manage to figure out a way of getting the table widths to always be 100% as maybe that will be less distracting .
We're explicitly not differentiating node graphs as from the POV of a user (or clean-room implementer) this is an implementation detail.

There normative description does indeed use equations to unambiguously define the operations of the nodes. In the cases where that would be onerous (e.g. fractals) we'll probbaly use pseudo-code. It sounds like you might be asking to include the normative descriptions in the MaterialX spec as well?

dbsmythe:
@anderslanglands Responses to your responses:
Re: using "port" to cover both inputs and outputs: I like this idea, and would support adding this term to the paragraph describing Nodes in the MaterialX Spec's "Definitions" section, and editing the Spec to use "port" wherever "inputs and outputs" are used together.  I'll try to take a crack at this.

Re: accepted values optional vs always: as long as "accepted values" are shown wherever they make an important difference, I'm okay with leaving "accepted values" empty if any value of the type for that input is acceptable, and removing that column if all ports for a node can accept any value.

Re: indicating nodegraph vs native implementations: I agree this is generally an implementation detail, though I thought it would be useful to those implementing new targets to know which nodes they'd need to implement explicitly for the target and which "come for free" via a nodegraph implementation.  But I'll defer to the community if this detail isn't worth the added clutter to the tables.

Re: adding equations/pseudocode descriptions to the Spec: I think this would worthwhile in some instances where there could be confusion, but would just be added clutter in most cases.  If the normative descriptions become part of (or are tightly linked to) the Spec, then no additional equations/pseudocode descriptions would be necessary.

ld-kerley:
@dbsmythe - Firstly thanks for all the feedback - I know you're super busy with production.

Regarding annotating the nodegraphs - I would pose that anyone who is implementing a new target would also be required to investigate the actual data library files themselves, and so would "hopefully" have the expertise to understand which nodegraph elements they could choose adopt "for free".  I'll note that implementors are free to choose to reimplemented, existing nodegraphs themselves as well.  The specification is complete enough to stand alone, without any dependence on the data library.

documents/Specification/MaterialX.StandardNodes.md
jstone-lucasfilm: Good catch, and fixed!

documents/Specification/MaterialX.StandardNodes.md
jstone-lucasfilm: Good catch, and fixed!

documents/Specification/MaterialX.StandardNodes.md
jstone-lucasfilm: Good catch, and fixed!

documents/Specification/MaterialX.StandardNodes.md
jstone-lucasfilm: For now, I'm leaving the informational content unchanged from the original nodes specification, since the interpretation of `lumacoeffs` is in flux, and we can return to this in a future PR.

documents/Specification/MaterialX.StandardNodes.md
jstone-lucasfilm: Good catch, and fixed!

jstone-lucasfilm:
Thanks @dbsmythe, and let's go ahead and merge this as a foundation for your upcoming improvements.

Thanks to my colleagues at the AOUSD Materials Working Group, whose work made these improvements possible!

PR #2521, opened 2025-08-13
Add an option to toggle eliding constant nodes

JGamache-autodesk:
Editing this node in a real time MaterialX editing software leads to constant shader recompilation because the value gets burned inside the shader text. On complex shading models, this has a noticeable perfomance impact.

Add an option to keep these as regular nodes that do not require recompiling the shader.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -980,7 +980,8 @@ void ShaderGraph::optimize(GenContext& context)
         {
             // Constant nodes can be elided by moving their value downstream.
             bool canElide = context.getOptions().elideConstantNodes;
-            if (!canElide) {
+            if (!canElide)
+            {
                 // We always elide filename constant nodes regardless of the
                 // option. See DOT below.
                 ShaderInput* in = node->getInput("value");
@@ -989,7 +990,8 @@ void ShaderGraph::optimize(GenContext& context)
                     canElide = true;
                 }
             }
-            if (canElide) {
+            if (canElide)
+            {
                 bypass(node, 0);
                 ++numEdits;
             }


source/MaterialXGenShader/ShaderNode.cpp
```diff
@@ -292,7 +292,7 @@ ShaderNodePtr ShaderNode::create(const ShaderGraph* parent, const string& name,
         newNode->_classification = Classification::VDF | Classification::CLOSURE;
     }
     // Second, check for specific nodes types
-    else if (nodeDef.getNodeString() == CONSTANT)
+    else if (nodeDef.getNodeString() == CONSTANT && context.getOptions().elideConstantNodes)
```
ld-kerley: Instead of checking the option here - I think it's clearer to apply the same logic inside `ShaderGraph::optimize()` when the constant nodes are actually being elided [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/ShaderGraph.cpp#L904).

I know we only use the `CONSTANT` classification for this at the moment, but if we ever used it for something else this would need to be refactored.

JGamache-autodesk: I was going to say it was a matter of taste, but you are actually correct since non-elided ND_constant_filename nodes will become an issue in GLSL, which means we need to change the condition in [ShaderGraph::optimize()](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/ShaderGraph.cpp#L910) from
```C++
else if (node->hasClassification(ShaderNode::Classification::DOT))
```
to
```C++
else if (node->hasClassification(ShaderNode::Classification::DOT) || 
         node->hasClassification(ShaderNode::Classification::CONSTANT))
```
Mentioning this also opens another can of worms, because that is a vital elision only when the filename is replaced by a sampler, and only if the shading language does not allow passing samplers as function parameters. But that is a discussion for another time.

So my recommendation is to wait until I have completed @ld-kerley 's request and added proper elision at least for GLSL.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -971,16 +971,28 @@ void ShaderGraph::disconnect(ShaderNode* node) const
...
     size_t numEdits = 0;
     for (ShaderNode* node : getNodes())
     {
         if (node->hasClassification(ShaderNode::Classification::CONSTANT))
         {
             // Constant nodes can be elided by moving their value downstream.
-            bypass(node, 0);
-            ++numEdits;
+            bool canElide = context.getOptions().elideConstantNodes;
+            if (!canElide) {
```
jstone-lucasfilm: I see some mixing of code styles here, and make sure to align all of your new code to the Allman-style braces of MaterialX before we merge this.

JGamache-autodesk: Done.

PR #2522, opened 2025-08-13
Assign source URI to generated OCIO NodeDefs

JGamache-autodesk:
Also Implementations. Without a source URI these can not be distinguished from user created nodes and filtered out when writing.


jstone-lucasfilm (approved):
This looks like a worthwhile improvement to the new OCIO support in 1.39.4, and it seems logical to include it in the same release.

PR #2523, opened 2025-08-13
GlslRender was hardcoding SRGB even is framebuffer was set to raw

JGamache-autodesk:


source/MaterialXRenderGlsl/GLFramebuffer.cpp
```diff
@@ -27,7 +27,7 @@ GLFramebuffer::GLFramebuffer(unsigned int width, unsigned int height, unsigned i
     _height(height),
     _channelCount(channelCount),
     _baseType(baseType),
-    _encodeSrgb(false),
+    _encodeSrgb(true),
```
JGamache-autodesk, line 30: Made it the default, since GlslRenderer was enforcing it for so long.

PR #2533, opened 2025-08-25
Simplify `UnitSystem::getNodeDef` function

ld-kerley:
Remove unnecessary loop over output ports from UnitSystem::getNodeDef() function.  Should be completely transparent, other than possibly slightly faster in  cases with no matching node def and multiple outputs.


jstone-lucasfilm (approved):
This looks like a good change to me, @ld-kerley, and I'm CC'ing @kwokcb in case he catches any issues that I'm missing.

PR #2536, opened 2025-08-28
MDL: Fix the use of the weight in the subsurface BSDF

krohmerNV:
The subsurface weight should not be applied to the expected color but instead enable/disable the effect in general.
This means the semantic of the weight is meant for composition rather than turning down the volume absorption and scattering itself.

With the performance optimizations for Standard Surface and OpenPBR this issue surfaced since weights are not 1.0 anymore.


krohmerNV:
Note, i added a few more materials to the tests because the pbrlib tests have been limited to surfaceshaders only. I'm not sure this was intentional.

jstone-lucasfilm (approved):

PR #2537, opened 2025-08-29
Consolidate `requiresLighting` in HwShaderGenerator

ld-kerley:
Both the GlslShaderGenerator and the MslShaderGenerator have identical functions to decide if lighting is required, so these can be consolidated to the parent class HwShaderGenerator.

Also in HwShaderGenerator there is the following piece of code

```
   bool lighting = graph->hasClassification(ShaderNode::Classification::SHADER | ShaderNode::Classification::SURFACE) ||
                    graph->hasClassification(ShaderNode::Classification::BSDF);
```
Which appears to have the same intention, but is missing the guard for not being `Classification::UNLIT`.  I believe this may be a small bug, so using the same `requiresLighting()` function here may make things a little more efficient too.


niklasharrysson (approved):
Nice cleanup! 

And yes, I think you're right that the missing check for UNLIT is a bug. The UNLIT classification was introduced later and we must have missed adding this here. Good catch! :)

jstone-lucasfilm (approved):

PR #2538, opened 2025-09-01
Modernize value string logic for C++17

jstone-lucasfilm:
This changelist modernizes the value string logic in `toValueString` and `fromValueString`, replacing C++11 SFINAE templates with C++17 `if constexpr` statements, and removing the intermediate stringToData and dataToString functions.


ld-kerley (approved):
Looks good to me - I'm assuming you'll merge this after the 1.39.4 RC phase has closed?

jstone-lucasfilm:
@ld-kerley Thanks for the review!  And yes, this seems like a good candidate for the early phases of 1.39.5, as we'll need wider testing than we can achieve within 1.39.4.

PR #2539, opened 2025-09-01
Restrict optimized BSDF graphs to hardware targets

jstone-lucasfilm:
This changelist restricts the optimized BSDF graphs for Standard Surface, glTF PBR, and OpenPBR Surface to hardware shading targets, falling back to the original unoptimized graphs on other targets.

The optimized graphs now live within a `libraries/bxdf/genglsl` folder with a `glsl` target assignment, and the generic graphs in `libraries/bxdf` have been restored to their pre-optimized versions.

In the future, we're hoping to move in the direction of automated optimization of shading model graphs, which would allow each renderer to opt in or out of this category of optimization through GenOptions settings.

### DIFF
--- a/libraries/bxdf/genglsl/gltf_pbr.mtlx
+++ b/libraries/bxdf/genglsl/gltf_pbr.mtlx
@@ -1,5 +1,8 @@
 <?xml version="1.0"?>
 <materialx version="1.39">
+  <!-- An optimized graph implementation for hardware shading languages,
+       improving the effectiveness of dynamic branches. -->
+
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm:
GitHub doesn't let me yet add @krohmerNV as a formal reviewer, so I'm CC'ing him here in the comments.

krohmerNV:
I like that approach. For intermediate solution it's perfect.
I tried to same pattern with OpenPBR layering changes in  #2215 and there it works good as well.

ld-kerley (approved):
looks neat and tidy to me - thanks for putting this together

PR #2541, opened 2025-09-03
Fix Python Interfaces For Image Loaders

### LINKED ISSUE #2528: Python Interface for ImageLoader and Image cannot support custom loaders
## Issue

The current Python (pybind11) interface for I`mageLoader` and `Image` needs modifications
to allow for support for custom  image loaders.

Namely:
- It is not possible to read and write `void*` buffers via the API so load, save, edit cannot work.
- There are no trampoline interfaces to allow calling into the Python implementation
- The supportedExtensions() api does allow references so no data can be returned.

`ImageHandler` seems to work properly so no changes would be required there.

## Application
- This has been tried by trying to use the Python package for OIIO to provide exr support when it's not part of the release build.

kwokcb:
## Updates

- Fixes #2528
- Patches to make image loaders declared in Python to work properly.
  - Fixes image data passing (void*) cannot resolve properly.
  - Virtualize extension support query and pass by reference to get data returned properly from derived classes.
  - Add in trampoline classes so that derived classes get called properly
  - Fix missing type enumerations in pybind bindings.

## Tests
- Sample OIIO Loader  test added along with sample image from [OpenEXR tests](https://openexr.com/en/latest/test_images/index.html)
   - Registers the custom load with `ImageHander` interface
   - Calls `aquireImage` which will check extensions EXR is not by default supported so will find the sample OIIO loader
   - Calls virtual `loadImage` override. Also displays the image using `matplotlib` if available
   - The MaterialX `Image` is cached in memory after load.
   - Calls `saveImage` on the `ImageHandler`  passing the MaterialX `Image` which will again check extensions and call back to the OIIO loader.
   - Calls `saveImage` override. Also displays the imsage using `matplotlib` if available.
   
## Results:
- EXR Image loaded 
image

- EXR Image saved
image


source/PyMaterialX/PyMaterialXRender/PyImage.cpp
```diff
@@ -10,11 +10,18 @@
 namespace py = pybind11;
 namespace mx = MaterialX;
 
+uintptr_t getResourceBuffer(const mx::Image& image)
```
kwokcb, line 13: Cannot pass back and forth void pointers for buffer data. 
Add a concrete type to wrap.

source/PyMaterialX/PyMaterialXRender/PyImageHandler.cpp
```diff
@@ -18,7 +18,39 @@ void bindPyImageHandler(py::module& mod)
         .def_readwrite("filterType", &mx::ImageSamplingProperties::filterType)
         .def_readwrite("defaultColor", &mx::ImageSamplingProperties::defaultColor);
 
-    py::class_<mx::ImageLoader, mx::ImageLoaderPtr>(mod, "ImageLoader")
+    // Trampoline class for Python overrides
```
kwokcb, line 21: Since this glue code was not added, Python loader code would never be called.

source/PyMaterialX/PyMaterialXRender/PyImage.cpp
```diff
@@ -10,11 +10,18 @@
...
+{
+    return reinterpret_cast<uintptr_t>(image.getResourceBuffer());
+}
+
 void bindPyImage(py::module& mod)
 {
     py::enum_<mx::Image::BaseType>(mod, "BaseType")
         .value("UINT8", mx::Image::BaseType::UINT8)
+        .value("INT8", mx::Image::BaseType::INT8)
```
kwokcb, line 22: Add missing types.

python/Scripts/oiio_loader.py
```diff
@@ -0,0 +1,429 @@
+"""
```
kwokcb, line 1: Sample code which works for EXR and other image types.

jstone-lucasfilm, line 1: Just to match the surrounding naming conventions, I'd recommend naming this script `oiio_loader.py`.

resources/Images/gamma_chart.exr
jstone-lucasfilm, line 1: Just to match the surrounding naming conventions, I'd recommend naming this image `gamma_chart.exr`.

jstone-lucasfilm (commented):
Overall this looks good to me, thanks @kwokcb, and I had just minor notes on following surrounding naming conventions for new files.

PR #2548, opened 2025-09-08
Add GenSlang, SlangRenderer, and RenderSlang

tdavidovicNV:
- Adds MaterialXGenSlang, matching GenGlsl as closely as possible.
    - The genslang `.mtlx` files were created by copying the genglsl files.
    - Various `mx_function.slang` files were created by copying `mx_function.glsl` files and doing the minimal required changes.
    - SlangShaderGenerator generates vertex and pixel shader, again to mimic the GenGlsl (and reuse as much of the existing Hw code).
    - Uniform defaults in the code are not supported by Slang, so SlangShaderGenerator generates custom user attributes, that the application needs to read and set.
        - This is the major compatibility issue that needs to be resolved. Another way would be a separate kernel call to move all uniforms into a struct, and have a separate kernel call to set the defaults. It would deviate from the existing Glsl code, and still wouldn't be completely self-contained shader.
    - Code is not using Strings, despite Slang supporting them. Can be added in the future, replacing the string-to-enum code.
    - It is configured as off by default.

- Adds MaterialXRenderSlang, matching the `ShaderRenderer` interface.
    - Used exclusively for testing.
    - Uses `slang-rhi` library to abstract the specific graphics API, allowing to run identical code on D3D12 and Vulkan.
        - Mac and WebGPU are also supported by both Slang and slang-rhi, but haven't been tested on this codebase yet.
    - Like the MaterialXRenderMsl, the immediate (OpenGL) interface of `ShaderRenderer` doesn't really fit the modern graphics API, so not all interfaces are implemented.
    - `SlangRederer::initialize` takes `const char*` of the required device type. If none is provided, default one for the given architecture is chosen instead.
    - `MaterialXRenderSlang` is explicitly taken out of `install` targets, because `slang-rhi` is currently not ready for that workflow.

- Adds `RenderSlang` that uses `MaterialXRenderSlang` to run tests.
    - Only uses the `Default` device type for now, i.e., `D3D12` on Windows and `Vulkan` on Linux.

### DIFF
--- a/documents/Specification/MaterialX.StandardNodes.md
+++ b/documents/Specification/MaterialX.StandardNodes.md
@@ -72,68 +72,118 @@ Standard Texture nodes:
 
 <a id="node-image"> </a>
 
-* **`image`**: samples data from a single image, or from a layer within a multi-layer image.  When used in the context of rendering a geometry, the image is mapped onto the geometry based on geometry UV coordinates, with the lower-left corner of an image mapping to the (0,0) UV coordinate (or to the fractional (0,0) UV coordinate for tiled images).
-The type of the &lt;image> node determines the number of channels output, which may be less than the number of channels in the image file, outputting the first N channels from the image file.  So a `float` &lt;image> would return the Red channel of an RGB image, and a `color3` &lt;image> would return the RGB channels of an RGBA image.  If the type of the &lt;image> node has more channels than the referenced image file, then the output will contain zero values in all channels beyond the N channels of the image file.
-    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.
-    * `layer` (uniform string): the name of the layer to extract from a multi-layer input file.  If no value for `layer` is provided and the input file has multiple layers, then the "default" layer will be used, or "rgba" if there is no "default" layer.  Note: the number of channels defined by the `type` of the `<image>` must match the number of channels in the named layer.
-    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;_geometry token_>, [_interface token_] or {_hostattr_} is included in the filename but no substitution value or default is defined, or if the resolved `file` URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
-    * `texcoord` (vector2): the name of a vector2-type node specifying the 2D texture coordinate at which the image data is read.  Default is to use the current u,v coordinate.
-    * `uaddressmode` (uniform string): determines how U coordinates outside the 0-1 range are processed before sampling the image; see below.  Default is "periodic".
-    * `vaddressmode` (uniform string): determines how V coordinates outside the 0-1 range are processed before sampling the image; see below.  Default is "periodic".
-    * `filtertype` (uniform string): the type of texture filtering to use; standard values include "closest" (nearest-neighbor single-sample), "linear", and "cubic".  If not specified, an application may use its own default texture filtering method.
+### `image`
+
+Samples data from a single image, or from a layer within a multi-layer image.  When used in the context of rendering a geometry, the image is mapped onto the geometry based on geometry UV coordinates, with the lower-left corner of an image mapping to the (0,0) UV coordinate (or to the fractional (0,0) UV coordinate for tiled images).
+
+The type of the `image` node determines the number of channels output, which may be less than the number of channels in the image file, outputting the first N channels from the image file.  So a `float` image would return the Red channel of an RGB image, and a `color3` image would return the RGB channels of an RGBA image.  If the type of the `image` node has more channels than the referenced image file, then the output will contain zero values in all channels beyond the N channels of the image file.
+
+The `file` input can include one or more substitutions to change the file name that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.  The `filtertype` input supports options `closest` (nearest-neighbor single-sample), `linear`, and `cubic`.
+
+The `layer` input is the name of the layer to extract from a multi-layer input file. If no value for `layer` is provided and the input file has multiple layers, then the "default" layer will be used, or "rgba" if there is no "default" layer. Note: the number of channels defined by the type of the `image` must match the number of channels in the named layer.
+
+The `default` input is the default value to use if the file reference can not be resolved (e.g. if a geometry token, interface token, or hostattr is included in the filename but no substitution value or default is defined, or if the resolved file URI cannot be read), or if the specified layer does not exist in the file. The default value must be the same type as the `image` element itself. If default is not defined, the default color value will be 0.0 in all channels.
+
+|Port            |Description                                                                                                      |Type                  |Default  |Accepted Values                  |
+|----------------|-----------------------------------------------------------------------------------------------------------------|----------------------|---------|---------------------------------|
+|`file`          |The URI of an image file                                                                                         |filename              |__empty__|                                 |
+|`layer`         |The name of the layer to extract from a multi-layer input file                                                   |string                |__empty__|                                 |
+|`default`       |A default value to use if the file reference can not be resolved                                                 |float, colorN, vectorN|__zero__ |                                 |
+|`texcoord`      |The 2D texture coordinate at which the image data is read                                                        |vector2               |_UV0_    |                                 |
+|`uaddressmode`  |Determines how U coordinates outside the 0-1 range are processed before sampling the image                       |string                |periodic |constant, clamp, periodic, mirror|
+|`vaddressmode`  |Determines how V coordinates outside the 0-1 range are processed before sampling the image                       |string                |periodic |constant, clamp, periodic, mirror|
+|`filtertype`    |The type of texture filtering to use                                                                             |string                |linear   |closest, linear, cubic           |
+|`framerange`    |A string 'minframe-maxframe', e.g. '10-99', to specify the range of frames that the image file is allowed to have|string                |__empty__|                                 |
+|`frameoffset`   |A number that is added to the current frame number to get the image file frame number                            |integer               |0        |                                 |
+|`frameendaction`|What to do when the resolved image frame number is outside the `framerange` range                                |string                |constant |constant, clamp, periodic, mirror|
+|`out`           |Output: the sampled texture value                                                                                |Same as `default`     |__zero__ |                                 |
 
 <a id="node-tiledimage"> </a>
 
-* **`tiledimage`** (NG): samples data from a single image, with provisions for tiling and offsetting the image across uv space.
-    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.
-    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;geomtoken>, [interfacetoken] or {hostattr} is included in the filename but no substitution value or default is defined, or if the resolved file URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
-    * `texcoord` (vector2): the name of a vector2-type node specifying the 2D texture coordinate at which the image data is read.  Default is to use the current u,v coordinate.
-    * `uvtiling` (vector2): the tiling rate for the given image along the U and V axes. Mathematically equivalent to multiplying the incoming texture coordinates by the given vector value. Default value is (1.0, 1.0).
-    * `uvoffset` (vector2): the offset for the given image along the U and V axes. Mathematically equivalent to subtracting the given vector value from the incoming texture coordinates. Default value is (0.0, 0.0).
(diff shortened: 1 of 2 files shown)


libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
```diff
@@ -32,7 +32,7 @@ vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 alpha, int distributio
         float VdotH = clamp(dot(V, H), M_FLOAT_EPS, 1.0);
 
         // Sample the environment light from the given direction.
-        vec3 Lw = tangentToWorld * L;
+        vec3 Lw = mx_matrix_mul(tangentToWorld, L);
```
jstone-lucasfilm, line 35: @tdavidovicNV This solution to the issue of cross-platform matrix multiplication looks good to me!

What would you think about the idea of splitting this change into its own PR, so that we can review and merge the cross-platform matrix multiplication support first, and then review and merge the current Slang PR?

tdavidovicNV: Sounds like a good idea.

I would also propose unifying at least some of the nodes that are unique to Msl and Glsl with no apparent reason

jstone-lucasfilm: That sounds great to me, and feel free to address any obvious examples of unnecessary MSL/GLSL splitting as you notice them.  If there are good reasons for any of these to exist, we can catch them in the review process.

kwokcb:
For this statement:

> Uniform defaults in the code are not supported by Slang, so SlangShaderGenerator generates custom user attributes, that the application needs to read and set.
> This is the major compatibility issue that needs to be resolved. Another way would be a separate kernel call to move all uniforms into a struct, and have a separate kernel call to set the defaults. It would deviate from the existing Glsl code, and still wouldn't be completely self-contained shader.

ESSL and Vulkan seem to be in the same situation. There is no additional meta data added in these cases. Thus I believe you can also consider them to not be self-contained. 

The code generator creates a "Shader graph" and the API for this contains things like default value are part of it's reflection structure. Would it make sense to handle default bindings that way -- at least for the renderer you provide ?

I will ping @ashwinbhat since he did build reflection data for DX12 so can provide better insights.

tdavidovicNV:
@kwokcb that is a good point. And I was wondering about it.

To me, the problematic case is this call:
```
    /// Load shader source from file.
    virtual bool loadSource(const FilePath& vertexShaderFile,
                            const FilePath& pixelShaderFile,
                            bool hasTransparency) = 0;
```
or (more Glsl specific, but exposed in Python bindings)
```
void GlslProgram::addStage(const string& stage, const string& sourceCode)
```

If you do *not* have the defaults baked in the source code and depend only on the values coming from the compiled shader, than this code simply cannot work correctly.

And I am generally very suspicious of two codepaths that seems like they should be doing the same thing not doing the same thing. Bottomline, my thinking was that if OpenGL backend renders the same whether it gets the shader from graph or via this call, all backends should render the same. (Or the call should be removed).

But I will rather happily remove that whole part, if the `loadSource` is not considered an integral part of MaterialX.

libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
tdavidovicNV: The suggested split is here: https://github.com/AcademySoftwareFoundation/MaterialX/pull/2556

The unified nodes (removing unnecessary Msl/Glsl splitting) are here: https://github.com/AcademySoftwareFoundation/MaterialX/pull/2555

The noted caveat is that the letter is obviously not using the syntax from the former, so depending on the order they go in, I will fix the other.

tdavidovicNV:
I've now updated to the new Hw nodes (so deleted all the Slang-specific nodes as they're no longer needed).

But I'd really like to discuss the broader topic of defaults for languages that can't have uniform defaults in the shader code, before we move forward.

tdavidovicNV:
I have updated the code to:
- Only get defaults from the MaterialX shader, not via the source code attribute hackery, as it overcomplicated the matter and wasn't universal anyway.
- Removed the torso of handling array uniforms, as it wasn't setting values correctly anyway, and there doesn't really seem to be much current support for it in the implementation. (I'd rather not push things that aren't needed and/or do not have tests running).

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "source/MaterialXGraphEditor/External/ImGuiNodeEditor"]
 	path = source/MaterialXGraphEditor/External/ImGuiNodeEditor
 	url = https://github.com/thedmd/imgui-node-editor
+[submodule "source/MaterialXRenderSlang/External/slang-rhi"]
```
jstone-lucasfilm: Instead of adding `slang-rhi` as a submodule of MaterialX, I might suggest adding instructions to our Per-Language Render Setup on how to install `slang-rhi`, following the pattern we use for OSL and MDL rendering:

https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/source/MaterialXTest#per-language-render-setup

This will give developers more control over when and how `slang-rhi` is installed on their local machines, and will remove this overhead for developers that aren't yet working with Slang.

tdavidovicNV: The slang-rhi is fundamentally different from OSL and MDL, tho. Those point to an external binary that is passed the shader which is then executed. 

SlangRhi is more akin to OpenGL or DX12 headers. It is the interface you use to set buffers and textures and so on. The SlangRenderer is more like the OpenGL or MSL renderer, directly invoking API calls, rather than the OSL or MDL renderer, that just shuffle an existing shader source code to an external binary.

Importantly, since Slang does not store the defaults in the source code, it cannot actually run from a standalone shader source at this point. I can put back in the defaultvalues attribute and spin off the whole RenderSlang into a separate binary but the actual generated shader won't be of much use to anyone not using that code.

jstone-lucasfilm: @tdavidovicNV This is an area where I definitely need to learn more, and let's plan to discuss this in today's TSC meeting.  Just providing my initial naive viewpoint, if SlangRhi is analogous to the OpenGL or DirectX headers, I would still imagine that a developer would want to take control over where these are located on their machine, so that no unintended duplication of these headers is introduced by MaterialX.  I'm sure I'm missing key details here, though, and let's continue the discussion so that I can better understand them.

tdavidovicNV: It is definitely possible to have it compiled as a stand-alone project and point to it and the slang DLLs. But we haven't really tested that workflow yet (it is not how we use the library internally).

ld-kerley: @jstone-lucasfilm - I'm curious how this differs from the existing git submodules - in terms of how the project thinks people should acquire its dependencies. It feels like there are two different conventions here.  Maybe its a good idea for the project to take a single stance around this.

tdavidovicNV:
After today's discussion, I definitely want to take another look at the split textures (separate texture and sampler), if only because my current way of handling it (packaged both in a struct) isn't compatible with Slang Playground.

jstone-lucasfilm:
Apologies @tdavidovicNV, we've recently merged PR https://github.com/AcademySoftwareFoundation/MaterialX/pull/2565, which has some overlap with your PR, and we'd appreciate your merging and resolving conflicts when you have a chance.

The gist of https://github.com/AcademySoftwareFoundation/MaterialX/pull/2565 is that all of the `MaterialXGenHw` source and header files have been moved to a dedicated `MaterialXGenHw` module, which should make it more straightforward to share code between hardware shading languages in the future.

tdavidovicNV:
@jstone-lucasfilm I've updated to the latest main. The move of those files was no big issue. 

I've updated:
- The new include paths.
- Added `AIRY_FRESNEL_ITERATIONS`
- Added `mx_float_bits_to_int` (not used by anything, it seems)
- Included the `mx_shadow_platform` (again, shadows are not actually tested)
- Removed the dedicated `SlangShaderGenerator::getImplementation`

I have decided to keep the `SampleTexture2D` struct as the representation for `Texture+SampleState`, copying the Metal's approach. Splitting them (like WGSL) would make the Slang Playground integration a bit easier, but having `Mx::FILENAME` correspond to a single variable is more Slang-like. 

I believe the code is pretty much ready to be merged in this form. The follow-up work will include:
- Slang team taking a look at the generated code w.r.t. different backends.
- Modifying the generated code to fit with Slang Playground better (e.g., emitting a single file, possibly emitting a harness that will run the material using a compute shader on analytical sphere).

jstone-lucasfilm:
Much appreciated, @tdavidovicNV, and I'll reserve some time to read through this in depth!

CMakeLists.txt
```diff
@@ -133,6 +135,15 @@ set(MATERIALX_OSL_BINARY_OSLC "" CACHE FILEPATH "Full path to the OSL compiler b
 set(MATERIALX_OSL_BINARY_TESTRENDER "" CACHE FILEPATH "Full path to the OSL test render binary.")
 set(MATERIALX_OSL_INCLUDE_PATH "" CACHE PATH "Full path to OSL shader includes (e.g. 'stdosl.h').")
 
+# Helpers for Slang validation
+set(MATERIALX_SLANG_RHI_SOURCE_DIR "" CACHE FILEPATH "Full path to the Slang RHI build directory.")
```
jstone-lucasfilm: This variable needs to be of type `PATH` rather than type `FILEPATH`, so that the user can assign it to a folder rather than a filename from the CMake GUI.

CMakeLists.txt
```diff
@@ -171,6 +182,7 @@ set(MATERIALX_LIBNAME_SUFFIX "" CACHE STRING "Specify a suffix to all libraries
 
 mark_as_advanced(MATERIALX_BUILD_DOCS)
 mark_as_advanced(MATERIALX_BUILD_GEN_GLSL)
+mark_as_advanced(MATERIALX_BUILD_GEN_SLANG)
```
jstone-lucasfilm, line 174: Make sure to also mark the new `MATERIALX_SLANG_RHI_SOURCE_DIR` as advanced, so that it doesn't appear at the top level of a user's CMake GUI interface.

CMakeLists.txt
```diff
@@ -425,7 +437,7 @@ function(mx_add_library MATERIALX_MODULE_NAME)
 
     target_compile_definitions(${TARGET_NAME} PRIVATE "-D${args_EXPORT_DEFINE}")
 
-    if(NOT SKBUILD)
+    if(NOT SKBUILD AND NOT args_SKIP_INSTALL)
```
jstone-lucasfilm, line 425: I'm not familiar with this use of the `args` prefix -- is it an intentional choice or a typo?

jstone-lucasfilm (commented):
This is looking really promising, @tdavidovicNV, and I'm highlighting a couple of hurdles that I encountered in my initial local tests.

CMakeLists.txt
ld-kerley: this is standard CMake convention for parsing arguments to function - and we're following it above in the rest of the the code for this function. 

`assign_source_group("Source Files" ${args_SOURCE_FILES})`

CMakeLists.txt
tdavidovicNV: Done

CMakeLists.txt
tdavidovicNV: I also took note from the MDL handling, and only set `MATERIALX_SLANG_RHI_SOURCE_DIR` when `MATERIALX_BUILD_GEN_SLANG` is on. (Saw that MDL had its extra options guarded by `MATERIALX_BUILD_GEN_MDL` when setting it as_advanced).

.gitmodules
tdavidovicNV: Just to conclude this thread. We arrived at the option of `slang-rhi` source just living outside of the MaterialX repository. So it is still compiles the same way it would as a submodule, it just isn't part of the repository itself, meaning it is not tied to any particular version.

The next step, having actually slang-rhi be distributed as binaries is some way off, and we're debating that the library should probably end up being an integral part of the Slang library sooner rather than later, so maybe it will never have its own binary distribution/install process and be either part of Slang, or a standalone source repo.

jstone-lucasfilm (commented):
@tdavidovicNV I've had a chance to test this work in my local Windows environment, and I think it's really good, and is practically ready to merge to `main`.

One thought on a potential improvement:

Could we enable `MATERIALX_BUILD_GEN_SLANG` _by default_ in MaterialX, and simply skip Slang render tests when the developer hasn't also provided a folder for `MATERIALX_SLANG_RHI_SOURCE_DIR`?

This would mirror our approach in OSL and MDL, where we allow artists and developers to generate OSL/MDL shaders from MaterialX documents in virtually all contexts, since that can be an important element in workflows that don't have immediate access to a rendering framework for those shaders.

tdavidovicNV:
> @tdavidovicNV I've had a chance to test this work in my local Windows environment, and I think it's really good, and is practically ready to merge to `main`.
> 
> One thought on a potential improvement:
> 
> Could we enable `MATERIALX_BUILD_GEN_SLANG` _by default_ in MaterialX, and simply skip Slang render tests when the developer hasn't also provided a folder for `MATERIALX_SLANG_RHI_SOURCE_DIR`?
> 
> This would mirror our approach in OSL and MDL, where we allow artists and developers to generate OSL/MDL shaders from MaterialX documents in virtually all contexts, since that can be an important element in workflows that don't have immediate access to a rendering framework for those shaders.

@jstone-lucasfilm that is no problem, but I think I will have to remove the current warning that: `"MATERIALX_SLANG_RHI_SOURCE_DIR must be set to build MaterialXRenderSlang."`, because we wouldn't want to warn in the default setup?

Can you think of a better idea there?

tdavidovicNV:
I've turned `MATERIALX_BUILD_GEN_SLANG` on by default, and removed the warning about RenderSlang

jstone-lucasfilm:
@tdavidovicNV That's a fine approach for now, and in the future we might consider noting the reason that RenderSlang was not built as an informational message, without making this a warning condition that developers should be concerned about.

PR #2550, opened 2025-09-10
Fixed image path in test suite example

tdavidovicNV:
The image path was missing one `../` to reach the required level, but instead of adding that I changed it to the "full path" to match the rest of the test cases.


jstone-lucasfilm (approved):

PR #2552, opened 2025-09-11
Fix GLSL type in LightCompoundData node

tdavidovicNV:
In prep for the Slang merge, I was looking how close are the remaining Nodes between GenMsl and GenGlsl. And I found that the GenGlsl one was using `float3` type instead of the `vec3`. I don't believe there is any alias for that, and even if there was, I think using the native type is the better idea. Also, it seems like there is no test coverage for this node, as I was running it with full tests (modified the `_options` to just point to `resources/Materials`) and this didn't come up.


ld-kerley:
Its always been a little unclear to me how the lights in MaterialX are intended to be used - so I was more asking if you have any concrete examples we can add - or have any insight as to how the lights are intended to be used, that would help me be able to construct some unit tests.

Issue #2551, opened 2025-09-10
`atan2` defaults are inconsistent across types


ld-kerley:
```xml
  <nodedef name="ND_atan2_float" node="atan2" nodegroup="math">
    <input name="iny" type="float" value="0.0" />
    <input name="inx" type="float" value="1.0" />
    <output name="out" type="float" defaultinput="iny" />
  </nodedef>
  <nodedef name="ND_atan2_vector2" node="atan2" nodegroup="math">
    <input name="iny" type="vector2" value="1.0, 1.0" />
    <input name="inx" type="vector2" value="0.0, 0.0" />
    <output name="out" type="vector2" defaultinput="iny" />
  </nodedef>
  <nodedef name="ND_atan2_vector3" node="atan2" nodegroup="math">
    <input name="iny" type="vector3" value="1.0, 1.0, 1.0" />
    <input name="inx" type="vector3" value="0.0, 0.0, 0.0" />
    <output name="out" type="vector3" defaultinput="iny" />
  </nodedef>
  <nodedef name="ND_atan2_vector4" node="atan2" nodegroup="math">
    <input name="iny" type="vector4" value="1.0, 1.0, 1.0, 1.0" />
    <input name="inx" type="vector4" value="0.0, 0.0, 0.0, 0.0" />
    <output name="out" type="vector4" defaultinput="iny" />
  </nodedef>
```

The default values for `iny` and `inx` should be consistent across the different types.

ld-kerley:
Do we feel like this is a small enough bug we can fix it now - or because it is a data library change, do we need for 1.40?

jstone-lucasfilm:
Good catch, @ld-kerley!

Although this is a tricky edge case where our specification and data libraries disagree, my intuition is that we can fix this in a point release, as the incorrect default values in these vector-typed `atan2` nodes are so rarely used, and their role in MaterialX is more of a formality than a practical technique in authored assets.

What are your thoughts?

ld-kerley:
I think I agree with your intuition - but since changing this would have the possible effect of changing someones render - I feel like we need to broadcast this change as widely and loudly as we can, to ensure we're not going to inadvertently have a negative impact somewhere.

I'll start a dedicated thread on the Materialx slack channel - and I think we should raise this as an important topic in the next TSC meeting.

PR #2554, opened 2025-09-11
Fix incorrect default values for `atan2`

ld-kerley:
This addressed #2551 

The `iny` and `inx` ports have the incorrect default values, for vector2/3/4 node definitions.

This PR corrects that.

This is technically a breaking change - so we want to ensure we get wide community consensus before merging this. I propose we highlight this change at the next MaterialX TSC.


ld-kerley:
Seems like there was no objection here - I'd vote to merge this now.

PR #2555, opened 2025-09-12
Unify remaining GLSL and MSL nodes as HW nodes

tdavidovicNV:
The diff between the Msl and Glsl nodes showed that the differences were almost exclusively in the types (e.g., `vec3` vs `float3`), which are already handled by `#define`s in the Msl codepath. I've unified all the nodes as HwKindNode in `MaterialXGenShader/Nodes` where all the other shared nodes are.

I have introduced a `makeClosureData` call with `$closureDataConstructor` implementation.

The reason is that Msl requires constructing structs via initializer lists like so:
```
{CLOSURE_TYPE_REFLECTION, L, V, N, P, occlusion}
```

While GLSL only allows this in versions since 4.6 (or 4.2, see https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Initializer_lists )

Please note that `LightCompoundNodeMsl` on main is generating this code:
```
shadergen.emitLine("ClosureData closureData = ClosureData(CLOSURE_TYPE_EMISSION, float3(0), -L, light.direction, float3(0), 0)", stage);
```
which is invalid.

The other significant change is Msl condition for emitting transmission ignored VDF
```
	if (bsdf->hasClassification(ShaderNode::Classification::BSDF_T)) {
		shadergen.emitLine("ClosureData closureData = {CLOSURE_TYPE_TRANSMISSION, L, V, N, P, occlusion}", stage);
		shadergen.emitFunctionCall(*bsdf, context, stage);
```
changed to (code emitted for VDF, not just transmission):
```
	if (bsdf->hasClassification(ShaderNode::Classification::BSDF_T) || bsdf->hasClassification(ShaderNode::Classification::VDF)) {
		shadergen.emitLine("ClosureData closureData = makeClosureData(CLOSURE_TYPE_TRANSMISSION, L, V, N, P, occlusion)", stage);
		shadergen.emitFunctionCall(*bsdf, context, stage);
```
I believe this might have been a genuine bug.


libraries/pbrlib/genglsl/lib/mx_closure_type.glsl
```diff
@@ -15,3 +15,8 @@ struct ClosureData {
     vec3 P;
     float occlusion;
 };
+
+ClosureData makeClosureData(int closureType, vec3 L, vec3 V, vec3 N, vec3 P, float occlusion)
```
ld-kerley, line 19: Instead of having a token substitution - which increases the coupling between the C++ generator code and the data library. 

Could we move this function to a per language defined file - similar to how we're dealing with mx_math.XXX 

I've been slowly trying to reduce the coupling between the C++ shader generator code and the data library definitions. 

It might be logistically simplest to just add it to mx_math - even though it doesn't feel like it belongs there from a naming perspective.

ld-kerley (commented):
I'll read this more carefully early next week - but I think this a good improvement overall. 

Leaning more in to the HwShaderGenerator idea - makes me start to wonder if we should create a separate MaterialX module `MaterialXGenHw` to house all this HW shader language specific stuff.  People who may build only for MDL or OSL don't need all the additional HWxxx classes.  

Certainly not proposing this as part of this PR - but just starting to socialize the idea for possible future refactoring work. 

I guess one tiny caveat we need to consider - unfortunately all of these things are part of the public API - so its possible that people are inheriting from the classes like `LightShaderNodeGlsl` - and this refactor would break that.

Per the previous PR - it's not clear to me how much any of the Light features in MaterialX are even used - so maybe thats a good community discussion to have.

libraries/pbrlib/genglsl/lib/mx_closure_type.glsl
tdavidovicNV: I thought about putting in `mx_math` because, as you already noted, while dirty on the naming, it is cleaner on the coupling.

But the problem is that `mx_math` gets included before the `ClosureData` definition (because other includes might need the math functions already defined when they are included). So you need the factory function to be defined after ClosureData is defined, and only if ClosureData is even used. Which feels somewhat similar to `$fileTransformUv`, so I went with that route. 

Ideally we would just assume GLSL 4.6+ and use the same initializer list between Msl and Glsl. (And I need to have a closer look at Slang in this, I think either syntax works)

ld-kerley: Interesting... Happy to live with the substitution for now then - as its seems like the easy path. Later once some of this dust has settled, I may see if I can find a way to refactor this away. 

+1 to moving to a GLSL syntax where everything aligns - but that feels like a wider community question that would need consensus.

jstone-lucasfilm:
This looks like a good change to me as well, but let's hold on merging this until @tdavidovicNV has a chance to integrate the usage of `mx_matrix_mul` into this PR.

tdavidovicNV:
I've updated this PR with the `mx_matrix_mul` now.

niklasharrysson:
Nice cleanup and consolidation :)

I like the idea of separating this out to a new `MaterialXGenHw` module. It's a similar separation and naming convention as already used with the `MaterialXRenderXyz` modules.

This PR is already API breaking by renaming and moving the classes, so would be a good opportunity to make that change at the same time, or at least soon after. On the other hand, moving all HW classes would be a bigger breakage, so need to consider that as well, and so it's probably better for another PR.

ld-kerley:
Now this is merged I'm gonna work on putting together a PR for a MaterialXGenHw module - just mentioning here so no one else starts...

PR #2556, opened 2025-09-12
Add `mx_matrix_mul` wrapper for HW shading

tdavidovicNV:
Replacing matrix `operator*` with `mx_matrix_mul` function call, as per discussion here https://github.com/AcademySoftwareFoundation/MaterialX/pull/2548

Please note that this PR is not cleanly compatible with https://github.com/AcademySoftwareFoundation/MaterialX/pull/2555 because https://github.com/AcademySoftwareFoundation/MaterialX/pull/2555 creates new nodes that include matrix multiplication (using operator `*`) that are not included in this PR.

So whichever PR goes in first, the other one will need to adapt, or a third PR will need to be created that fixes the new nodes.


source/MaterialXGenShader/Nodes/HwTransformNode.cpp
```diff
@@ -42,14 +42,17 @@ void HwTransformNode::emitFunctionCall(const ShaderNode& node, GenContext& conte
 
         shadergen.emitLineBegin(stage);
         shadergen.emitOutput(output, true, false, context, stage);
-        shadergen.emitString(" = (", stage);
 
         const string toSpace = getToSpace(node);
         const string fromSpace = getFromSpace(node);
         const string& matrix = getMatrix(fromSpace, toSpace);
         if (!matrix.empty())
```
ld-kerley: Is this if/else clause the right way around?  if `matrix` is not empty - then you're not using `matrix`?

tdavidovicNV: You're right, I was going back and forth on inverting the condition and ended up in the wrong place.
Will fix.

ld-kerley (approved):
Looks good to me - thanks

jstone-lucasfilm (approved):

Issue #2547, opened 2025-09-07
Web Viewer : Packages are incompatible using Node version 16.


kwokcb:
## Issue

The current setup for the Web Viewer is using an incompatible version of Node (16) for builds using webpack (the Viewer).
Unknown how this might affect the release.

Current CI runs have these messages showing up:
```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@sindresorhus/merge-streams@2.3.0',
npm WARN EBADENGINE   required: { node: '>=18' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'bundle-name@4.1.0',
npm WARN EBADENGINE   required: { node: '>=18' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'copy-webpack-plugin@12.0.2',
npm WARN EBADENGINE   required: { node: '>= 18.12.0' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'default-browser@5.2.1',
npm WARN EBADENGINE   required: { node: '>=18' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'default-browser-id@5.0.0',
npm WARN EBADENGINE   required: { node: '>=18' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'globby@14.1.0',
npm WARN EBADENGINE   required: { node: '>=18' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'open@10.2.0',
[...]

Issue #2545, opened 2025-09-05
WebPack Packages are incompatible (won't run) for MaterialXView


kwokcb:
## Issue

The current package versions used for `webpack` are incompatible and won't build run 
on Windows.

```
"webpack": "^5.100.2",
"webpack-cli": "^4.10.0",
"webpack-dev-server": "^5.2.2"
```

- webpack-cli@4 is not compatible with webpack-dev-server@5.
- webpack-dev-server@5 requires webpack-cli@5.

This is an issue with different versions of Node including: 16.20.2 and more recent 18.20.5.

## Proposed fix:

```
npm install --save-dev webpack-cli@^5.1.4
```
which will update `package.json`:
```
"devDependencies": {
  "copy-webpack-plugin": "^12.0.2",
  "html-webpack-plugin": "^5.6.3",
  "webpack-cli": "^5.1.4",
  "webpack-dev-server": "^5.2.2"
}
```

PR #2557, opened 2025-09-12
Patch JavaScript build packages

kwokcb:
## Updates

- Fix the packages used for MaterialXTest and MaterialXView so that they are compatible with the current version of NodeJS used (16.20.2).  
- The default emscripten 2.0.20 utility script sets NodeJS version to a version which is too old and will result in possible incompatibilities.
- Lock the packages versions until such time as NodeJS version is changed. 
  - This will avoid new package versions from accidently being installed (via npm install) which **has already occurred**.
  - For example the build / tests / installs using packages for NodeJS 18 will not work as the current code is not compatible with changes from 16 to 18.

- *Note* This does not try to bump any version of emcripten or NodeJS, it just fixes so that the appropriate packages are now used.

- Fixes: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2547,  https://github.com/AcademySoftwareFoundation/MaterialX/issues/2545

## Tests

- Cleaned out node modules and rebuild locked package manifest.
- Rebuilt, tested and run on Windows local and via CI. CI no longer shows any warnings / error messages.

## Documentation

- Updated with appropriate NodeJS version information
- Fix documentation for emscripten version as not any verson can be used. Note the locked version require.


kwokcb:
@ashwinbhat , @jstone-lucasfilm : I'm not sure if it's too late for this change, but I think it would be better to patch this to have a known safe version of the  packages used for the JS build.

javascript/MaterialXTest/package.json
kwokcb, line 1: Lock these to avoid version "creep". New versions can make it so that tests won't run.

javascript/MaterialXView/package.json
kwokcb, line 1: Fix this as version "creep" has already occurred. This is 100% compatible with the NodeJS version we are locked on for builds (16.20.2). I set this to test locally on Windows.

javascript/build_javascript_win.bat
```diff
@@ -5,8 +5,8 @@
 @rem Edit the following paths to match your local locations for the Emscripten and MaterialX projects.
 set EMSDK_LOCATION=C:/GitHub/emsdk
 set MATERIALX_LOCATION=C:/GitHub/MaterialX
-call %EMSDK_LOCATION%/emsdk.bat install latest
-call %EMSDK_LOCATION%/emsdk.bat activate latest
+call %EMSDK_LOCATION%/emsdk.bat install 2.0.20
```
kwokcb, line 8: This should never have been used as newer versions can be incompatible and not build the C++ code due to API changes.

jstone-lucasfilm:
This change sounds very reasonable to me, @kwokcb, though since it's a significant number of edits, I'd feel more confident waiting until 1.39.5 to move forward with this.

PR #2562, opened 2025-09-16
Update shared library macros for templates on clang-cl

boberfly:
I've been recently building various things in the MSVC-compatible clang-cl and came across some dllimport/dllexport and templates issue with them.

I've tested this with clang-cl 19.1 and MSVC 14.44 and both compile with this change. This was with 1.38.10-OpenPBR and 1.39.3 but I haven't tested on the main branch...


jstone-lucasfilm (approved):
Thanks for the fix, @boberfly, and this approach looks good to me.

PR #2563, opened 2025-09-16
Separate Shader Generation Exception header

ld-kerley:
Create separate header for the shader generation exception. This reduces the dependency on the shader generator header.

Note - the exception header is still included in the main shader generation header for backwards compatibility


source/MaterialXGenMdl/Nodes/CustomNodeMdl.cpp
```diff
@@ -8,7 +8,7 @@
 #include <MaterialXGenMdl/MdlShaderGenerator.h>
 
 #include <MaterialXGenShader/GenContext.h>
-#include <MaterialXGenShader/ShaderGenerator.h>
+#include <MaterialXGenShader/Exception.h>
```
jstone-lucasfilm, line 10: Minor: Within each module block, the individual includes should remain in alphabetical order.

source/MaterialXGenMdl/Nodes/SourceCodeNodeMdl.cpp
```diff
@@ -8,7 +8,7 @@
 #include <MaterialXGenMdl/MdlShaderGenerator.h>
 
 #include <MaterialXGenShader/GenContext.h>
-#include <MaterialXGenShader/ShaderGenerator.h>
+#include <MaterialXGenShader/Exception.h>
```
jstone-lucasfilm, line 10: Minor: Within each module block, the individual includes should remain in alphabetical order.

source/MaterialXGenShader/ColorManagementSystem.cpp
```diff
@@ -6,8 +6,7 @@
 #include <MaterialXGenShader/ColorManagementSystem.h>
 
 #include <MaterialXGenShader/GenContext.h>
-#include <MaterialXGenShader/ShaderGenerator.h>
-#include <MaterialXGenShader/Nodes/SourceCodeNode.h>
+#include <MaterialXGenShader/Exception.h>
```
jstone-lucasfilm, line 8: Minor: Within each module block, the individual includes should remain in alphabetical order.

source/MaterialXGenMdl/Nodes/CustomNodeMdl.cpp
ld-kerley: Thats a little confusing to me - since the `.clang-format` file is setup to not sort the includes alphabetically.

```
SortIncludes: false
```

If we want to adopt this as a convention, should the `.clang-format` file be updated?

jstone-lucasfilm: There may be newer features in `clang-format` that I'm not yet aware of, but the last time I looked closely at its feature set, it did not have the level of flexibility that would be needed to capture and enforce the MaterialX conventions for header includes.

Just picking an arbitrary source file, you can see the current MaterialX convention [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXRenderGlsl/GlslProgram.cpp), where the top-level blocks represent modules, order from highest to lowest, and within each block the includes are ordered alphabetically.

```
#include <MaterialXRenderGlsl/External/Glad/glad.h>
#include <MaterialXRenderGlsl/GlslProgram.h>
#include <MaterialXRenderGlsl/GLTextureHandler.h>
#include <MaterialXRenderGlsl/GLUtil.h>

#include <MaterialXRender/LightHandler.h>
#include <MaterialXRender/ShaderRenderer.h>

#include <MaterialXGenShader/HwShaderGenerator.h>
#include <MaterialXGenShader/Util.h>
```

If this convention doesn't work well for us, I'm open to ideas for new ones, where any new convention would be applied globally to the entire codebase, rather than applied piecemeal to files.

ld-kerley:
I think changing .clang-format from 
```
SortIncludes: false
```
to 
```
SortIncludes: CaseSensitive
IncludeBlocks: Preserve
```

would perform the sorting that I think you're outlining.

I also think it would be really helpful if these coding conventions were either enforced rigorously by clang-format or that they were documented somewhere.

.clang-format
```diff
@@ -12,7 +12,7 @@ IndentPPDirectives: BeforeHash
 IndentWidth: 4
 LambdaBodyIndentation: OuterScope
 PointerAlignment: Left
-SortIncludes: false
```
jstone-lucasfilm, line 15: I'd probably recommend against changing this line in `clang-format`, as I've run into cases where it will have unintended side effects that affect compilation.

Here's one example from `MaterialXFormat/File.cpp`, where the current include order correctly places `windows.h` and `unistd.h` first in their respective blocks:

```
#if defined(_WIN32)
    #define WIN32_LEAN_AND_MEAN
    #include <windows.h>
    #include <direct.h>
#else
    #include <unistd.h>
    #include <sys/stat.h>
    #include <dirent.h>
#endif
```

But with `SortIncludes` set to `true`, this gets reformatted to the following:

```
#if defined(_WIN32)
    #define WIN32_LEAN_AND_MEAN
    #include <direct.h>
    #include <windows.h>
#else
    #include <dirent.h>
    #include <sys/stat.h>
    #include <unistd.h>
#endif
```

ld-kerley: I would still like to find a path forwards to being able to use clang-format.  A lot of reviews get hung up on formatting issues that could be avoided if we could use the tools to enforce the formatting. As well as if we use a formatting tool, it makes rebasing really old branches a LOT easier.

for cases like this - we could easily wrap the section of includes - or any other small part of the code - we don't want auto-formatted in the comments that disable the formatter.

```
// clang-format off
<code that is not reformatted>
// clang-format on
```

This would theoretically give us the best of both worlds.

PR #2569, opened 2025-09-17
Add ShaderGen option for Airy Fresnel iterations

jstone-lucasfilm:
This changelist adds a ShaderGen option for the number of iterations used to compute Airy Fresnel reflections for iridescent materials in hardware shading languages.

The visual difference between iteration counts turns out to be very small for many iridescent materials, and this option should give applications more control over the tradeoff between performance and reference-quality accuracy.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -374,7 +374,7 @@ vec3 mx_fresnel_airy(float cosTheta, FresnelData fd)
 
     // Reflectance term for m>0 (pairs of diracs)
     Cm = Rs - T121.x;
-    for (int m=1; m<=2; m++)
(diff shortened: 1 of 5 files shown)


jstone-lucasfilm:
Performance tests were conducted on an NVIDIA RTX A6000 at 4K resolution, and the following timing results were seen for the OpenPBR Pearl example material:

One Airy Fresnel Iteration: 19ms

OpenPbrSurface_Pearl_AiryIterations_1

Two Airy Fresnel Iterations: 21ms

OpenPbrSurface_Pearl_AiryIterations_2

Four Airy Fresnel Iterations: 24ms

OpenPbrSurface_Pearl_AiryIterations_4

Eight Airy Fresnel Iterations: 30ms

OpenPbrSurface_Pearl_AiryIterations_8

Note that the current default iteration count remans unchanged at two, though we could consider adjusting this default in the future.

niklasharrysson (approved):
Looks good to me!

PR #2572, opened 2025-09-19
Add derivation for OSL field of view

jstone-lucasfilm:
Historically, we've aligned our OSL render tests with GLSL and other languages using an empirically determined field of view, as we didn't yet understand why a default 45-degree field of view didn't produce the expected results in OSL `testrender`.

We now have a better sense of the underlying issue, where OSL `testrender` uses screen-space coordinates in the [-0.5, 0.5] range, while GLSL and other languages use screen-space coordinates in the [-1, 1] range.

This changelist adds a mathematical derivation for the OSL field of view, and replaces the original empirical value with an exact computed value.


fpsunflower:
This is not really intentional. I looked over the code and its indeed using a slightly odd convention that would not align with most renderers.

We should probably fix it - its just a quirk of the current implementation.

jstone-lucasfilm:
@fpsunflower That's good to know, and we'd be happy to handle this setting differently in new builds of OSL, with the compensation factor only used in older builds.

kwokcb:
I'm curious if we can set camera `screen_window` to the desired [-1, 1] range to avoid the computed value ?

fpsunflower:
Yes, but that change needs to be in the `testrender` source itself.

jstone-lucasfilm:
I believe @fpsunflower is correct that we'd need updates on the `testrender` side before we could address this issue with a `screen_window` setting, and at that point it likely makes sense just to fix the interpretation of screen coordinates in `testrender`.

As of today, this change just provides a mathematical explanation for the magic number that we've been using in our `testrender` template, without creating any significant change in visual output.

@kwokcb @niklasharrysson Does this seem like a worthwhile improvement that we should integrate into our codebase for MaterialX 1.39.5?

PR #2577, opened 2025-09-23
Align unused parameter warnings across compilers

ld-kerley:
As noted in by @bernardkwok in [#2385](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2385). We have divergent compiler warnings and errors. This makes developing on a platform with more relaxed rules than others more difficult. 

This PR adds the `-Wunused-parameter` compiler flat to non-windows platforms to bring things in line.  As well as fixing the warnings raised in the code not previously compiled on Windows.

We could make this change in the opposite direction, where the Windows compile flags are relaxed. I'm open to that, but don't develop on Windows so can't do that work.  If that is the desired direction I would hope we can post a PR for that immediately, or merge this and then revert this change when someone is able to relax the Windows warnings.  The current difference between platforms feels like the worst situation for us all to be in.


CMakeLists.txt
```diff
@@ -287,7 +287,7 @@ if(MSVC)
         add_compile_options(/WX)
     endif()
 else()
-    add_compile_options(-Wall -Wno-missing-braces)
+    add_compile_options(-Wall -Wno-missing-braces -Wunused-parameter)
```
jstone-lucasfilm: From a high level, I don't believe the `unused-parameter` warning provides much value to developers, and I'd be more aligned with the idea of _disabling_ it on all platforms, but that's a broader discussion to have with the MaterialX community.

ld-kerley: I'm very happy to have a wider discussion around what is the best set of warnings - but that will take some time - and while we wait for that I would like to propose that we at least align to something. 

As I said above - I'm very happy if you want to put up a PR for removing the the warning on the windows side, instead of this PR - if thats the immediate step we can take to align around.

jstone-lucasfilm:
@ld-kerley I'm very sympathetic to your observation that the diversity of warning behaviors across different compilers can be a headache for developers.

The one that always catches me when I'm working on cross-platform projects is the following warning from Clang, which is not triggered in MSVC builds:

```
error: field '_var1' will be initialized after field '_var2' [-Werror,-Wreorder-ctor]
```

I view the broader landscape differently from you, though, in that I see this diversity of warning behaviors as an _advantage_ for the robustness of a cross-platform project such as MaterialX, as it allows us to see potential issues that only one compiler in the industry happens to catch.

This is one of the main reasons that we focus so strenuously on minimizing _build times_ for our per-commit GitHub Actions workflows, so that developers are encouraged to enable GitHub Actions in their forks of MaterialX, and they can quickly see if one of their proposed changes triggers a warning on a platform other than their own.

Although it's not always the case, I've often found that these compiler-specific warnings hint at a subtle issue that I hadn't noticed and is worth considering, and I can address the underlying issue in my fork of MaterialX before writing up a full PR to the MaterialX project.

I'm very interested in thoughts from other developers in the MaterialX community, to get a sense of whether this diversity of compiler warnings has been a benefit or a hindrance to their development work, and how we can improve our approach going forward.

ld-kerley:
If it's possible to align compiler warnings across development platforms - which it is in this specific case. 

What is the advantage to having some developers (Windows) seeing that warning locally - and forcing others to wait until they run the CI - if they do at all?

ld-kerley:
Just to be completely clear - I am trying to make all platforms equally strict. Bringing the unused parameter warning to non-Windows workflows.

source/MaterialXRenderHw/SimpleWindowIOS.cpp
```diff
@@ -21,8 +21,8 @@ SimpleWindow::SimpleWindow() :
     windowCount++;
 }
 
-bool SimpleWindow::initialize(const char* title,
-                              unsigned int width, unsigned int height,
+bool SimpleWindow::initialize(const char* /*title*/,
```
kwokcb, line 24: I've been making this explcit by adding to the function:
```
(void)arg;
```
but maybe we can use:
```
[[maybe_unused]] arg
```
as I think we are on C++17 and above now ?

kwokcb: Minor: removing the arg on methods that are parsed with Doxygen will cause warnings.

CMakeLists.txt
kwokcb: Beyond the scope of this patch but it seems it would be better to have globals which define the list of flags and reference those flags from each compile options instance. 

List would be based on platform building with:
```
if (MSVC)
   list(APPEND MX_WARNING_FLAGS
       /W4
       /WX
 else()
    list(APPEND MX_WARNING_FLAGS
       -Wall
       -Werror
       -Wunused_parameter
       etc
)
endiif()
```

kwokcb (commented):
Thanks for putting this together. 

It's a good step forward, but have left suggestions on how to mark unused args and how we could possibly have global warning flags that could be consistently used everywhere.

The `(void)arg` is the way I've historically marking unused args but the C++17 way seems better if we use that. 

Thoughts on both appreciated. Thx.

CMakeLists.txt
ld-kerley: Yes I totally agree - I believe there is a "modern CMake" approach to this by creating a target that carries the arguments - and that target is just linked to the relevant projects.  I'm happy to do a pass on this refactor, but I think that makes more sense as a separate CMake clean up PR - where we're just refactoring and not changing behavior at the same time - CMake is hard sometimes :)

source/MaterialXRenderHw/SimpleWindowIOS.cpp
ld-kerley: TIL about `[[maybe_unused]]` - that seems to make most sense to me, assuming we decide to stick with being strict across all platforms.  

If we can get consensus - I'm happy to do a pass over the whole codebase and conform this - either in this PR - or again, maybe a separate PR makes sense here too.

CMakeLists.txt
kwokcb: Yes, this was just a forward thinking sug. 
( I believe we can probably use [target_compile_options](https://cmake.org/cmake/help/latest/command/target_compile_options.html) with Cmake 3.13+ ).

jstone-lucasfilm: Just a minor note on this one line:

I would include the new `-Wunused-parameter` argument before the existing `-Wno-missing-braces` argument, so that we're declaring _enabled_ warnings before declaring _disabled_ warnings.

Issue #2588, opened 2025-09-25
Failing to compile certain codegen'd OSL shaders


JoshuaSenouf:
Hi,

As part of an investigation for https://github.com/AcademySoftwareFoundation/MaterialX/issues/2575, I tried to codegen the `NodeDef` from all the libraries to individual `.osl` files, but some of them are failing to compile, such as:

- `lama_surface`, as it generates the following function, but `surface_material_out` is not defined anywhere:

```
void NG_lama_surface(BSDF materialFront, BSDF materialBack, float presence, output MATERIAL out)
{
    out = surface_material_out;
}
```

- All the `..._filename` shaders, likely because of a syntax error, using `()` instead of `{}` when defining a `textureresource` parameter:

```
reference/osl/geompropvalueuniform_filename.osl:40: warning: Comma operator inside parenthesis is probably an error -- it is not a vector/color.

reference/osl/geompropvalueuniform_filename.osl:40: error: Cannot initialize struct textureresource out = string

reference/osl/geompropvalueuniform_filename.osl:45: error: No matching function call to 'mx_geompropvalue_string (string, struct textureresource, struct textureresource)'

  Candidates are:
    reference/osl/geompropvalueuniform_filename.osl:19    void mx_geompropvalue_string (string, string, string)
```

jstone-lucasfilm:
Thanks to @JoshuaSenouf for this original report, as well as to @ld-kerley for the fix in #2590!

PR #2590, opened 2025-09-25
Fix OSL ShaderGen for `textureresource` and `LamaSurface`

ld-kerley:
As reported in #2588. 

The OSLShaderGenerator fails to generate the correct syntax for shader inputs of type `textureresource` aka `Type::FILENAME`.  Which makes me really curious if we're testing any texture code-path in the OSL unit tests.

Any node with a nodegraph implementation that output a `surfacematerial` type, including`LamaSurface` was also failing to generate because of a mis-match guard check. 

I am also updating the OSL reference generation to generate all nodes in the entire data library to ensure we don't regress in the future. I removed specific exceptions that were added so get the tests to pass, and expanded nods loaded from being only `stdlib` to the entire data library.

### DIFF
--- a/libraries/stdlib/genosl/mx_geompropvalue_filename.osl
+++ b/libraries/stdlib/genosl/mx_geompropvalue_filename.osl
@@ -1,9 +1,9 @@
 void mx_geompropvalue_filename(string geomprop, textureresource defaultVal, output textureresource out)
 {
-    // initialize to the default
+    // Initialize to the default
     out = defaultVal;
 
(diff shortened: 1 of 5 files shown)


JoshuaSenouf:
Tested on my side, all the `NodeDef` are now both codegen'd to `.osl` and compiled to `.oso` as expected (those lacking an OSL implementation aside, which are rightfully skipped), and the test suite is all green, thanks for this!

jstone-lucasfilm (approved):

PR #2591, opened 2025-09-25
Support alpha in Standard Surface to glTF PBR translation

### LINKED ISSUE #2579: Support alpha in Standard Surface to glTF PBR translation
From issue #2578:

### Alpha
**glTF docs:** [Metallic-Roughness Material](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#metallic-roughness-material) (in the docs alpha is the forth component of base color)
**Standard surface doc:** [3.3 Transparency](https://autodesk.github.io/standard-surface/#closures/transparency)
**gltf_pbr inputs:** `alpha` 
**standard surface inputs:** `opacity`

bowald:
closes #2579

This converts opacity to alpha as implemented in `standard_surface.mtlx`, i.e with the limitation of monochromatic opacity.

[standard_surface.mtlx#L411-L424](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/standard_surface.mtlx#L411-L424)
```xml
    <!-- Surface construction with opacity -->
    <!-- Node <surface> only supports monochromatic opacity so use the luminance of input opacity color -->
    <luminance name="opacity_luminance" type="color3">
      <input name="in" type="color3" interfacename="opacity" />
    </luminance>
    <extract name="opacity_luminance_float" type="float">
      <input name="in" type="color3" nodename="opacity_luminance" />
    </extract>
```

**Opacity 1, 0, 0** 
|<br>Standard Surface |<br>Translated to gltf_pbr, **_with alpha mode manual set to BLEND_** | **Diff**
|:-:|:-:|:-:|

### DIFF
--- a/libraries/bxdf/translation/standard_surface_to_gltf_pbr.mtlx
+++ b/libraries/bxdf/translation/standard_surface_to_gltf_pbr.mtlx
@@ -42,6 +42,7 @@
     <output name="normal_out" type="vector3" />
     <output name="tangent_out" type="vector3" />
     <output name="alpha_out" type="float" />
+    <output name="alpha_mode_out" type="integer" />
   </nodedef>
 
   <nodegraph name="NG_standard_surface_to_gltf_pbr" nodedef="ND_standard_surface_to_gltf_pbr">
@@ -225,6 +226,13 @@
       <input name="index" type="integer" value="0" />
     </extract>
 
+    <ifequal name="alpha_mode_from_opacity_luminance" type="integer">
+      <input name="value1" type="float" nodename="opacity_luminance_float" />
+      <input name="value2" type="float" value="1.0" />
+      <input name="in1" type="integer" value="0" />
+      <input name="in2" type="integer" value="2" />
+    </ifequal>
+
     <output name="base_color_out" type="color3" nodename="base_color" />
     <output name="metallic_out" type="float" nodename="metallic" />
     <output name="roughness_out" type="float" nodename="roughness" />
@@ -243,6 +251,7 @@
     <output name="normal_out" type="vector3" nodename="normal" />
     <output name="tangent_out" type="vector3" nodename="tangent" />
     <output name="alpha_out" type="float" nodename="opacity_luminance_float" />
+    <output name="alpha_mode_out" type="integer" nodename="alpha_mode_from_opacity_luminance" />
 
   </nodegraph>
 </materialx>


bowald:
Since `alpha_mode` is required to be set to BLEND in gltf_pbr to be render correctly if opacity is used. Should the translation graph also set alpha mode to BLEND if alpha not eq to 1?

Marking this PR as draft until we decide if the translation graph should handle alpha_mode as well.

jstone-lucasfilm:
Just leaving a note for @bowald for when he returns from leave, and there's no hurry to address this:

I think your suggestion to set `alpha_mode` to the appropriate value in the translation graph is the best approach, since it can be determined entirely from the value of `opacity`.

bowald:
> Just leaving a note for @bowald for when he returns from leave, and there's no hurry to address this:
> 
> I think your suggestion to set `alpha_mode` to the appropriate value in the translation graph is the best approach, since it can be determined entirely from the value of `opacity`.

I agree. `alpha_mode_out` have been added to the translation graph.

Here is test setup, where the sphere gets a green color if `alpha_mode_out` is 2 (i.e BLEND) else blue ( in this case it's 0. i.e OPAQUE)


https://github.com/user-attachments/assets/cbbba4a6-b8a4-49eb-9cd5-4bac0f506380

```xml
<?xml version="1.0"?>
  <standard_surface name="SR_marble1" type="surfaceshader">
    <input name="base" type="float" value="1" />
    <input name="base_color" type="color3" output="out" nodename="ifequal_color3I" />
    <input name="specular_roughness" type="float" value="0.1" />
    <input name="subsurface" type="float" value="0.4" />
    <input name="subsurface_color" type="color3" output="out" nodename="ifequal_color3I" />
  </standard_surface>
  <surfacematerial name="Marble_3D" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_marble1" />
  </surfacematerial>
  <standard_surface_to_gltf_pbr name="standard_surface_to_gltf_pbr" type="multioutput">
    <output name="alpha_mode_out" type="integer" />
    <input name="opacity" type="color3" value="1, 1, 1" />
  </standard_surface_to_gltf_pbr>
  <ifequal name="ifequal_color3I" type="color3">
    <input name="value2" type="integer" value="2" />
    <input name="in1" type="color3" value="0, 1, 0" />
    <input name="in2" type="color3" value="0, 0, 1" />
    <input name="value1" type="integer" output="alpha_mode_out" nodename="standard_surface_to_gltf_pbr" />
  </ifequal>
</materialx>
```

kwokcb:
A couple of quick questions here:
1. Should "transmission" also be considered to enable alpha blend ?  It is one of the channels on std surface which is considered for the MaterialX heuristic to test for "transparency" .
2. Is it assumed that alpha-cutoff doesn't affect translation and does not need to be modified from it's default (0.5) ? I believe so but just double checking.
Thanks.

bowald:
Great points @kwokcb 

1. I don't think transmission should affect alpha mode, from a glTF perspective alpha_mode only cares about the alpha value and not the transmission extension. 

From [KHR_materials_transmission](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_materials_transmission#blend-mode)

>  If alpha-as-coverage is not being used, the blend mode of the material should be set to "OPAQUE" even though it is transparent.

2. I believe so as well, at least in the MaterialX implementation of standard surface there is nothing relating to a threshold for opacity similar to alpha_cutoff. So I think it makes sense to have `BLEND` if there is opacity and `OPAQUE` when not (i.e white)

Issue #2546, opened 2025-09-06
Provide Complete Python Bindings for Vector and Matrix Types


kwokcb:
## Issue

Currently a number of binary and unary operations are specified in C++ but are not
exposed in Python.

This includes operators such as { `/=`, `*=`, `+=`, `-=`, `-` and `+` }

## Code to Address

See `MaterialXCore/Types.h` for operations and `PyMaterialXCore\PyTypes.cpp` for wrappers.
There are macros that need to be updated for vector and matrix types. Not that matrix operations such as `/` are not simply component-wise operations.

## Suggested Change

The following is rough pseudo code which uses the built in identifiers, and `py::self` as 
appropriate for vector.

```c++
  // Compound assignment operators still need explicit binding
  .def("__iadd__", &Vector::operator+=, py::return_value_policy::reference_internal)
  .def("__isub__", &Vector::operator-=, py::return_value_policy::reference_internal)
  .def("__imul__", &Vector::operator*=, py::return_value_policy::reference_internal)
  .def("__itruediv__", &Vector::operator/=, py::return_value_policy::reference_internal)
  
  // Unary operators
  .def(-py::self)  // Unary negation
  .def(+py::self)  // Unary plus
  ```

dougbtinoco:
Hi y'all! I'll be tackling this on the upcoming dev days!

jstone-lucasfilm:
Welcome, @dougbtinoco, and I'll assign you this task!

jstone-lucasfilm:
Thanks to @kwokcb for proposing this issue, as well as to @dougbtinoco for addressing it in #2594!

PR #2594, opened 2025-09-26
Adding python bindings for Vector and Matrix Types

dougbtinoco:
This solves https://github.com/AcademySoftwareFoundation/MaterialX/issues/2546

PR includes support for operators: `/=`, `*=`, `+=`, `-=`, `-` and `+` . 
I have also added on the cpp support for the missing unary.


source/PyMaterialX/PyMaterialXCore/PyTypes.cpp
```diff
@@ -24,9 +24,15 @@ using IndexPair = std::pair<size_t, size_t>;
 .def(py::self == py::self)                              \
 .def(py::self != py::self)                              \
 .def(py::self + py::self)                               \
+.def(-py::self)
```
jstone-lucasfilm: Looking at the build errors, I believe many of them are just from missing backslashes at the ends of these newly added lines!

dougbtinoco: Yeah! Thank you so much for the feedback @jstone-lucasfilm !

dougbtinoco: Im slowly cleaning up the quick run down I did during the day :+1:

dougbtinoco:
I'm a bit confused on why that specific one build is failing

ld-kerley:
I'm not quite sure why - but adding parenthesis as below, seems to fix things for me. In `source/PyMaterialX/PyMaterialXCore/PyTypes.cpp` in your diff
change 
```
.def("__iadd__", &V::operator+=, py::return_value_policy::reference_internal)
```
in to 
```
.def("__iadd__", (&V::operator+=), py::return_value_policy::reference_internal)
```
And similarly on lines 29, 31, 60 and 62.
 
I think whatever parser the static analysis tool is using doesn't like the `=` preceding the `,` perhaps.  I did notice just below these changes you already had some of these operators wrapped in parentheses.

Hope this helps :)

ld-kerley:
It's failing in the static analysis step - which is not performed for all builds - that's why only this build is failing.

dougbtinoco:
Thank you so much @ld-kerley, you suggestion really resolved the problem, I'll keep that in mind for the future!

python/MaterialXTest/main.py
```diff
@@ -82,6 +82,12 @@ def test_Vectors(self):
         self.assertTrue(v1 * 2 == v2)
         self.assertTrue(v2 / 2 == v1)
 
+        # Unary operation
```
kwokcb, line 85: Is an update also required for testing matrices ?

dougbtinoco: ahh, thanks for pointing that out @kwokcb! I will add for the matrices as well

jstone-lucasfilm: @kwokcb Perhaps this is already addressed in the test code below?  I actually see tests for unary operations on both vectors and matrices in this code, with the matrix tests in the main section on matrix operators.

PR #2596, opened 2025-09-30
Use "_oslOutputFilePath" in "OslRenderer::compileOSL()" to build the output path of the OSO file if it has been defined

JoshuaSenouf:
In `OslRenderer::compileOSL()`, we always use the input file path of the OSL file we want to compile to define the output path of the resulting OSO file no matter what.

Now, we make use of `_oslOutputFilePath` if it is has been defined.


ld-kerley:
I think this looks good to me, it's certainly an improvement. Looking at the code it does feel likely that maybe the RenderOSL module is primarily used of the MaterialX unit tests - as bundling source and compiled shaders and rendered images together feels like it might be less useful in general. 

If you're making this suggestion because you're using this for something other than the unit tests then I almost wonder if we should go a step further and add API to the OSLRender module to allow users to explicitly set the directories that each of following types of files are written to.
* OSL source
* compiled OSO
* rendered image

If this is all you need for now - I'm very happy for this to be merged as is.

JoshuaSenouf:
This is a good point, we could definitely extend the capabilities of `OslRenderer` even more in that regard and make it more flexible.

I could do soon, but in terms of my immediate needs, this PR should do! :)

jstone-lucasfilm:
This looks like a good improvement, thanks @JoshuaSenouf, and on our side we should validate this with the full MaterialX render test suite before merging.

jstone-lucasfilm:
I've just tested our render suite with this change, and everything checks out!

[MaterialXRenderTests_10_16_2025_GitHub.pdf](https://github.com/user-attachments/files/22959488/MaterialXRenderTests_10_16_2025_GitHub.pdf)

Issue #2428, opened 2025-05-30
Graph Editor : Crashes when selecting nodes in functional graphs


kwokcb:
## Issue

Graph Editor will crash if you open the implementation graph (functional) for a node and then select any node in that graph.

hybridherbst:
I've seen this sporadically as well but wasn't able to reliably reproduce. Do you happen to have repro steps?

kwokcb:
> I've seen this sporadically as well but wasn't able to reliably reproduce. Do you happen to have repro steps?

I've put in a fuller description.

MeghaS94:
Hello! I tried to look at this issue, the crash occurs because `graphDoc->getDescendant(...)` returns a nullptr when called from nodes inside functional/implementation graphs, because _childMap in the Element class does not have entries for these graph names.
Most nodes seem to avoid this by checking if the parent graph has a nodeDef and returning early https://github.com/AcademySoftwareFoundation/MaterialX/blob/dd699d89009e9f39399c063c5ba64347e23b02ef/source/MaterialXGraphEditor/Graph.cpp#L786
Output nodes don't have this check. Should output nodes also check for parent graph nodeDef and return early? This seems to fix the crash, but I wanted to check if this made sense before submitting a fix.

weta-sfinnie:
Hi I'd like to pick this up for dev days.

jstone-lucasfilm:
Welcome, @weta-sfinnie, and it's all yours!

weta-sfinnie:
I agree with @MeghaS94 analysis on this problem. The getNodeDef() check seems to be called to catch when something is in a functional graph, but the check only applies to NodePtr which the output node is not. 
And so something inside a functional graph is getting past the check and triggering a nullptr error down the line. Adding the check in solves the problem.

PR #2597, opened 2025-09-30
Graph Editor : Solved crashes when selecting nodes in functional graphs

weta-sfinnie:
Closes: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2428

### Issue
There was an issue where when you click on an output node in Functional Graphs in the materialX graph editor, it would crash.

### Fix
The reason for this was that output nodes return a nullptr on getNode() calls, and so were being missed by the check for that determines if the node was inside a functional graph. I've added this check for output nodes, and now this crash doesn't occur.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -792,6 +792,14 @@ void Graph::setRenderMaterial(UiNodePtr node)
         }
         else if (mtlxNodeGraph)
         {
+            // As above, there is no logic to support traversing from inside a functional graph.
+            // We add a check for output nodes to make sure it's accounted for in this case.
+            if(mtlxOutput){     
+                if (mtlxNodeGraph->getNodeDef())
+                {
+                    return;
+                }
+            }
             testPaths.insert(mtlxNodeGraph->getNamePath());
         }
 


jstone-lucasfilm:
Thanks for putting together this proposal, @weta-sfinnie!

I'm CC'ing @lfl-eholthouser for her thoughts on this work, since she has the deepest knowledge of this codebase.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -792,6 +792,14 @@ void Graph::setRenderMaterial(UiNodePtr node)
         }
         else if (mtlxNodeGraph)
         {
+            // As above, there is no logic to support traversing from inside a functional graph.
+            // We add a check for output nodes to make sure it's accounted for in this case.
+            if(mtlxOutput){
```
jstone-lucasfilm: This is a very minor note, but MaterialX follows the convention of [Allman-style braces](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/.clang-format#L5), so this would be better written in the following way:

```
            if (mtlxOutput)
            {
                if (mtlxNodeGraph->getNodeDef())
                {
                    return;
                }
            }
```

lfl-eholthouser:
I know @jstone-lucasfilm mentioned the small style change for the brace but besides that I think the fix looks great.

PR #2598, opened 2025-10-01
Improve `Syntax::makeValidName` function

ld-kerley:
This PR improves the logic around how valid names are generated with consideration for reserved words defined in specific Syntax objects.

Currently the shader generator calls `Syntax::makeIdentifier()` to ensure that variable names are legal. This compares the current name against a map of already used identifiers. The reserved words for a syntax are prepopulated in this map.

There are some places, namely [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/Nodes/CompoundNode.cpp#L40) and [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/Nodes/SourceCodeNode.cpp#L70), where we generate names that need to be legal, but do not call `makeIdentifier()` but instead just call `makeValidName()`.  Which got me to thinking that make `makeValidName()` should be doing the reserved word check.  

Moving the check to `makeValidName()` means we do not have to copy all of the reserved words in to the identifier map for each `ShaderGraph` object we create, as well as plugging the potential bug of being able to generate illegal function names.

I didn't add a warning - but I wonder if we should emit a warning if a reserved word is used as a variable name?

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -18,15 +18,10 @@ MATERIALX_NAMESPACE_BEGIN
 // ShaderGraph methods
 //
 
-ShaderGraph::ShaderGraph(const ShaderGraph* parent, const string& name, ConstDocumentPtr document, const StringSet& reservedWords) :
+ShaderGraph::ShaderGraph(const ShaderGraph* parent, const string& name, ConstDocumentPtr document, const StringSet& /*reservedWords*/) :
     ShaderNode(parent, name),
     _document(document)
 {
-    // Add all reserved words as taken identifiers
-    for (const string& n : reservedWords)
-    {
-        _identifiers[n] = 1;
-    }
 }
 
 void ShaderGraph::addInputSockets(const InterfaceElement& elem, GenContext& context)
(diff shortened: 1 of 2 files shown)


niklasharrysson:
This looks like a great change. Feels much better and safer to keep the reserved words check fully contained within the `Syntax` class.

source/MaterialXGenShader/ShaderGraph.cpp
```diff
@@ -18,15 +18,10 @@ MATERIALX_NAMESPACE_BEGIN
 // ShaderGraph methods
 //
 
-ShaderGraph::ShaderGraph(const ShaderGraph* parent, const string& name, ConstDocumentPtr document, const StringSet& reservedWords) :
+ShaderGraph::ShaderGraph(const ShaderGraph* parent, const string& name, ConstDocumentPtr document, const StringSet& /*reservedWords*/) :
```
niklasharrysson, line 21: Perhaps remove the `reservedWords` parameter completely since it's not used now? 

This constructor is only ever called from within ShaderGraph.cpp, in the `ShaderGraph::create` methods, and I doubt anyone has ever used it explicitly from an integration. It's not exposed in python. So I think it should be safe to change the signature.

ld-kerley:
Thanks @niklasharrysson - yeah I was debating removing it from the interface too - but wasn't sure how worried I needed to be about backwards compatibility.  

I think I'll maybe it's better to merge this PR (just need someone to hit the approve button), and then I can put up a cleanup PR afterwards, that way if the clean up causes someone an issue - we can easily revert it.

jstone-lucasfilm (approved):
This looks fine to me as well, thanks @ld-kerley!

PR #2599, opened 2025-10-02
Upgrade to emsdk 4.0.8 and fix missing readFromXmlString

wingfiring:
This PR upgrade emsdk to 4.0.8, and upgrade node to version 22.
Also add the missing export function `readFromXmlString`, it's called in `Web Viewer`.


ashwinbhat:
hi @bernardkwok @scotbrew could you provide your input on this PR? It updates emsdk to 4.0.8.

kwokcb:
I tried clean building this on Windows local and noticed 

- **`package.json` needs to be udpated since it using deprecated packages and has security warnings.**
- if you delete `package-lock.json` and `node_modules` from MaterialXTest will still run but MaterialXView viewer will not run:
```
> MaterialXView@1.0.0 start
> webpack serve

[webpack-cli] Invalid options object. Dev Server has been initialized using an options object that does not match the API schema.
 - options has an unknown property '_assetEmittingPreviousFiles'. These properties are valid:
   object { allowedHosts?, bonjour?, client?, compress?, devMiddleware?, headers?, historyApiFallback?, host?, hot?, ipc?, liveReload?, onListening?, open?, port?, proxy?, server?, app?, setupExitSignals?, setupMiddlewares?, static?, watchFiles?, webSocketServer? }
```
  - upgrading packages using `npm audit fix` causes the same issue.
- I've tried node: 22.16.0 (emscripten version) and 22.19.0 (what happens to be used in CI).
 - **For CI can we specify 22.16.0 vs 22 to get an exact match with emscripten.**
 
 It would be useful to have this tested on Mac for supported versions as it's been found that some versions may not be supported on different platforms. CI is Linux only I think.

javascript/MaterialXTest/package.json
```diff
@@ -9,7 +9,7 @@
     "pretest": "npm run clean && npm run copyBuild",
     "test": "npm run mocha",
     "test:browser": "npm run karma -- --browsers ChromeHeadless --singleRun true",
-    "mocha": "mocha '*.spec.js' --require @babel/register --timeout 5000",
+    "mocha": "mocha *.spec.js --require @babel/register --timeout 5000",
```
wingfiring, line 12: remove single quote so it can launch on windows.

javascript/MaterialXTest/package.json
```diff
@@ -26,7 +26,7 @@
     "karma-chrome-launcher": "^3.2.0",
     "karma-mocha": "^2.0.1",
     "karma-mocha-reporter": "^2.2.5",
-    "mocha": "^10.8.2",
+    "mocha": "^11.7.4",
```
wingfiring, line 29: Reduce deprecaed module

javascript/MaterialXView/package.json
```diff
@@ -17,7 +17,7 @@
   "devDependencies": {
     "copy-webpack-plugin": "^12.0.2",
     "html-webpack-plugin": "^5.6.3",
-    "webpack-cli": "^4.10.0",
+    "webpack-cli": "^6.0.1",
```
wingfiring, line 20: Fix MaterialXView viewer run issue.

source/JsMaterialX/JsMaterialXFormat/JsXmlIo.cpp
```diff
@@ -39,6 +39,8 @@ EMSCRIPTEN_BINDINGS(xmlio)
     BIND_FUNC_RAW_PTR("_readFromXmlFile", mx::readFromXmlFile, 2, 4, mx::DocumentPtr, mx::FilePath,
         mx::FileSearchPath, const mx::XmlReadOptions*);
     BIND_FUNC_RAW_PTR("writeToXmlFile", mx::writeToXmlFile, 2, 3, mx::DocumentPtr, const mx::FilePath&, const mx::XmlWriteOptions *);
+    BIND_FUNC_RAW_PTR("readFromXmlString", mx::readFromXmlString, 2, 4, mx::DocumentPtr,
```
wingfiring, line 42: Missing, required by MaterialXView.

source/JsMaterialX/CMakeLists.txt
```diff
@@ -61,7 +61,6 @@ string(APPEND JS_LINK_FLAGS_CORE " -I${SOURCE_FOLDER} \
     -s WASM=1 \
     -s DISABLE_EXCEPTION_CATCHING=0 \
     -s MODULARIZE=1 \
-    -s USE_ES6_IMPORT_META=0 \
```
wingfiring, line 64: No longer support.

wingfiring:
Thank you for your review.
According to your comments, I did those` changes:
- Fixed security warnings, and reduced deprecated modules.
- Fixed MaterialXView viewer.
- No vulnerabilities in `npm audit fix`
- Specified node version to 22.16.0
- Verified on MacOS, including both MaterialXTest and MaterialXView viewer
Please help to review again. Thanks!

scotbrew:
I reviewed the code and also perfomed a build and test on my Mac M1 system.  Here are the results.

**Code Review:**
The code changes look reasonable and what would be expected for updating emsdk and node.  Note that there is a non-blocking warning during compilation.

**Version Review**
It may be worthwhile to check the exact versions required, especially on the Mac.
-  emsdk: 4.0.4  (that is the latest one when running `emsdk list` on my Mac system)
-  node: v22.14.0 (that is the one listed as "default" when running `nvm list` on my Mac system)

**Compiling**
Success, with 1 warning for `getVersionIntegers()`
```
wasm-ld: warning: function signature mismatch: _ZNK17MaterialX_v1_39_516InterfaceElement18getVersionIntegersEv
>>> defined as () -> void in CMakeFiles/JsMaterialXCore.dir/JsMaterialXCore/JsElement.cpp.o
>>> defined as (i32, i32) -> void in ../../lib/libMaterialXCore.a(Interface.cpp.o)
```

**Testing**
Test1 <SUCCESS>: `npm run test`

Test2 <FAILURE>: `npm run test:browser`
```
(node:20105) [DEP0060] DeprecationWarning: The `util._extend` API is deprecated. Please use Object.assign() instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
LOG: 'Generating shader for essl...'
  Generate Shaders
    ✖ Compile Shaders
      Chrome Headless 140.0.0.0 (Mac OS 10.15.7)
    RuntimeError: null function or function signature mismatch
        at _build/JsMaterialXGenShader.wasm:wasm-function[5291]:0x11f877
        at _build/JsMaterialXGenShader.wasm:wasm-function[5292]:0x11f896
        at Object.getExceptionMessage (eval at newFunc (_build/JsMaterialXGenShader.js:8:113219), <anonymous>:5:10)
        at Context.<anonymous> (browser/shaderGenerator.spec.js:102:34)
```

Test3: <SUCCESS> `cd MaterialXView; npm install; npm run build; emrun --browser chrome dist/index.html`

**Documentation Suggestion:**
* The minimum versions needed for node and emdsk should be listed, likely along with the THREE version
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/javascript/README.md

wingfiring:
> **Version Review** It may be worthwhile to check the exact versions required, especially on the Mac.
> 
> * emsdk: 4.0.4  (that is the latest one when running `emsdk list` on my Mac system)
> * node: v22.14.0 (that is the one listed as "default" when running `nvm list` on my Mac system)

I installed emsdk via https://github.com/emscripten-core/emsdk.git, just pulled and checked the main branch, I found the latest emsdk version is 4.0.16. and latest node is node-24.7.0. 

> 
> **Compiling** Success, with 1 warning for `getVersionIntegers()`
> 
> ```
> wasm-ld: warning: function signature mismatch: _ZNK17MaterialX_v1_39_516InterfaceElement18getVersionIntegersEv
> >>> defined as () -> void in CMakeFiles/JsMaterialXCore.dir/JsMaterialXCore/JsElement.cpp.o
> >>> defined as (i32, i32) -> void in ../../lib/libMaterialXCore.a(Interface.cpp.o)
> ```

I haven't fully grasped the mechanism behind this issue yet. `InterfaceElement::getVersionIntegers` is a virtual function and do not need to link it actually.  
This warning looks harmless and exists for a long time, and caused by '-flto' option, 
An earlier report: https://github.com/emscripten-core/emscripten/issues/10831
Explaination: https://lld.llvm.org/WebAssembly.html#function-signatures

> Test2 : `npm run test:browser`
> 
> ```
> (node:20105) [DEP0060] DeprecationWarning: The `util._extend` API is deprecated. Please use Object.assign() instead.
> (Use `node --trace-deprecation ...` to show where the warning was created)
> LOG: 'Generating shader for essl...'
>   Generate Shaders
>     ✖ Compile Shaders
>       Chrome Headless 140.0.0.0 (Mac OS 10.15.7)
> ```

Fixed this issue on MacOS, and verified on Window and Linux. Looks like the default ChromeHeadless disabled gpu or webgl on MacOS only. 

> **Documentation Suggestion:**
> 
> * The minimum versions needed for node and emdsk should be listed, likely along with the THREE version
>   https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/javascript/README.md

Thanks. Updated the README.md. I should follow this document instead of main.yml

jstone-lucasfilm:
This change looks excellent to me, @wingfiring, and I'm eager to hear from @kwokcb, @ashwinbhat, and @scotbrew whether it looks ready to merge for 1.39.5.

javascript/MaterialXTest/browser/karma.conf.js
```diff
@@ -33,5 +33,15 @@ module.exports = function (config)
             'karma-mocha',
             'karma-mocha-reporter',
         ],
+        customLaunchers: {
+            ChromeHeadlessGL: {
```
wingfiring, line 37: Looks like the 'ChromeHeadless' on MacOS disable gpu or webgl, so define a customized Chrome headless and enable gpu and webgl.

wingfiring:
> **Compiling** Success, with 1 warning for `getVersionIntegers()`↳
> 
> ```
> wasm-ld: warning: function signature mismatch: _ZNK17MaterialX_v1_39_516InterfaceElement18getVersionIntegersEv
> >>> defined as () -> void in CMakeFiles/JsMaterialXCore.dir/JsMaterialXCore/JsElement.cpp.o
> >>> defined as (i32, i32) -> void in ../../lib/libMaterialXCore.a(Interface.cpp.o)
> ```
Found the root cause and fixed. 

The flag '-flto' is enabled for JsMaterialXCore and JsMaterialXGenShader (defined in source/JsMaterialX/CMakeLists.txt), but not set in other project, particularly MaterialXCore. '-flto'  needs to be enabled for either all compile unit or none.

But the fix discovered a serials of issue.

1. In debug build, the tests reports memory leak and recommend to invoke .delete() manually after used. I fixed all of those memory leaks, that's why I added lots of 'xxx.delete()'.
2. New version Node changed the module initialization process, the original workflow is broken. That's why I renamed `source/JsMaterialX/post.js` to `source/JsMaterialX/pre.js`, so the file name can match emcc flag '--pre-js'. This files is also modified for this item and item 4.
3. Exception pointer is no longer a integer number, so fixed `getExceptionMessage` to handle this.
4. `ems::constant(name, some_variable);` accesses variable immediately. A statement like`EMSCRIPTEN_BINDINGS(xxx)` defines a global variable, and  it calls `ems::constant` immediately in the ctor. So the global variable `EmBindInit_##xxx##_instance` in the following code piece depends on `some_variable`. As a well known trap, the initialization order of global variable is undeterminable, so the `some_variable` may not be initialized when the `ems::constant` is called. In order to fix this issue, I use `ems::function` and lambda function to defer the access to global `some_variable`, and assign the public constants at runtime in pre.js. 

```C++
#define EMSCRIPTEN_BINDINGS(name)                                              \
  static void embind_init_##name();                                            \
  static struct EmBindInit_##name : emscripten::internal::InitFunc {           \
    EmBindInit_##name() : InitFunc(embind_init_##name) {}                      \
  } EmBindInit_##name##_instance;                                              \
  static void embind_init_##name()
```

My local test matrix:
 `Cartesian product([Debug, Release],[Windows, Linux, MacOS], [MaterialXView{build, start}, MaterialXTest{test, test:browser}])`

PR #2601, opened 2025-10-02
Remove NodeGraph dependency in ShaderTranslation

ld-kerley:
While working on some side experiments - I bumped in to a case where the shader translation tests were failing because I did not have a nodegraph to implement the translation node. 

I don't think this should be a requirement, and we only seem to be using the NodeGraph to find the outputs, which are also available on the NodeDef object.

This appears to pass all the tests, but I do have a question about the code [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/ShaderTranslator.cpp#L106). The comment suggests perhaps a NodeGraph is a requirement - but I haven't seen this documented anywhere, and I'm not familiar enough with the context the comment is referring to to understand what it means.  If there is some requirement in the system like this, it would be great to have a unit tests that validates it, so we ensure we don't regress.

### DIFF
--- a/source/MaterialXGenShader/ShaderTranslator.cpp
+++ b/source/MaterialXGenShader/ShaderTranslator.cpp
@@ -19,7 +19,7 @@ void ShaderTranslator::connectTranslationInputs(NodePtr shader, NodeDefPtr trans
     std::set<OutputPtr> origOutputs;
     for (InputPtr shaderInput : origInputs)
     {
-        if (translationNodeDef->getInput(shaderInput->getName()))
+        if (translationNodeDef->getActiveInput(shaderInput->getName()))
         {
             InputPtr input = _translationNode->addInput(shaderInput->getName(), shaderInput->getType());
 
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
I'm additionally CC'ing @bowald for his thoughts, as IKEA also leverages this codebase.

bowald:
I think this change looks good. We have our own implementation of `connectTranslationOutputs`, where we are using the outputs from nodedef similar to the change @ld-kerley suggest. The biggest difference in our custom function is that we are taking inherence in mind for the nodedef, i.e. we are traversing the translationgraphs nodedef, using `nodedef->getActiveOutputs()` instead of `nodedef->getOutputs()`.  

Currently no translation graphs are using inherence, but we had an idea of doing some custom ones using inherence of the official translation graphs, but havn't gotten around to it.

ld-kerley:
Thats a great point and seems a good future proofing thing to include, I'll update the PR thx.

lfl-eholthouser:
This change looks great to me!

Issue #2575, opened 2025-09-23
OSL Shader Node Networks


ogulbrandsen:
At Imageworks we rely heavily on OSL for shading, where the renderer ingests a material as a network of compiled shaders (oso files) parameterized and wired together in a format very close to OSL's ShaderGroups:

(from OSL's testrender testsuite)
```
<ShaderGroup>
  param float scale_s 20;
  param float scale_t 20;
  param color Ca 0.1 0.1 0.1;
  param color Cb 0.5 0.5 0.5;
  shader checkerboard tex;
  shader matte layer1;
  connect tex.Cout layer1.Cs;
</ShaderGroup>
```

We're interested in exploring a MaterialX solution that lets us both generate a render-consumable network of standard library MaterialX nodes (from their OSL implementation), and also use it for networks of our own production OSL shaders (and possibly mix and match the two).
The project will likely consist of three main parts:

* A tool that generates buildable OSL files (shaders) for each of the MaterialX node-definitions.
* A shader generator that translates a MaterialX network to an OSL command string (something close to the ShaderGroup mentioned above).
* A test framework.

While codegen works very well for the standard library nodes, we have found some challenges when trying to use it with existing production shaders:

* the use of network-inspecting functions like "isconnected" (is there an upstream shader output driving a parameter) has to be reconsidered when using codegen.
* parameter initialization: OSL supports input and output parameters being initialized by global variables and its output parameters can also be initialized with an input parameter as a shorthand.
* primvar driven input parameters: OSL lets you read geometry attributes if a shader input parameter has the same name as a geometry attribute, making it very sensitive to parameter renaming/prefixing.

Most of these can be worked around by re-writing the shaders, but OSL as a language has so far evolved strongly around the idea that each shader exists as a standalone, inspectable unit in a network, so their use is quite widespread (at least in our shaders).

In terms of performance and where the shader/network optimziation really should happen, it is definitely an interesting question how a codegen'ed monolithic OSL shader that can be optimized at compile time, performs compared to a (possibly very large) network of compiled (oso) nodes, where most of the optimization burden is left up to the runtime optimizer. If we get this project up and running, we can actually start performing apples-to-apples comparisons using the materialx standard library.

kwokcb:
A note for the tooling step that generates references for every definitions osl/oso:

[...]

PR #2603, opened 2025-10-06
Implement OSL command string code generator

chellmuth:
This PR implements a new OSL code generator that emits shader group command strings as described in #2575. The work is from a Dev Days collaboration at Imageworks between myself, @JoshuaSenouf, and @MrigankaArya.

The three main parts are a tool to generate .oso's from MaterialX nodedefs, a ShaderGenerator class that generates the command string, and tests that run the existing OSL testrender cases with the new codegen.

At Imageworks we would like to use this to investigate performance differences between the two code generation styles, and to explore mixing and matching our internal shaders with MaterialX's.

### DIFF
--- a/source/MaterialXGenOsl/OslNetworkShaderGenerator.cpp
+++ b/source/MaterialXGenOsl/OslNetworkShaderGenerator.cpp
@@ -12,7 +12,6 @@
 #include <MaterialXGenShader/TypeDesc.h>
 #include <MaterialXGenShader/ShaderStage.h>
 
-
 MATERIALX_NAMESPACE_BEGIN
 
 const string OslNetworkShaderGenerator::TARGET = "genoslnetwork";
@@ -56,20 +55,22 @@ ShaderPtr OslNetworkShaderGenerator::generate(const string& name, ElementPtr ele
     std::set<std::string> osoPaths;
 
     // Walk the node graph, emitting shaders and param declarations.
-    for (auto&& node : graph.getNodes()) {
+    for (auto&& node : graph.getNodes())
+    {
         const string& nodeName = node->getName();
 
-        for (auto&& input : node->getInputs()) {
+        for (auto&& input : node->getInputs())
+        {
             string inputName = input->getName();
             _syntax->makeValidName(inputName);
 
             const ShaderOutput* connection = input->getConnection();
-            if (!connection || connection->getNode() == &graph) {
+            if (!connection || connection->getNode() == &graph)
+            {
                 if (!input->hasAuthoredValue())
                     continue;
 
-                if (input->getName() == "backsurfaceshader"
-                    || input->getName() == "displacementshader")
+                if (input->getName() == "backsurfaceshader" || input->getName() == "displacementshader")
                     continue; // FIXME: these aren't getting pruned by hasAuthoredValue
 
                 string value = _syntax->getValue(input);
@@ -78,30 +79,37 @@ ShaderPtr OslNetworkShaderGenerator::generate(const string& name, ElementPtr ele
 
                 // TODO: Figure out how to avoid special-casing struct-types in the generator, perhaps in the syntax?
                 auto inputType = input->getType();
-                if (inputType == Type::VECTOR2) {
+                if (inputType == Type::VECTOR2)
+                {
                     auto parts = splitString(value, " ");
                     emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".x", parts[0]), stage, false);
                     emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".y", parts[1]), stage, false);
                 }
-                else if (inputType == Type::VECTOR4) {
+                else if (inputType == Type::VECTOR4)
+                {
                     auto parts = splitString(value, " ");
                     emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".x", parts[0]), stage, false);
                     emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".y", parts[1]), stage, false);
                     emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".z", parts[2]), stage, false);
-                    emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".w", parts[3]), stage, false);}
-                else if (inputType == Type::COLOR4) {
+                    emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".w", parts[3]), stage, false);
+                }
+                else if (inputType == Type::COLOR4)
+                {
                     auto parts = splitString(value, " ");
                     emitLine(paramString(_syntax->getTypeName(Type::COLOR3), inputName + ".rgb", parts[0] + " " + parts[1] + " " + parts[2]), stage, false);
                     emitLine(paramString(_syntax->getTypeName(Type::FLOAT), inputName + ".a", parts[3]), stage, false);
                 }
-                else {
+                else
+                {
                     emitLine(paramString(_syntax->getTypeName(input->getType()), inputName, value), stage, false);
                 }
-            } else {
+            }
+            else
+            {
                 string connName = connection->getName();
                 _syntax->makeValidName(connName);
 
-                string connect = connectString(connection->getNode()->getName(), connName,  nodeName, inputName);
+                string connect = connectString(connection->getNode()->getName(), connName, nodeName, inputName);
                 // Save connect emits for the end, because they can't come
                 // before both connected shaders have been declared.
                 connections.push_back(connect);
@@ -121,24 +129,26 @@ ShaderPtr OslNetworkShaderGenerator::generate(const string& name, ElementPtr ele
         lastNodeName = nodeName;
     }
 
-    if (!lastOutput) {
+    if (!lastOutput)
+    {
         printf("Invalid shader\n");
         return nullptr;
     }
 
-    for (auto&& connect : connections) {
+    for (auto&& connect : connections)
+    {
         emitLine(connect, stage, false);
     }
 
     // During unit tests, wrap a special node that will add the output to Ci.
-    if (context.getOptions().oslNetworkConnectCiWrapper) {
+    if (context.getOptions().oslNetworkConnectCiWrapper)
+    {
         emitLine("shader setCi root ;", stage, false);
         string connect = connectString(
             lastNodeName,
             lastOutput->getName(),
             "root",
-            lastOutput->getType().getName() + "_input"
-        );
+            lastOutput->getType().getName() + "_input");
         emitLine(connect, stage, false);
     }
 
@@ -158,7 +168,6 @@ ShaderPtr OslNetworkShaderGenerator::generate(const string& name, ElementPtr ele
 
     return shader;
 }
-
 
 ShaderPtr OslNetworkShaderGenerator::createShader(const string& name, ElementPtr element, GenContext& context) const
 {
@@ -197,7 +206,6 @@ ShaderPtr OslNetworkShaderGenerator::createShader(const string& name, ElementPtr
     return shader;
 }
 
-
 namespace OSLNetwork
 {
 
@@ -206,6 +214,6 @@ const string UNIFORMS = "u";
 const string INPUTS = "i";
 const string OUTPUTS = "o";
 
-} // namespace OSL
+} // namespace OSLNetwork
 
 MATERIALX_NAMESPACE_END
--- a/source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
+++ b/source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
@@ -311,7 +311,7 @@ bool OslShaderRenderTester::runRenderer(const std::string& shaderName,
 
                     const std::string& outputName = output->getVariable();
                     const std::string& outputType = typeSyntax.getTypeAlias().empty() ? typeSyntax.getName() : typeSyntax.getTypeAlias();
-                    const std::string& sceneTemplateFile = (_useOslCmdStr ? "scene_template_oslcmd.xml": "scene_template.xml");
+                    const std::string& sceneTemplateFile = (_useOslCmdStr ? "scene_template_oslcmd.xml" : "scene_template.xml");
 
                     // Set shader output name and type to use
                     _renderer->setOslShaderOutput(outputName, outputType);
@@ -323,7 +323,7 @@ bool OslShaderRenderTester::runRenderer(const std::string& shaderName,
                     auto osoPathAttr = shader->getAttribute("osoPath");
                     if (osoPathAttr)
                     {
-                        _renderer->setDataLibraryOSOPath( osoPathAttr->getValueString() );
+                        _renderer->setDataLibraryOSOPath(osoPathAttr->getValueString());
                     }
 
                     // Set scene template file. For now we only have the constant color scene file
--- a/source/MaterialXGenOsl/LibsToOso.cpp
+++ b/source/MaterialXGenOsl/LibsToOso.cpp
@@ -21,7 +21,6 @@
 
 namespace mx = MaterialX;
 
-
 const std::string setCiOslNetworkSource = R"(
 
 #include "mx_funcs.h"
@@ -92,8 +91,6 @@ shader setCi (
     Ci = Out_Ci;
 }
 )";
-
-
 
 const std::string options =
     "    Options: \n"
@@ -313,7 +310,8 @@ int main(int argc, char* const argv[])
     catch (mx::ExceptionRenderError& exc)
     {
         std::cout << "Encountered a codegen/compilation related exception for the "
-                   "following node: " << std::endl;
+                     "following node: "
+                  << std::endl;
         std::cout << exc.what() << std::endl;
 
         // Dump details about the exception in the log file.
@@ -327,12 +325,11 @@ int main(int argc, char* const argv[])
     // Catch any other exceptions
     catch (mx::Exception& exc)
     {
-        std::cout << "Failed to codegen/compile the following node to OSL: "  << std::endl;
+        std::cout << "Failed to codegen/compile the following node to OSL: " << std::endl;
         std::cout << exc.what() << std::endl;
 
         hasFailed = true;
     }
-
 
     // We create and use a dedicated `NodeGraph` to avoid `NodeDef` names collision.
     mx::NodeGraphPtr librariesDocGraph = librariesDoc->addNodeGraph("librariesDocGraph");
@@ -434,7 +431,7 @@ int main(int argc, char* const argv[])
         }
     }
 
-    mx::writeToXmlFile(implMtlxDoc,  implMtlxDocFilePath);
+    mx::writeToXmlFile(implMtlxDoc, implMtlxDocFilePath);
 
     // If something went wrong, return an appropriate error code.
     if (hasFailed)
--- a/source/MaterialXGenOsl/OslNetworkShaderGenerator.h
+++ b/source/MaterialXGenOsl/OslNetworkShaderGenerator.h
@@ -51,7 +51,6 @@ class MX_GENOSL_API OslNetworkShaderGenerator : public ShaderGenerator
   protected:
     /// Create and initialize a new OSL shader for shader generation.
     virtual ShaderPtr createShader(const string& name, ElementPtr element, GenContext& context) const;
-
 };
 
 namespace OSLNetwork
@@ -62,7 +61,7 @@ extern MX_GENOSL_API const string UNIFORMS;
 extern MX_GENOSL_API const string INPUTS;
 extern MX_GENOSL_API const string OUTPUTS;
 
-} // namespace OSL
+} // namespace OSLNetwork
 
 MATERIALX_NAMESPACE_END
 


jstone-lucasfilm:
This looks very promising, @chellmuth, and I'm looking forward to a deeper discussion with @niklasharrysson, @kwokcb, and @ld-kerley on this PR.

As a very first observation, would it perhaps be more natural to leverage the existing `OslShaderGenerator` class and `genosl` target in MaterialX, and to select this new variant of OSL shader generation through a boolean in [GenOptions](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/GenOptions.h#L75)?

ld-kerley:
Full disclosure - I was supporting these guys a little with some MaterialX questions through this process - so I'm already a little familiar with the work.

@jstone-lucasfilm - The need to pre-compile the OSOs means that we need to have a separate `<implementation>` to describe the oso itself.  The PR slightly abuses the current syntax for implementation, and we may want to look at formalizing this a little better - but given the time it takes to add new syntax I recommended they worked with what they had access to for now.   But this alternate `<implementation>` means we need a new shader generation target name to select it - hence `genoslnodes` and not extending `genosl`.

CMakeLists.txt
```diff
@@ -244,6 +258,10 @@ if (MATERIALX_BUILD_GEN_MDL)
     add_definitions(-DMATERIALX_INSTALL_MDL_MODULE_PATH=\"${MATERIALX_INSTALL_MDL_MODULE_PATH}\")
 endif()
 
+if (MATERIALX_BUILD_GEN_OSL_NODES)
+    set(MATERIALX_BUILD_DATA_LIBRARY ON)
```
ld-kerley: Is it possible to move this up to line 143 ish ?

libraries/CMakeLists.txt
```diff
@@ -28,6 +28,33 @@ if(MATERIALX_BUILD_DATA_LIBRARY)
     add_custom_target(MaterialXBuildData ALL
         DEPENDS ${MATERIALX_DATA_LIBRARY_BUILD_FILES})
 
+    if (MATERIALX_BUILD_GEN_OSL_NODES AND MATERIALX_OSL_BINARY_OSLC)
```
ld-kerley: Instead of guarding this with `MATERIALX_OSL_BINARY_OSLC` - we might want to raise an error if `MATERIALX_OSL_BINARY_OSLC` is empty - because I'm assuming the oso builder won't be able to function if it doesn't have access to the osl compiler.

source/MaterialXTest/MaterialXRenderOsl/Utilities/scene_template_oslcmd.xml
```diff
@@ -0,0 +1,32 @@
...
+   <!-- Background quad with raytype test. -->
+   <ShaderGroup>
+      color Cin %background_color%;
+      %environment_shader_parameter_overrides%
+      shader raytype_background layer1;
+   </ShaderGroup>
+   <Quad corner="-340, -340, -800" edge_x="680, 0, 0" edge_y="0, 680, 0" />
+
+   <!-- Shader graph for routing to output layer:
+   input_shader_parameter_overrides : are parameter overrides for the input shader
```
ld-kerley: this comment is out of date.

source/MaterialXTest/MaterialXRenderOsl/Utilities/test_node.osl
```diff
@@ -0,0 +1,8 @@
+surface
+
+test_node(
```
ld-kerley: I don't think this is used anywhere.

source/MaterialXGenShader/ShaderNode.h
```diff
@@ -500,6 +511,8 @@ class MX_GENSHADER_API ShaderNode
     ShaderNodeImplPtr _impl;
     ShaderMetadataVecPtr _metadata;
 
+    string _nodeDefName;
```
ld-kerley: Storing the node def name here probably needs a little more investigation - I think its a great way to get something working, but we might want to look at a more decoupled way to accessing the necessary implementation information.  

I'm not sure what that is right now - perhaps some specialization of the `ShaderNodeImpl` class that could collect the necessary information about the corresponding OSO.

source/MaterialXTest/MaterialXRenderOsl/Utilities/test_node.osl
chellmuth: Whoops!

source/MaterialXTest/MaterialXRenderOsl/Utilities/scene_template_oslcmd.xml
chellmuth: thanks, fixed

lgritz:
@chellmuth I've added you to SPI's CLA authorization list for this project.

kwokcb:
Concur that this is impressive and fits in quite well with the existing system!

Just a cursory read over with mostly clarification questions so I understand things better :)
Sorry for the long spam -- I'm mostly curious about items in *italic* . Thanks.

----

- Does it make sense to have 2 separate build dependencies:
  - One to build the oso's only. BTW, how often are the oso's built ? 
  - One to use the available oso for code generation
  - Currently the "code gen" module has render dependencies which gives the impression of a circular dependency.

- Items that pop to mind:
  - *Can an existing OSL generator derivation use this approach*
  - *Can renderers which have additional nodedefs not in the std library use this approach ?*
  - Seems like the new key dependency the oso library path vs just having the osl path ?
  - Is the oso building command reusable as is ? 

- Implementation details:
  - How much overlap is there between the current and new OSL generators. Does not need to be addressed now
    but just curious as to the amount of code duplication. e.g. is there syntax duplication?  
  - *Noticed there are CI / unit test code insertions in the generator which are triggered via generation options.
    It does feel a bit out of place.*
  - *Not sure I understand the need for the nodedef name. I assume of oso association ? Unsure if there there is any brittleness
  is using ids.*
  - Minor: New genoptions need exposure in Python. 
  
- Test:
  - Not sure if the current image comparison script can compare current osl vs oslnode. Would be good to get
    a run comparison.
  - *Can you provide a sample generation output. Would be nice to see what the ShaderGroup output looks like.*

ld-kerley:
I think the circular dependency is because the "build" tool is included in the "generate" module - I think it were separated out to its own `MaterialXBuildTools` module - then it would disappear.  I think the build tool actually only needs the `MaterialXRenderOsl` module to re-use the `CompileOSL()` function - and it might be cleaner to just re-implement the necessary pieces of that directly in the tool itself.

I think the intention is to build the standard library nodes as osos once, at build time.  There are lots of open questions about custom node defs, or if other people want to bring their own data library.  They all feel solvable, but likely need some input and thought from others about how we all want it to work.  I think the current tool could certainly be refined to support these cases if desired - if it doesn't already.

The nodedef name is only used to derive the oso name - and I've been chatting with Chris about this offline - I think there's a path where a new `OsoNode` derived from `ShaderNodeImpl` can be used to work around the direct storing of the nodedef name in the `ShaderNode` class.  

I did run the osl vs oslnodes tests. 
[osl_vs_oslcmd.pdf](https://github.com/user-attachments/files/22731933/osl_vs_oslcmd.pdf)

There are some differences around normal mapping - but I think that might be related to the customization in the OSL test frame for tangent and bitangent [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp#L26)

chellmuth:
Thanks for taking a look @kwokcb.

> * Does it make sense to have 2 separate build dependencies:
>   
>   * One to build the oso's only. BTW, how often are the oso's built ?
>   * One to use the available oso for code generation
>   * Currently the "code gen" module has render dependencies which gives the impression of a circular dependency.

Yes, this sounds reasonable to me. Right now the oso's rebuild when you change the existing OSL generator, so like @ld-kerley mentioned above, it would also be an improvement to remove that dependency entirely.


> * Items that pop to mind:
>
>   * _Can an existing OSL generator derivation use this approach_

Do you mind clarifying what you mean on this one? Like, can users switch from the existing OSL method to the command strings? Or is this about the implementation?

>   * _Can renderers which have additional nodedefs not in the std library use this approach ?_

Hopefully. We have nodedef wrappers for many of our internal OSL shaders that we'd like to use in this pipeline, but we haven't experimented with it yet.

>   * Seems like the new key dependency the oso library path vs just having the osl path ?

Right.

>   * Is the oso building command reusable as is ?

I believe so. @joshuasenouf can chime in if I'm misrepresenting.

> * Implementation details:
>   
>   * How much overlap is there between the current and new OSL generators. Does not need to be addressed now
>     but just curious as to the amount of code duplication. e.g. is there syntax duplication?

The big duplication is in the syntax object. I think a big chunk of the duplicated syntax code can be deleted, but I wanted to post this PR and get some feedback before diving into that. There's also copy-paste in the python bindings, and `OslNodesShaderGenerator::createShader`.

>   * _Noticed there are CI / unit test code insertions in the generator which are triggered via generation options.
>     It does feel a bit out of place._

That's right, I assume you mean this code:
```
    // During unit tests, wrap a special node that will add the output to Ci.
    if (context.getOptions().oslNodesConnectCiWrapper) {
        emitLine("shader setCi root ;", stage, false);
        string connect = connectString(
            lastNodeName,
            lastOutput->getName(),
            "root",
            lastOutput->getType().getName() + "_input"
        );
        emitLine(connect, stage, false);
    }
```
Agreed, it doesn't feel great. Happy to implement any suggestions here.


>   * _Not sure I understand the need for the nodedef name. I assume of oso association ? Unsure if there there is any brittleness
>     is using ids._

Looks like @ld-kerley answered this already. He also opened #2604 to help me improve this.

>   * Minor: New genoptions need exposure in Python.

Thanks for pointing that out!

> * Test:
>   
>   * Not sure if the current image comparison script can compare current osl vs oslnode. Would be good to get
>     a run comparison.

Looks like @ld-kerley covered this one as well.

>   * _Can you provide a sample generation output. Would be nice to see what the ShaderGroup output looks like._

Sure, here's a short example from the CarPaint test:
```
param string space "world" ;
shader normal_vector3 geomprop_Nworld ;
param string space "world" ;
param int index 0 ;
shader tangent_vector3 geomprop_Tworld ;
param float base 0.5 ;
param color base_color 0.103779 0.59212 0.850649 ;
param float specular 1 ;
param color specular_color 1 1 1 ;
param float specular_roughness 0.4 ;
param float specular_anisotropy 0.5 ;
param float coat 1 ;
param float coat_roughness 0 ;
shader standard_surface_surfaceshader SR_carpaint ;
shader surfacematerial Car_Paint ;
connect geomprop_Nworld.out SR_carpaint.coat_normal ;
connect geomprop_Nworld.out SR_carpaint.normal1 ;
connect geomprop_Tworld.out SR_carpaint.tangent ;
connect SR_carpaint.out Car_Paint.surfaceshader1 ;
shader setCi root ;
connect Car_Paint.out root.material_input ;
```

Hopefully that mostly addresses your questions, let me know if you'd like more detail on anything.

niklasharrysson:
Great to see this work!

I think it makes sense to keep this new generator separate from the `OslShaderGenerator`, since they don't have much in common. Perhaps the syntax classes could be unified more, but the generators seems quite different, as the old one emits OSL shader code while the new one emits ShaderGroup statements / command strings.

To emphasize this would it make sense to rename the new generator? Perhaps a name like `OslShaderGroupsGenerator` or `OslCommandStringGenerator` would be more descriptive?

From a first pass reading the PR here are some other high level notes.

We should be careful with extending the shadergen core data model, adding new data and API, unless it's really really needed or serves a generally useful purpose, since this is shared among all generators and integrations.

- Addition of the nodedef string to the `ShaderNode` instances. This you have already discussed above, and the proposed solution to avoid this sounds good, introducing a new `OsoNode` implementation node instead.

- Addition of the boolean (`_isDefault`) to all ports on `ShaderNodes`. If I understand correctly this is to track if the port's value is at it's default or not? But it's just a bool tracking if the value has been set more than once, so can't track for sure if the value is still default or not. I wonder if it would be better to handle this explicitly in the generate() method where `isDefault()` is called, and here compare the value against the nodedef's value to see if it's really default or not. We then avoid introducing this extra data/api and it seems more robust.

`GenOptions.oslImplicitSurfaceShaderConversion`: Just curious why this new options is needed, since the old "OSL Reference" unit test is doing OSL generation for individual nodes, and is not using such a flag.

`GenOptions.oslNodesConnectCiWrapper`: This seems to be used to select a special case for running unit tests. Would it be possible to handle that special case in the actual unit test code instead? For example, appending the extra `setCi root shader` lines on the unit test side instead?

chellmuth:
Hey @niklasharrysson, thanks for the questions.

> To emphasize this would it make sense to rename the new generator? Perhaps a name like `OslShaderGroupsGenerator` or `OslCommandStringGenerator` would be more descriptive?

I'm don't think we have a preference here, happy to follow any consensus on appropriate naming.
 
> * Addition of the boolean (`_isDefault`) to all ports on `ShaderNodes`. If I understand correctly this is to track if the port's value is at it's default or not? But it's just a bool tracking if the value has been set more than once, so can't track for sure if the value is still default or not. I wonder if it would be better to handle this explicitly in the generate() method where `isDefault()` is called, and here compare the value against the nodedef's value to see if it's really default or not. We then avoid introducing this extra data/api and it seems more robust.

Agreed with the problem, but when we switch to OsoNode I believe we're going to lose the NodeDef. So I'm not sure how to improve this as it will be. Like you noted, the term "default" is inaccurate, but the intention was that a value overridden in the xml would always appear in the command string, if only to try to make the two representations as 1:1 as possible. Open to feedback here as well, probably a minor point compared to other issues.

> `GenOptions.oslImplicitSurfaceShaderConversion`: Just curious why this new options is needed, since the old "OSL Reference" unit test is doing OSL generation for individual nodes, and is not using such a flag.

We didn't want the generated oso library to convert surfaceshader outputs to surfacematerial. Under the OSL Reference tests, if the root shader is a surfaceshader output it gets converted to surfacematerial so that OSL gets a closure, but intermediate surfaceshaders are preserved. That breaks under the command string paradigm, where the command string may be incorrectly trying to connect modified surfacematerial outputs to surfaceshader inputs in the middle of the graph (eg `mix_surfaceshader`).

> `GenOptions.oslNodesConnectCiWrapper`: This seems to be used to select a special case for running unit tests. Would it be possible to handle that special case in the actual unit test code instead? For example, appending the extra `setCi root shader` lines on the unit test side instead?

That's a good idea, I'll make that update.

ld-kerley:
@niklasharrysson - at the moment it doesn't look like the `ShaderNode`/`ShaderPort` objects have any way to determine what the default value is.  I guess we could add an entirely new `ValuePtr defaultVal` to store this and populate it when we're creating the initial ports from the NodeDef.  Do you think this would be preferable to just storing a boolean value indicating if the current stored value is the default or not.

I do think I agree with @chellmuth here that an explicitly authored value, even if it happens to have the same value as the default, should be written in to the OSL command string output. So I think we're really trying to determine if the value stored in the `ShaderPort` came from the original NodeDef or if it was authored on the Node.

niklasharrysson:
Thanks for the explanations @chellmuth.

My main concern with `isDefault` was that when adding new API to the core data model, we need to make sure it's usable and robust for anyone using it anywhere in the future. And as noted it's not robustly tracking if the value is a default value or not. But you're right, it could be solved by just changing the terminology.

Perhaps a better name would be `hasAuthoredValue()`.

Also, the implementation could be done using a `ShaderPortFlag` instead of the extra boolean.

ld-kerley:
@niklasharrysson I was chewing over the `GenOptions.oslNodesConnectCiWrapper part a bit more....

> GenOptions.oslNodesConnectCiWrapper: This seems to be used to select a special case for running unit tests. Would it be possible to handle that special case in the actual unit test code instead? For example, appending the extra setCi root shader lines on the unit test side instead?

I think there might actually be legitimate reasons for this to be a "real" generator option, I think there are use-cases for wanting to generate a shader with and without this additional node that will write to `Ci`. 

* `ON`
  * I would want to generate the osl command string for a MaterialX element that might represent a sub-graph (not a final renderable material) and just pass it directly to OSLs `testshade`, `testrender` or any other renderer that can consume the command string, and render an image, say for a swatch in a nodegraph editor. 
* `OFF`
  * I bring my own data library of my studios custom OSL nodes (this is the Imageworks use-case), and I know that my terminal node will already write to `Ci`.
  * I'm using a renderer that doesn't even `Ci`, but instead pulls on outputs from the OSL `ShaderGroup` directly.

chellmuth:
I've updated the PR with code from #2604 to replace the ShaderNode's nodeDefName field with the new OsoNode ShaderNodeImpl technique.

source/MaterialXGenShader/ShaderNode.h
chellmuth: Addressed in e437b2b.

CMakeLists.txt
chellmuth: Fixed in e914f00.

chellmuth:
> I think it makes sense to keep this new generator separate from the OslShaderGenerator, since they don't have much in common. Perhaps the syntax classes could be unified more, but the generators seems quite different, as the old one emits OSL shader code while the new one emits ShaderGroup statements / command strings.
> 
> To emphasize this would it make sense to rename the new generator? Perhaps a name like OslShaderGroupsGenerator or OslCommandStringGenerator would be more descriptive?

This came up at the OSL TSC today, it seemed there was consensus around OSL Network over eg OSL ShaderGroup, so I've replaced OslNodesShaderGenerator with OslNetworkShaderGenerator in ff60644.

ld-kerley:
I think XXXNetworkShaderGenerator makes a good precedent for possible other nodegraph based output generators to follow.

source/MaterialXGenOslNetwork/OslNetworkShaderGenerator.h
```diff
@@ -0,0 +1,69 @@
...
+
+#include <MaterialXGenShader/ShaderGenerator.h>
+
+MATERIALX_NAMESPACE_BEGIN
+
+using OslNetworkShaderGeneratorPtr = shared_ptr<class OslNetworkShaderGenerator>;
+
+/// @class OslNetworkShaderGenerator
+/// Base class for OSL (Open Shading Language) shader generators.
+/// A generator for a specific OSL target should be derived from this class.
```
niklasharrysson: Doc string needs an update.

source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
```diff
@@ -365,6 +387,22 @@ TEST_CASE("Render: OSL TestSuite", "[renderosl]")
     mx::FileSearchPath searchPath = mx::getDefaultDataSearchPath();
     mx::FilePath optionsFilePath = searchPath.find("resources/Materials/TestSuite/_options.mtlx");
 
-    OslShaderRenderTester renderTester(mx::OslShaderGenerator::create());
+    OslShaderRenderTester renderTester(mx::OslShaderGenerator::create(), false);
+    renderTester.validate(optionsFilePath);
+}
+
+TEST_CASE("Render: OSL Nodes TestSuite", "[renderoslnetwork]")
```
niklasharrysson: Nitpick: "Nodes" -> "Network"

CMakeLists.txt
```diff
@@ -40,6 +40,7 @@ option(MATERIALX_BUILD_DOCS "Create HTML documentation using Doxygen. Requires t
 
 option(MATERIALX_BUILD_GEN_GLSL "Build the GLSL shader generator back-end." ON)
 option(MATERIALX_BUILD_GEN_OSL "Build the OSL shader generator back-end." ON)
+option(MATERIALX_BUILD_GEN_OSL_NETWORK "Build the OSL nodes shader generator back-end." ON)
```
niklasharrysson: Nitpick: "nodes" -> "network"

source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
```diff
@@ -299,11 +311,21 @@ bool OslShaderRenderTester::runRenderer(const std::string& shaderName,
...
-                    const std::string& sceneTemplateFile = "scene_template.xml";
+                    const std::string& sceneTemplateFile = (_useOslCmdStr ? "scene_template_oslcmd.xml": "scene_template.xml");
 
                     // Set shader output name and type to use
                     _renderer->setOslShaderOutput(outputName, outputType);
 
+                    // Get oslcmdstr
+                    _renderer->setOSLCmdStr(oslCmdStr);
+                    _renderer->useOslCommandString(_useOslCmdStr);
+                    // Osl Nodes Shaders record the oso path in an attribute
```
niklasharrysson: Nitpick: "Nodes" -> "Network"

source/MaterialXGenOslNetwork/OslNetworkSyntax.h
```diff
@@ -0,0 +1,41 @@
...
+/// OSL syntax class
+
+#include <MaterialXGenOslNetwork/Export.h>
+
+#include <MaterialXGenShader/Syntax.h>
+
+MATERIALX_NAMESPACE_BEGIN
+
+/// @class OslSyntax
+/// Syntax class for OSL (Open Shading Language)
```
niklasharrysson: Doc string needs an update

niklasharrysson:
> I think XXXNetworkShaderGenerator makes a good precedent for possible other nodegraph based output generators to follow.

Sounds great! 👍

niklasharrysson:
> I think there might actually be legitimate reasons for this to be a "real" generator option, I think there are use-cases for wanting to generate a shader with and without this additional node that will write to `Ci`.

Allrighty! Then it makes sense to keep it. Perhaps then just rename it to oslNetworkConnectCiWrapper for consistency.

chellmuth:
Thanks for pointing out those missed renames and doc-string copy-pastes, I pushed some updates to use the OSL Network naming.

chellmuth:
> My main concern with `isDefault` was that when adding new API to the core data model, we need to make sure it's usable and robust for anyone using it anywhere in the future. And as noted it's not robustly tracking if the value is a default value or not. But you're right, it could be solved by just changing the terminology.
> 
> Perhaps a better name would be `hasAuthoredValue()`.
> 
> Also, the implementation could be done using a `ShaderPortFlag` instead of the extra boolean.

This is implemented in the latest round of commits.

source/MaterialXGenOslNetwork/LibsToOso.cpp
```diff
@@ -0,0 +1,450 @@
...
+        {
+            std::cout << "MaterialXGenOslNetwork - LibsToOso version " << mx::getVersionString() << std::endl;
+            std::cout << options << std::endl;
+
+            return 0;
+        }
+        else
+        {
+            std::cout << "Unrecognized command-line option: " << token << std::endl;
+            std::cout << "Launch the graph editor with '--help' for a complete list of supported "
```
niklasharrysson: Help string needs an update: "graph editor"

libraries/targets/genoslnetwork.mtlx
```diff
@@ -0,0 +1,14 @@
...
+<materialx version="1.39">
+
+  <!--
+    Copyright Contributors to the MaterialX Project
+    SPDX-License-Identifier: Apache-2.0
+  -->
+
+  <!-- ======================================================================== -->
+  <!-- Base target definition for OSL targets using shader generation.          -->
+  <!-- ======================================================================== -->
```
niklasharrysson, line 11: Doc string needs an update.

source/MaterialXGenOsl/OslNetworkSyntax.cpp
```diff
@@ -0,0 +1,464 @@
...
+const StringVec OslNetworkSyntax::VECTOR2_MEMBERS = { ".x", ".y" };
+const StringVec OslNetworkSyntax::VECTOR4_MEMBERS = { ".x", ".y", ".z", ".w" };
+const StringVec OslNetworkSyntax::COLOR4_MEMBERS = { ".rgb[0]", ".rgb[1]", ".rgb[2]", ".a" };
+
+//
+// OslNetworkSyntax methods
+//
+
+OslNetworkSyntax::OslNetworkSyntax(TypeSystemPtr typeSystem) : Syntax(typeSystem)
+{
```
niklasharrysson, line 223: There seems to be a lot of code duplication with this class and the `OslSyntax` class. Perhaps you could derive this class from the `OslSyntax` class, and only override the type syntax registration for the types that differ in syntax?

If the base class (`OslSyntax`) constructor is called first you can then call `registerTypeSyntax()` again for the types that has a different syntax, it will replace any registration done by the base class.

niklasharrysson:
This is coming along nicely! I just had a few minor comments about, and also a suggestion to derive the new syntax class from the OslSyntax class in order to avoid code duplication.

source/MaterialXGenOsl/OslNetworkSyntax.cpp
ld-kerley: I wonder if we might log a GitHub issue for this work and allow it to land as clean up later.  This was an epic scale dev-days project. If we merge this knowing there are things that can be improved like this, it will allow others to lend a hand get this fully wrapped up. 

I would be happy to help here - but I'm not able to contribute to someone else's PR.

niklasharrysson: Oh absolutely! Indeed this is an epic PR for a dev-days project, and we can log and fix the remaining issues once this PR has been merged.

jstone-lucasfilm:
@niklasharrysson @chellmuth Following our existing pattern in [MaterialXGenGlsl](https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/source/MaterialXGenGlsl), where generator and syntax variants such as ESSL and WGSL are maintained within the same `MaterialXGenGlsl` module, it seems logical to take a similar approach here, where these new classes for OSL network generation would be maintained within `MaterialXGenOsl`.

That should additionally make it more straightforward to harmonize these two generators and syntaxes in the future, as they will be built together, and be guaranteed access to each others' codebases.

Does that seem like a reasonable suggestion to you both?

ld-kerley:
@jstone-lucasfilm - I had actually considered that right back at the beginning of these discussions. I think the significant distinction here is that all the GLSL based generators are all dependent on the same data library input, where as the OSL network generator requires a completely different data library input, one that quite reasonably is opt'ed in to separately with a build flag.  I think there will likely be consumers that want one, and not the other, both ways around. If we bundle them together, then I think we would need to build the OSOs during the build for all OSL customers, which wouldn't be necessary.

I don't think the two generators really need to be harmonized all that much to be honest. The Syntaxes likely have some overlap, and that area needs to be investigated. 

I would like to propose we merge this PR as is, and move forward with any further refactors/improvements as secondary pieces of work.

source/MaterialXGenOsl/OslNetworkSyntax.cpp
jstone-lucasfilm: As noted by @niklasharrysson here, we definitely want to harmonize this with the existing `OslSyntax` class, rather than duplicating the code.

I'm OK with deferring that to a future PR, and moving this code into the existing `MaterialXGenOsl` module should make that much more straightforward when we get to that work in the future.

jstone-lucasfilm:
@ld-kerley I think we'll ultimately need these new generators to live within `MaterialXGenOsl`, allowing us to improve the code sharing between them in subsequent PRs, and the main question in my mind is whether we should merge them now or in a future improvement.

@chellmuth Do you feel that it would be too heavy of a lift to make this change in your initial PR, so that we don't add a brand new MaterialX module and then subsequently remove it?

niklasharrysson:
I agree with @ld-kerley here. The two generators are not mutually related. The new network generator is dependent on the OslShaderGenerator in order to generate the OSOs. But there’s no relation the other way around, and I think most OSL targets will not use, and don’t need to build, the network generator.

Perhaps in the future though when we have thoroughly tested the OSL network approach, and this turns out to be more performant, maybe this will be the default way to do OSL generation, and merging to a single module makes sense.

jstone-lucasfilm:
@niklasharrysson Wouldn't that be analogous to our ESSL and WGSL generators, though, which have a one-directional dependence upon the GLSL generator?

My goal isn't to require developers to build both at once, but rather to allow the new network generator to assume the presence of the existing OSL generator, and it would be fine to provide an independent build option for the network generator if that's valuable.

niklasharrysson:
@jstone-lucasfilm To my mind it’s not the same as GLSL/ESSL/WGSL, as the new network generator is emitting a command string rather than shader code, and requires the extra build step of generating all the OSOs. 

However, as long as this can be disabled with a build option I think it’s fine to keep it all in one module :)

chellmuth:
> @chellmuth Do you feel that it would be too heavy of a lift to make this change in your initial PR, so that we don't add a brand new MaterialX module and then subsequently remove it?

That shouldn't be a problem, I'm happy to make that update in this PR.

jstone-lucasfilm:
@chellmuth I'd really appreciate that, if it's not too much of a pain, and I wouldn't ask if I didn't think this was a valuable feature that we should integrate fully into our shader generation architecture.

@niklasharrysson That completely makes sense, and maintaining the build option for this feature seems important.

If the build-time cost for the network generator is significant, should its build option default to OFF, and we can set it to ON in our GitHub CI tests?

Let me know if I'm misreading the PR, but it looks like it's currently defaulting to ON in our root-level CMake configuration.

ld-kerley:
I think its possible to build the OSL network generator and not need the OSL source generator at all - in the situation where you're bring your own data library and OSOs from another source.

I think there is investigation needed to see if we even need to share the Syntax or not - I suspect initially it was just "lets get this working" level of effort. Which is why I think a deeper more considered investigation is warranted.  The output requirements for the data types (the largest responsibility of the syntax objects) is not necessarily the same between the two systems. 

@jstone-lucasfilm - I would love to understand more concretely what you think the benefits are to merging here. You pose the question as to if its a good idea to do this now or later, I would propose we do it later - when we resolve concretely if there is actually a legitimate dependency between the two generators. 

I agree with @niklasharrysson that this is a very different situation to the GLSL/ESSL/WGSL case, the deltas between those drives shader generation systems and the parent (GLSL) one are really small.  For these two OSL based generators, as Niklas points out, they are inherently different.   I might actually pose that we could have easily called this `MaterialXGenOsoNetwork` as really it is networking together the compiled OSOs. 

There is currently a dependency between `MaterialXGenOslNetwork` and `MaterialXGenOsl` to support the OSO building tool - but I actually think perhaps this should be extracted from this module and built somewhere else (but we should do this work in a separate PR). Once thats done I think perhaps the dependency between two might no longer be needed. 

I think most of the existing precendent for MaterialX cmake flags `MATERIALX_BUILD_XXX` are targeting specific library modules, and not features inside those modules.  I worry it becomes complex to know if an installation has been built with or without OSL network support.  Currently thats trivially determined by inspecting the built shared libraries. But if these two OSL generators live in the same module, and hence the same shared library, then that determination becomes more complicated. 

I do think we should default building the OSL Network generator (and the associated OSOs for the data library) to `OFF`, for now at least.

jstone-lucasfilm:
Two thoughts on this, @ld-kerley:

- I usually think of MaterialX C++/Python modules as _permanent institutions_ that are unlikely to be renamed or removed in future releases, so the bar for adding a new one should remain very high.  There are some great new ideas in this PR, but they are clearly still fluid and subject to change in the short term, so it seems best to keep these new files within an existing module such as `MaterialXGenOsl` until they have fully settled.
- Since this new work is just a starting point for OSL network generation, I completely understand that some code duplication is warranted in this first pass, but I think it's very important that we make it straightforward for future developers to remove that duplication.  Keeping the new files in the same module as the code that they duplicate seems valuable in achieving this goal, and the path for developers to harmonize these two OSL generation approaches should remain clear as the code evolves.

From that perspective, if @chellmuth believes it's reasonable to propose this new work within `MaterialXGenOsl`, I'd very much appreciate his taking that approach.

ld-kerley:
I would vote to leave them separate for now - and merge them in a PR before we make the next release - if we find that there really is a code to be shared here, and it is determined that its important that they live in the same module.  I'm not aware we're discussing making a new MaterialX release anytime soon - so we have a window to figure this out. 

So far it seems that the benefit to merging is the possibility that the code might be shared - but we don't know this yet.  I will be happy to work through the investigation to find the common points once this PR has landed.  I'll also note It is still very possible to share code across MaterialX modules - that's what we're doing with `MaterialXGenShader` and `MaterialXGenHW` now too.  

@jstone-lucasfilm  - If they are merged - you do propose to address the confusion situation of cmake build arguments not aligning with easily inspectable elements in the final delivery? As I outlined above. 

I'll also note that I am aware of multiple other network generator experiments currently happening, that may (hopefully) end up being contributed to the public project, and those also are concretely different to their source generating counterpart generators.

niklasharrysson:
I don't have a strong opinion here, and I think there are good arguments for either approach of placing the new files. The only detail I find important is that the OSO generation gets wrapped in a build option that is off by default, so we don't run this unless specifically requested.

Looking at the changes again I would have to agree with Jonathan that the simplest, cleanest approach for now is too just add the new files to the MaterialXGenOsl module. The pros being that we avoid adding a new MaterialXGenOslNetwork modules as well as a PyMaterialXGenOslNetwork module, that we might want to remove again later anyway.

It's just a few new files and I don't think the build time would be affected, as long as the OSO generation is handled separately. It also keeps all OSL-related code in one place. But if we later find it advantageous to split things out we can of course create the new modules then.

Again, just my 2 cents. If Chris has the time and bandwidth to make this move, perhaps it's better to do it directly before merging the PR. But otherwise I think we could do these changes directly in his PR as well.

ld-kerley:
If thats the consensus then, I'm happy to go with it. I do think they will end up not relying on each other in the long run.

If we only use the cmake argument to guard the creation of the OSOs as @niklasharrysson suggests. Then we don't bifurcate the potential contents of the MaterialXGenOSL module, and we'll always know what's inside. Which addresses one of my concerns around merging them together. 

We will need to be careful to handle the unit testing correctly, if we're redefining what the cmake arguments mean.

chellmuth:
I've pushed a series of commits making the discussed changes, at least how  I understand them.

* Move all the code in MaterialXGenOslNetwork/ in to MaterialXGenOsl/
* Remove the MaterialXGenOslNetwork module, and include the new work in the MaterialXGenOsl module
* Remove the MATERIALX_BUILD_OSL_NETWORK cmake option. All new compilation is driven by the existing MATERIALX_BUILD_OSL option
* Add a new option, MATERIALX_BUILD_OSOS. This controls whether or not we generate the .oso files during build. It defaults to OFF. It also controls whether we execute the OslNetwork unit tests.

Finally - I also removed the python bindings. Our internal build isn't set up to install them properly, so I wasn't able to properly test them. Seems like an easy follow-up.

As a newbie to the project, I figure there's a high probability I misinterpreted some details of the discussion and conclusions above. Let me know, and I can push another round of updates to get closer to what the group wants.

niklasharrysson:
Thanks a lot @chellmuth! This looks great to me! 

And let's defer the possible merging of the syntax classes to another PR.

chellmuth:
Thanks for taking another look!

I've pushed one more commit addressing an issue CI identified, when we're building the gen-osl module but not the render module, the new LibsToOso code failed to link because my refactor didn't make its dependency explicit.

As @ld-kerley brought up earlier (https://github.com/AcademySoftwareFoundation/MaterialX/pull/2603#issuecomment-3374481805), this dependency can be removed with a little refactoring, hopefully outside of this PR, so I went with a quick kludgey fix to get CI passing.

source/MaterialXGenOsl/OslNetworkShaderGenerator.cpp
```diff
@@ -0,0 +1,211 @@
...
+    ConstDocumentPtr document = element->getDocument();
+
+    string lastNodeName;
+    ShaderOutput* lastOutput = nullptr;
+    std::vector<string> connections;
+
+    std::set<std::string> osoPaths;
+
+    // Walk the node graph, emitting shaders and param declarations.
+    for (auto&& node : graph.getNodes()) {
```
jstone-lucasfilm: Just a minor note, but make sure to run `clang-format` on this new source file, so that it's aligned with the Allman braces style of the broader codebase.

Issue #2365, opened 2025-05-04
Graph Editor / Viewer : Add Token Support to the UI


kwokcb:
## Issue

If you have a document using tokens, these do not show up anywhere in the UI. 

## Proposal

Allow them to at least show up in the property panel for each as editable / non-connectable values.
Suggest that they do not show up as pins on nodes in the editor as they can't be connected.

xipaja:
Hello, I'm interested in taking this on for Dev Days!

jstone-lucasfilm:
Welcome, @xipaja!

kwokcb:
I've split this issue into 2 parts so we can close the first part and leave the second (editing) open. @xipaja what your working is only the first part.

jstone-lucasfilm:
Thanks to @xipaja for implementing the first task in this series, and I'll close out this task group, so that a developer can select the second part for our next Dev Days event in May.  (You're more than welcome to select this task if you'd like, @xipaja!)

PR #2605, opened 2025-10-08
Graph editor viewer UI token support

xipaja:
This PR adds fallback token resolution logic to the Graph Editor and Viewer to display available tokens and their source in the UI.

For the Graph Editor, if the current element has no tokens, it will check the nodegraph, then the nodedef. For the Viewer, if a material has no tokens, it traverses upstream nodes and their definitions to collect tokens from nodedefs and nodegraphs.

This handles the cases outlined within the [token_graph.mtlx](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/stdlib/texture/token_graph.mtlx) and the [token_graph_material.mtlx](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/stdlib/texture/token_graph_material.mtlx) test cases.

Fixes #[#2611](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2611)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     paths-ignore:
       - '**.md'
+  schedule:
+    # Nightly build
+    - cron: '30 10 * * *'
   workflow_dispatch:
 
 jobs:
@@ -14,6 +17,8 @@ jobs:
   build:
     name: Build
     runs-on: ${{ matrix.os }}
+    env:
+      IS_EXTENDED_BUILD: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
     strategy:
       fail-fast: false
       matrix:
@@ -109,6 +114,7 @@ jobs:
           python: 3.13
           cmake_config: -G "Visual Studio 17 2022" -A "x64"
           test_shaders: ON
+          extended_build_oiio: ON
 
         - name: Windows_VS2022_x64_SharedLibs
           os: windows-2025
@@ -156,11 +162,10 @@ jobs:
         Add-Content $env:GITHUB_PATH "$PWD/build/installed/bin"
 
     - name: Install OpenImageIO
-      if: matrix.build_oiio == 'ON' && runner.os == 'Windows'
+      if: env.IS_EXTENDED_BUILD == 'true' && matrix.extended_build_oiio == 'ON' && runner.os == 'Windows'
       run: |
-        C:/vcpkg/vcpkg install openimageio --triplet=x64-windows-release
-        Add-Content $env:GITHUB_PATH "C:/vcpkg/installed/x64-windows-release/bin"
-        Add-Content $env:GITHUB_PATH "C:/vcpkg/installed/x64-windows-release/share/openimageio"
+        C:/vcpkg/vcpkg install openimageio --triplet=x64-windows
+        Add-Content $env:GITHUB_PATH "C:/vcpkg/installed/x64-windows/bin"
 
     - name: Install Python ${{ matrix.python }}
       if: matrix.python != 'None'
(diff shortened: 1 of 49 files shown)


jstone-lucasfilm:
Thanks for putting this proposal together, @xipaja!

I'm linking the original GitHub Issue from https://github.com/AcademySoftwareFoundation/MaterialX/issues/2365, and I'm CC'ing @kwokcb and @lfl-eholthouser for their thoughts on this proposal.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3696,6 +3696,114 @@ void Graph::propertyEditor()
...
+            {
+                // Check nodedef if no tokens in nodegraph
+                mx::NodeDefPtr nodedef = nodegraph->getNodeDef();
+                if (nodedef != nullptr)
+                {
+                    tokens = nodedef->getActiveTokens();
+                    source = nodedef;
+                }
+
+                if (tokens.empty())
```
kwokcb: I think this should just look for immediate token children and not tokens under child nodes. When you select the node then you will get it's tokens.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3696,6 +3696,114 @@ void Graph::propertyEditor()
...
+                            childrenWithTokens.push_back(childNode->getName());
+                        }
+                    }
+                }
+            }   
+        }
+
+        if (!tokens.empty())
+        {
+            ImGui::Text("Tokens: %zu", tokens.size());
```
kwokcb: I'm not sure how this looks formatted. It would be good to add a Tokens section and format it the same as input string values are. If / when edits are allow then they would be made editable.

source/MaterialXView/Editor.cpp
```diff
@@ -711,6 +711,123 @@ void PropertyEditor::updateContents(Viewer* viewer)
             }
         }
     }
+
+    std::vector<mx::TokenPtr> tokens = {};
```
kwokcb: I think you want something akin to `StringResolver::addTokenSubstitutions()` 
which builds a token/value map for you.

source/MaterialXGraphEditor/Graph.cpp
xipaja: Thank you for the feedback! I'll work on those updates shortly.

source/MaterialXGraphEditor/Graph.cpp
xipaja: This is how it currently looks! I'm happy to update the layout/formatting.

image

kwokcb: I think it's better to have a separator with a Tokens title, then list the tokens the same way that inputs are listed.
You don't need the Tokens: #, and Tokens Source as it should always just be itself. See my not above.

source/MaterialXGraphEditor/Graph.cpp
kwokcb: As noted in formatting comment, it should always be the node or nodegraph itself which is being examined so no need to explicitly distinguish this and hence no need to show in the UI. 

It should be okay to show any tokens on the nodedef with this check-in but ideally you would only want to expose the tokens for the instance and have a "Show all tokens" akin to "Show all inputs".  Just leaving as a note for posterity. I'll update the issue appropriately with this info.

xipaja:
Here is how the updated UI looks:

Graph Editor:
graph_editor_tokens_ui
Material Viewer:
viewer_token_ui

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3696,6 +3697,64 @@ void Graph::propertyEditor()
...
+                int count = _currUiNode->_showAllTokens ? static_cast<int>(tokens.size()) : 1;
+                if (count)
+                {
+                    ImVec2 tableSize(0.0f, TEXT_BASE_HEIGHT * std::min(SCROLL_LINE_COUNT, count));
+                    bool haveTable = ImGui::BeginTable("tokens_node_table", 2, tableFlags, tableSize);
+                    if (haveTable)
+                    {
+                        ImGui::SetWindowFontScale(_fontScale);
+
+                        // Show first token until show all is selected
```
kwokcb: This is isn't the correct logic as it should be show all for instances and "show all" for any on nodedefs.

I think to keep it simple let's just remove this toggle and show them all. Part of the reason is that since you've unified the approach to use the string resolver there is no simple way to tell the difference between them unless we change the API.

xipaja: Removed the toggle! Here is the Graph Editor UI now:

image

kwokcb (approved):
Thanks for the change. This looks good to me.

jstone-lucasfilm (approved):
Great work, @xipaja, and thanks for this Dev Days contribution!

Issue #2606, opened 2025-10-08
Inconsistent background color in MaterialXViewer


ashwinbhat:
MaterialXViewer background color is inconsistent between Windows and Mac due to renderer differences (GL vs Metal). 
Screenshot below show Mac (left) and Windows (right)

<img width="2578" height="892" alt="Image" src="https://github.com/user-attachments/assets/4f0da962-9c83-499f-846a-d2f315256a91" />

From a quick test below forcing the background screen color to a linear color addresses this inconsistency, but I think It seems to be related to color linearization in the Viewer's metal rendering pipeline that needs further investigation.

<img width="1434" height="986" alt="Image" src="https://github.com/user-attachments/assets/1f0ce6db-be56-4ba9-81f6-41426478be21" />

<img width="1606" height="1088" alt="Image" src="https://github.com/user-attachments/assets/8cb4be90-dd7e-44df-bc66-5f6d9d850ed7" />

aabhinavvvvvvv:
I can try fixing this issue

jstone-lucasfilm:
We'd appreciate that, @aabhinavvvvvvv, and let us know if you have the bandwidth to take this on!

aabhinavvvvvvv:
Hi @jstone-lucasfilm, I think the issue is that the metal renderer uses an MTLPixelFormatRGBA16Float (linear) framebuffer. macOS applies gamma correction when displaying linear float values, so passing the sRGB background color (0.3, 0.3, 0.32) directly as the clear color makes it appear lighter on Mac.

The existing linearToSrgb lambda was also converting in the wrong direction, making things worse. The fix would replace it with srgbToLinear so the stored linear value gets gamma-corrected back to the intended sRGB appearance on display — matching Windows.

PR: #2800

PR #2607, opened 2025-10-08
Restore platform-specific shadow code in HW generation

ld-kerley:
A small bug was introduced in #2555 cauing the texture coordinates used in Metal when applying the shadow map to be incorrect, leading to a darker image.

This was originally noted by @ashwinbhat [here](https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1759876675993069).


jstone-lucasfilm:
Thanks for putting this fix together, @ld-kerley, and I'm CC'ing @tdavidovicNV to confirm that this conveys his original intent.

ashwinbhat:
I've verified this fix locally. The differences in background are unrelated to this and logged as https://github.com/AcademySoftwareFoundation/MaterialX/issues/2606
PR2607

tdavidovicNV:
Thanks for the fix. However, do we have any clue why this didn't break on Linux? It doesn't look like Mac specific code to me

jstone-lucasfilm:
@ld-kerley @ashwinbhat One note of caution is that this PR seems to break shadow rendering on Linux, and you can see the difference in the render artifacts generated by our per-commit GitHub CI:

Before this PR:

Before

After this PR:

After

ld-kerley:
That is interesting - @ashwinbhat tested this on Windows - so I'm now wondering where the bug is - or if perhaps we're hitting an OpenGL different in implementation on different platforms?

jstone-lucasfilm:
I like this updated approach, @ld-kerley, and it looks like our Linux CI render tests now match the output of the main branch.

Does this generate the expected results in your local MacOS tests, @ashwinbhat?

ld-kerley:
Not as part of this PR - but if we pursue this approach more - I'd love to explore the idea of moving more of the GLSL/MSL code out of the c++ code base and in to the data library.  the `HwSurfaceNode.cpp` is one of the largest remaining sources.

ashwinbhat:
> I like this updated approach, @ld-kerley, and it looks like our Linux CI render tests now match the output of the main branch.
> 
> Does this generate the expected results in your local MacOS tests, @ashwinbhat?
Here are the results from my local windows build:
image
image (5)

jstone-lucasfilm:
This looks good to me, @ld-kerley, though I believe you'll need to resolve conflicts before this can be merged.

libraries/pbrlib/genglsl/lib/mx_shadow_platform.glsl
```diff
@@ -0,0 +1,13 @@
...
+    $texSamplerSignature,
+    mat4 shadow_matrix,
+    vec3 world_position
+)
+{
+    vec4 shadowCoord4 = mx_matrix_mul(shadow_matrix, vec4(world_position, 1.0));
+    vec3 shadowCoord = shadowCoord4.xyz / shadowCoord4.w;
+    shadowCoord = shadowCoord * 0.5 + 0.5;
+    vec2 shadowMoments = texture($texSamplerSampler2D, shadowCoord.xy).xy;
+    return  mx_variance_shadow_occlusion(shadowMoments, shadowCoord.z);
```
jstone-lucasfilm, line 12: One thought on this:

We already have a helper file named `mx_shadow`, which defines the `mx_variance_shadow_occlusion` function that you call here.

Since this new helper file is also dedicated to shadowing, but is platform-specific rather than universal, what if we were to call this new helper file `mx_shadow_platform`?

That would clarify that it's a platform-specific extension of the existing `mx_shadow`, and not a new universal module dedicated to lighting.

ld-kerley: happy to go with that if you want - I opted for this because I thought that over time we might add more "lighting" functions to it - extracting more of the code from the C++ nodes in to the data library - marching ever more towards the goal of being fully data-driven :).

Another vote in the column of having a more general name would be that this file, currently, needs to be included in the shader generator C++ code, so it actually introduces a new non-data-driven dependency, which would be another vote for a more general name so we can re-use it for other things. 

If you don't think those are strong enough arguments - I'm happy to rename it to your suggestion.

We could also call it `mx_lighting_platform.<ext>` if the goal here is to indicate that it's not a general purpose file, but platform specific. Though I'd note that we don't do that for `mx_math.<ext>` - which is similarly specialized between GLSL and MSL.

In the `genmsl` source locations its pretty clear - because there aren't many files :), but I do see that in the `genglsl` location that differentiation between general source and platform specific source is less clear.

I wonder actually if maybe we should try and consolidate all the "platform" specific pieces somehow - either in a separate folder, or perhaps just in a single file `mx_platform.<ext>`?   Having a single file like this would be nice in terms of reducing the coupling with the C++ shader generator side, though we would probably have to be careful to make sure any refactor like this is backwards compatible.

jstone-lucasfilm:
@ld-kerley Completely agreed that we should aim to move more code from our C++ logic to these shading language helper files over time, and also agreed that we can do a better job in clarifying which files are universal and which are platform-specific.

I think the strongest argument for naming this new file `shadow_platform` is the following logic in the MSL generator:

```
        if (shadowing)
        {
            emitLibraryInclude("pbrlib/genglsl/lib/mx_shadow.glsl", context, stage);
            emitLibraryInclude("pbrlib/genmsl/lib/mx_lighting.metal", context, stage);
        }
```

Since this helper file is only emitted when shadowing is active, and not in the broader `if (lighting)` code path, I think it makes sense to consider this a platform-specific shadowing file, and further in the future we can create parallel helper files named `lighting` and `lighting_platform` as required.

ld-kerley:
I worry a little about knock on implications for downstream derived shader generators if we're adding new files at a regular cadence.  Thats where the desire to try and be "one and done" with the additional file comes from.

I'm very open to moving where the file is included - if that means we can make more use of the platform specific file. 

Just looking to confirm you understand the concern with impacting downstream integrators of MaterialX. 

The more I think about it - I really like the idea of a `mx_platform.<ext>` file that would get included - likely in a similar place to where we included `mx_math` today, that would hold all the platform specific differences.  I'm totally fine if we want that file to then include a number of other source files - so things are broken down in to smaller logical blocks etc.  

But I really think it would be a big positive step forwards to reduce the coupling with the C++ shader generator and the data library, not increase it.

jstone-lucasfilm:
From a downstream dependency perspective, I can see the benefits in a single `mx_platform.<ext>` approach, but there are major tradeoffs involved, as shader compilation times scale with the number of lines of code that we pass to the compiler.  That's the reason that we currently focus on very _granular_ inclusion of shading code for each language, in order to keep shader compilation as fast as possible.

This is a bigger discussion than we can resolve here, but I definitely see the downstream dependency benefits of what you're suggesting, and we just need to manage these benefits against the compilation performance goals for our end users.

ld-kerley:
I've renamed the files as you're asking here - but I don't think this is the right direction for the project to move.  I think we should be weighing system stability higher than possible compile time performance. If we continue this way, we are making the lives of our OpenUSD friends more difficult.

jstone-lucasfilm:
That's a completely valid perspective, @ld-kerley, and we could make a global shift from "granular includes" to "monolithic includes" in the future, assuming the full community is behind this idea.

Since we're using the "granular includes" approach consistently in our code today, I think maintaining it in your update makes sense, but there's nothing wrong with reconsidering this approach globally in the future.

Issue #2595, opened 2025-09-28
Removal of SurfaceNodeGlsl.h


limann77:
[OpenUSD](https://github.com/PixarAnimationStudios/OpenUSD/commits/dev/) (even the dev version) can't be build without SurfaceNodeGlsl.h which was removed after 1.39.4 in https://github.com/AcademySoftwareFoundation/MaterialX/commit/5818e06e79091f4fbb4219da8951fb996f1c4ce3

```
FAILED: pxr/imaging/hdSt/CMakeFiles/hdSt.dir/materialXShaderGen.cpp.o 
[...]

PR #2610, opened 2025-10-08
Add backwards compatibility SurfaceNode headers

ld-kerley:
It was reported in #2595 that the current MaterialX codebase won't compile with OpenUSD

This PR restores two files being included by the current OpenUSD build that were removed, along with the necessary alias to allow OpenUSD to build.

### DIFF
--- a/source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.h
+++ b/source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.h
@@ -10,11 +10,8 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
-// This alias from SurfaceNodeGlsl to HwSurfaceNode is
-// a backwards compatibility affordance for OpenUSD.
-// The old SurfaceNodeGlsl class was removed in MatX 1.39.5
-// once that becomes the minimum version for MaterialX
-// in OpenUSD we can remove this.
+// Backward compatibility header for OpenUSD and other external
+// projects referencing files that were removed in MaterialX 1.39.5
 using SurfaceNodeGlsl = HwSurfaceNode;
 
 MATERIALX_NAMESPACE_END
--- a/source/MaterialXGenMsl/Nodes/SurfaceNodeMsl.h
+++ b/source/MaterialXGenMsl/Nodes/SurfaceNodeMsl.h
@@ -10,11 +10,8 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
-// This alias from SurfaceNodeMsl to HwSurfaceNode is
-// a backwards compatibility affordance for OpenUSD.
-// The old SurfaceNodeGlsl class was removed in MatX 1.39.5
-// once that becomes the minimum version for MaterialX
-// in OpenUSD we can remove this.
+// Backward compatibility header for OpenUSD and other external
+// projects referencing files that were removed in MaterialX 1.39.5
 using SurfaceNodeMsl = HwSurfaceNode;
 
 MATERIALX_NAMESPACE_END


source/MaterialXGenGlsl/Nodes/SurfaceNodeGlsl.h
```diff
@@ -0,0 +1,21 @@
+//
+// Created by Lee Kerley on 10/8/25.
```
jstone-lucasfilm: Just a note that we should use the standard MaterialX header here!

Here's an example of the standard header, which contains the collective copyright and license identifier for the MaterialX project:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Element.h

ld-kerley: oops - thanks for catching - I'll fix this.

ld-kerley:
@jstone-lucasfilm could you elaborate a little more on what your concrete concerns are with the "shim" headers is?  They seem like they would have very minimal impact to the MaterialX codebase - and would provide out OpenUSD colleagues a much easier path to continue to stay on current versions of MaterialX. 

I would much rather we absorb these compatibility affordances, especially if they don't come at significant detriment on the MaterialX side, rather than ask the OpenUSD team to increase their already complex build matrix to support different versions of MaterialX. It would be really helpful to understand your perspective if you think they are a significant detriment.

jstone-lucasfilm:
Taking this idea forward, I have a few specific suggestions for improvements, as it seems likely that we'll need to repeat this pattern many times in the future, and taking the time to establish a good template seems worthwhile:

- I would move the includes that _validate_ the shim headers from `MaterialXGen*` to `MaterialXTest`, so that they are only compiled in contexts that include the MaterialX test suite.
- I would simplify the comment language in each shim header, since we'll likely need to repeat this language many times in the future.  Here's an initial suggestion, and feel free to refine this further: `Backward compatibility header for OpenUSD and other external projects referencing files that were removed in MaterialX 1.39.5`.
- We can safely remove the comment after the `#endif` in each shim header, as this isn't a convention for MaterialX headers.

ld-kerley:
I moved the shim includes to the test files - so long as we always continue to build the tests in CI - this performs the same purpose.
I remove the errant closing `#endif` comment - I think that was auto created by my IDE.

I would like to keep the more verbose language for the shim comment - I think it's helpful to record the future action in the code (as well as potentially a GitHub issue). Otherwise we may just continue to carry this indefinitely, I know I have come across code in the project that was no longer needed, or could be refactored because of natural evolution. I feel like this comment will help our future selves (or future maintainers) to know when the right time is to be able to remove these additional headers.

jstone-lucasfilm:
This is looking promising, though I would say that the comment descriptions are still too focused on a single external client (OpenUSD), in addition to being a bit wordy.

One motivation for my alternative text was to highlight the value of these shim headers to _all external projects_ that build against MaterialX, where OpenUSD is just the first one that has been brought to our attention.  Even once OpenUSD no longer supports MaterialX versions before 1.39.5, we'll still need a broader conversation with the community before these shim headers can safely be removed.

Here's my initial proposal for that revised language, and of course you should feel free to edit this as needed:

```
Backward compatibility header for OpenUSD and other external projects referencing files that were removed in MaterialX 1.39.5
```

ld-kerley:
How about 

```
Backward compatibility header for OpenUSD and other external projects referencing files that were removed in MaterialX 1.39.5. 
This file can be removed when the minimum required MaterialX version exceeds 1.39.5.
```

or

```
Backward compatibility header for external projects referencing files that were removed. 
This file can be removed when the minimum required MaterialX version exceeds 1.39.5.
```

I would love to leave a note in the code reminding us when it's safe to remove this in the future. Documentation in the code is intended to help our future selves :)

jstone-lucasfilm:
To me, that second sentence begs the question -- when the minimum required MaterialX version _of what project_ exceeds 1.39.5?

It wouldn't be sufficient for the minimum required MaterialX version of _OpenUSD_ to reach this milestone, as other external projects have custom MaterialX shader generators as well, and we'd really need to reach out to the entire community before making such a change.

That's the motivation, to my mind, for focusing on the concrete purpose of the shim header, without setting too low of a bar for the complex set of steps that would be required to remove it.

ld-kerley:
I don't agree - I don't think it attempts to infer any definition of when the requirement might be met - just that once we meet it - this file is no longer needed. 

I really don't understand what the objection to one additional comment line in two files is - but its not worth the effort to me to continue to debate this, so I will just change the comment to exactly what you request so we can move on.

PR #2614, opened 2025-10-11
Remove duplicate includes

jstone-lucasfilm:
This changelist removes duplicate includes across the MaterialX C++ codebase, specifically targeting cases where a header has been included both transitively and directly within the same source file.

No functional changes to the codebase have been made, and no changes in build behavior are expected on any platform, though build times may be slightly improved.

### DIFF
--- a/source/JsMaterialX/JsMaterialXCore/JsElement.cpp
+++ b/source/JsMaterialX/JsMaterialXCore/JsElement.cpp
@@ -7,11 +7,6 @@
 #include <JsMaterialX/Helpers.h>
 
 #include <MaterialXCore/Document.h>
-#include <MaterialXCore/Geom.h>
-#include <MaterialXCore/Look.h>
-#include <MaterialXCore/Material.h>
-#include <MaterialXCore/Node.h>
-#include <MaterialXCore/Traversal.h>
 
 #include <emscripten/bind.h>
 
--- a/source/JsMaterialX/JsMaterialXGenShader/JsGenContext.cpp
+++ b/source/JsMaterialX/JsMaterialXGenShader/JsGenContext.cpp
@@ -7,7 +7,6 @@
 
 #include <MaterialXCore/Unit.h>
 #include <MaterialXGenShader/GenContext.h>
-#include <MaterialXGenShader/ShaderGenerator.h>
 #include <MaterialXGenShader/DefaultColorManagementSystem.h>
 #include <MaterialXFormat/Util.h>
 
--- a/source/JsMaterialX/JsMaterialXGenShader/JsShader.cpp
+++ b/source/JsMaterialX/JsMaterialXGenShader/JsShader.cpp
@@ -6,7 +6,6 @@
 #include <JsMaterialX/Helpers.h>
 
 #include <MaterialXGenShader/Shader.h>
-#include <MaterialXGenShader/ShaderStage.h>
 
 #include <emscripten/bind.h>
 
--- a/source/JsMaterialX/JsMaterialXGenShader/JsShaderGenerator.cpp
+++ b/source/JsMaterialX/JsMaterialXGenShader/JsShaderGenerator.cpp
@@ -4,7 +4,6 @@
 //
 
 #include <MaterialXGenShader/GenContext.h>
-#include <MaterialXGenShader/ShaderGenerator.h>
 
 #include <emscripten/bind.h>
 
--- a/source/MaterialXCore/Element.cpp
+++ b/source/MaterialXCore/Element.cpp
@@ -6,7 +6,6 @@
 #include <MaterialXCore/Element.h>
 
 #include <MaterialXCore/Document.h>
-#include <MaterialXCore/Util.h>
 
 #include <iterator>
 
--- a/source/MaterialXCore/Node.cpp
+++ b/source/MaterialXCore/Node.cpp
@@ -6,7 +6,6 @@
 #include <MaterialXCore/Node.h>
 
 #include <MaterialXCore/Document.h>
-#include <MaterialXCore/Material.h>
 
 #include <deque>
 
--- a/source/MaterialXFormat/XmlIo.cpp
+++ b/source/MaterialXFormat/XmlIo.cpp
@@ -7,8 +7,6 @@
 
 #include <MaterialXFormat/External/PugiXML/pugixml.hpp>
 
-#include <MaterialXCore/Types.h>
-
 #include <cstring>
 #include <fstream>
 #include <sstream>
--- a/source/MaterialXGenGlsl/EsslShaderGenerator.cpp
+++ b/source/MaterialXGenGlsl/EsslShaderGenerator.cpp
@@ -6,8 +6,6 @@
 #include <MaterialXGenGlsl/EsslShaderGenerator.h>
 #include <MaterialXGenGlsl/EsslSyntax.h>
 
-#include <MaterialXGenHw/HwConstants.h>
-
 MATERIALX_NAMESPACE_BEGIN
 
 const string EsslShaderGenerator::TARGET = "essl";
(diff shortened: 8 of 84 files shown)


ld-kerley:
I certainly want to read this more closely - but I'm curious if you can explain more concretely what your strategy is here.

Generally I've always been told you should "include what you use". I'm sure after many refactors there are errant headers around, but are we leaving all the headers necessary for the elements used locally? or are you relying on transitive includes? I would suggest we shouldn't do that, because it makes future edits more fragile. Which can be bigger issue for projects with downstream consumers - like we have in OpenUSD.

jstone-lucasfilm:
Good questions, @ld-kerley, and yes -- this optimization is designed to be completely transparent to clients such as OpenUSD and Unreal Engine, where they can order their MaterialX includes any way that they prefer, and will encounter no changes in build behavior.

The key property that we're preserving in this PR is the _self-sufficiency_ of each source and header file in the MaterialX codebase, where it must include every header needed to support its own code, without relying upon the client to have any knowledge of its dependencies.

In terms of the strategy used in developing this PR, I created a _graph_ of the include dependencies across the entire MaterialX codebase, and then analyzed this graph to determine which headers were being included twice from the same source or header file.  With this data, it was straightforward to remove just the _duplicate_ includes, leaving the build behavior unchanged for both internal tools and external clients.

ld-kerley:
Is this graph - and whatever tool you used to create it - in any form you can share? that would make this PR easier tor review.

I'm a little confused by the term "duplicated includes" here - if we are "including what you use" then this would only really mean the same header was included twice in a given file - wouldn't it?

As I understand it "Include what you use" means include locally all the things you need locally - not relying on transitive includes that could change and invalidate local include requirements? 

In terms of downstream (OpenUSD) users - I think "Include what you use" is really important to adhere to. It would be ideal if there are any analysis tools that enforce local inclusion of required headers.

jstone-lucasfilm:
@ld-kerley You bring up some interesting points here, though I think your view of the ideal approach for header inclusion in MaterialX is very different from my own, and indeed very different from the current pattern we use in the MaterialX codebase.

Here's an example of a transitive include structure that we commonly use in MaterialX, which seems like a good pattern from my way of thinking, but which doesn't match what you're advocating:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Material.h

In this header, we're only including `Node.h`, but we're referencing the transitively included `OutputPtr`, which is only made available through the pathway from `Node.h` to `Definition.h` to `Interface.h`.

To me, this transitive inference seems natural, as support for `Node` _implies_ support for `Input` and `Output` as well, since it's not possible to meaningfully define the behavior of nodes without also defining their inputs and outputs.  But these concepts are defined in different header files, and it requires an embrace of transitive dependencies to achieve this through a single include in `Material.h`.

I can see where there's plenty of room for different opinions here, and I won't claim that my view on this is any more _correct_ than yours, but it definitely matches our current approach across the MaterialX codebase.

ld-kerley:
Thank you for the helpful explanation - I would still love to better understand whatever analysis tool you're using - it sounds super helpful.

Another case I'm curious to understand from you is demonstrated here:

[https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Traversal.h](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Traversal.h)

Where instead of including the necessary header to expose `Element` we just use a forward declaration. Perhaps we should lean in to this approach more instead? 

I guess I'm hoping to understand the motivations behind these changes, and see a corresponding addition to the developer guidelines, so it's easier for people to consistently follow the intended style of the project.

jstone-lucasfilm:
That's a great question, @ld-kerley, and the motivation for the use of forward declarations in `Traversal.h` is to avoid creating an _include cycle_, as `Element.h` needs to include `Traversal.h` in order to provide the interface for its `traverseGraph` method.  We needed to choose just _one_ direction for inclusion between these two files, and the inclusion of `Element.h` from `Traversal.h` can easily be avoided through the use of forward declarations.

I'd be happy to share the graph analysis tools that I'm using in a private thread, since they're not battle-tested production tools that I'd yet want to share publicly in an open source project.  Essentially they are just a collection of Python scripts that can be used to analyze the include dependencies in a C++ project, and if they turn out to be useful beyond MaterialX, we can improve them to the level that they might be worthwhile to share further.

I'm always interested in improving our developer guidelines, and perhaps that's a project that we can collaborate on in the future, as documentation often turns out to be much more challenging to write well than code.

ld-kerley:
Maybe we can use this piece of work to concretely improve the developer guidelines now?  Much as we do for proposing new syntax to the specification, if we're taking action to improve the codebase using some specific approach here - perhaps we can document it at the same time? Having definitive language around the expectations here would also make the specific PR easier to review too.

jstone-lucasfilm:
@ld-kerley Are you imagining new developer documentation on how we use forward declarations to avoid include cycles in the MaterialX codebase, how transitive includes can be used to simplify and clarify code, and other recommendations along those lines?

I like the idea of adding that documentation to assist future MaterialX developers, but my sense is that it's a much larger project than the minor optimization I've proposed here.

I wasn't thinking that this optimization would be very controversial, since the only possible impact is improved build times, but if folks have concerns about this, we can hold off on this PR and leave our duplicated includes untouched.

ld-kerley:
I was just thinking that if we don't start trying to improve the documentation now - we might never around to doing it :).  This improvement seems helpful, but doesn't seem like something that is time critical, so maybe its a good point to lean in to improving the documentation at the same time - much as we do for the specification and new syntax. 

I was curious to see what your proposed text would be for describing all the rules around both include structure and forward declaration for the project, and knowing concretely what those rules are in a form that could be added to the documentation is what would make reviewing this PR a lot easier.  I'd be happy to review the documentation rules with you and help refine them to something we think will be helpful to everyone.

If I'm missing the point here - and in fact this PR is time critical for some reason, then I'm happy to look at it more closely now, but then it would be helpful to share the tool - even if its not production ready, as the codified version of the rules would also be helpful to understand.

jstone-lucasfilm:
I wish I could say that I had a unified set of concrete rules, but my own understanding of best practices for header includes is really just a set of _intuitions_, built in the trenches of software development with the great teams at ILM RnD and the Lucasfilm ADG.

Just focusing on the concepts that we've been discussing in this thread, here are a few intuitions that come to mind:

- Include blocks should be ordered _hierarchically_, with high-level modules listed before low-level modules (e.g. `MaterialXGenShader`, followed by `MaterialXFormat`, followed by `MaterialXCore`).  This maximizes opportunities to catch missing dependencies in the high-level modules, which might otherwise be hidden at build time.
- Within include blocks, individual includes should be ordered _alphabetically_, providing a simple canonical order that is easy for both developers and automated systems to check.
- In the interest of avoiding include cycles, developers are free to leverage _forward declarations_ of classes that are trivially referenced within another header.
- In the interest of clarity and efficiency, developers are free to leverage _transitive header includes_, where low-level headers that have already been included by a high-level header do not need to be restated individually.

jstone-lucasfilm:
Perhaps I should add these notes to a new section of our [Developer Guide](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/DeveloperGuide/MainPage.md), titled `Developer Guidelines`?

We could launch this new section with a single subsection titled `Include Headers` for now, with a plan to add more sections as we recognize their value in the future.

jstone-lucasfilm:
@ld-kerley Ok, I've followed your advice, and here is an initial set of proposed Developer Guidelines for MaterialX!

https://github.com/AcademySoftwareFoundation/MaterialX/pull/2632

I've included guidelines for Header Includes, along with many other guidelines that we strive to follow in the MaterialX project.

Let me know your thoughts!

jstone-lucasfilm:
Since this proposal is now a bit out-of-date with respect to `main`, I'll close this original PR out, and we can revisit the idea of graph-based analysis and optimization of includes in the future.

PR #2615, opened 2025-10-12
Simplify hextiling implementations

jstone-lucasfilm:
This changelist simplifies the hextiling implementations in GLSL and MDL, with the goal of making them more maintainable and extensible to new languages in the future.  In GLSL testing, no changes to rendered visuals or performance have been detected.

One natural next step will be to implement the equivalent hextiling logic in OSL, using the simplified approach from this changelist as a template.


jstone-lucasfilm:
Hextiled Onyx Before:

OnyxHextiled_Old

Hextiled Onyx After (no visual or performance changes):

OnyxHextiled_New

jstone-lucasfilm:
I'm especially interested in thoughts from @msuzuki-nvidia on this change, but GitHub doesn't let me add him as a reviewer, so I'm CC'ing him here.

jstone-lucasfilm:
I'm additionally CC'ing @krohmerNV and @jreichel-nvidia for their MDL expertise, since our work on GitHub CI validation of MDL syntax is still in progress.

I've taken great care to follow the syntax rules of MDL in these updates, but I certainly could have missed an important detail.

ld-kerley:
Were you able to measure any performance difference before/after this change - so of the modifications look like they may have sped things up...

jstone-lucasfilm:
Although many of the improvements in this PR have _potential_ performance benefits, including vectorization of `sin` and `cos` calculations that were formerly repeated multiple times, I ran a large number of tests on an NVIDIA RTX A6000, and the performance before and after this PR was identical in all tests.

It's hard to be sure as to why this is the case, and my theory is that the graphics drivers were already able to detect these vectorization optimizations on their own, and this change merely simplifies the work that they need to perform at shader compilation time.

ld-kerley:
If there are no measurable benefits here - then we should make sure that all the affected platforms don't have any negative effects. I don't expect they will - but the mix bsdf optimization was unexpectedly impactful to the MDL generator - so we should make sure we're building process to make sure we're validating these things :)

msuzuki-nvidia:
The reason I used non-array variables for those is because the MDL compiler had an issue with arrays in struct at the point I wrote the code. I _think_ the bug has been fixed now but I need to run tests on my end. Please allow me take some time to clarify.

jstone-lucasfilm:
@msuzuki-nvidia Since we still support MDL versions back to 1.6 in MaterialX, I wonder if perhaps this simplification will not end up being effective across the full range of MDL versions?

If that's the case, then we can hold off on this PR, as we would need to rethink the category of simplifications that will work consistently across GLSL, OSL, and MDL.

msuzuki-nvidia:
@jstone-lucasfilm The issue I encountered was _not_ caused by a particular version of the MDL spec. It was simply a bug in the MDL compiler. I believe MDL 1.6 (or later) spec supports arrays in struct, so this simplification shouldn't be a problem in 1.6 code.

source/MaterialXGenMdl/mdl/materialx/hextile.mdl
```diff
@@ -25,15 +25,27 @@ export float mx_schlick_gain(float x, float r)
...
+    }
+    return w;
+}
+
 export struct mx_hextile_data
 {
-    float2 coord1;
-    float2 coord2;
-    float2 coord3;
+    float2 coords[3];
```
msuzuki-nvidia: Correct syntax in MDL is:
```
float2[3] coords;
```

msuzuki-nvidia:
The issue in MDL compiler I had before has been fixed, so this change should be fine.

jstone-lucasfilm:
Following up on this PR, now that the automated MDL shader testing work from @jreichel-nvidia and @krohmerNV has been merged to MaterialX main, I've addressed the typo in my implementation, and all of the extended MDL tests pass:

https://github.com/jstone-lucasfilm/MaterialX/actions/runs/19165806216/job/54786084974

I'm pinging @msuzuki-nvidia, @niklasharrysson, and @ld-kerley to check whether this looks good from their perspective, and if all teams are in agreement, we can move forward with these simplifications.

msuzuki-nvidia:
LGTM!

PR #2617, opened 2025-10-12
Fix shadow and environment edge cases in GLSL generation

craouette:
shadow map projection for spot light (divide by shadowCoord.w) 
shadow map only for the first light (occlusion = 1 after the first light),
environment_radiance in case u_envRadianceSamples is 0 (in library returns 0)

### DIFF
--- a/libraries/stdlib/genmsl/lib/mx_math.metal
+++ b/libraries/stdlib/genmsl/lib/mx_math.metal
@@ -6,27 +6,27 @@
 #define mx_asin metal::asin
 #define mx_acos metal::acos
 
-vec2 mx_matrix_mul(vec2 v, mat2 m) { return v * m; }
-vec3 mx_matrix_mul(vec3 v, mat3 m) { return v * m; }
-vec4 mx_matrix_mul(vec4 v, mat4 m) { return v * m; }
-vec2 mx_matrix_mul(mat2 m, vec2 v) { return m * v; }
-vec3 mx_matrix_mul(mat3 m, vec3 v) { return m * v; }
-vec4 mx_matrix_mul(mat4 m, vec4 v) { return m * v; }
-mat2 mx_matrix_mul(mat2 m1, mat2 m2) { return m1 * m2; }
-mat3 mx_matrix_mul(mat3 m1, mat3 m2) { return m1 * m2; }
-mat4 mx_matrix_mul(mat4 m1, mat4 m2) { return m1 * m2; }
+float2 mx_matrix_mul(float2 v, float2x2 m) { return v * m; }
+float3 mx_matrix_mul(float3 v, float3x3 m) { return v * m; }
+float4 mx_matrix_mul(float4 v, float4x4 m) { return v * m; }
+float2 mx_matrix_mul(float2x2 m, float2 v) { return m * v; }
+float3 mx_matrix_mul(float3x3 m, float3 v) { return m * v; }
+float4 mx_matrix_mul(float4x4 m, float4 v) { return m * v; }
+float2x2 mx_matrix_mul(float2x2 m1, float2x2 m2) { return m1 * m2; }
+float3x3 mx_matrix_mul(float3x3 m1, float3x3 m2) { return m1 * m2; }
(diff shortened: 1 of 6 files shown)


jstone-lucasfilm:
Thanks for putting this proposal together, @craouette!

I'm CC'ing @kwokcb for his expertise with this logic, with an eye towards making sure that we're not overlapping any existing functionality or introducing any performance regressions with the change.

libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
```diff
@@ -2,6 +2,8 @@
 
 vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 alpha, int distribution, FresnelData fd)
 {
+    if ($envRadianceSamples == 0)
```
jstone-lucasfilm: Instead of adding an early out here, which is likely to generate a full dynamic branch in the GPU code, it may be better to compute `envRadianceSamples` in the following way:

```
int envRadianceSamples = max($envRadianceSamples, 1);
```

craouette: yes, make sense...  returning 0 seems more relevant to the function parameters (and avoid texture sampling)... not sure which one is faster, but I take your word for it.
May be modifying the last line is a the best option:
form
 return radiance * $envLightIntensity;
 to
  return ($envRadianceSamples == 0 ? vec3(0.0) : radiance) * $envLightIntensity;
  
  what do you think?
  Do I make the modif or do you do it?

jstone-lucasfilm: That's a great proposal as well, and let's go with your suggested solution.

Feel free to make the change on your side!

source/MaterialXGenHw/Nodes/HwSurfaceNode.cpp
```diff
@@ -270,6 +271,9 @@ void HwSurfaceNode::emitLightLoop(const ShaderNode& node, GenContext& context, S
         shadergen.emitComment("Accumulate the light's contribution", stage);
         shadergen.emitLine(outColor + " += lightShader.intensity * " + bsdf->getOutput()->getVariable() + ".response", stage);
 
+        shadergen.emitComment("Clear shadow factor for next light", stage);
```
jstone-lucasfilm, line 276: Maybe we should have a call to `shadergen.emitLineBreak(stage);` before this new comment, so that this block of code is visually separated in the GLSL output?

craouette: Currently, the generated code look likes:
```
        for (int activeLightIndex = 0; activeLightIndex < numLights; ++activeLightIndex)
        {
            sampleLightSource(u_lightData[activeLightIndex], vd.positionWorld, lightShader);
            L = lightShader.direction;

            // Calculate the BSDF response for this light source
            ClosureData closureData = makeClosureData(CLOSURE_TYPE_REFLECTION, L, V, N, P, occlusion);

           ...

            BSDF thin_film_layer_attenuated_out = BSDF(vec3(0.0),vec3(1.0));
            mx_multiply_bsdf_color3(closureData, metalness_blend_out, coat_attenuation_out, thin_film_layer_attenuated_out);
            BSDF coat_layer_out = BSDF(vec3(0.0),vec3(1.0));
            mx_layer_bsdf(closureData, coat_bsdf_out, thin_film_layer_attenuated_out, coat_layer_out);

            // Accumulate the light's contribution
            shader_constructor_out.color += lightShader.intensity * coat_layer_out.response;
            // Clear shadow factor for next light
            occlusion = 1.0;
        }
```
        
        it looks ok for me, but I can add a new line if you want... just tell me and I will do it.

jstone-lucasfilm: Yes, I think the newline would be helpful here, so that readers don't assume the entire block is focused on accumulating the light contribution.

A minor detail, for sure, but I think it's worthwhile update.

craouette: done...

PR #2619, opened 2025-10-14
Fix cmake rebuild for MaterialXRenderGlsl. Fix PyTinyObjectLoader.load. 

Javantea:
You need to take care about adding files from now on but that is normal.
Fix PyTinyObjLoader.load so that it returns meshes instead of not returning meshes.
These are the first patches toward getting Python rendering working which should be a lot of fun.

### DIFF
--- a/source/PyMaterialX/PyMaterialXRender/PyTinyObjLoader.cpp
+++ b/source/PyMaterialX/PyMaterialXRender/PyTinyObjLoader.cpp
@@ -6,14 +6,24 @@
 #include <PyMaterialX/PyMaterialX.h>
 
 #include <MaterialXRender/TinyObjLoader.h>
+#include <utility>
 
 namespace py = pybind11;
 namespace mx = MaterialX;
+
+std::pair<bool, mx::MeshList> load_meshes_wrapper_tuple(mx::TinyObjLoader& self,
+                                                        const mx::FilePath& filePath,
+                                                        bool texcoordVerticalFlip)
+{
+    mx::MeshList meshList;
+    bool success = self.load(filePath, meshList, texcoordVerticalFlip);
+    return {success, meshList};
+}
 
 void bindPyTinyObjLoader(py::module& mod)
 {
     py::class_<mx::TinyObjLoader, mx::TinyObjLoaderPtr, mx::GeometryLoader>(mod, "TinyObjLoader")
         .def_static("create", &mx::TinyObjLoader::create)
         .def(py::init<>())
-        .def("load", &mx::TinyObjLoader::load);
+        .def("load", &load_meshes_wrapper_tuple);
 }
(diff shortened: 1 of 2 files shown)


source/PyMaterialX/PyMaterialXRender/PyTinyObjLoader.cpp
```diff
@@ -6,14 +6,24 @@
...
+    return {success, meshList};
+}
+
 void bindPyTinyObjLoader(py::module& mod)
 {
     py::class_<mx::TinyObjLoader, mx::TinyObjLoaderPtr, mx::GeometryLoader>(mod, "TinyObjLoader")
         .def_static("create", &mx::TinyObjLoader::create)
         .def(py::init<>())
-        .def("load", &mx::TinyObjLoader::load);
+        .def("load", &load_meshes_wrapper_tuple);
```
kwokcb, line 28: I'm guessing this is an attempt to fix that the meshlist is not being updated since it's a referenced shared point list ? We should not change an existing signature to handle this.

I think the easiest fix in this case is to expose Python wrapper for `MeshList` in `PyMesh.cpp`. e.g:
```py::bind_vector<mx::MeshList>(mod, "MeshList");```
I think this was just a oversight. 

As "load" in in the base class Python wrapper I think these derived class "load"s can be removed as well. There is one for `CgltfLoader` as well as `TinyObjLoader`. 

Any new / user defined Python derived loaders will would also work properly.

To test this, the workflow of adding loaders to a handler should also be tested. There load and get are separated into two different methods. (BTW, Working via a handler is the desired pattern to follow to allow per loader extension handling.)

It would be good to have a unit test for this as well. e.g. of handler workflow:
```python
import MaterialX as mx
import MaterialX.PyMaterialXRender as mxr

# Test obj, gltf should also be tested
location = "./resources/Geometry/plane.obj"
meshlist = [] # This would create a Python wrapper MeshList instead.
flipUVs = False

cgltf = mxr.CgltfLoader.create()
objl = mxr.TinyObjLoader.create()
# Individua loader test
loaded = objl.load(location, meshlist, flipUVs)

handler = mxr.GeometryHandler.create()
handler.addLoader(cgltf)
handler.addLoader(objl)

loaded = handler.loadGeometry(location, flipUVs)
handler.getGeoemetry(meshlist, location)
```

Javantea: I'll give that a try. I'm much more flexible in changing signature when it is necessary but you have figured out what I could not on MeshList. Thanks for your help in looking at this issue that was blocking rendering in python (which is now working I might add, with a few more changes which will be coming soon). I like the idea of a unit test for this behavior.

jstone-lucasfilm:
@Javantea I wanted to follow up on this PR, as I think it's a valuable goal to make MaterialX rendering accessible from Python, and your work here is a good start.

I agree with @kwokcb's suggestion that the best path forward would be adding `py::bind_vector<mx::MeshList>` in PyMesh.cpp. That would fix the reference-passing issue at the root for all loaders (TinyObjLoader, CgltfLoader, and any future ones) without needing per-loader wrapper functions, and it would keep the Python API consistent with C++.

Let us know if you might have the bandwidth to continue work on this PR, and if it helps in focusing the work, I think it would be fine to consider the MeshList opaque binding change and CMake glob change as separate PRs.

PR #2625, opened 2025-10-15
Update C++ Hw nodes to emit the syntax based type names

ld-kerley:
Following on from #2622. Improving the direct support of Metal syntax. 

The C++ Hw nodes currently have some hardcoded GLSL'isms in them, and because we're generating this code dynamically we can lean on the MaterialX Shader Generator Syntax system to actually emit the correct type names, and not rely on the string post-processing to correct the syntax. 

This will set us up well to support any other derived languages that might have different syntax.

### DIFF
--- a/source/MaterialXGenHw/Nodes/HwSurfaceNode.cpp
+++ b/source/MaterialXGenHw/Nodes/HwSurfaceNode.cpp
@@ -135,10 +135,10 @@ void HwSurfaceNode::emitFunctionCall(const ShaderNode& node, GenContext& context
             if (context.getOptions().hwAmbientOcclusion)
             {
                 ShaderPort* texcoord = vertexData[HW::T_TEXCOORD + "_0"];
-                shadergen.emitLine(syntax.getTypeName(Type::VECTOR2)+" ambOccUv = " + prefix + texcoord->getVariable(), stage);
+                shadergen.emitLine(vec2+" ambOccUv = " + prefix + texcoord->getVariable(), stage);
                 if (context.getOptions().fileTextureVerticalFlip)
                 {
-                    shadergen.emitLine("ambOccUv = "+syntax.getTypeName(Type::VECTOR2)+"(ambOccUv.x, 1.0 - ambOccUv.y)", stage);
+                    shadergen.emitLine("ambOccUv = "+vec2+"(ambOccUv.x, 1.0 - ambOccUv.y)", stage);
                 }
                 shadergen.emitLine("occlusion = mix(1.0, texture(" + HW::T_AMB_OCC_MAP + ", ambOccUv).x, " + HW::T_AMB_OCC_GAIN + ")", stage);
             }


source/MaterialXGenHw/Nodes/HwSurfaceNode.cpp
```diff
@@ -89,10 +90,10 @@ void HwSurfaceNode::emitFunctionCall(const ShaderNode& node, GenContext& context
...
         shadergen.emitScopeBegin(stage);
 
-        shadergen.emitLine("vec3 N = normalize(" + prefix + HW::T_NORMAL_WORLD + ")", stage);
-        shadergen.emitLine("vec3 V = normalize(" + HW::T_VIEW_POSITION + " - " + prefix + HW::T_POSITION_WORLD + ")", stage);
-        shadergen.emitLine("vec3 P = " + prefix + HW::T_POSITION_WORLD, stage);
-        shadergen.emitLine("vec3 L = vec3(0,0,0);", stage);
+        shadergen.emitLine(syntax.getTypeName(Type::VECTOR3)+" N = normalize(" + prefix + HW::T_NORMAL_WORLD + ")", stage);
+        shadergen.emitLine(syntax.getTypeName(Type::VECTOR3)+" V = normalize(" + HW::T_VIEW_POSITION + " - " + prefix + HW::T_POSITION_WORLD + ")", stage);
+        shadergen.emitLine(syntax.getTypeName(Type::VECTOR3)+" P = " + prefix + HW::T_POSITION_WORLD, stage);
+        shadergen.emitLine(syntax.getTypeName(Type::VECTOR3)+" L = "+syntax.getValue(Type::VECTOR3, HW::zeroVec3), stage);
```
niklasharrysson: Should we trust the compiler to optimize repeated function calls, or perhaps proactively remove the redundant calls ourselves? This pattern applies to a few other places in this PR as well.

source/MaterialXGenHw/HwConstants.h
```diff
@@ -247,6 +249,13 @@ extern MX_GENHW_API const string USER_DATA_BINDING_CONTEXT;
 
 /// Type Descriptor for closure context data.
 extern MX_GENHW_API const TypeDesc ClosureDataType;
+
+/// Constant Values
+extern MX_GENHW_API const TypedValue<Vector2> zeroVec2;
+extern MX_GENHW_API const TypedValue<Vector2> oneVec2;
+extern MX_GENHW_API const TypedValue<Vector3> zeroVec3;
+extern MX_GENHW_API const TypedValue<Vector3> oneVec3;
```
niklasharrysson: Being constants perhaps these should be all upper case for consistency with coding convention?

niklasharrysson (commented):
This looks like a great change to make this more dynamic, and I just added a few comments above.

source/MaterialXGenHw/Nodes/HwSurfaceNode.cpp
ld-kerley: I like this idea - doing this might actually make the embedded shader code a little easier to read too. 

.....

I went with `vec3_zero` because I'm not generally a fan of numbers in variable names, but it does occur to me that we could do

```
const string& vec3_0 = syntax.getValue(Type::VECTOR3, HW::VEC3_ZERO);
```

and then a line line 
```
        shadergen.emitLine(vec3+" L = "+vec3_zero, stage);
``` 
would actually read as 
```
        shadergen.emitLine(vec3+" L = "+vec3_0, stage);
```
which is a little closer to the original GLSL of 
```
        shadergen.emitLine(vec3+" L = vec3(0)", stage);
```

What do you think?

source/MaterialXGenHw/Nodes/HwLightNode.cpp
```diff
@@ -51,8 +40,15 @@ void HwLightNode::emitFunctionCall(const ShaderNode& node, GenContext& context,
     DEFINE_SHADER_STAGE(stage, Stage::PIXEL)
     {
         const HwShaderGenerator& shadergen = static_cast<const HwShaderGenerator&>(context.getShaderGenerator());
+        const Syntax& syntax = shadergen.getSyntax();
+
+        const string& vec3 = syntax.getTypeName(Type::VECTOR3);
+        const string& vec3_zero = syntax.getValue(Type::VECTOR3, HW::VEC3_ZERO);
```
niklasharrysson: Note that `Syntax::getValue` returns the string by value. Storing it by const ref is valid, but I think it's more consistent with coding convention to store it by const value instead. (same for all calls to getValue below)

niklasharrysson (approved):
This looks good to me (just added a comment on style above)

source/MaterialXGenHw/Nodes/HwLightNode.cpp
ld-kerley: Thanks - I think I just copy/pasted from the `getTypeName()` call. fixed that now.

PR #2626, opened 2025-10-15
Add output default value to unified noise nodes

ld-kerley:
As a result of working on the automatic documentation effort, I noticed that the unified noise nodes are missing default values for the output.


jstone-lucasfilm (approved):
Thanks for this fix, @ld-kerley!

PR #2631, opened 2025-10-16
Fix error adding default to unifiednoise nodes

ld-kerley:
I made a mistake in #2626 - where the default was erroneously added as a `value` and not a `default`.


jstone-lucasfilm (approved):
Good catch, @ld-kerley!

PR #2632, opened 2025-10-17
Initial developer guidelines

jstone-lucasfilm:
This changelist expands upon the original Coding Style and Unit Tests sections of CONTRIBUTING.md, adding an initial, more comprehensive set of Developer Guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -259,10 +259,10 @@ the root source folder (e.g. `#include <MaterialXCore/Element.h>`). This
 ensures consistent include paths across the entire codebase, regardless of the
 location of the referencing file.
 
-In implementation files (.cpp), the corresponding header file should be included
-first (e.g. the first include in `Element.cpp` should be `Element.h`).
-This ensures that the header file is self-contained and doesn't accidentally
-depend on includes from other headers.
+Each implementation file should include its corresponding header file first,
+so the first include in `Element.cpp` should be `Element.h`. This ensures that
+the header file is self-contained and doesn't accidentally depend on includes
+from other headers.
 
 After the corresponding header, include blocks should be ordered hierarchically,
 with high-level modules listed before low-level modules (e.g.


ld-kerley:
This is really great - and is gonna help new people join the project with less friction - I look forward to exploring areas where we can incrementally improve this over time - as we learn more lessons, but its great to have a place to start from.

CONTRIBUTING.md
```diff
@@ -189,7 +189,69 @@ in the ASWF Slack or in TSC meetings before any PR is submitted, in order to
...
+
+The following guidelines represent coding standards that we strive to follow
+in the MaterialX project. While not all existing code may adhere to these
+standards yet, we encourage all new contributions to follow these practices,
+and we welcome incremental improvements to bring existing code into alignment
+with these guidelines.
+
+#### Naming Conventions
+
+Member variables should use an underscore prefix, such as `_name` or `_parent`.
```
kwokcb: This is for private / protected members. Public members do not use an underscore prefix.

CONTRIBUTING.md
```diff
@@ -189,7 +189,69 @@ in the ASWF Slack or in TSC meetings before any PR is submitted, in order to
...
+Constants should be written in UPPER_CASE with underscores separating words,
+as in `EMPTY_STRING` or `CATEGORY`. Method names should use camelCase starting
```
kwokcb: Note sure how to say this generally but : method names, local variables, arguments, and member variables also should be camelCase. Maybe all identifiers should use camelCase with class names starting with uppercase ?

CONTRIBUTING.md
```diff
@@ -189,7 +189,69 @@ in the ASWF Slack or in TSC meetings before any PR is submitted, in order to
...
+with a lowercase letter, for example `getNode` or `setName`. Class names should
+use PascalCase, such as `NodeGraph` or `ShaderGenerator`. Type aliases should
+append appropriate suffixes to indicate their purpose, using `Ptr` for pointers,
```
kwokcb: Is it worth to mention that the `using` syntax is used for aliases ?

CONTRIBUTING.md
```diff
@@ -189,7 +189,69 @@ in the ASWF Slack or in TSC meetings before any PR is submitted, in order to
...
+`Vec` for vectors, `Map` for maps, and `Set` for sets.
+
+#### Static Constants and Class Organization
+
+Class members should be organized in order of decreasing visibility: public,
+protected, then private. Static constants should be placed at the end of their
+respective visibility section. String constants should be defined in implementation
+files rather than headers to avoid One Definition Rule violations. The
+`EMPTY_STRING` constant should be used instead of empty string literals (`""`)
+for clarity and consistency.
```
kwokcb, line 218: Is it worthwhile to state to use or add string constants for reusable strings vs hard-coded strings.
e.g. folks using the string "nodegraph" instead of `PortElement::NODE_GRAPH_ATTRIBUTE`.

CONTRIBUTING.md
```diff
@@ -189,7 +189,69 @@ in the ASWF Slack or in TSC meetings before any PR is submitted, in order to
...
+
+#### Smart Pointer Conventions
+
+Heap-allocated objects in the public API should always use `shared_ptr` for
+memory management. Type aliases should be defined for all shared pointers,
+following the pattern of `ElementPtr` for `shared_ptr<Element>`. Both mutable
+and const versions of these type aliases should be provided, e.g. `ElementPtr`
+and `ConstElementPtr`. Raw pointers should be avoided except when representing
+non-owning references within implementation details.
```
kwokcb, line 227: Is it worth noting that it's not required to create lists, sets, maps as shared pointers ?
Also, I've never been clear as to when we should add lists, sets, maps -- I guess as needed ?
(I know this doesn't cover Python but type aliases should be exposed in Python type wrappers as needed)

CONTRIBUTING.md
```diff
@@ -200,7 +262,36 @@ file to automatically align the code to MaterialX conventions. When modifying
 existing code, follow the surrounding formatting conventions so that new or
 modified code blends in with the current code.
 
-### Unit Tests
+#### Header Includes
+
+In implementation files (.cpp), the corresponding header file should always be
```
kwokcb: Is it worth mentioning includes are relative to the  `source` folder ? e.g. you never do #include "./foo/bar.h"

CONTRIBUTING.md
```diff
@@ -200,7 +262,36 @@ file to automatically align the code to MaterialX conventions. When modifying
...
+_transitive header includes_, where low-level headers that have already been
+included by a high-level header do not need to be restated individually.
+
+#### Documentation Standards
+
+All public API methods and classes should be documented with Doxygen comments.
+Documentation should include `@param`, `@return`, and `@throws` tags where
+applicable. Related methods should be grouped together using `/// @name GroupName`
+sections to improve readability. File-level documentation should be placed
+immediately after the copyright header using the `/// @file` directive.
```
kwokcb: Think `@class` should also be mentioned.

CONTRIBUTING.md
jstone-lucasfilm: Good point, and fixed!

CONTRIBUTING.md
jstone-lucasfilm: Agreed that this rule is broader than just class method names, though more thought will be needed to state this clearly and concisely.

CONTRIBUTING.md
jstone-lucasfilm: You're right that we strongly prefer the C++11 `using` syntax in MaterialX, though if we had any old-school `typedef` aliases in our codebase, they would use the same convention!

CONTRIBUTING.md
jstone-lucasfilm: That's a more subtle rule to capture clearly, but I think you're right that we should tackle this in the future.

CONTRIBUTING.md
jstone-lucasfilm: I was aiming to capture that distinction with the phrase "heap-allocated objects", as lists, sets, and maps are usually allocated on the stack.

CONTRIBUTING.md
jstone-lucasfilm: Good point, and fixed!

CONTRIBUTING.md
jstone-lucasfilm: Good point, and fixed!

PR #2633, opened 2025-10-17
Update `FilePath::getFilesInDirectory` to allow getting all files

ld-kerley:
While working on some other internal projects I found myself using the `FilePath` object, and bumped up against the fact that `getFilesInDirectory` is unable to query all the files in the provided directory.

This PR provides a simple extension to the existing code to allow accessing all the files in a directory if no extension is provided to match against.

NOTE: I am not a Windows developer, so I would greatly appreciate someone more familiar with the Windows API checking what I'm proposing here will work.

### DIFF
--- a/source/MaterialXFormat/File.h
+++ b/source/MaterialXFormat/File.h
@@ -189,7 +189,7 @@ class MX_FORMAT_API FilePath
 
     /// Return a vector of all files in the given directory with the given extension.
     /// If extension is empty all files are returned.
-    FilePathVec getFilesInDirectory(const string& extension="") const;
+    FilePathVec getFilesInDirectory(const string& extension = EMPTY_STRING) const;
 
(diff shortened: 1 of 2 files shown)


source/MaterialXFormat/File.h
```diff
@@ -188,7 +188,8 @@ class MX_FORMAT_API FilePath
     bool isDirectory() const;
 
     /// Return a vector of all files in the given directory with the given extension.
-    FilePathVec getFilesInDirectory(const string& extension) const;
+    /// If extension is empty all files are returned.
+    FilePathVec getFilesInDirectory(const string& extension="") const;
```
kwokcb: `extension = EMPTY_STRING` instead to follow coding guidelines :)

kwokcb (commented):
Maybe add a unit test for empty string arg? 
That's how I test for Linux since I don't have that.

PR #2637, opened 2025-10-20
Fix some obscure swizzle upgrade bugs

JGamache-autodesk:
Was working on upgrading a MaterialX material when it is written using UsdShade data model, and that required reviewing the current upgrade code to convert it to Python/USD. Also was working on a unit test, which forced me to put on my QA hat and start devising devious ways to fail the code.

This resulted in the following two fixes for extremely rare corner cases.


source/MaterialXCore/Version.cpp
```diff
@@ -1307,7 +1347,7 @@ void Document::upgradeVersion()
                             else
                             {
                                 combineInInput->setConnectedNode(separateNode);
-                                combineInInput->setOutputString(combineInInput->isColorType() ? "outr" : "outx");
+                                combineInInput->setOutputString(inInput->isColorType() ? "outr" : "outx");
```
JGamache-autodesk: The `combineInInput` is of type `float` and can not be used here to check for color. We still have the original `inInput` that gives us the right answer when queried.

source/MaterialXCore/Version.cpp
```diff
@@ -1187,7 +1187,47 @@ void Document::upgradeVersion()
                         }
                     }
 
-                    if (inInput->hasValue())
+                    // We convert to a constant node if "in" input is a constant value:
+                    bool convertToConstantNode = inInput->hasValue();
+                    // We also convert to a constant node if every destination
+                    // channel is constant:
+                    //   eg: "ND_swizzle_color3_color3" node with
+                    //       "010" in the "channels" input.
```
JGamache-autodesk, line 1194: Ridiculous case, but I don't see why it should not be handled.

JGamache-autodesk:
Hmmm, while writing test cases I found out that a freshly created swizzle node gets completely ignored. Not good. We need to convert that into a constant `0` node of the right type since that node has a default value for `in` of zero and for `channels` something like "always first channel".

JGamache-autodesk:
Tested against 
[upgrade_to_1_39_test_cases.zip](https://github.com/user-attachments/files/23028063/upgrade_to_1_39_test_cases.zip)
Which hits 100% coverage of the node section of the 1.38 to 1.39 update and resolves as expected.

Still one remaining issue of the nitpick type. The `{ "rgb", 4 }` and `{ "xyz", 4 }` in CHANNEL_CONVERT_PATTERNS fail in a very subtle way when the source type is `color3` or `vector3`. In these cases, the `swizzle` resolves the missing last channel to `r` or `x` while the convert resolves to `1`. I can create a scene where the upgrade will cause a look change, but I don't think we will ever see this case in the wild.

jstone-lucasfilm (approved):
Thanks for diving into these tricky edge cases, @JGamache-autodesk, and the changes look good to me!

PR #2639, opened 2025-10-22
Remove global constructor for Traversal constants

ld-kerley:
When we ship MaterialX in apple ecosystem we maintain a number of patches to avoid global constructors and destructors on exit. (`-Wglobal-constructors -Wexit-time-destructor`)

This PR proposed one of these patches, if this is merged, I'll propose some others moving forwards. This will help us lower the burden of ongoing maintenance for MaterialX integration in the apple ecosystem.

### DIFF
--- a/source/MaterialXCore/Element.cpp
+++ b/source/MaterialXCore/Element.cpp
@@ -509,7 +509,7 @@ GraphIterator Element::traverseGraph() const
 
 Edge Element::getUpstreamEdge(size_t) const
 {
-    return NULL_EDGE;
+    return getNullEdge();
 }
 
 ElementPtr Element::getUpstreamElement(size_t index) const
--- a/source/MaterialXCore/Traversal.cpp
+++ b/source/MaterialXCore/Traversal.cpp
@@ -9,11 +9,23 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
-const Edge NULL_EDGE(nullptr, nullptr, nullptr);
+const Edge& getNullEdge() {
+    static const auto ret = new Edge(nullptr, nullptr, nullptr);
+    return *ret;
+}
 
-const TreeIterator NULL_TREE_ITERATOR(nullptr);
-const GraphIterator NULL_GRAPH_ITERATOR(nullptr);
-const InheritanceIterator NULL_INHERITANCE_ITERATOR(nullptr);
+const TreeIterator& getNullTreeIterator() {
+    static const auto ret = new TreeIterator(nullptr);
+    return *ret;
+}
+const GraphIterator& getNullGraphIterator() {
+    static const auto ret = new GraphIterator(nullptr);
+    return *ret;
+}
+const InheritanceIterator& getNullInheritanceIterator() {
+    static const auto ret = new InheritanceIterator(nullptr);
+    return *ret;
+}
 
 //
 // Edge methods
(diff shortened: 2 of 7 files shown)


source/MaterialXCore/Traversal.cpp
```diff
@@ -9,19 +9,31 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
-const Edge NULL_EDGE(nullptr, nullptr, nullptr);
+const Edge& getNullEdge() {
```
jstone-lucasfilm: Just adding a reminder to run `clang-format` on your new code, so that it inherits the Allman braces style of the project.

ld-kerley: Thanks for the reminder - if only we were running clang-format in the CI... :) Then I wouldn't forget.

source/MaterialXCore/Element.cpp
```diff
@@ -713,18 +719,17 @@ const string& ValueElement::getActiveUnit() const
...
     // Perform value comparisons
     bool performedValueComparison = false;
     if (options.performValueComparisons)
     {
-        const StringSet uiAttributes = 
-        {
-           ValueElement::UI_MIN_ATTRIBUTE, ValueElement::UI_MAX_ATTRIBUTE,
-           ValueElement::UI_SOFT_MIN_ATTRIBUTE, ValueElement::UI_SOFT_MAX_ATTRIBUTE,
-           ValueElement::UI_STEP_ATTRIBUTE
+        const StringSet uiAttributes = {
```
jstone-lucasfilm: Not related to your PR, but does anyone know how to get `clang-format` to maintain Allman braces consistently in data initializers as well as functions?

The MaterialX code style expects Allman braces _everywhere_, but `clang-format` seems to treat data initializers as a special case.

jstone-lucasfilm:
This looks good to me, @ld-kerley, though I'd recommend reverting the style changes that don't relate to your work on Traversal constants, so that we're not interleaving functional and style changes in a single PR.

I'm still curious as to whether `clang-format` can be instructed to use Allman braces _universally_, matching the existing style of the MaterialX codebase, rather than using one convention for functions and another convention for data initializers.  But we don't need to solve that in your PR, and it's a project that we can follow up on another day.

ld-kerley:
Thats pretty frustrating... Being asked to run `clang-format` and then revert those changes.

I think we really need to improve that aspect of the developer experience.  It would save a lot of time if we could just decide on a `clang-format` configuration, and then apply it universally - even if that led to a change in formatting conventions, it would be applied across the entire project, and would save many hours of time, in both manual fixes, and also time spent reviewing code for formatting.

ld-kerley:
I reverted the clang-format changes.

ld-kerley:
Thanks for merging this - I'll likely have some more changes in this vein to help us manage future MaterialX upgrades internally.

Full `clang-format` automation would be a huge win IMHO. Every project I've worked on that has taken that leap has seen a significant improvement in velocity, code cleanliness and also interestingly quality of code review.  Once you don't have to think about formatting anymore your brain is able to focus on the meaningful pieces.

I was poking around the internet and couldn't find a formal definition of Allman braces written down anywhere.  Is it perhaps possible that `clang-format` is correct and we're misinterpreting in the MaterialX source code? 

Either way, I really don't have any strong preference as to what formatting style is used, if it's enforced by the tools and I don't have to think about it. I would propose that its beneficial to the community to morph the MaterialX coding standards to meet `clang-format` if we can't find a way to get `clang-format` to follow our coding standards.  To me the productivity upside outweighs the historical formatting choices - but thats just a personal opinion.

Issue #2570, opened 2025-09-18
[Wgsl] Validation errors due to use of `boolean` uniforms


ashwinbhat:
The glsl generated by WebGPU ShaderGen has some validation errors due to the use of `bool` uniforms. 

Steps to repro:

### Generate wgsl (glsl) shader
python generateshader.py resources\Materials\TestSuite\libraries\metal\brass_wire_mesh.mtlx --target wgsl 
---------- Generate code for file:  TestSuite\libraries\metal\brass_wire_mesh.mtlx --------------------
- Set up CMS ...
- Set up Units ...
- Shader output path: TestSuite\libraries\metal
-- Generate code for element: Brass_Wire_Mesh
--- Wrote pixel shader to: TestSuite\libraries\metal/Brass_Wire_Mesh.wgsl.frag
--- Wrote vertex shader to: TestSuite\libraries\metal/Brass_Wire_Mesh.wgsl.vert
--- Validation passed for element: Brass_Wire_Mesh

### Validate and convert to `.wgsl` using `naga`
naga Brass_Wire_Mesh.frag.glsl wgsl\Brass_Wire_Mesh.frag.wgsl
error: Global variable [0] '' is invalid
   ┌─ Brass_Wire_Mesh.frag.glsl:15:36
   │
15 │ layout (std140, binding=1) uniform PrivateUniforms_pixel
   │                                    ^^^^^^^^^^^^^^^^^^^^^ naga::ir::GlobalVariable [0]
   │
   = Alignment requirements for address space Uniform are not met by [9]
   = The type is not host-shareable

Generating wgsl output requires validation to succeed, and it failed in a previous step

By temporarily replacing the `bool` uniforms with `int` the conversion is successful. See attached glsl and final wgsl output

[Brass_Wire_Mesh.frag-updated.glsl.txt](https://github.com/user-attachments/files/22396491/Brass_Wire_Mesh.frag-updated.glsl.txt)
[Brass_Wire_Mesh.wgsl.frag-orig.glsl.txt](https://github.com/user-attachments/files/22396490/Brass_Wire_Mesh.wgsl.frag-orig.glsl.txt)

Output:
[Brass_Wire_Mesh.frag.wgsl.txt](https://github.com/user-attachments/files/22396549/Brass_Wire_Mesh.frag.wgsl.txt)

crystalqdm:
Hi - hope I'm not too late! I'd like to pick this issue up for dev days 🙂

[...]

PR #2640, opened 2025-10-22
Handle boolean type uniforms in WGSL

crystalqdm:
Issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2570 

Description:

WGSL does not support uniform bool types, so we must declare them as int instead. However, we cannot blanket change every bool into int, as WGSL also does not support implicitly converting non-boolean types (like integers or objects) to booleans in a conditional context; WGSL is strictly typed and requires an explicit bool type for if conditions/ ternary operations.
eg. 
```
bool myBool = True
if ( myBool ) // Complies

int myBool = 1
if ( myBool ) // Does not compile
```
As a solution to get around this, this PR defines a macro immediately after the uniform variable is declared which turns any usage of the uniform to be wrapped in a bool cast. This should work in all cases, since you can't re-define a uniform variable anyways.

Testing:

`python python/Scripts/generateshader.py resources/Materials/Examples/StandardSurface/standard_surface_brass_tiled.mtlx --target wgsl`
`python python/Scripts/generateshader.py resources/Materials/TestSuite/libraries/metal/brass_wire_mesh.mtlx --target wgsl`

You can use naga-cli to verify the results:
naga Brass_Wire_Mesh.frag.glsl wgsl\Tiled_Brass.frag.wgsl
naga Brass_Wire_Mesh.frag.glsl wgsl\Brass_Wire_Mesh.frag.wgsl

Example outputs:

[Tiled_Brass.wgsl.frag.txt](https://github.com/user-attachments/files/23046834/Tiled_Brass.wgsl.frag.txt)
[Brass_Wire_Mesh.wgsl.frag.txt](https://github.com/user-attachments/files/23046832/Brass_Wire_Mesh.wgsl.frag.txt)


jstone-lucasfilm:
Thanks so much for this Dev Days PR, @crystalqdm!

I'm CC'ing @ashwinbhat as a potential reviewer, though I believe he's currently out of the office, and hopefully the Linux Foundation support links above can help you get set up with your CLA.

kwokcb:
Could we install `naga` as part of CI and use it to validate the output via `generateShader.py` ?  Should be a short install for Linux I believe. It would be good to test with `tint` as well but AFAIK this has to be built + has a long build time.

ld-kerley:
We should start looking using the ASWF CI containers - even if only for linux.  The CI group are happy to help deliver in to the container anything we need, so we could get `tint` and `naga` pre-installed for this.  

I think we need to prioritize these containers as a way to improve our general CI coverage.

kwokcb:
I did a local patch to the windows runner here:

https://github.com/AcademySoftwareFoundation/MaterialX/compare/main...kwokcb:MaterialX:shadergen_validation?expand=1.

The `generateshader` is issues and actually does not fail properly nor handle non-ansi characters coming back from validators. I fixed this in the patch.

This is the run that I run this patch with that fails currently:
https://github.com/kwokcb/MaterialX/actions/runs/18854067962/job/53797421292

---
This should be put into another runner to spread out the time required as the current runner also build `glslangValidator` from vcpkg.  

@ld-kerley, 
- I'm not sure about priorities but it would be good to get some kind of proper validation on WebGPU (WGSL) as it's producing invalid code currently (There are other issues that @scotbrew , @ashwinbhat and I have found) so we should at least clarify messaging on this.
- However time can be amortized getting code gen validation seems to be a smaller reasonable step to get running with containers even if we cannot get to GPU runners yet. 
`tint` would be a good candidate since it's easy to build but takes a while.

My 2 cents.

jstone-lucasfilm:
This looks great to me, @crystalqdm, and I believe we just need you to resolve the remaining conflicts with the `main` branch before we can merge this PR.

crystalqdm:
@jstone-lucasfilm Merge conflicts have been resolved and branch is ready 🙂

PR #2641, opened 2025-10-22
glTF PBR Clearcoat interaction with Emission

### LINKED ISSUE #2592: Emission and Clearcoat interaction in gltf_pbr
I was looking at the gltf_pbr shader implementation of emission and it seems to be unaffected by clearcoat. The resulting emission color is plugged into a `uniform_edf` and then into the [surface node](https://github.com/AcademySoftwareFoundation/MaterialX/blob/b5501a57af67aaa4b02b9eca68459fc05761cc4a/documents/Specification/MaterialX.PBRSpec.md#pbr-shader-nodes).

Image

Based on the spec of the [KHR_materials_clearcoat](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md#emission), the clearcoat should preferably be layered on top of the emission. According to the spec it should be darkened by the Fresnel term of the clearcoat. 

```
coated_emission = emission * (1 - clearcoat * clearcoat_fresnel)
```
where **clearcoat_fresnel** is **Schlick Fresnel** using the clearcoat normal and f0 as `0.04`

bowald:
closes #2592 

This PR will darken the emission color by a weighted clearcoat fresnel factor from [KHR_materials_clearcoat](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md#emission)

```
coated_emission = emission * (1 - clearcoat * clearcoat_fresnel)
```

Screenshots of a red emissive material with a clearcoat. 
<table width="100%">
  <thead>
    <tr>
      <th width="33.33%">
        MaterialXGraphEditor_djwiexXuWH
        </th>
      <th width="33.33%">MaterialXGraphEditor_lBI0QFpkee</th>
      <th width="33.33%">chrome_XpDlrBBniD</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td width="33.33%">Before PR: No clearcoat fresnel interaction</td>
      <td width="33.33%">After PR: Emission darken by clearcoat_fresnel</td>
      <td width="33.33%">Reference rendered using<a href="https://github.khronos.org/glTF-Sample-Viewer-Release/"> gltf_sample_viewer </a> </td>
    </tr>
  </tbody>
</table>

Material
```xml
  <gltf_pbr name="gltf_pbr_surfaceshader" type="surfaceshader">
    <input name="emissive" type="color3" value="1, 0, 0" />
    <input name="base_color" type="color3" value="0, 0, 0" />
    <input name="clearcoat" type="float" value="1" />
    <input name="clearcoat_roughness" type="float" value="0" />
  </gltf_pbr>
  ```
  
##  Discussion:

I ended up implementing Schlick Fresnel returning floats using nodes. I was planning to use the generalized_schlick node, however they returned distribution functions, is there an option to use the included schlick nodes or is how I have dont it in the PR a good way?

### CODE REVIEW

File: libraries/bxdf/genglsl/gltf_pbr.mtlx [hunk 1]
```diff
@@ -259,15 +259,73 @@
...
     </multiply>
 
+    <multiply name="coated_emission" type="color3">
+      <input name="in2" type="float" nodename="clearcoat_emission_scaler" />
+      <input name="in1" type="color3" nodename="emission_color" />
+    </multiply>
+
     <uniform_edf name="emission" type="EDF">
-      <input name="color" type="color3" nodename="emission_color" />
+      <input name="color" type="color3" nodename="coated_emission" />
```
jstone-lucasfilm, line 328: This looks like a thoughtful implementation of the Fresnel effect of the coating, though I wonder if we could leverage the standard `generalized_schlick_edf` node instead:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.PBRSpec.md#node-generalized-schlick-edf

bowald: I agree, that was my first idea, and use `multiply_edfC` with the `emission_color` or pipe the `uniform_edf` into the base-input.

But there are some implementation details of `generalize_schlick` that I think differ from the gltf_spec formula, as well as not allowing me to supply the normal (as it needs to be the clearcoat normal).

This is the fresnel described by the glTF spec, [KHR_materials_clearcoat](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md#layering).
```
clearcoat_fresnel = 0.04 + (1 - 0.04) * (1 - abs(VdotNc))^5
```

In the mx_generalized_schlick_edf glsl implementation it seems that [NdoV is clamped](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/mx_generalized_schlick_edf.glsl#L9) and not the absolute value.

also the [mx_fresnel_schlick](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet.glsl#L43-L48) function in glsl uses `mix` with the arguments in another order then expected for clearcoat interaction.
```glsl
float mx_fresnel_schlick(float cosTheta, float F0, float F90, float exponent)
{
    float x = clamp(1.0 - cosTheta, 0.0, 1.0);
    return mix(F0, F90, pow(x, exponent));
}
```
According to [GLSLangSpec.4.40.pdf, page 146](https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.40.pdf), this is the definition of mix :
```
mix(x,y,a) = x⋅(1−a)+ y⋅a
```
to be equalent to glTF implementation the mix should be called with 
```
mix(pow(x, exponent), F90, F0) = 0.04 + (1 - 0.04) * (1 - abs(VdotNc))^5
```

I should say that I havn't tested the code (because I couldn't figure out how to supply the clearcoat normal to the edf func).

bowald: Regarding the use of `clamp` vs `abs` as in the spec. It seems like the gltf-sample-renderer uses `clamp` here as well and not `abs` as in the spec. [Code reference](https://github.com/KhronosGroup/glTF-Sample-Renderer/blob/44cbb275f9fa075e41596ba6577d7d55fc8967b0/source/Renderer/shaders/pbr.frag#L157). So that may not be an issue.

bowald: Also realized that there is a `Standard Schlick Fresnel` in the [mx_microfacet ](https://github.com/AcademySoftwareFoundation/MaterialX/blob/b91ec7ae4d32a4567f905608dfd83e64969af43e/libraries/pbrlib/genglsl/lib/mx_microfacet.glsl#L15-L21) lib. That is what I'm after, it will sort out the "mix"-problem I wrote about in an earlier reply. But I cannot see a way to access it from `generalized_schlick_edf`, is that possible?

jstone-lucasfilm: All good points, @bowald, and I'm CC'ing @niklasharrysson for his expertise with the definition and implementation of `generalized_schlick_edf`.

Our original goal for the `generalized_schlick_edf` node was to handle the very situation you've encountered here, where the shape of an EDF lobe is influenced by the clearcoat layer above it, so if there are limitations in our current definition I'm hoping we can address them to encompass the needs of `gltf_pbr` as well.

As a second reference point, here's where we use `generalized_schlick_edf` in the definition of `standard_surface` to handle a similar situation:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/standard_surface.mtlx#L399

niklasharrysson: Hej Johan!

Sorry for the late reply.

An explicit implementation of the Fresnel effect using the `viewdirection` is certainly possible, as shown in your PR. However, there are a few important considerations to keep in mind:

**Portability:**
Relying on the view direction explicitly within a graph can introduce portability issues for certain targets. In particular, for bidirectional rendering techniques, using the camera-based view direction becomes problematic, since it may be undefined when tracing paths from a light source (or EDF). This is why the `viewdirection` node lives in the NPR library rather than the PBR library. Ideally, it should only be used for effects that are strictly camera-dependent, such as facing-ratio adjustments or other NPR-style shading.

**Performance:**
Distribution functions have known analytic shapes that renderers can importance-sample efficiently. When reproducing the same behavior through an explicit graph, this advantage is lost, which can make the result significantly less efficient on some targets.

In general, we should prefer distribution functions whenever possible. In this particular case, I believe the main limitation is that we don’t currently support using a custom normal to steer the distribution, the coat normal in this case. However, since the underlying EDF is diffuse, I suspect this has minimal impact on the visible result.

The proposed setup with `generalized_schlick_edf` is already used for both the Standard Surface and OpenPBR graphs in MaterialX, and we haven’t received any reports of issues related to not accounting for the coat normal in the emission. With that in mind, it might make sense to hold off on further changes to `generalized_schlick_edf` until there is a demonstrated need.

bowald: Hej @niklasharrysson :)

No worries, there is no rush to get this into MaterialX from our part.

I agree that portability and performance is preferred over staying close to the glTF spec.
We should not go forward merging this PR as is.

There was two limitations I couldn't fix using the `generalized_schlick_edf` to implement clearcoat interaction.
1. Not being able to supply the clearcoat normal to `generalized_schlick_edf` (as you stated)
2. Not being able to call [Standard Schlick Fresnel](https://github.com/AcademySoftwareFoundation/MaterialX/blob/b91ec7ae4d32a4567f905608dfd83e64969af43e/libraries/pbrlib/genglsl/lib/mx_microfacet.glsl#L15-L21a) (Without the F90) from the `generalized_schlick_edf` to get the correct mix-function for the clearcoat interaction.

I think having the limitation to not account for the clearcoat normalmap, and use the surface normalmap, would be a good middle way.

Do you know if there is a way to call the [Standard Schlick Fresnel](https://github.com/AcademySoftwareFoundation/MaterialX/blob/b91ec7ae4d32a4567f905608dfd83e64969af43e/libraries/pbrlib/genglsl/lib/mx_microfacet.glsl#L15-L21a) (Without the F90) from the `generalized_schlick_edf` or if there is another pbr node that could be used?

Where we get this instead: 
`clearcoat_fresnel = 0.04 + (1 - 0.04) * (1 - abs(VdotN))^5`
(Using, N = surface normal, skipping the N__clearcoat_ from the [glTF clearcoat extension spec](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_clearcoat/README.md#layering))

niklasharrysson: Just set `F90` to 1.0 (white), and `exponent` to 5. The Generalized Schlick Fresnel is then reduced to the Standard Schlick Fresnel.

The generalized equation:

```
float x = clamp(1.0 - cosTheta, 0.0, 1.0);
fresnel = mix(F0, F90, x^exponent);
```
is reduce to:

```
fresnel = mix(F0, 1.0, x^5));

fresnel = (1 - x^5) * F0 + x^5 * 1.0

fresnel = F0 - x^5 * F0 + x^5 * 1.0

fresnel = F0 + (1.0 - F0) * x^5
```
You can then hardcode F0 to 0.04 if that is desired (this corresponds to an IOR of 1.5, is that always fixed in glTF's coat layer?)

bowald: You are right. I screwed up the math. The second limitation I listed is not an issue.

> this corresponds to an IOR of 1.5, is that always fixed in glTF's coat layer?

Yes, currently IOR of clearcoat is always 1.5, there is a proposal to the glTF spec allowing IOR to be supplied to the coat layer https://github.com/KhronosGroup/glTF/pull/2530 but its not yet part of the glTF spec.

I can update the PR with the `generalized_schlick_edf`, I think the limitation with surface normal is better then no interaction.

niklasharrysson: Excellent! :)

PR #2645, opened 2025-10-25
XML validation

Tellusim:
& symbol cannot be used in XML.

### DIFF
--- a/libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
+++ b/libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
@@ -732,7 +732,7 @@
   <!-- Logical operator nodes                                                   -->
   <!-- ======================================================================== -->
 
-  <implementation name="IM_logical_and_genglsl" nodedef="ND_logical_and" target="genglsl" sourcecode="{{in1}} && {{in2}}"/>
+  <implementation name="IM_logical_and_genglsl" nodedef="ND_logical_and" target="genglsl" sourcecode="{{in1}} &amp;&amp; {{in2}}"/>
   <implementation name="IM_logical_or_genglsl" nodedef="ND_logical_or" target="genglsl" sourcecode="{{in1}} || {{in2}}"/>
   <implementation name="IM_logical_not_genglsl" nodedef="ND_logical_not" target="genglsl" sourcecode="!{{in}}"/>
 
--- a/libraries/stdlib/genmdl/stdlib_genmdl_impl.mtlx
+++ b/libraries/stdlib/genmdl/stdlib_genmdl_impl.mtlx
@@ -744,7 +744,7 @@
   <!-- Logical operator nodes                                                   -->
   <!-- ======================================================================== -->
 
-  <implementation name="IM_logical_and_genmdl" nodedef="ND_logical_and" target="genmdl" sourcecode="{{in1}} && {{in2}}"/>
+  <implementation name="IM_logical_and_genmdl" nodedef="ND_logical_and" target="genmdl" sourcecode="{{in1}} &amp;&amp; {{in2}}"/>
   <implementation name="IM_logical_or_genmdl" nodedef="ND_logical_or" target="genmdl" sourcecode="{{in1}} || {{in2}}"/>
   <implementation name="IM_logical_not_genmdl" nodedef="ND_logical_not" target="genmdl" sourcecode="!{{in}}"/>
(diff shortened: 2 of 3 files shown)


kwokcb:
These files should be loaded every time the standard libraries are loaded.
Would you have any idea why the XML parser (pugi) does not catch this as an error already ?

It would be good to prevent these kind of errors from ever being introduced at all.
 
 I see this in `Xmlio.cpp` which I assume should throw an error.
 ```c++
}

void validateParseResult(const xml_parse_result& result, const FilePath& filename = FilePath())
{
    if (result)
    {
        return;
    }

    if (result.status == xml_parse_status::status_file_not_found ||
        result.status == xml_parse_status::status_io_error ||
        result.status == xml_parse_status::status_out_of_memory)
    {
        throw ExceptionFileMissing("Failed to open file for reading: " + filename.asString());
    }

    string desc = result.description();
    string offset = std::to_string(result.offset);
    string message = "XML parse error";
    if (!filename.isEmpty())
    {
        message += " in " + filename.asString();
    }
    message += " (" + desc + " at character " + offset + ")";

    throw ExceptionParseError(message);
} 
``

ld-kerley:
Like @bernardkwok I'm surprised this can sneak thru - unless our internal copy of pugixml has a specific carve out to allow this - but I wasn't aware it did.

For this specific fix - personally I find the `&amp;&amp;` less friendly to my shader writer mind, and I'd rather see us move this inlined `sourcecode` attribute to an external source code file, where we can retain the original syntax.

Tellusim:
We are using our own material generation code based on MTLX and GenGLSL definitions from the libraries. Our parser failed on the " && " sequence during the latest MaterialX.git merge. It might make sense to run xmllint or another XML validator to catch such issues earlier.

kwokcb:
I dug around a bit and this is what seems to be the situation:

1. There is an option to disallow unescaped characters.
2. For MaterialX parser this appears to be off by default. Thus allowing characters like `&`.
3. When turned off it will **read** unescaped characters but on **write** will escape them. i.e. it does not preserve breaking XML.
4. MaterialX has a modified version of pugi which also allowed `<` explicitly to support UDIM syntax. This will always break XML validation even on write.

Suggested Implementation change:
1. Add the escape option in. Leave it off by default to preserve current behaviour.
2. Turn on by default for next desired release for read. Write is not affected since it always escapes.
3. [Optional] Update to current version pugi. I don't think it's need so an option.

Suggested process sanity check change:
1. Run all files through read / write. We already run `upgrade` on all files. I'm not sure if we missed these files by accident.

I have no idea what is desired with `UDIMs`. This feels "hacky" though it does make sense from a user persective.

Alternatives:
- Switch to use CDATA.

kwokcb:
BTW. Here's the test I did:

```python
import MaterialX as mx

test_string = '<?xml version="1.0"?><materialx version="1.39"><implementation name="IM_logical_and_genglsl" nodedef="ND_logical_and" target="genglsl" sourcecode="{{in1}} && < {{in2}}"/></materialx>'

doc = mx.createDocument()
mx.readFromXmlString(doc, test_string)
impl = doc.getImplementation("IM_logical_and_genglsl")
print('Source code: :', impl.getAttribute("sourcecode"))
out_string = mx.writeToXmlString(doc)
print('Output XML: :', out_string)
```

Result:
```shell
Source code: : {{in1}} && < {{in2}}
Output XML: : <?xml version="1.0"?>
<materialx version="1.39">
  <implementation name="IM_logical_and_genglsl" nodedef="ND_logical_and" target="genglsl" sourcecode="{{in1}} &amp;&amp; < {{in2}}" />
</materialx>
```

kwokcb:
I've put in a fix to `mxformat.py` which "fixes" unescaped characters when write occurs. See this [PR](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2647) which is run against `libraries`.

A separate [issue](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2648) has been logged to fix PugiXML options usage.

jstone-lucasfilm:
For this specific fix, I like @ld-kerley's suggestion to move the logical-and code into external shader source files, where the `&` character is syntactically valid.

jstone-lucasfilm:
Thanks for tracking this issue down, @Tellusim, and this fix has been included by @kwokcb in his broader set of fixes for https://github.com/AcademySoftwareFoundation/MaterialX/pull/2647.

PR #2647, opened 2025-10-28
Fixes for MaterialX format checker

kwokcb:
## Changes

The utility  `mx_format.py` was not traversing to sub-folders so when applied to `libraries` it would miss some files.
This fixes some files which had invalid XML (unescaped characters) as `PugiXML` will "fix" this on write.

Two additional options have been added:
- Check for XML syntax validity
- MaterialX validation checking
Both are off by default.

## Results

- Ran this again `libraries` folder. Changes are mostly cosmetic except for the mentioned fixes to missed files.
- Ran this against `resources` folder. Found a file which had duplicate elements. Fixed manually.


resources/Materials/TestSuite/stdlib/conditional/conditional_logic.mtlx
```diff
@@ -24,7 +24,7 @@
     </xor>
     <output name="out" type="boolean" nodename="xor" />
   </nodegraph>
-  <nodegraph name="and_boolean">
+  <nodegraph name="not_boolean">
```
kwokcb, line 27: and_boolean was declared twice so this graph was never test since it would not load in.

python/Scripts/mxformat.py
```diff
@@ -6,33 +6,43 @@
 
 import argparse
 import os
+import xml.etree.ElementTree as ET
 
 import MaterialX as mx
 
+def is_well_formed(xml_string):
```
kwokcb, line 13: Allows for a basic "lint" test.

python/Scripts/mxformat.py
```diff
@@ -6,33 +6,43 @@
...
+    parser.add_argument('-x', '--xml_syntax', dest='xml_syntax', action="store_true", help='Check XML syntax after reformatting.')
     parser.add_argument(dest="inputFolder", help="An input folder to scan for MaterialX documents.")
     opts = parser.parse_args()
 
     validDocs = dict()
     for root, dirs, files in os.walk(opts.inputFolder):
-        for file in files:
-            if file.endswith('.mtlx'):
-                filename = os.path.join(root, file)
+        for filename in files:
```
kwokcb, line 31: This logic was not checking subfolder > 1 depth.

libraries/bxdf/open_pbr_surface.mtlx
```diff
@@ -3,90 +3,48 @@
...
-    <input name="base_color" type="color3" value="0.8, 0.8, 0.8" uimin="0,0,0" uimax="1,1,1" uiname="Base Color" uifolder="Base"
-           doc="Color of the reflection from the diffuse and metallic base." />
-    <input name="base_diffuse_roughness" type="float" value="0.0" uimin="0.0" uimax="1.0" uiname="Base Diffuse Roughness" uifolder="Base" uiadvanced="true"
-           doc="Roughness of the diffuse reflection. Higher values cause the surface to appear flatter." />
-    <input name="base_metalness" type="float" value="0.0" uimin="0.0" uimax="1.0" uiname="Base Metalness" uifolder="Base"
-           doc="Specifies how metallic the base material appears (dials the base from pure dielectric to pure metal)." />
-    <input name="specular_weight" type="float" value="1.0" uimin="0.0" uisoftmax="1.0" uiname="Specular Weight" uifolder="Specular"
-           doc="Multiplies the specular reflectivity." />
-    <input name="specular_color" type="color3" value="1, 1, 1" uimin="0,0,0" uimax="1,1,1" uiname="Specular Color" uifolder="Specular"
-           doc="Color of the specular reflection (controls the physical edge-tint for metals, and a non-physical overall tint for dielectrics)." />
```
jstone-lucasfilm, line 19: Would it be visually clearer to maintain our existing layout for shading model inputs, where the `doc` string is placed on its own line?

To my eye, the original formatting with the `doc` string on its own line seems clearer and more understandable, though it's true that our MaterialX format checker doesn't yet handle this automatically.

libraries/bxdf/lama/lama_iridescence.mtlx
```diff
@@ -72,7 +64,8 @@
     <rotate3d name="tangent_rotate" type="vector3">
       <input name="in" type="vector3" interfacename="anisotropyDirection" />
       <input name="amount" type="float" nodename="tangent_rotate_degree_offset" />
-      <input name="axis" type="vector3" nodename="tangent_normal" /> />
+      <input name="axis" type="vector3" nodename="tangent_normal" />
+      <:anonymous name="1" />
```
ld-kerley: This looks like a typo in the original source? 

I'm a little surprised this passes validation - are we confident that the validator is processing all of the files too?

libraries/bxdf/open_pbr_surface.mtlx
ld-kerley: Personally I don't find it any clearer with the new line break - as to me it hints that it is only the "doc" attribute on that line, but we can see in the example above that "uiName" comes after, which my eyes fail to parse because of the new line break.

Ultimately I don't have a super strong opinion here, other than I think the formatter tools rules should be the ultimate truth. If we want to enforce any formatting convention the it should be implemented in the formatting tool. I feel that any human post-editing of the files is potentially error prone, and burdensome. 

If we are going to enforce "doc" on a new line - are we applying that everywhere?  does it make sense everywhere?  If we do can we also ensure it is the last attribute in the element?

I would cast a soft vote against making a new line, only because I worry there will be other corner cases where we don't want the additional lines. 

For instance here - https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/stdlib_defs.mtlx#L109

do we feel it would be beneficial to introduce new lines for each "doc" attribute, including the one in the "nodedef" element?

libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
```diff
@@ -732,9 +732,9 @@
   <!-- Logical operator nodes                                                   -->
   <!-- ======================================================================== -->
 
-  <implementation name="IM_logical_and_genglsl" nodedef="ND_logical_and" target="genglsl" sourcecode="{{in1}} && {{in2}}"/>
-  <implementation name="IM_logical_or_genglsl" nodedef="ND_logical_or" target="genglsl" sourcecode="{{in1}} || {{in2}}"/>
-  <implementation name="IM_logical_not_genglsl" nodedef="ND_logical_not" target="genglsl" sourcecode="!{{in}}"/>
+  <implementation name="IM_logical_and_genglsl" nodedef="ND_logical_and" target="genglsl" sourcecode="{{in1}} &amp;&amp; {{in2}}" />
```
ld-kerley, line 735: Note  - unrelated to this PR - I strongly dislike this - and would like to propose (in a separate PR) we refactor this to an external source file.

libraries/stdlib/genmdl/stdlib_genmdl_impl.mtlx
```diff
@@ -266,8 +266,10 @@
   <implementation name="IM_divide_vector3FA_genmdl" nodedef="ND_divide_vector3FA" sourcecode="{{in1}} / {{in2}}" target="genmdl" />
   <implementation name="IM_divide_vector4_genmdl" nodedef="ND_divide_vector4" sourcecode="{{in1}} / {{in2}}" target="genmdl" />
   <implementation name="IM_divide_vector4FA_genmdl" nodedef="ND_divide_vector4FA" sourcecode="{{in1}} / {{in2}}" target="genmdl" />
-  <implementation name="IM_divide_matrix33_genmdl" nodedef="ND_divide_matrix33" sourcecode="{{in1}}" target="genmdl" /> <!-- TODO: Implement properly -->
-  <implementation name="IM_divide_matrix44_genmdl" nodedef="ND_divide_matrix44" sourcecode="{{in1}}" target="genmdl" /> <!-- TODO: Implement properly -->
+  <implementation name="IM_divide_matrix33_genmdl" nodedef="ND_divide_matrix33" sourcecode="{{in1}}" target="genmdl" />
```
ld-kerley: Might be good to move these comments above the lines they apply to - by hand.

libraries/stdlib/stdlib_defs.mtlx
```diff
@@ -1583,7 +1583,8 @@
     The current time in seconds as defined by the host environment.
   -->
   <nodedef name="ND_time_float" node="time" nodegroup="application">
-    <input name="fps" type="float" value="24.0" /> <!-- An unused input, to be removed in a future specification version -->
+    <input name="fps" type="float" value="24.0" />
+    <!-- An unused input, to be removed in a future specification version -->
```
ld-kerley, line 1587: same here - comment on its own line is a little unclear which input its referring too... or even better we should be more explicit in the comment text so its cannot be ambiguous.

ld-kerley (commented):
This looks like a great change to me - anything we can do to improve our ability to keep formatting consistent across the project gets my vote.

Left a couple of small things that probably need to be fixed up locally below.

libraries/stdlib/genmdl/stdlib_genmdl_impl.mtlx
kwokcb: Sure thing.

libraries/stdlib/genglsl/stdlib_genglsl_impl.mtlx
kwokcb: This would avoid this remapping but I believe the purpose of inlining was to reduce the previous large number of file reads.

libraries/stdlib/stdlib_defs.mtlx
kwokcb: I really dislike inline these comments since they are non-standard XML, but can try to patch up the original.

libraries/bxdf/lama/lama_iridescence.mtlx
kwokcb: Hmm. The std library generally get's loaded into docs and this get's validated. Strange it's never been caught. Can patch the original and log a new parsing issue.

libraries/bxdf/open_pbr_surface.mtlx
kwokcb: Even though I believe one aim is make original XML as user readable for things like comments and doc strings this makes it so that non-standard "extra" XML rules keep on having to be introduced. Additionally this promotes hand-editing XML which from this PR can introduce unnoticed user errors.

My vote is thus to not introduce more special case handling which no other XML parser will ever handle. If we want formatted document strings then this should be introduced properly and no rely on newlines embedded into a text file.

kwokcb:
As a general comment, I will revert all formatting changes which are under discussion and just keep the obvious XML error fixes. This should make the PR self-contained to just enhancing the `mxformat` utility.

libraries/stdlib/genmdl/stdlib_genmdl_impl.mtlx
```diff
@@ -266,8 +266,10 @@
   <implementation name="IM_divide_vector3FA_genmdl" nodedef="ND_divide_vector3FA" sourcecode="{{in1}} / {{in2}}" target="genmdl" />
   <implementation name="IM_divide_vector4_genmdl" nodedef="ND_divide_vector4" sourcecode="{{in1}} / {{in2}}" target="genmdl" />
   <implementation name="IM_divide_vector4FA_genmdl" nodedef="ND_divide_vector4FA" sourcecode="{{in1}} / {{in2}}" target="genmdl" />
-  <implementation name="IM_divide_matrix33_genmdl" nodedef="ND_divide_matrix33" sourcecode="{{in1}}" target="genmdl" /> <!-- TODO: Implement properly -->
-  <implementation name="IM_divide_matrix44_genmdl" nodedef="ND_divide_matrix44" sourcecode="{{in1}}" target="genmdl" /> <!-- TODO: Implement properly -->
+  <!-- TODO: Implement IM_divide_matrix33_genmdl properly -->
```
kwokcb, line 269: I actually think this is very bad practice, especially if it's not documented anywhere. Someone may use these nodes and not know that they are a no-op. Leaving a note for discussion elsewhere.

jstone-lucasfilm (approved):
Thanks for this investigation work, @kwokcb, and these improvements look good to me!

JGamache-autodesk:
Nice! I see we used the correct XML form and changed those `&&` to `&amp;&amp;` in some inline implementations.

There still remains a definite XML format break [enforced here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXFormat/External/PugiXML/pugixml.cpp#L3949). If we use UDIM tags, the resulting file will break the Python ElementTree parser:
```xml
   <input name="file" type="filename" value="diffuse<UDIM>.jpg" />
```
is illegal.

Correct XML representation is:
```xml
   <input name="file" type="filename" value="diffuse&lt;UDIM>.jpg" />
```
Looks like we chose readability by humans over strict adherence to W3 specifications by overriding PUGI code directly. This will be surprising to any TD wishing to apply XML based Python scripts to a collection of MaterialX documents. This hit me once this year as I use ET to check unit test result files and they started failing after I added UDIMs into the test mixture.

ld-kerley:
I would cast a vote to move to a completely conformant XML parser.

Issue #2608, opened 2025-10-08
CI : Support rendering using Metal for Mac


kwokcb:
## Requirements

- Update Github actions workflow to allow for headless rendering using Metal on Mac
- The current entry points used for rendering are `MaterialXView` and `MaterialXGraphEditor`. Each can be incrementally added to CI.
- Note: `MaterialXTest` render unit tests have never been added to CI so is not part of the scope of this issues.
- There are a few tests which work on runners without GPUs (software) which run for Linux only.

## Proposed changes

- Add test_render option to appropriate runners.
- Add setup to allow software rendering. 
```yaml
  - name: Initialize Headless Rendering (macOS)
      if: matrix.test_render == 'ON' && runner.os == 'macOS'
      run: |
        # Force software rendering for Metal backend 
        echo "MTL_HARDWARE_RENDERING=0" >> $GITHUB_ENV
        # Enable Metal debug layer for better debugging in CI
        echo "MTL_DEBUG_LAYER=1" >> $GITHUB_ENV
        # Force software rendering for OpenGL backend compatibility
        echo "LIBGL_ALWAYS_SOFTWARE=1" >> $GITHUB_ENV
```

### MaterialXView

This appears to work for MaterialXView with these results:

<img width="128" height="128" alt="Image" src="https://github.com/user-attachments/assets/dff8bfb8-6e24-45d5-a70c-dbfa5afdd45a" />

<img width="128" height="128" alt="Image" src="https://github.com/user-attachments/assets/f61db9ba-720d-477c-8df3-042ca2a71631" />

### MaterialXGraphEditor

Does not run with this error:

 MaterialXGraphEditor[18931:75242] Metal API Validation Enabled
Glfw Error 65545: NSGL: Failed to find a suitable pixel format

Root cause analysys is required to determine the cuase. It appears to be ImGUI setup related?

jstone-lucasfilm:
Thanks @kwokcb, and this is addressed in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2673.

PR #2651, opened 2025-10-30
Fix mx_shadow_platform.metal filename

ld-kerley:
I made a mistake when addressing notes in #2607.  This file was missed when renaming the source files.

This file is being included [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenMsl/MslShaderGenerator.cpp#L934).

I'm really surprised, and a little concerned that the unit tests didn't catch this.  We should investigate what is necessary to improve the coverage here.


jstone-lucasfilm:
Good catch, @ld-kerley, and I'll bet we don't yet have an example of _shadowed rendering on Metal_ in our GitHub CI test suite.

@kwokcb has proposed https://github.com/AcademySoftwareFoundation/MaterialX/issues/2608, which would add MaterialX Viewer renders on Metal to our GitHub CI, and I believe that would catch this issue in the future.

PR #2652, opened 2025-10-30
Improve ShaderGenerator API

ld-kerley:
In #2604 two new API calls for the `ShaderGenerator` class were added to allow customization of the creation of `ShaderNodeImplementation` objects for `NodeGraph` and `Implementation` elements respectively.

I think the API is cleaner, and more robust if we pass the actual implementation objects, and not the associated `NodeDef`.  In some cases these function have to re-acquire the implementation object from the `NodeDef`, but at the call sight we already have it.

#2604 has not been included in a MaterialX release yet - so I don't believe we have any problems changing the API.

### DIFF
--- a/source/MaterialXGenHw/HwShaderGenerator.h
+++ b/source/MaterialXGenHw/HwShaderGenerator.h
@@ -63,7 +63,7 @@ class MX_GENHW_API HwShaderGenerator : public ShaderGenerator
     /// Determine the prefix of vertex data variables.
     virtual string getVertexDataPrefix(const VariableBlock& vertexData) const = 0;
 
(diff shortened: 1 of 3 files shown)


source/MaterialXGenHw/HwShaderGenerator.h
```diff
@@ -64,7 +64,7 @@ class MX_GENHW_API HwShaderGenerator : public ShaderGenerator
     virtual string getVertexDataPrefix(const VariableBlock& vertexData) const = 0;
 
     /// Create the shader node implementation for a nodedef that has a NodeGraph implementation.
```
niklasharrysson: Just need to update the doc string as well

jstone-lucasfilm (approved):
Ship it!

PR #2656, opened 2025-11-05
Improve data library tests for OSL

ashwinbhat:
OSL Tests fail when nodes with namespaces are used. The root cause is the test tries to determine the nodename from the nodedef by stripping out "ND_" prefix. This change reworks the implementation to query the nodestring from the nodedef and ensure shadergen works for various input/output variations of the node.

### DIFF
--- a/documents/Specification/MaterialX.StandardNodes.md
+++ b/documents/Specification/MaterialX.StandardNodes.md
@@ -72,68 +72,118 @@ Standard Texture nodes:
 
 <a id="node-image"> </a>
 
-* **`image`**: samples data from a single image, or from a layer within a multi-layer image.  When used in the context of rendering a geometry, the image is mapped onto the geometry based on geometry UV coordinates, with the lower-left corner of an image mapping to the (0,0) UV coordinate (or to the fractional (0,0) UV coordinate for tiled images).
-The type of the &lt;image> node determines the number of channels output, which may be less than the number of channels in the image file, outputting the first N channels from the image file.  So a `float` &lt;image> would return the Red channel of an RGB image, and a `color3` &lt;image> would return the RGB channels of an RGBA image.  If the type of the &lt;image> node has more channels than the referenced image file, then the output will contain zero values in all channels beyond the N channels of the image file.
-    * `file` (uniform filename): the URI of an image file.  The filename can include one or more substitutions to change the file name (including frame number) that is accessed, as described in the [Filename Substitutions](./MaterialX.Specification.md#filename-substitutions) section in the main Specification document.
-    * `layer` (uniform string): the name of the layer to extract from a multi-layer input file.  If no value for `layer` is provided and the input file has multiple layers, then the "default" layer will be used, or "rgba" if there is no "default" layer.  Note: the number of channels defined by the `type` of the `<image>` must match the number of channels in the named layer.
-    * `default` (float or color<em>N</em> or vector<em>N</em>): a default value to use if the `file` reference can not be resolved (e.g. if a &lt;_geometry token_>, [_interface token_] or {_hostattr_} is included in the filename but no substitution value or default is defined, or if the resolved `file` URI cannot be read), or if the specified `layer` does not exist in the file.  The `default` value must be the same type as the `<image>` element itself.  If `default` is not defined, the default color value will be 0.0 in all channels.
-    * `texcoord` (vector2): the name of a vector2-type node specifying the 2D texture coordinate at which the image data is read.  Default is to use the current u,v coordinate.
-    * `uaddressmode` (uniform string): determines how U coordinates outside the 0-1 range are processed before sampling the image; see below.  Default is "periodic".
-    * `vaddressmode` (uniform string): determines how V coordinates outside the 0-1 range are processed before sampling the image; see below.  Default is "periodic".
-    * `filtertype` (uniform string): the type of texture filtering to use; standard values include "closest" (nearest-neighbor single-sample), "linear", and "cubic".  If not specified, an application may use its own default texture filtering method.
+### `image`
+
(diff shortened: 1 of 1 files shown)


source/MaterialXTest/MaterialXRenderOsl/GenReference.cpp
```diff
@@ -72,6 +72,18 @@ TEST_CASE("GenReference: OSL Reference", "[genreference]")
     {
         std::string nodeName = nodedef->getName();
         std::string nodeNode = nodedef->getNodeString();
+ 
+        //TODO: this test needs to be updated to reflect newer workflows.
+        // strip namespaces
+        if (nodedef->hasNamespace())
```
kwokcb: If you do this then if you have 2 defs with same nodename, but one is namespaced won't one clobber the other when you write ? 

Curious why you cannot use namespaced names ?

kwokcb: Forgot to add would using getQualifiedName() work since it includes the namespace. Currently I don't think getName() is correct here?

ashwinbhat:
@kwokcb here is short 30s debug capture of the issue.
https://github.com/user-attachments/assets/4be5b8f0-0262-4fc7-8e8c-345ae97ce98f

source/MaterialXTest/MaterialXRenderOsl/GenReference.cpp
ashwinbhat: I attached a short 30s debug session video to this PR. I still think we need a way to enumerate all available nodes in the library instead of trying to derive it off the nodedef name. My fix is just a workaround to the exception that get's thrown. Note that we do :
`mx::NodePtr node = datalib->addNodeInstance(nodedef, nodeName);` and then after shadergen
remove it `datalib->removeChild(node->getName());`

ashwinbhat: maybe we should be using `nodedef->getNodeString();` instead?

jstone-lucasfilm: @ashwinbhat That's the API call that I was imagining we would use here, rather than applying manual string operations on element names.

kwokcb: I think this could work:

- `NodeDef::getNodeString()` will return you the category which isn't unique.
- You could call `NodeDef::getType()` to find the output type. If it's `multioutput` then you could find all output types and concat. Basically a variant on `getType()`.
- To handle namespaces you could call `nodedef->getQualifiedName(()` on the value returned from `getNodeString()`. If you want to embed the namespace Otherwise some other mechanism could be used. e.g . Save it into a sub-folder.

This should give you something similar to the hack of removing the "ND_" string.

e.g. if your using the adsk def:
```
nodedef name="ND_adsk_bitmap_color4" node="bitmap" nodegroup="texture2d" version="1.0.1" isdefaultversion="true" namespace="adsk">
```

The hack gives you `adsk_bitmap_color4` which is could be generated as `<namespace>_<node name>_<type>`.

```
<nodedef name="ND_adsk_bitmap_float" node="bitmap" nodegroup="texture2d" version="1.0.1" isdefaultversion="true" namespace="adsk">
 ```
 becomes: `adsk_bitmap_float` following the same computation.

Comment (ashwinbhat) [reply source/MaterialXTest/MaterialXRenderOsl/GenReference.cpp]: @kwokcb I've reworked the changes lmk your thoughts.

File: source/MaterialXTest/MaterialXRenderOsl/GenReference.cpp [hunk 2]
```diff
@@ -70,20 +70,18 @@ TEST_CASE("GenReference: OSL Reference", "[genreference]")
...
     {
-        std::string nodeName = nodedef->getName();
-        std::string nodeNode = nodedef->getNodeString();
-        if (nodeName.size() > 3 && nodeName.substr(0, 3) == "ND_")
-        {
-            nodeName = nodeName.substr(3);
-        }
-
+        // determine the corresponding node for the nodedef
+        std::string nodeName = nodedef->getQualifiedName(nodedef->getNodeString());
```
jstone-lucasfilm, line 74: This looks much better to me, @ashwinbhat!

Just as a minor style note, we should make sure your newly added comments start with a capital letter, so that they match the surrounding style of the codebase.

source/MaterialXTest/MaterialXRenderOsl/GenReference.cpp
kwokcb: It's not quite enough as you'll get repeated names for output files.

I have this as a Python script and had to do this. Minimum you need the output type. 
Since defs can be polymorphic, you sometimes need to add the input name.

```python
        nodeString = nodedef.getQualifiedName(nodedef.getNodeString())
        outputs = nodedef.getOutputs()
        for output in outputs:
            nodeString += '_' + output.getType()
        if nodeString in names_used:
            orig_nodeString = nodeString
            inputs = nodedef.getInputs()
            for input in inputs:
                nodeString += '_' + input.getType()
                if nodeString not in names_used:
                    logger.debug(f'Resolve node string collision {orig_nodeString} -> {nodeString}')
                    break
        # Check if windows. If so clamp length to smaller than max size 
        if os.name == 'nt':
            max_path = 200
            nodeString = nodeString[:max_path]
        names_used.add(nodeString)
        remap_table[nodedef.getName()] = nodeString
```        

The entire file is this. You need to use  `nodeString` vs `nodedef.getName()`.
[generateReferenceShaders.py](https://github.com/user-attachments/files/23401707/generateReferenceShaders.py)

ashwinbhat: @kwokcb I've added enumeration of nodes by output and input type.

PR #2661, opened 2025-11-11
Improve OSL render testing

ld-kerley:
Currently the render unit tests are constrained to a small set of the mtlx files in resources.  I think we should run as many as we can - but thats a discussion for another day.... 

Locally if I enable all of the `Materials/Examples`, and all of the `Materials/TestSuite` folder there are a number of failures for OSL - this is because the shader it outputting an incompatible type with what `testrender` is expecting.

While working on the OSL network shader generator, @chellmuth added a new osl node `setCi` that can accept a number of different inputs and output a `closure color`. 

This PR generalizes, extends and replaces that idea. 

I'm adding a new MaterialX node definition specific for OSL use - which is why I've placed it in the implementation specific file - but I'm open to other ideas there.  This new node is largely the same as the node Chris wrote, but instead of leveraging the `isconnected()` mechanism, it uses an input integer to select which input to export from the node.  This is because in the OSL shader generator we don't have OSL connections. 

Both OSL shader generators then use a generator option to control if this node should be added inline with the outputs.  This ensures that all outputs are converted to `closure color` which means they should all be renderable with `testrender`. 

The GLSL shader generate was updated to correctly emit a boolean value to the image, previously we would always emit black. 

As previously requested - clang-format has been run on all the files and unit tests have been updated to skip testing this node, as it is not a "real" MaterialX node.


source/MaterialXGenOsl/OslUtil.h
```diff
@@ -0,0 +1,28 @@
...
+
+class ShaderGraph;
+class Document;
+using ConstDocumentPtr = shared_ptr<const Document>;
+class TypeSystem;
+using TypeSystemPtr = shared_ptr<TypeSystem>;
+class GenContext;
+
+/// Write docs
+MX_GENOSL_API void addSetCiTerminalNode(ShaderGraph& graph, ConstDocumentPtr document, TypeSystemPtr typeSystem, GenContext& context);
```
niklasharrysson: This function seems to only be called from within the generators. Would it make sense to place it as a protected method in the `OslShaderGenerator`, to avoid exposing it as new API? 

I see that `OslNetworkShaderGenerator` is not inheriting from `OslShaderGenerator`, but this would be a good reason to change that since they now share logic. As we have discussed before there are more things that can be unified in the future, like syntax handling.

libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
```diff
@@ -7,6 +7,27 @@
     Declarations for OSL implementations of standard nodes included in the MaterialX specification.
   -->
```
niklasharrysson, line 9: Maybe document here what this node is used for.

source/MaterialXGenOsl/OslSyntax.h
```diff
@@ -36,6 +40,15 @@ class MX_GENOSL_API OslSyntax : public Syntax
     static const StringVec COLOR4_MEMBERS;
 };
 
+/// Specialization of TypeSyntax for aggregate types.
+class MX_GENOSL_API OslStructTypeSyntax : public StructTypeSyntax
```
niklasharrysson: Could this class be moved/hidden in the .cpp file?

source/MaterialXGenOsl/OslSyntax.cpp
```diff
@@ -89,12 +89,12 @@ class OslIntegerArrayTypeSyntax : public OslArrayTypeSyntax
 
 // In OSL vector2, vector4, and color4 are custom struct types and require a different
 // value syntax for uniforms. So override the aggregate type syntax to support this.
-class OslStructTypeSyntax : public AggregateTypeSyntax
+class OslVecTypeSyntax : public AggregateTypeSyntax
```
niklasharrysson, line 92: Nice! A much better separation of concerns now.

resources/Materials/TestSuite/_options.mtlx
```diff
@@ -62,7 +62,8 @@
     <input name="extraLibraryPaths" type="string" value="" />
 
     <!-- List of document paths for render tests -->
-    <input name="renderTestPaths" type="string" value="resources/Materials/Examples/StandardSurface,resources/Materials/TestSuite/stdlib/convolution,resources/Materials/TestSuite/stdlib/color_management,resources/Materials/TestSuite/stdlib/procedural,resources/Materials/TestSuite/pbrlib/surfaceshader,resources/Materials/TestSuite/nprlib,resources/Materials/TestSuite/pbrlib/bsdf" />
+<!--    <input name="renderTestPaths" type="string" value="resources/Materials/Examples/StandardSurface,resources/Materials/TestSuite/stdlib/convolution,resources/Materials/TestSuite/stdlib/color_management,resources/Materials/TestSuite/stdlib/procedural,resources/Materials/TestSuite/pbrlib/surfaceshader,resources/Materials/TestSuite/nprlib,resources/Materials/TestSuite/pbrlib/bsdf" />-->
+    <input name="renderTestPaths" type="string" value="resources/Materials/Examples,resources/Materials/TestSuite" />
```
niklasharrysson: For the final merge perhaps remove the commented out part here.

niklasharrysson:
This looks like an excellent change! Nice improvement and generalization. I just added a few comments above.

source/MaterialXGenOsl/OslUtil.h
ld-kerley: I took this other approach because my intuition is that we shouldn't be inheriting one from the other.  I have experimental work  - not ready for PR yet - where very little is actually shared between the two generator syntaxes. Which kind of makes sense when you think the two are just emitting two completely forms of text. One is emitting OSL shader language, and the other is emitting the shader group text - even the parameter formatting is different between the two in some cases.

ld-kerley: If its ok with you'd like to try going down this route first - and if it turns out we do need to end up sharing - we can refactor this in to protected methods later.

Concretely I have some work where the OSLCompile call and other pieces including the generation of the shader code for the OSO end up in this OslUtil module, and I'm not sure they would really make sense as protected methods.

resources/Materials/TestSuite/_options.mtlx
ld-kerley: For the final merge I was actually intended to revert to the existing set of paths to render - I know in the past Jonathan has had concerns around the tests taking too long to run - and this change significantly increases the run time.   I just included it here for anyone who was testing the PR locally.

source/MaterialXGenOsl/OslSyntax.h
ld-kerley: yep - I'll move it

source/MaterialXGenOsl/OslUtil.h
ld-kerley: Actually - for now let's just add it as a protected method.  But it's possible I may suggest refactoring this again later.

libraries/stdlib/genosl/mx_set_ci.osl
```diff
@@ -0,0 +1,70 @@
...
+            c = color(input_matrix44[0][0], input_matrix44[1][1], input_matrix44[2][2]);
+            a = input_matrix44[4][4];
+        } else if (output_type == 14) {
+            c = color(input_displacementshader);
+        }
+        out_ci = c * a * emission() + (1-a) * transparent();
+    }
+
+    Ci = out_ci;
+}
```
jstone-lucasfilm: Two minor notes:

1. Perhaps this new shader source file should be located within `stdlib/genosl/lib` for consistency?
2. I'd recommend adding a final newline to the file!

source/MaterialXGenOsl/OslShaderGenerator.cpp
```diff
@@ -191,8 +196,7 @@ void OslShaderGenerator::registerShaderMetadata(const DocumentPtr& doc, GenConte
     }
 
     // Rename the standard metadata names to corresponding OSL metadata names.
-    const StringMap nameRemapping =
-    {
+    const StringMap nameRemapping = {
```
jstone-lucasfilm: Minor note: I think this may just be `clang-format` converting data initializers from Allman style to K&R style, rather than a functional change.

jstone-lucasfilm (commented):
I like this change overall, and had just a few minor notes for improvements.

libraries/stdlib/genosl/mx_set_ci.osl
ld-kerley: I'll move it if you'd like - but it sits along side all the other OSL node implementations currently - and this is also an OSL node implementation - not a header with helper functions - which would feel more natural under `stdlib/genosl/lib` .

source/MaterialXGenOsl/OslShaderGenerator.cpp
ld-kerley: I'm really starting to wonder if Allman style doesn't define the brace rules for data initializers - is there any official documentation anywhere on what Allman style is - if clang-format is wrong - then perhaps you could submit a bug report for it?

jstone-lucasfilm: This seems to be a known issue with braced initializer lists in `clang-format`, as described here:

https://github.com/llvm/llvm-project/issues/134539

I can't imagine any project would want Allman braces for code and K&R braces for initializer lists, so this seems like an important bug to fix!

jstone-lucasfilm: Indeed it's pretty hard to find a formal definition of Allman or K&R braces that covers this specific case, but here's what I get when I ask Gemini:

> Yes, the Allman style generally applies to data initializers that use braces (e.g., array, object, or struct initialization) by placing the opening and closing braces on their own new lines, aligned with the current indentation level. 
> The core principle of Allman style is that a brace should be on a new line and aligned vertically with the statement or declaration it belongs to, which provides visual clarity and makes it easier to match braces.

jstone-lucasfilm (approved):
Ship it!

Issue #2663, opened 2025-11-11
Unit test fails with hidden files: getFilesInDirectory()


kwokcb:
## Failure Condition.

The File.cpp unit test file has a check for call getFilesInDirectory() on a folder. 
On Mac, I happen to get a .DS_Store file autocreated. Thus there is a mismatch
in files found since this is not on the list of "all files" to test.

## Resolution

I'm not sure how to resolve this other than filtering out .* files if this makes sense
in the unit test.

ld-kerley:
`.DS_Store` file gets created when a folder is browsed to by the Finder application - which explains why the unit test worked for me - as I tend to work in the terminal more. (edit) I think its more nuanced than this because I wasn't able to create a `.DS_Store` at all on my machine....

~~I think the right thing to do here is just to add an exceptions file list for this unit test - unclear to me if other OS might have similar files etc.~~

I actually think maybe we should just invert the check and validate that the expected files exist - rather than all the returned files match the expected files - I'm gonna put up a PR for this and we can discuss there.

PR #2666, opened 2025-11-11
Improve get all files in directory unit test

ld-kerley:
As reported in #2663 the get all files in directory unit test is not robust enough to OS generated files.

This PR inverts the validation logic to make it more robust.

Looks like this idea fails the windows CI - but im not sure why and don't have a windows machine to investigate.

### DIFF
--- a/source/MaterialXFormat/File.cpp
+++ b/source/MaterialXFormat/File.cpp
@@ -185,7 +185,7 @@ FilePathVec FilePath::getFilesInDirectory(const string& extension) const
 
 #if defined(_WIN32)
     WIN32_FIND_DATAA fd;
-    FilePath query = extension.empty() ? *this : (*this / ("*." + extension));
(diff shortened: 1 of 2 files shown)


kwokcb:
There is an issue with the logic in `FilePath::getFilesInDirectory()` for Windows:

You need this change since on `FindFirstFileA` will try to match that specific file (not the contents of the folder).
```c++
    FilePath query = extension.empty() ? (*this / "*") : (*this / ("*." + extension));
    //FilePath query = extension.empty() ? *this : (*this / ("*." + extension)); // finds nothing
    HANDLE hFind = FindFirstFileA(query.asString().c_str(), &fd);
```

Also suggest to put in a `REQUIRE(results.size() > 0)` in the unit test to check for this.

I ran with these changes locally and it seems to work if you have extra files in the folders being checked.

Issue #2664, opened 2025-11-11
Metal test suite rendering : image files are not found


kwokcb:
## Issue

I tried running Metal render tests and found that the image files are not found.
A comparison run using the same build does find image files.

## Test output

Here is a snippet from the logged output run using `MaterialXTest "[rendermsl]"
```sh
Validating MSL rendering for: /Users/bernardkwok/work/bernard_materialx/build/bin/resources/Materials/Examples/StandardSurface/standard_surface_chess_set.mtlx
Image file not found: chess_set/king_white_base_color.jpg
Image file not found: chess_set/king_white_roughness.jpg
Image file not found: chess_set/king_white_normal.jpg
Validating MSL rendering for: /Users/bernardkwok/work/bernard_materialx/build/bin/resources/Materials/Examples/StandardSurface/standard_surface_chess_set.mtlx
Image file not found: chess_set/knight_black_base_color.jpg
Image file not found: chess_set/knight_black_roughness.jpg
Image file not found: chess_set/knight_black_normal.jpg
```
Example render comparison:

<img width="1596" height="1158" alt="Image" src="https://github.com/user-attachments/assets/5ba5eccf-ae8b-47db-a72d-5aa729e7c189" />

PR #2667, opened 2025-11-12
Fix image search path for Metal render test suite

kwokcb:
## Issue

Fixes: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2664.
where some tests would not find images in relative paths.

### Change
Pass image search path to image handler search path for MSL test renderer.

### Results

Example results. There are no image missing warnings anymore.

image

### Notes
- This is a subtle issue related to unused args as well as it is not caught due to the arg
being commented out.
- Recommend that an missing images should be tagged as render failures instead of warnings in test suite.


jstone-lucasfilm (approved):
Good catch, thanks @kwokcb!

ld-kerley (commented):
LGTM too

Issue #2672, opened 2025-11-13
MaterialX GraphEditor : Support Metal Rendering


kwokcb:
## Proposal

Currently it seems like the Graph Editor is using OpenGL to render with.
The proposal is to switch it to use Metal (by default, if there is any option).

### Notes

- MaterialXView does use Metal so theoretically it should render the same but could be confusing to users.
- This would help CI rendering as well. Currently I cannot find any way to run this headless.

PR #2673, opened 2025-11-13
Enable CI Rendering on Mac

kwokcb:
## Change

- Turn on headless rendering on Mac for MaterialXView
- Added to the MacOS_Xcode_16_Python313 workflow as that already does shader validation testing.

Results:

Viewer_BrassAverage
Viewer_CarpaintTranslated

Notes:

- MaterialxGraphEditor rendering not enabled as it's using OpenGL which has issues running headless. (A[ new issue](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2672) has been logged to support Metal rendering here. If / when it is available then it can be turned on)
- The Metal environment set up here is the same pre-requisite as for this [PR for full test suite enabling](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2675).

### DIFF
--- a/source/MaterialXFormat/File.cpp
+++ b/source/MaterialXFormat/File.cpp
@@ -185,7 +185,7 @@ FilePathVec FilePath::getFilesInDirectory(const string& extension) const
 
 #if defined(_WIN32)
     WIN32_FIND_DATAA fd;
-    FilePath query = extension.empty() ? *this : (*this / ("*." + extension));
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
jstone-lucasfilm: Since you've split the steps for the Viewer and Graph Editor, I'd say we can go further and call this step "Viewer Tests", and call the next step "Graph Editor Tests (Linux)".

kwokcb: Sounds good.

PR #2675, opened 2025-11-13
Enable test suite rendering on Mac

kwokcb:
## Change

- Turn on rendering of test suite  for Mac CI runners

## Changes

- Add in a CMake flag (`MATERIALX_RENDER_MSL_ONLY`) on render in paths supprting `Metal`. This allows CI to run without also trying to run GLSL renders which are not supported.
- When set this will enable the existing MATERIALX_TEST_RENDER flag to trigger test suite rendering.
- Add in an archive and upload test of the `resources` folder results.

## Results

- Run through entire test suite succesfully.
- The renders are very fast (total unit test time is about 1.5 min with render time a percentage of that)
- Packaged downloaded resources area contains rendered images. Below is a snapshot of the downloaded renders unzipped locally:

image

- Downloaded tests are the same as local rendering with the exception of `resources/Materials/TestSuite/pbrlib/surfaceshader/usd_normal_map` which for some reason is rendering the shader from the incorrect path. This occurred 1 out of 3 runs and left to be resolved separate from activation. One other test has a RMS diff of 0.00001

image
 
 Some other sample renderings (cannot attach full pdf results into github as file is too large)
 
image


.github/workflows/main.yml
```diff
@@ -203,7 +204,13 @@ jobs:
             EXTENDED_BUILD_CONFIG="$EXTENDED_BUILD_CONFIG -DMATERIALX_MDL_SDK_DIR=C:/vcpkg/installed/x64-windows"
           fi
         fi
-        cmake -S . -B build -DMATERIALX_BUILD_PYTHON=ON -DMATERIALX_BUILD_VIEWER=ON -DMATERIALX_BUILD_GRAPH_EDITOR=ON -DMATERIALX_BUILD_TESTS=ON -DMATERIALX_TEST_RENDER=OFF -DMATERIALX_WARNINGS_AS_ERRORS=ON $EXTENDED_BUILD_CONFIG ${{matrix.cmake_config}}
+        TEST_RENDER_CONFIG="-DMATERIALX_TEST_RENDER=OFF"
+        if [ "${{ matrix.test_render }}" == "ON" ]; then
+          if [ "${{ runner.os }}" == "macOS" ]; then
+            TEST_RENDER_CONFIG="$TEST_RENDER_CONFIG -DMATERIALX_TEST_RENDER=ON -DMATERIALX_RENDER_MSL_ONLY=ON"
```
kwokcb: Key change: Turn on test rendering for MSL but not GLSL.

.github/workflows/main.yml
```diff
@@ -315,6 +337,25 @@ jobs:
         name: Renders_${{ matrix.name }}
         path: build/render/*.png
 
+    - name: Archive Resources (macOS)
+      if: matrix.test_render == 'ON' && runner.os == 'macOS'
```
kwokcb: Post results.

CMakeLists.txt
```diff
@@ -112,6 +113,17 @@ if (MATERIALX_BUILD_JS)
     set(MATERIALX_BUILD_TESTS OFF)
 endif()
 
+# Validate MSL-only rendering option
+if(MATERIALX_RENDER_MSL_ONLY)
```
kwokcb, line 117: Turn off GLSL rendering. For some reason there a renderGLSL module dependency for renderMSL. Not touching that dependency here.

source/MaterialXTest/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ endif()
 if(MATERIALX_BUILD_RENDER)
   add_subdirectory(MaterialXRender)
   target_link_libraries(MaterialXTest MaterialXRender)
-  if(MATERIALX_BUILD_GEN_GLSL)
+  if(MATERIALX_BUILD_GEN_GLSL AND NOT MATERIALX_RENDER_MSL_ONLY)
```
kwokcb, line 72: Turn off GLSL rendering if want MSL only

ld-kerley (approved):
This looks like a great step forwards to me - the most testing we're able to do the better.

I hope we can even build on top of this to introduce regression testing for renders

.github/workflows/main.yml
```diff
@@ -331,6 +337,25 @@ jobs:
...
+      run: |
+        mkdir -p build/artifacts
+        if [ ! -d build/bin/resources ]; then
+          echo "Missing build/bin/resources"
+          echo "Contents of build/bin (if present):"
+          ls -la build/bin || true
+          exit 1
+        fi
+        zip -r build/artifacts/resources.zip build/bin/resources
```
jstone-lucasfilm: What is the motivation for _manually_ generating a zip archive of the folder, rather than leveraging the built-in zip functionality in `actions/upload-artifact`?

Would it make sense to align this with our other upload steps (e.g. Upload Reference Shaders, Upload Renders) and use the existing system in `actions/upload-artifact`?

.github/workflows/main.yml
jstone-lucasfilm: Perhaps we can merge the existing `EXTENDED_BUILD_CONFIG` with your new `TEST_RENDER_CONFIG`, naming the combined variable `ADDITIONAL_BUILD_CONFIG` for generality?

That should help to keep the final `cmake build` call as compact and clear as possible, as it's becoming a bit hard to read!

.github/workflows/main.yml
kwokcb: Good catch. I had this "cherry picking" images at one point but reverted that as it's better to produce a "results" folder as part of test. Will revert this to the simpler rule.

.github/workflows/main.yml
```diff
@@ -193,6 +193,7 @@ jobs:
     - name: CMake Generate
       shell: bash
       run: |
+        EXTENDED_BUILD_CONFIG=""
```
kwokcb, line 196: Small fix: This was not initialized so would error out if none of the "extended" conditions were set.

kwokcb (commented):
Review changes in. 

Note: I will need to get some help to triage why the single case that renders the 
the incorrect shader sporadically. 2 out of 3 runs were good.

Issue #2609, opened 2025-10-08
Proposal: MaterialX Flake3d Node


msuzuki-nvidia:
# MaterialX Flake3d Node Proposal

## Summary

Flake generation provides a procedural method to create sparkling, metallic flake effects commonly seen in automotive paints, cosmetics, and other materials with embedded reflective particles. This document proposes a new MaterialX node `<flake3d>` based on the algorithm developed by **Matthias Raab** that generates procedurally distributed flake normals with controllable size, roughness, and coverage.

This node is particularly valuable for creating realistic car paint materials, glittery surfaces, and other effects that require procedural generation of micro-scale reflective particles without requiring high-resolution texture maps.

## Key Features

The `<flake3d>` node provides the following key features:
- **Procedural generation**: Creates infinite variations of flake patterns without requiring texture assets
- **Physically-based distribution**: Uses microfacet theory (GGX) for realistic normal perturbation
- **Efficient evaluation**: Optimized algorithm suitable for real-time rendering
- **Artistic control**: Intuitive parameters for size, roughness, and coverage density

### `<flake3d>`

| Parameter      | Type     | Range      | Default  | Description                                                          |
| -------------- | :------: | :--------: | :------: | -------------------------------------------------------------------- |
| **size**       | float    | [0.0, 5.0] | 0.01     | Size of individual flakes                                            |
| **roughness**  | float    | [0.0, 1.0] | 0.1      | Roughness of the flake normal distribution                           |
| **coverage**   | float    | [0.0, 1.0] | 0.5      | Density of flake coverage (0 = no flakes, 1 = full coverage)         |
| **position**   | vector3  |            | `Pworld` | 3D position for flake generation                                     |
| **normal**     | vector3  |            | `Nworld` | Base surface normal                                                  |
| **tangent**    | vector3  |            | `Tworld` | Surface tangent vector                                               |
| **bitangent**  | vector3  |            | `Bworld` | Surface bitangent vector                                             |

## Outputs

The node returns multiple outputs to enable advanced shading workflows:

| Output       | Type    | Range          | Description                                                                    |
| ------------ | :-----: | :-----------:  | ------------------------------------------------------------------------------ |
| **id**       | int     | [0, INT32_MAX] | Unique identifier for each flake (0 for no flake)                              |
| **rand**     | float   | [0.0, 1.0]     | Random value per flake for additional variation (0 for no flake)               |
[...]

PR #2676, opened 2025-11-13
Add flake node implementations

msuzuki-nvidia:
This PR adds GLSL and MDL implementations for the flake nodes proposed in #2609.

Nodes to add:

- `<flake2d>`
- `<flake3d>`

### DIFF
--- a/source/MaterialXGenMdl/mdl/materialx/flake.mdl
+++ b/source/MaterialXGenMdl/mdl/materialx/flake.mdl
@@ -75,7 +75,6 @@ export struct mx_flake__result
     float mxp_rand;          // random float [0, 1], 0 for no flake
     float mxp_presence;      // flake presence [0, 1], 0 for no flake
     float3 mxp_flakenormal;  // flake normal
-
 };
 
 export mx_flake__result mx_flake(
@@ -88,7 +87,6 @@ export mx_flake__result mx_flake(
     float3 bitangent
 )
 {
-
     float probability = mx_flake_density_to_probability(math::saturate(coverage));
     const float flake_diameter = 1.5 / math::sqrt(3.0);
 
@@ -143,7 +141,6 @@ export mx_flake__result mx_flake(
             }
         }
     }
-
 
     mx_flake__result flake_out(0, 0.0, 0.0, normal);
 


libraries/stdlib/genglsl/lib/mx_math.glsl
```diff
@@ -1,4 +1,5 @@
 #define M_FLOAT_EPS 1e-8
+#define M_PI 3.1415926535897932
```
ld-kerley, line 2: Metal defines `M_PI_F` along with other constants - if you'd be to defining it that way we could avoid specifying it all in `mx_math.metal`.

libraries/stdlib/genglsl/lib/mx_flake.glsl
```diff
@@ -0,0 +1,158 @@
...
+            for (int k = -1; k < 2; ++k)
+            {
+                uint seed = mx_flake_init_seed(base_P_int + ivec3(i, j, k));
+
+                seed = mx_flake_xorshift32(seed);
+                if (mx_uint_to_01(seed) > probability)
+                    continue;
+
+                seed = mx_flake_xorshift32(seed);
+                float priority = mx_uint_to_01(seed);
```
ld-kerley, line 104: Minor suggestion : feel free to ignore....

I wonder if we might be able to replace this with

`float priority = mx_cell_noise_float(base_P_int + ivec3(i, j, k));`

If this gives a similar enough result - then it would perhaps make things simpler to port to OSL - as we already know (by design) the noise functions match between GLSL and OSL.

libraries/stdlib/genglsl/mx_flake2d.glsl
```diff
@@ -0,0 +1,19 @@
...
+    vec3 normal,
+    vec3 tangent,
+    vec3 bitangent,
+    out int id,
+    out float rand,
+    out float presence,
+    out vec3 flakenormal
+)
+{
+    vec3 position = vec3(texcoord.x, texcoord.y, 0.0);
```
ld-kerley, line 18: I think this 2d->3D trick is totally valid - but it might be worth leaving a comment that there might be a future performance improvement to write a 2D specific version of mx_flake().  I don't think we should write that yet - but a comment might leave a clue for our future selves for improvements that might be valid.

libraries/stdlib/genglsl/lib/mx_math.glsl
msuzuki-nvidia: `M_PI` is already used [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet.glsl) (but can't reference it from stdlib) so I think we want to keep this for consistency.

libraries/stdlib/genglsl/lib/mx_flake.glsl
msuzuki-nvidia: I was trying to use it actually, but `mx_hash_int()` is much more complex than xorshift. This shader needs new hash 7 times per shade so decided not to use it for the performance.

libraries/stdlib/genglsl/mx_flake2d.glsl
msuzuki-nvidia: Sounds good, added a comment.

libraries/stdlib/genglsl/lib/mx_math.glsl
ld-kerley: OK - then maybe it makes sense in `mx_math.metal` to say 
```
#define M_PI M_PI_F
```

msuzuki-nvidia: Sounds good, done.

jstone-lucasfilm:
Apologies for the delay in reviewing this, @msuzuki-nvidia, and I'd like to return community focus to this PR.

Would you mind addressing the merge conflicts, so that we can begin testing this in the latest MaterialX codebase?

msuzuki-nvidia:
@jstone-lucasfilm Resolved. I hope we can get reviews on PRs more regularly before they expire.

jstone-lucasfilm:
That's my aim, @msuzuki-nvidia, to gather sufficient community focus on this PR that it can be reviewed and merged.  Bringing it up for our next MaterialX TSC seems like a good first step along that path.

documents/Specification/MaterialX.StandardNodes.md
```diff
@@ -512,6 +512,43 @@ A single node supporting 3D Perlin, Cell, Worley or Fractal noise in a unified i
...
+|`size`       |The size of individual flakes, with smaller values producing larger flakes         |float  |0.01    |
+|`roughness`  |The surface roughness of individual flakes, controlling the variation in normal    |float  |0.1     |
+|`coverage`   |The density of flakes in the pattern, ranging from 0.0 (no flakes) to 1.0 (maximum)|float  |0.5     |
+|`texcoord`   |The 2D texture coordinate at which the flake pattern is evaluated                  |vector2|_UV0_   |
+|`normal`     |The surface normal vector used as the base for flake normal perturbations          |vector3|_Nworld_|
+|`tangent`    |The surface tangent vector, used to construct the tangent space                    |vector3|_Tworld_|
+|`bitangent`  |The surface bitangent vector, used to construct the tangent space                  |vector3|_Bworld_|
+|`id`         |Output: unique identifier for each flake. 0 for no flake                           |integer|0       |
+|`rand`       |Output: random value per flake for additional variation. 0.0 for no flake          |float  |0.0     |
+|`presence`   |Output: presence value per flake. 0.0 for no flake                                 |float  |0.0     |
```
crydalch: This is the tiniest of notes, but in the spec doc, I think it'd be helpful to also mention `presence` is a depth-based value (for both 2d/3d entries, of course).

crydalch:
Added a small note on the spec doc, but it otherwise looks good. No further comments from us since looking at the proposal in October, this will be welcome/exciting! Sorry we didn't transfer our thumbs-up to the PR when it came online in November.

Thanks @msuzuki-nvidia !

documents/Specification/MaterialX.StandardNodes.md
msuzuki-nvidia: good idea, updated!

fpliu:
are there any more comments? Is this good to merge @jstone-lucasfilm

PR #2682, opened 2025-11-16
Restore render tests in standard OSL

jstone-lucasfilm:
This changelist restores a handful of render tests in standard OSL, allowing them to be included in GLSL/OSL/MDL comparisons.

These render tests were marked as skipped as part of the new OSO building functionality, so this changelist leaves them as skipped when `MATERIALX_BUILD_OSOS` is set.

### DIFF
--- a/resources/Materials/TestSuite/pbrlib/surfaceshader/normalmapped_surfaceshader.mtlx
+++ b/resources/Materials/TestSuite/pbrlib/surfaceshader/normalmapped_surfaceshader.mtlx
@@ -17,7 +17,7 @@
       <input name="scale" type="float" value="1.1" />
     </normalmap>
     <tiledimage name="tiledimage" type="vector3">
-      <input name="file" type="filename" value="resources/images/mesh_wire_norm.png" />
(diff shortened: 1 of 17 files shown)


jstone-lucasfilm:
Here's a new GLSL/OSL render comparison using the code in this changelist, where previous render comparisons were missing OSL images for several material examples that are supported in both GLSL and OSL:

[MaterialXRenderTests_11_16_2025_GitHub.pdf](https://github.com/user-attachments/files/23570728/MaterialXRenderTests_11_16_2025_GitHub.pdf)

jstone-lucasfilm:
Good suggestion, @chellmuth, and I've made that change.

jstone-lucasfilm:
@ld-kerley I've added the new pbrlib tests to this change as recommended in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2683, and let me know how this looks to you.

ld-kerley (approved):
looks good to me - I have a few more updates cooking that I hope will end up reducing this list again

PR #2683, opened 2025-11-16
Restore OSL Network test pass

ld-kerley:
This PR skips the files that currently fail the OSL Network MaterialX tests.

I have some ideas about how we might fix these - but for now we really should have the tests pass.

If we don't want to take this approach, then we should revert the addition of the `pbrlib` to the render test.

### DIFF
--- a/source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
+++ b/source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp
@@ -113,6 +113,10 @@ class OslShaderRenderTester : public RenderUtil::ShaderRenderTester
         _skipFiles.insert("network_surfaceshader.mtlx");
         _skipFiles.insert("sheen.mtlx");
         _skipFiles.insert("toon_shade.mtlx");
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This looks good, thanks @ld-kerley, though we should integrate it with the related fix in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2682.

If you'd like, you can just copy the two lines from that change into this one, so that we can then merge both fixes in a single PR.

ld-kerley:
go for it

jstone-lucasfilm:
@ld-kerley Now that these fixes have been integrated into #2682, reviewed, and merged, are there any remaining steps that you'd like to see for this PR?

ld-kerley:
I think we can close this out - I will note that I think your fix accidentally included the hextile test for the OSL source generator - but we don't yet have an OSL implementation of that node.

I have a work that I'm trying to get tidied up that brings the OSL network generator in to 1:1 parity with the OSL source generator - in that PR I'll be removing the carve out here separating the two different generators.

PR #2684, opened 2025-11-17
Handle backsurface in MDL generator

jreichel-nvidia:


jreichel-nvidia:
CI run with extended build: https://github.com/jreichel-nvidia/MaterialX/actions/runs/19424555798

PR #2687, opened 2025-11-19
Fix CMake build rule for resource installation

kwokcb:
## Issue

The post build step for MaterialXTest does no trigger so resources are not copied to the
build area on a build. 

This means developers must manually copy this over when they are changing any files affecting the resources folder. If the lack of copy is not noticed / done then it can get confusing as to why source changes are not reflected in the build tests. 

## Change

Add an explicit target for resource copy and make MaterialXTest dependeng on it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -115,7 +115,6 @@ jobs:
           python: 3.13
           cmake_config: -G "Visual Studio 17 2022" -A "x64"
           test_shaders: ON
-          extended_build_oiio: ON
           extended_build_mdl_sdk: ON
 
         - name: Windows_VS2022_x64_SharedLibs
@@ -123,6 +122,7 @@ jobs:
           architecture: x64
           python: None
           cmake_config: -G "Visual Studio 17 2022" -A "x64" -DMATERIALX_BUILD_SHARED_LIBS=ON
+          extended_build_oiio: ON
           upload_shaders: ON
 
     steps:
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -70,6 +70,7 @@ The current Stakeholders of the MaterialX TSC are:
 - Lutz Kettner - NVIDIA
 - Chris Kulla - Epic Games
 - Bernard Kwok - Khronos Group
+- Karen Lucknavalai - Pixar USD
 - André Mazzone - ILM
 - Magnus Pettersson - IKEA
 - Dimitar Toshev - Chaos
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -811,7 +811,7 @@ void Graph::setRenderMaterial(UiNodePtr node)
             for (const std::string& testPath : testPaths)
             {
                 mx::ElementPtr testElem = _graphDoc->getDescendant(testPath);
-                mx::NodePtr testNode = testElem->asA<mx::Node>();
+                mx::NodePtr testNode = testElem ? testElem->asA<mx::Node>() : nullptr;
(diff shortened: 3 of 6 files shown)


jstone-lucasfilm:
This looks very promising, thanks @kwokcb, and I'll just add a note that we should time our build process before and after this change, to make sure we're maintaining build performance.

kwokcb:
It doesn't have any affect on the initial build since resources are always installed so CI build time won't change. 

To check however, I put a timer in the build step and get this so it's pretty negligable.

```sh
Copying OSL utilities to build directory...
Elapsed time (seconds): 0.00699933
[ 68%] Built target MaterialXTestOslUtilities
[ 68%] Copying resources to build directory
Copying resources to build directory...
Elapsed time (seconds): 0.0909475
```

I could check in the timer code but don't think it's needed.

kwokcb:
After chatting with @lkerley on Slack, I've made the small change to only copy if files have changed to minimize the time spent in case the file system is slower.

source/MaterialXTest/CMakeLists.txt
```diff
@@ -110,13 +110,22 @@ if(MATERIALX_BUILD_GEN_MDL)
         DESTINATION ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${MATERIALX_INSTALL_STDLIB_PATH}/mdl)
 endif()
 
-add_custom_command(TARGET MaterialXTest POST_BUILD
-    COMMAND ${CMAKE_COMMAND} -E copy_directory
-    ${CMAKE_CURRENT_SOURCE_DIR}/../../resources ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/resources)
+# Copy resources to build directory - always runs during build
+add_custom_target(MaterialXTestResources ALL
+    COMMAND ${CMAKE_COMMAND} -E copy_directory_if_different
```
kwokcb, line 115: Copy only files which have changed.

kwokcb:
I've done some timings on my machine which is a 4 yr old Alienware personal laptop.

Clean builds show about 2 seconds for resources. libraries which we already copy is about a second. 
Without this change this is already the overhead when building with clean resources and libraries.

Incremental builds are about 1 second.

```sh
// Clean build
[MaterialXTestData] Copying data libraries...
  Elapsed time (seconds): 0.944269
  Copying OSL utilities to build directory
  [MaterialXTestOslUtilities] Copying OSL utilities...
  Elapsed time (seconds): 0.0731214
  Copying resources to build directory
  [MaterialXTestResources] Copying resources...
  Elapsed time (seconds): 1.86759

   [MaterialXTestData] Copying data libraries...
  Elapsed time (seconds): 0.951778
  Copying OSL utilities to build directory
  [MaterialXTestOslUtilities] Copying OSL utilities...
  Elapsed time (seconds): 0.0723189
  Copying resources to build directory
  [MaterialXTestResources] Copying resources...
  Elapsed time (seconds): 1.87808 

// Incremental build
  [MaterialXTestData] Copying data libraries...
  Elapsed time (seconds): 0.206294
  Copying OSL utilities to build directory
  [MaterialXTestOslUtilities] Copying OSL utilities...
  Elapsed time (seconds): 0.0529004
  Copying resources to build directory
  [MaterialXTestResources] Copying resources...
  Elapsed time (seconds): 0.728033

  [MaterialXTestData] Copying data libraries...
  Elapsed time (seconds): 0.194733
  Copying OSL utilities to build directory
  [MaterialXTestOslUtilities] Copying OSL utilities...
  Elapsed time (seconds): 0.0547186
  Copying resources to build directory
  [MaterialXTestResources] Copying resources...
  Elapsed time (seconds): 0.705728
  ```

PR #2690, opened 2025-11-19
Improve memory handling in traversal singletons

jstone-lucasfilm:
This changelist implements improvements to the memory handling for traversal singletons, making sure that all allocated memory is released when the application is closed.

### DIFF
--- a/source/MaterialXCore/Traversal.cpp
+++ b/source/MaterialXCore/Traversal.cpp
@@ -9,21 +9,25 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
-const Edge& getNullEdge() {
-    static const auto ret = new Edge(nullptr, nullptr, nullptr);
+const Edge& getNullEdge()
+{
+    static const std::unique_ptr<Edge> ret = std::make_unique<Edge>(nullptr, nullptr, nullptr);
     return *ret;
 }
 
-const TreeIterator& getNullTreeIterator() {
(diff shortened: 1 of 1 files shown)


ld-kerley:
The original code here shouldn't leak memory, as it should be recovered to the system as the process is torn down.

Its important for us at Apple that we don't raise warnings with `-Wexit-time-destructors`, which is still a work in progress, but my initial PR proposing the code you're changing here was the first step.

The `unique_ptr` has a destructor that it's calling at exit here - so introducing this code defeats the purpose of my original PR.

jstone-lucasfilm:
Ah, I didn't realize _exit-time destructors_ were the problem for Apple builds, and I'd mistakenly thought that the problem was constructors at static initialization time in Apple builds.

In that case, I think you're right that we need to leave these pseudo-memory-leaks in the MaterialX codebase, unless another developer can propose a better solution.

Leak analysis tools will still flag this code as the source of "memory leaks", but I think we can live with this, as long as we're aware that it's a tradeoff we're intentionally making to better support Apple builds.

ld-kerley:
I'd be happy if we wanted to provide some sort of #ifdef guarded solution here - so we can just opt in easily.  Ultimately it would be really helpful (after a number of PRs with more patches land) to have a CI run where we build with the stricter compile flags we need. Being able to build direct from source without maintaining patches would really help us stay on the latest MatX version in the OS

PR #2691, opened 2025-11-20
Refactor LibToOSO tool

ld-kerley:
* Localize the compileOSL() code, this removes the dependency on MaterialXRender
* Add optional dependency on liboslcomp, which allows compiling of the osos without using std::system. This significantly speeds up compilation time, on my machine dropping from about a minute for the entire data library to about 15 seconds.
* With liboslcomp, we can avoid writing the OSL source to disk - saving space. This is controlled by a command line argument
* Add command line arguments to specify data library location
* Add argument to avoid writing the MaterialX document for the implementation elements, this may be helpful for those not using MaterialX shader generation.


niklasharrysson (approved):
LGTM!

jstone-lucasfilm (approved):

PR #2693, opened 2025-11-22
Update string constants to avoid global constructors - Option 1

ld-kerley:
As part of the work to make adopting MaterialX easier for us and lowering the number of internal patches we have to carry for security and performance reasons, this PR proposes a method we could use to avoid the global constructor/destructor calls for std::string constants.

There are actually two different possible approaches here - the other is in #2694 

This PR only applies the change to a single source file `Property.cpp/h`, but if/when one of the two options is selected it would be my intention to apply the same pattern at least to the rest of `MaterialXCore`, and then eventually elsewhere. 

This approach is probably my preferred one, it uses `constexpr std::string_view` instead of `const std::string`, but it does need API change to the attribute interface on `Element`. Everything compiles, so I believe it to be compatible with the existing source.  Also some changes were necessary in the Javascript bindings, which I'm not familiar with  - hoping experts there can highlight any problems, if they exist.

### DIFF
--- a/source/MaterialXCore/Element.h
+++ b/source/MaterialXCore/Element.h
@@ -479,19 +479,19 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
     /// @{
 
     /// Set the value string of the given attribute.
-    void setAttribute(const string& attrib, const string& value);
+    void setAttribute(std::string_view attrib, const string& value);
 
     /// Return true if the given attribute is present.
-    bool hasAttribute(const string& attrib) const
+    bool hasAttribute(std::string_view attrib) const
     {
-        return _attributeMap.count(attrib) != 0;
+        return _attributeMap.count(string(attrib)) != 0;
     }
 
     /// Return the value string of the given attribute.  If the given attribute
     /// is not present, then an empty string is returned.
-    const string& getAttribute(const string& attrib) const
+    const string& getAttribute(std::string_view attrib) const
     {
-        StringMap::const_iterator it = _attributeMap.find(attrib);
+        StringMap::const_iterator it = _attributeMap.find(string(attrib));
         return (it != _attributeMap.end()) ? it->second : EMPTY_STRING;
     }
 
@@ -528,7 +528,7 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
     }
 
     /// Remove the given attribute, if present.
-    void removeAttribute(const string& attrib);
+    void removeAttribute(std::string_view attrib);
 
     /// @}
     /// @name Self And Ancestor Elements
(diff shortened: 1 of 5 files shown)


source/MaterialXCore/Element.h
```diff
@@ -479,19 +479,19 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
     /// @{
 
     /// Set the value string of the given attribute.
-    void setAttribute(const string& attrib, const string& value);
+    void setAttribute(std::string_view attrib, const string& value);
```
lgritz, line 482: Why not string_view `value` as well?

ld-kerley: just trying to make as minimal API change as possible - I think there's a good improvement to move to more string_view based API everywhere - but here (And in the other PR) I'm just trying to explore a strategy to avoid global constructor/destructor, moving the const string to string_view - it seemed to make the most sense to change the API here to accept a string_view for the attribute name.

lgritz: I'm a huge fan of string_view.

kwokcb:
I see you have put in safety wrappers for Python (pybind).

This is also necessary for Javascript  (emscripten) bindings since for both the strings are temporary (in temp buffers) which are destroyed after calls and future access will be to invalid memory.

If we do this, I guess there are two ways I can think of handling this: Keep the "safe" API for public API bindings and directly call them, or use wrappers. For either it needs to be simple and made required somehow to avoid possible regressions or future issues.  Maybe write wrapper templates ? 

I'm not sure if build or analysis tools catch this, but AFAIK no public Python/JS can expose string_view directly.

I think either way the performance should not change, but it would be good to regression test this.

source/MaterialXCore/Element.h
ld-kerley: I hear ya :) 

I'm much more familiar with the Shader Gen modules than I am the Core and the requirements there to maintain compatibility.

I do know that string operations from the core are measurable impacts from a performance perspective while loading and querying the MaterialX document, so I to think it would be good for someone to take a closer look.

ld-kerley:
This is why I wanted to post the two possible options - to explore what concerns people maybe have for each approach.  I certainly haven't kicked the tires here at all for either of the bindings - we don't really use them.

(Note : I didn't touch the python bindings - because there was no CI error there - only the JS bindings)

jstone-lucasfilm:
I'm a big fan of the `string_view` approach for this change to MaterialX, and it's an idea that @niklasharrysson and I have been considering for a long while.

The main question in my mind is how we should handle _backwards compatibility_ for our C++ API once the change is made.  To my mind, it seems ideal to make this change _globally_ across the MaterialX codebase, with appropriate support and communication to the developer community, allowing them to navigate the global change in a clear and straightforward way.

Once we have a good strategy for backwards compatibility, the actual code changes should be relatively mechanical, and should make the MaterialX codebase clearer and more efficient overall.

lgritz:
> The main question in my mind is how we should handle _backwards compatibility_ for our C++ API once the change is made.

Since you can pass either a `char*` or a `std::string` to a `string_view` parameter, you should trivially have API compatibility in the sense that downstream source code should not have to change. But of course, it's an ABI/link compatibility break, unless you use overloading to have both `foo(std::string_view)` and `foo(const std::string&)` versions of things, with the latter advertised to disappear an a future release that allows a bigger compatibility break.

PR #2698, opened 2025-11-26
Align writing of generated Slang shaders with GLSL

tdavidovicNV:
This brings GenSlang tests to be in line with GenGlsl, and disables the writing of tested shaders to disk.

The main motivation is that the tests (both Slang and GLSL) actually currently produce invalid code for the `NG_lightcompoundtest` node, which is confusing.

### DIFF
--- a/source/MaterialXGenGlsl/GlslShaderGenerator.cpp
+++ b/source/MaterialXGenGlsl/GlslShaderGenerator.cpp
@@ -621,31 +621,6 @@ void GlslShaderGenerator::emitLightFunctionDefinitions(const ShaderGraph& graph,
     }
 }
 
-void GlslShaderGenerator::toVec4(TypeDesc type, string& variable)
-{
-    if (type.isFloat3())
-    {
-        variable = "vec4(" + variable + ", 1.0)";
-    }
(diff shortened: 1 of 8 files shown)


kwokcb:
Can we add (to Slang) update (GLSL) `addSkipNodeDefs()` tester override to include `ND_lightcompoundtest` ?

tdavidovicNV:
> Can we add (to Slang) update (GLSL) `addSkipNodeDefs()` tester override to include `ND_lightcompoundtest` ?

I am sorry, I am not exactly sure what you mean by those two parentheses.
I've added it to the GenSlang `addSkipNodeDefs()`. Should I add it to GenGlsl as well?

jstone-lucasfilm:
@tdavidovicNV @kwokcb Let's keep this change focused on Slang, and we can address the GLSL exception in a separate PR.  Based on recent discussions on the MaterialX Slack, I'm getting the sense that we'll ultimately want to remove the light compound node entirely, as it's no longer used by any of the teams in our community, so that may end up being the next step for that specific exception in Slang and GLSL render tests.

jstone-lucasfilm (approved):

PR #2711, opened 2025-12-03
Modify OIIOImageLoader CMake 

ld-kerley:
This PR fixes a small error in the CMakeList.txt to set the `MATERIALX_BUILD_OIIO` compile flag.

`${TARGET_NAME}` needs to be used when modifying a target after being declared, this is to ensure the monolithic build operates correctly too.

We also have some internal build systems that are less happy about how the guards are structured - and it turns out the approach take for the `OCIONode` works well - so I'm proposing here we align the `OIIOImageLoader` guard with that one.

Tagging @JGamache-autodesk - because it looks like you touched this most recently.

### DIFF
--- a/source/MaterialXRender/OiioImageLoader.cpp
+++ b/source/MaterialXRender/OiioImageLoader.cpp
@@ -3,6 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+#ifdef MATERIALX_BUILD_OIIO
+
 #include <MaterialXRender/OiioImageLoader.h>
 
 #if defined(_MSC_VER)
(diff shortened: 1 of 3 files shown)


source/MaterialXRender/OiioImageLoader.cpp
```diff
@@ -116,3 +118,5 @@ ImagePtr OiioImageLoader::loadImage(const FilePath& filePath)
 }
 
 MATERIALX_NAMESPACE_END
+
+#endif
```
jstone-lucasfilm, line 122: A super minor note, but there should be a newline at the end of the source file!

jstone-lucasfilm (approved):
Ship it!

PR #2713, opened 2025-12-05
Consolidate shared CMake logic in PyMaterialX

jstone-lucasfilm:
This changelist consolidates shared CMake logic in PyMaterialX, defining a new `mx_add_python_module` function to reduce duplication and improve maintainability.


source/PyMaterialX/CMakeLists.txt
```diff
@@ -35,6 +35,61 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
...
+        "${multiValueArgs}"
+        ${ARGN})
+
+    # Default to .cpp source files if not specified
+    if (NOT args_SOURCE_EXTENSION)
+        set(args_SOURCE_EXTENSION "cpp")
+    endif()
+
+    # Gather source and header files
+    file(GLOB module_source "${CMAKE_CURRENT_SOURCE_DIR}/*.${args_SOURCE_EXTENSION}")
```
ld-kerley, line 56: Might be cleaner - and simpler in the cmake - if we leave the file globing in the respective modules - as we do for the regular MaterialX library modules. 

We wouldn't need the code to handle different file extensions, or excluded files then. 

This makes it easier if there are ever special case file exceptions etc, as well as following the existing convention.

As some point in the future, personally, I think we should follow CMake recommendation, and not glob for the source files at all - but instead list them explicitly - but thats a debate for another day.

source/PyMaterialX/CMakeLists.txt
```diff
@@ -35,6 +35,61 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
     endif()
 endif()
 
+function(mx_add_python_module MODULE_NAME)
```
ld-kerley, line 38: When I added the similar functions for the main MaterialX modules it was suggested these functions should all live in the main CMakeLists.txt file for easy location.  So maybe we should follow that convention.

Personally I would prefer to move all the function to a new `cmake/utils.cmake` file - and simplify the amount of code that live in the main CMakeLists.txt file.  Certalizing all the functions - we might find places we can re-use code even more - perhaps.

source/PyMaterialX/PyMaterialXGenMdl/CMakeLists.txt
```diff
@@ -1,25 +1,3 @@
...
-target_link_libraries(
-    PyMaterialXGenMdl
-    PUBLIC MaterialXGenMdl
-    PRIVATE ${CMAKE_DL_LIBS})
-
-install(TARGETS PyMaterialXGenMdl
-        DESTINATION "${MATERIALX_PYTHON_FOLDER_NAME}")
+mx_add_python_module(PyMaterialXGenMdl
+    LIBRARIES
+        MaterialXGenMdl)
```
JGamache-autodesk, line 3: Possibly add PyMaterialXGenShader for consistency with the other generators?

The simplification makes it easy to spot inconsistencies in this review.

source/PyMaterialX/PyMaterialXRenderOsl/CMakeLists.txt
```diff
@@ -1,26 +1,4 @@
...
-    PUBLIC PyMaterialXRender
-           MaterialXRenderOsl
-    PRIVATE ${CMAKE_DL_LIBS})
-
-install(TARGETS PyMaterialXRenderOsl
-        DESTINATION "${MATERIALX_PYTHON_FOLDER_NAME}")
+mx_add_python_module(PyMaterialXRenderOsl
+    LIBRARIES
+        PyMaterialXRender
+        MaterialXRenderOsl)
```
JGamache-autodesk, line 4: Inconsistent with others, would need:
  PyMaterialXGenOsl
  MaterialXGenOsl

ld-kerley (commented):
Looks like its heading in a good direction to me - sorry I had written this when you posted it but forgot to hit send

jstone-lucasfilm:
Although this change would provide a small consolidation win, I think it's more valuable as a proof of concept than as a merged PR, so I'll close this out and we can return to the idea further in the future.

Issue #2701, opened 2025-11-28
Regression in implementation search logic for node definitions


kwokcb:
## Issue

This is a regression from 1.39.4 to 1.39.5.

1. In 1.39.4: Nodedef.getImplementation() will return the NodeGraph when the assocaition is stored in an implementation. This is the case for both standard surface definitions.
2. In 1.39.5: Nodedef.getImplemetnation() will return the Implementation association instead. It no longer finds the `NodeGraph` defined in the association

## Steps:

This script will supports scanning the libraries either from a install location or a specification location. 
It was used to test a 1.39.4 install and the current 1.39.5 build.

```python
 import MaterialX as mx
import os
import argparse

parser = argparse.ArgumentParser(description="Build MaterialX NodeDef Introspection JSON")
parser.add_argument('-ip', '--input_path', type=str, default='', help="Input path for MaterialX libraries. If not specified will use default search path.")
args = parser.parse_args()

input_path = args.input_path
# Convert opts.library_folder from set to list
library_list = []

path_list = []
if not input_path:

    searchPath = mx.getDefaultDataSearchPath().asString() 
    searchPath = os.path.normpath(searchPath)
    # Append 'libraries' to the search path
    searchPath = mx.FileSearchPath(os.path.join(searchPath, 'libraries'))
    path_list.append(searchPath.asString())

    libraryPath = mx.getDefaultDataSearchPath()
    libraryFolders = mx.getDefaultDataLibraryFolders()

    version = mx.getVersionString()

else:
    libraryFolders = ['libraries']
    searchPath = os.path.normpath(input_path)
    path_list.append(searchPath)
    for lib in libraryFolders:
        path_list.append(os.path.join(searchPath, lib))
    
    libraryPath = mx.FileSearchPath(searchPath)

print("Library folders", libraryFolders)
print("Search paths:", path_list)
print("Library path:", libraryPath.asString())
stdlib = mx.createDocument()
libFiles = mx.loadLibraries(libraryFolders, libraryPath, stdlib)
#for l in libFiles:
#    print("Loaded library file:", l)

# Create main document and import the library document
doc = mx.createDocument()
doc.importLibrary(stdlib)

for nodedef in doc.getNodeDefs():
    nodedef_name = nodedef.getName()

    # Add implementation children if they exist
    implementation = nodedef.getImplementation()
    if implementation:

        # Add nodegraph
        if implementation.isA(mx.NodeGraph):
[...]

PR #2722, opened 2025-12-12
Reinstate prior API behavior for `NodeDef::getImplementation`

ld-kerley:
Reinstate prior API behavior for NodeDef.getImplementation() and introduce new accessor.

Based on feedback from Bernard here #2701.

Add new method to access the implementation elements directly.

### DIFF
--- a/source/MaterialXCore/Definition.h
+++ b/source/MaterialXCore/Definition.h
@@ -138,13 +138,23 @@ class MX_CORE_API NodeDef : public InterfaceElement
     /// @{
 
     /// Return the first implementation for this nodedef, optionally filtered
-    /// by the given target name.
+    /// by the given target name. Resolving any indirection chain in the
+    /// implementations.
     /// @param target An optional target name, which will be used to filter
     ///    the implementations that are considered.
     /// @return An implementation for this nodedef, or an empty shared pointer
     ///    if none was found.  Note that a node implementation may be either
     ///    an Implementation element or a NodeGraph element.
     InterfaceElementPtr getImplementation(const string& target = EMPTY_STRING) const;
+
+    /// Return the first implementation for this nodedef, optionally filtered
+    /// by the given target name.
+    /// @param target An optional target name, which will be used to filter
+    ///    the implementations that are considered.
+    /// @return An implementation for this nodedef, or an empty shared pointer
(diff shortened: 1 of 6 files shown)


source/MaterialXCore/Document.cpp
```diff
@@ -88,12 +89,26 @@ class Document::Cache
                     {
                         if (interface->isA<NodeGraph>())
                         {
-                            implementationMap[interface->getQualifiedName(nodeDefString)].push_back(interface);
+                            implementationDirectMap[interface->getQualifiedName(nodeDefString)].push_back(interface);
```
kwokcb, line 92: The only differing approach I had thought about was to store the direct + indirect in the direct map (which was previously just called the `implementationMap`, and keep the indirect in a new map. 

This works as well and the API docs are explicit about what you get.

ld-kerley: I opt'ed for storing the two complete sets to make it easier on the query side. Otherwise you have to make two separate calls to get to the indirect answer for all cases.

source/MaterialXCore/Definition.h
```diff
@@ -138,14 +138,24 @@ class MX_CORE_API NodeDef : public InterfaceElement
...
     InterfaceElementPtr getImplementation(const string& target = EMPTY_STRING) const;
 
+    /// Return the first implementation for this nodedef, optionally filtered
+    /// by the given target name.
+    /// @param target An optional target name, which will be used to filter
+    ///    the implementations that are considered.
+    /// @return An implementation for this nodedef, or an empty shared pointer
+    ///    if none was found.  Note that a node implementation may be either
+    ///    an Implementation element or a NodeGraph element.
+    InterfaceElementPtr getUnmappedImplementation(const string& target = EMPTY_STRING) const;
```
jstone-lucasfilm, line 157: @ld-kerley @kwokcb Oh, I think I ought to have looked more closely at this part of your PR before approving and merging it.  Is there perhaps a better way to express this than two separate API calls, e.g. as a flag on `getImplementation` that allows the user to select the new "unmapped" behavior?

jstone-lucasfilm: Ah, I have an idea for refactoring and simplifying this fix.  Let me give it a try locally, and if it works out, I'll propose a follow-up PR.

PR #2726, opened 2025-12-13
Simplify `NodeDef::getImplementation`

jstone-lucasfilm:
This changelist builds upon recent work on `NodeDef::getImplementation`, simplifying the method and its underlying caching strategy for clarity, while maintaining all aspects of its behavior.  Two important cases that were considered here are shader generation and LibsToOso.cpp, each of which requires different handling of graph indirection.

### DIFF
--- a/source/MaterialXCore/Definition.cpp
+++ b/source/MaterialXCore/Definition.cpp
@@ -63,22 +63,18 @@ InterfaceElementPtr NodeDef::getImplementation(const string& target, bool resolv
     vector<InterfaceElementPtr> secondary = getDocument()->getMatchingImplementations(getName());
     interfaces.insert(interfaces.end(), secondary.begin(), secondary.end());
 
-    // If requested, follow indirections from implementations to node graphs.
+    // If requested, resolve Implementation elements to their linked NodeGraph elements.
     if (resolveNodeGraph)
     {
         for (size_t i = 0; i < interfaces.size(); ++i)
         {
             ImplementationPtr impl = interfaces[i]->asA<Implementation>();
-            if (impl)
+            if (impl && impl->hasNodeGraph())
             {
-                const string& nodeGraphString = impl->getNodeGraph();
-                if (!nodeGraphString.empty())
+                NodeGraphPtr nodeGraph = getDocument()->getNodeGraph(impl->getNodeGraph());
+                if (nodeGraph)
                 {
-                    NodeGraphPtr nodeGraph = impl->getDocument()->getNodeGraph(nodeGraphString);
-                    if (nodeGraph)
-                    {
-                        interfaces[i] = nodeGraph;
-                    }
+                    interfaces[i] = nodeGraph;
                 }
             }
         }


jstone-lucasfilm:
I'm very interested in your thoughts, @ld-kerley and @kwokcb, and I'm linking this to a handful of related PRs and issues over the course of the development of 1.39.5:

- https://github.com/AcademySoftwareFoundation/MaterialX/pull/2681
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/2701
- https://github.com/AcademySoftwareFoundation/MaterialX/pull/2722

jstone-lucasfilm:
Thanks @JGamache-autodesk, and I'm very interested in  thoughts from the experts on this thread, as well as the broader community.

Based on my own local testing, this PR preserves the exact behaviors required by both shader generation and LibsToOso.cpp, with a simpler interface that aligns with the broader approach of the MaterialX API.

kwokcb:
Note that what was previously a direct cache lookup now needs to search for the nodegraph. As this is for standard surface it would be good to know what if any performance implications there were. An alternative is to precache these only -- my original thought on this, though Lee's approach is also valid though just caches a complete set of references.

jstone-lucasfilm:
@kwokcb Although your intuition about the performance of searching through interfaces is correct on the surface, there are important details that aren't captured by this analysis.

What we're adding in this changelist is one additional loop through _matching interfaces_, where there are already two such loops in the existing function.  These loops are, by design, very efficient, as they are restricted to just those interfaces that are already known to match the given nodedef name, in contrast with the loops through the _document_ that are performed at cache generation time.

So with this change, we're removing one loop through the _document_ at cache generation time, which is a potentially significant time savings.  And we're only adding one loop through _matching interfaces_ at lookup time, which is designed to be our most performant lookup step.

If we were to measure the overall performance of MaterialX graph edits, before and after this change, my expectation would be that the new version would be very slightly _faster_, though the main benefits of the change are focused on developer clarity rather than performance.

source/MaterialXCore/Definition.cpp
```diff
@@ -57,12 +57,42 @@ const string& NodeDef::getType() const
...
-{
-InterfaceElementPtr selectImplementation(const vector<InterfaceElementPtr>& interfaces, const TargetDefPtr targetDef)
+InterfaceElementPtr NodeDef::getImplementation(const string& target, bool graphIndirect) const
 {
+    vector<InterfaceElementPtr> interfaces = getDocument()->getMatchingImplementations(getQualifiedName(getName()));
+    vector<InterfaceElementPtr> secondary = getDocument()->getMatchingImplementations(getName());
+    interfaces.insert(interfaces.end(), secondary.begin(), secondary.end());
+
+    // If requested, follow indirections from implementations to node graphs.
+    if (graphIndirect)
```
ld-kerley: Introducing this extra compute all the time seems counterintuitive to me. 

We call getImplementation quite a bit during shader gen - and we're already trying to speed that up, I'm a little concerned this is a step in the opposite direction.

source/MaterialXCore/Definition.cpp
```diff
@@ -57,12 +57,42 @@ const string& NodeDef::getType() const
     }
 }
 
-namespace
-{
-InterfaceElementPtr selectImplementation(const vector<InterfaceElementPtr>& interfaces, const TargetDefPtr targetDef)
+InterfaceElementPtr NodeDef::getImplementation(const string& target, bool graphIndirect) const
```
ld-kerley: The name `graphIndirect` doesn't really tell me what it's trying to do here. As a boolean, I would hope we can find a clearer name, if we pursue this PR.

source/MaterialXCore/Document.cpp
```diff
@@ -87,28 +86,9 @@ class Document::Cache
...
-                            {
-                                NodeGraphPtr nodeGraph = impl->getDocument()->getNodeGraph(nodeGraphString);
-                                if (nodeGraph)
-                                    implementationIndirectMap[interface->getQualifiedName(nodeDefString)].push_back(nodeGraph);
-                            }
-                            else
-                            {
-                                implementationIndirectMap[interface->getQualifiedName(nodeDefString)].push_back(interface);
-                            }
+                            implementationMap[interface->getQualifiedName(nodeDefString)].push_back(interface);
```
ld-kerley, line 91: Feels better to me to front load any additional work in the cache population call - that we hopefully call as minimally as possible during critical code paths.

Here we're removing work that can be precomputed and moving it to a place that will need to be evaluated many more times during code generation.

ld-kerley (commented):
Overall I'm a little confused why  we're trying to propose this here.  Maybe you could elaborate a little more on the motivations?

"with a simpler interface that aligns with the broader approach of the MaterialX API"

This doesn't really feel like a simpler interface to me, and I'm not really sure what you mean when you say " broader approach of the MaterialX API" - I would really love to dig in to that some more, so we can all get on the same page.  Perhaps my misunderstanding of that approach is why I'm struggling to wrap my head around this change?

jstone-lucasfilm:
This is a great question, @ld-kerley, and I'll do my best to clarify why I believe this PR is a better fit for MaterialX than its predecessors.

In MaterialX, we traditionally use a Document-level cache to store lists of matching ports, nodedefs, and implementations, which can then be searched very efficiently at runtime, saving the far greater time it would take to search the entire document for a match.  The list of matching implementations for a nodedef string, for example, never grows beyond a small handful of entries, while the list of all implementations in the standard library can extend to hundreds or thousands of entries, making this optimization hugely valuable in practice.

In the common cases where _variations_ on a match are required, instead of creating multiple caches for each permutation, we simply iterate over the small list of matching entries at runtime, applying the appropriate filter or modification for the specific use case that is required.

Here's a good example of this pattern in `Node::getNodeDef`, where the returned value supports both "rough matches" and "exact matches" through a boolean named `allowRoughMatch`:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/v1.39.4/source/MaterialXCore/Node.cpp#L82

This approach extends very naturally to the new use case for `NodeDef::getImplementation`, where two variations are required for the differing needs of shader generation and LibsToOsos.cpp.  Rather than creating a _second cache_ for this particular permutation, requiring existing API methods to be split in two, we can simply pass an additional flag to the method, with the appropriate modification being applied to the small list of matching entries at runtime.

I'm completely open to alternative ideas for the name of the `graphIndirect` boolean, and one alternative that I might suggest is `resolveNodeGraph`, which more clearly describes the difference between the two variations that this method supports.

jstone-lucasfilm:
Let's move forward with this simplification, as I believe it's an important one for the MaterialX API and document caching system, and we can continue to refine these ideas in future PRs.

PR #2733, opened 2025-12-23
Use OBJ in OSL render testing

ld-kerley:
Update OSL render testing to use sphere.obj instead of the built in sphere primitive in test shade - thanks to @fpsunflower  for adding OBJ reading support to OSL testrender.

In order to align with GLSL/MSL testing we remove texture coordinate flipping in GLSL/MSL.  After talking to Jonathan, he thinks perhaps we've been rendering the unit tests upside-down since the flip was introduced.  We may decide we need to invert the texture reading to re-instate texture orientation. We will re-evaluate this after this is merged. 

Remove local OSL render test override tangent/bitangent nodes. Now we're using OBJ we get the correct result for the geometry directly from test render.

Update OSL render template instantiation to write absolute paths to make debugging externally easier.


jstone-lucasfilm (commented):
This looks really promising, thanks @ld-kerley, and my sense is that we'll learn a great deal by running a new set of reference quality GLSL/OSL render tests, to see how the latest set compares with previous ones.  I can run these tests on my side when time permits, or you potentially give them a try in your own environment.

One interesting outcome to watch for is whether this change _improves_ the parity of our GLSL/OSL render tests, e.g. by providing more consistent surface normals and tangent frames across the two rendering environments.

chellmuth (approved):
LGTM!

PR #2734, opened 2025-12-23
Add OSL hextiling implementations

jstone-lucasfilm:
This changelist adds implementations of `hextiledimage` and `hextilednormalmap` to OSL, making these nodes available across all shading languages that MaterialX supports.

The new OSL implementations closely follow the existing GLSL versions, with minor improvements to take better advantage of OSL channel indexing and loops.

Additionally, the `standard_surface_onyx_hextiled.mtlx` example has been moved to `Examples/StandardSurface`, making it more visible to users of the MaterialX Web Viewer and other tools.


jstone-lucasfilm:
Here's a new set of reference quality GLSL/OSL render comparisons, including the comparison of `standard_surface_onyx_hextiled.mtlx`, using the latest work in this PR:

[MaterialXRenderTests_12_22_2025_GitHub.pdf](https://github.com/user-attachments/files/24319296/MaterialXRenderTests_12_22_2025_GitHub.pdf)

libraries/stdlib/genosl/mx_hextiledimage_color3.osl
```diff
@@ -0,0 +1,57 @@
...
+    vector cw;
+
+    for (int i = 0; i < 3; i++)
+    {
+        float filter_width = length(tile_data.ddx[i]) + length(tile_data.ddy[i]);
+        c[i] = texture(file.filename, tile_data.coords[i].x, tile_data.coords[i].y,
+                       "swidth", filter_width, "twidth", filter_width,
+                       "missingcolor", default_value,
+                       "swrap", "periodic", "twrap", "periodic"
+#if OSL_VERSION_MAJOR >= 1 && OSL_VERSION_MINOR >= 14
```
lgritz: This is going to fail for a 2.0 release, if there ever is one.

jstone-lucasfilm: Good point, @lgritz.

Here I'm following our pattern for checking the OSL version in [MaterialX texture nodes](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/genosl/mx_image_color3.osl#L19), so it's likely that we should fix this globally in the MaterialX project.

Is there a monotonically increasing OSL version index that we ought to be checking here, or is the best solution just to check these two indices using the traditional three-comparison pattern that handles this case?

lgritz: There's a monotonic one in the C++ header, but these are definitions we directly set up in the C preprocessor as it filters the osl source, so... probably just these symbols.

Yes, should be changed everywhere. No rush, there's no 2.0 imminent, but best to track them all down and fix while it's on your mind and eliminate the "wrong way" from anything people may cut and paste in the future.

libraries/stdlib/genosl/lib/mx_hextile.osl
```diff
@@ -0,0 +1,189 @@
...
+    return (x < 0.5) ? x / (a + 1.0) : (a - x) / (a - 1.0);
+}
+
+struct HextileData
+{
+    vector2 coords[3];
+    vector weights;
+    vector rotations;
+    vector2 ddx[3];
+    vector2 ddy[3];
```
msuzuki-nvidia: I'm not an OSL expert, but I think you don't need to compute the derivatives. dPdu and dPdv may just work.

jstone-lucasfilm: That's a great point, @msuzuki-nvidia, and I've updated the implementation to use the native texture coordinate derivatives in OSL.

msuzuki-nvidia:
LGTM!

libraries/stdlib/genosl/mx_hextiledimage_color4.osl
```diff
@@ -30,17 +30,15 @@ void mx_hextiledimage_color4(
...
     color c[3];
     float alpha[3];
     vector cw;
 
     for (int i = 0; i < 3; i++)
     {
-        float filter_width = length(tile_data.ddx[i]) + length(tile_data.ddy[i]);
         c[i] = texture(file.filename, tile_data.coords[i].x, tile_data.coords[i].y,
                        "alpha", alpha[i],
-                       "swidth", filter_width, "twidth", filter_width,
```
lgritz: I want to point out that in OSL, the optional texture parameters "swidth" and "twidth" DO NOT supply the texture derivatives. Instead, they are *multipliers* of the implicitly computed derivatives of the passed texture coordinates. That is, the defaults for swidth and twidth are 1.0.

lgritz: You can supply explicit derivatives of the coordinates, but NOT THIS WAY.

lgritz: If this is a widespread misunderstanding and not just a one-off mistake, maybe it's worth looking for other places in the MatX OSL code where "swidth" and "twidth" may be getting used incorrectly.

jstone-lucasfilm: Thanks for that catch, @lgritz, and I believe it's just an error on my part in this PR, rather than a broader pattern in the codebase.  If we run into future situations where we need to override texture derivatives, we'll use the `dsdx`, `dtdx`, `dsdy` and `dtdy` arguments.

jstone-lucasfilm:
Here's a new set of reference quality GLSL/OSL render comparisons, using the latest updates to this PR:

[MaterialXRenderTests_01_07_2026_GitHub.pdf](https://github.com/user-attachments/files/24477309/MaterialXRenderTests_01_07_2026_GitHub.pdf)

PR #2735, opened 2025-12-23
Add `-Wshadow` to align with windows compiler warnings

ld-kerley:
Continuing the work proposed in #2385.  I noticed this as more compiler warnings were raised in other work for windows CI.

Add `-Wshadow` to non windows compile path. This aligns us a little closer with the strict Windows path warnings, making it easier to catch build errors locally before hitting CI.

Fixed some shadowing issues in Metal code-path that were missed previously because of mis-matched compiler warnings.

Need to add `-Wno-shadow` for dependency builds (IMGui/NanoGui).

Also adding `-Wno-nontrivial-memcall` because of recent clang warnings building IMGui.


CMakeLists.txt
```diff
@@ -312,7 +312,7 @@ if(MSVC)
         add_compile_options(/WX)
     endif()
 else()
-    add_compile_options(-Wall -Wunused-parameter -Wno-missing-braces)
+    add_compile_options(-Wall -Wunused-parameter -Wno-missing-braces -Wshadow)
```
jstone-lucasfilm: Minor note: Comparing this against other warning lists in MaterialX, we usually state all of the warning _additions_ first, followed by the warning _subtractions_, making it straightforward for readers to find the detail they're looking for in the list.

jstone-lucasfilm (approved):
This looks like a net win to me, thanks @ld-kerley!

PR #2736, opened 2025-12-27
Improvements to thread safety 

jstone-lucasfilm:
- Fix a race condition in the Document cache, where the `valid` flag could be simultaneously written and read from different threads.
- Use a `std::shared_mutex` pattern in the Document cache for greater robustness and clearer guarantees.
- Add an initial Thread Safety section to the Developer Guidelines.

### DIFF
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -5,6 +5,7 @@
 
 #include <MaterialXCore/Document.h>
 
+#include <mutex>
 #include <shared_mutex>
 
 MATERIALX_NAMESPACE_BEGIN


source/MaterialXCore/Document.cpp
```diff
@@ -29,82 +30,120 @@ class Document::Cache
...
-        valid(false)
+        _valid(false)
     {
     }
     ~Cache() = default;
 
+    void setDocument(weak_ptr<Document> document)
+    {
+        _doc = document;
+        invalidate();
```
JGamache-autodesk: What happens if another thread tries to fetch cached data in the miniscule moment after `_doc` is set, but before `invalidate()` has had the time to set `_valid` to false?
I would recommend using a C++17 std::shared_mutex instead for robustness. It allows multiple-reader/single-writer semantics.

jstone-lucasfilm: That's a really good suggestion, @JGamache-autodesk, and I've updated the PR to use this approach.

source/MaterialXCore/Document.cpp
```diff
@@ -29,82 +30,128 @@ class Document::Cache
...
+        }
 
-            // Traverse the document to build a new cache.
-            for (ElementPtr elem : doc.lock()->traverseTree())
-            {
-                const string& nodeName = elem->getAttribute(PortElement::NODE_NAME_ATTRIBUTE);
-                const string& nodeGraphName = elem->getAttribute(PortElement::NODE_GRAPH_ATTRIBUTE);
-                const string& nodeString = elem->getAttribute(NodeDef::NODE_ATTRIBUTE);
-                const string& nodeDefString = elem->getAttribute(InterfaceElement::NODE_DEF_ATTRIBUTE);
+        lock.unlock();
```
JGamache-autodesk, line 82: Thanks! That is the best currently possible with C++ (up to 23). Rechecking the `_valid` condition after re-acquiring the lock makes sure we handle the case where another thread pre-emptively rebuilds in the small window after the shared lock is released, especially since this thread now has to wait for all shared locks to be released before proceeding.

LGTM!

JGamache-autodesk (approved):

PR #2738, opened 2026-01-02
Apply `comparenodedefs.py` to PBR Nodes

jstone-lucasfilm:
This changelist adds a new GitHub CI step to compare the PBR Nodes specification and PBR data libraries using comparenodedefs.py.

Based on the results of the new GitHub CI step, several minor improvements to the PBR and Standard Nodes specifications have been made:

- Update the data type of `absorption` inputs from `color3` to `vector3` to match the data libraries.
- Update the name of `absorption` outputs to match the data libraries.
- Update the data type of the `intensity` input of the `light` node to match the data libraries.
- Simplify the data tables for the `displacement` node, but leave it in the Standard Nodes specification.

Additionally, this changelist adds support for the `or` conjunction in `comparenodedefs.py`, interpreting it in the same fashion as a comma.

### DIFF
--- a/documents/Specification/MaterialX.PBRSpec.md
+++ b/documents/Specification/MaterialX.PBRSpec.md
@@ -439,19 +439,6 @@ If the `edf` input is left unconnected, no emission will occur from the surface.
 |`thin_walled`|Set to true to make the surface thin-walled   |boolean      |false   |
 |`out`        |Output: the computed surface shader           |surfaceshader|        |
 
-<a id="node-displacement"> </a>
-
-### `displacement`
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm:
That's a good point, @niklasharrysson, and @dbsmythe mentioned the same issue in a side thread.

For now, I'd be happy to leave the `displacement` node in its current location within the Standard Nodes specification, and we can consider the move between data libraries to be work for a future PR.

With that approach, our `comparenodedefs.py` output can be considered a reminder that we still need to take this step in the future.

Issue #2680, opened 2025-11-14
Lobe Pruning optimization in MaterialX proper


ppenenko:
## Introduction

This proposal is to port the Lobe Pruning optimization from external renderers to MaterialX core, enabling runtime GPU performance improvements for all MaterialX-based renderers.

## Motivation

The MaterialX community has identified shader optimization as a key priority - see the call for action in issue [#2480](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2480).

Complex physically-based shading models like OpenPBR and Standard Surface generate shader code with numerous BSDF lobes (subsurface, transmission, coat etc.) When material parameters disable certain lobes — for example, by setting `subsurface_weight` to 0 — the generated shader still contains all the evaluation code for those unused features. This leads to:

- **Shader code bloat** which complicates navigation and debugging
- **Pushing texture and uniform limits** on platforms like WebGPU
- **Wasted GPU cycles** evaluating unused BSDFs and their input networks
- **Increased register pressure** reducing GPU occupancy, which severely degrades performance

The above problems are especially acute on mobile platforms and on integrated GPUs.

**Lobe Pruning** addresses these problems by analyzing the shader graph before code generation and eliminating branches that contribute nothing to the final result. When lobe weight parameters are **compile-time constants** (known at shader generation time, analogous to `constexpr` in C++) with values of 0 or 1, entire subgraphs can be safely removed.

### Expected Performance Impact

An implementation of the Lobe Pruning algorithm in OpenUSD PR [#3525](https://github.com/PixarAnimationStudios/OpenUSD/pull/3525) demonstrates **runtime performance improvements of up to x4**. We should expect similar results for the MaterialX implementation.

## Prior Work

There are currently two open-source implementations of the Lobe Pruning optimization by @JGamache-autodesk, which have been battle-tested and demonstrate significant performance benefits.

### Core Algorithm

Both implementations share the same optimization algorithm with three pruning rules:

1. **Mix nodes**: If the mix factor is 0 or 1, forward one input and delete the node
2. **Multiply nodes**: If either input is 0, propagate the zero upstream and delete the node
3. **BSDF nodes**: If a weight parameter is 0, substitute with a no-op "dark" BSDF and delete the original node

[...]

PR #2742, opened 2026-01-08
Add Perfetto tracing and a test output directory option

ppenenko:
## Summary

This PR adds a new `MaterialXTrace` module with optional Perfetto tracing infrastructure, plus a configurable test output directory, laying the groundwork for performance analysis and optimization work.

## Background and motivation

MaterialX doesn't currently offer a tracing/instrumentation infrastructure sufficient for the following goals:

- Profiling shader generation and optimization passes
- Measuring the impact of proposed optimizations (like those in #2499), and in particular per-material measurements
- Correlating MaterialX performance with USD pipeline traces
- Generating detailed flame graphs for performance analysis

In comparison, OpenUSD supports tracing in the [Chrome Tracing](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview) format—the original format used by Chrome DevTools, viewable via `chrome://tracing`. It's JSON-based, and OpenUSD implements serialization without external dependencies.

[Perfetto](https://perfetto.dev/) is the successor to Chrome Tracing which relies on a more optimal binary Protobuf-based serialization format. https://ui.perfetto.dev/ is capable of opening both Perfetto captures and legacy JSON Chrome Tracing captures. It has better performance than `chrome://tracing`, can open larger captures and supports more advanced analysis workflows, e.g. SQL queries.

Perfetto UI showing MaterialX trace

I've chosen Perfetto for the implementation in this PR mostly because of the format's superior performance. For context, using new scene index workflows with Hydra Storm, it's possible to generate multi-gigabyte Chrome Tracing captures that are too big to load in either profiler UI. The implementation relies on the [Perfetto SDK](https://perfetto.dev/docs/instrumentation/tracing-sdk).

At the same time, the design remains open for a potential integration of MaterialX tracing with USD tracing, by abstracting out the tracing implementation behind an interface. Such an integration would make it possible to analyze MaterialX shader gen performance in Hydra Storm holistically, in the same profiler UI session or script.

The Lobe Pruning optimization effort (#2680) is the immediate application for this infrastructure.

### 1. Perfetto Tracing (`MATERIALX_BUILD_TRACING`)

- **Dedicated `MaterialXTrace` module** – keeps tracing infrastructure separate from `MaterialXCore`
- **Zero overhead when disabled** – entire module excluded from build when `MATERIALX_BUILD_TRACING=OFF`
- **Abstract sink interface** (`Tracing::Sink`) – designed for future USD `TraceCollector` integration
- **Enum-based categories** (`Tracing::Category`) for type safety and efficient dispatch:
  - `Render` – rendering operations
  - `ShaderGen` – shader generation
  - `Optimize` – optimization passes
  - `Material` – material identity markers
- **Template-based scope guard** (`Tracing::Scope<Category>`) – zero stack overhead for category storage
- **RAII helpers** (`MX_TRACE_FUNCTION`, `MX_TRACE_SCOPE`) and `Dispatcher::ShutdownGuard`
- **Instrumentation** in shader generation tests (`GenShaderUtil.cpp`) and render tests (`RenderGlsl.cpp`) demonstrating per-material trace markers
- **Perfetto SDK** fetched via CMake `FetchContent` (v43.0)
- **CI integration** – extended builds (nightly/manual) enable tracing and upload `.perfetto-trace` artifacts

### 2. Test Output Directory (`outputDirectory` in `_options.mtlx`)

- Redirects all test artifacts (logs, images, shaders, traces) to a user-specified directory
- Prevents artifacts from different test runs overwriting each other
- Prints effective output path at end of test run for easy IDE navigation

### 3. Runtime Tracing Toggle (`enableTracing` in `_options.mtlx`)

- Enable/disable tracing without rebuilding (requires `MATERIALX_BUILD_TRACING=ON` at build time)
- Default is `false` to avoid overhead when not profiling

### Enable Tracing (Build Time)
```bash
cmake -DMATERIALX_BUILD_TRACING=ON ...
```

### Enable Tracing (Runtime)
In `resources/Materials/TestSuite/_options.mtlx`:
```xml
<input name="enableTracing" type="boolean" value="true" />
```

### Configure Output Directory
```xml
<input name="outputDirectory" type="string" value="/path/to/test/output" />
```

### Analyze Traces
Open `.perfetto-trace` files at https://ui.perfetto.dev

For CI builds, download trace artifacts from the GitHub Actions run page (extended builds only).

## Related

- #2680 – Lobe Pruning optimization proposal
- #2499 – Shader Generator ND_mix_bsdf optimization (@ld-kerley)

---

```
Assisted-by: Claude (Anthropic) via Cursor IDE
Signed-off-by: Pavlo Penenko <pavlo.penenko@autodesk.com>
```

### DIFF
--- a/libraries/README.md
+++ b/libraries/README.md
@@ -75,5 +75,4 @@ This folder contains the standard data libraries for MaterialX, providing declar
     - point, directional, spot
 - Shader generation does not currently support:
     - `displacementshader` and `volumeshader` nodes for hardware shading targets (GLSL, MSL).
-    - `hextiledimage` and `hextilednormalmap` for OSL and MDL.
     - `blur` the implementation passes through `in` unmodified in all shading languages.
--- a/libraries/stdlib/genosl/lib/mx_hextile.osl
+++ b/libraries/stdlib/genosl/lib/mx_hextile.osl
@@ -1,0 +1,176 @@
+// https://www.shadertoy.com/view/4djSRW
+vector2 mx_hextile_hash(vector2 p)
+{
+    vector pp3 = fmod(vector(p.x, p.y, p.x) * vector(0.1031, 0.1030, 0.0973), 1.0);
+    pp3 += dot(pp3, vector(pp3[1], pp3[2], pp3[0]) + 33.33);
+    return fmod(vector2(pp3[0] + pp3[1], pp3[0] + pp3[2]) * vector2(pp3[2], pp3[1]), 1.0);
+}
+
+// Christophe Schlick. "Fast Alternatives to Perlin's Bias and Gain Functions".
+// In Graphics Gems IV, Morgan Kaufmann, 1994, pages 401-403.
+// https://dept-info.labri.fr/~schlick/DOC/gem2.html
+float mx_schlick_gain(float x, float r)
+{
+    float rr = clamp(r, 0.001, 0.999);  // to avoid glitch
+    float a = (1.0 / rr - 2.0) * (1.0 - 2.0 * x);
+    return (x < 0.5) ? x / (a + 1.0) : (a - x) / (a - 1.0);
+}
+
+struct HextileData
+{
+    vector2 coords[3];
+    vector weights;
+    vector rotations;
+};
+
+// Helper function to compute blend weights with optional falloff
+vector mx_hextile_compute_blend_weights(vector luminance_weights, vector tile_weights, float falloff)
+{
+    vector w = luminance_weights * pow(tile_weights, 7.0);
+    w /= (w[0] + w[1] + w[2]);
+
+    if (falloff != 0.5)
+    {
+        w[0] = mx_schlick_gain(w[0], falloff);
+        w[1] = mx_schlick_gain(w[1], falloff);
+        w[2] = mx_schlick_gain(w[2], falloff);
+        w /= (w[0] + w[1] + w[2]);
+    }
+    return w;
+}
+
+// Morten S. Mikkelsen, Practical Real-Time Hex-Tiling, Journal of Computer Graphics
+// Techniques (JCGT), vol. 11, no. 2, 77-94, 2022
+// http://jcgt.org/published/0011/03/05/
+HextileData mx_hextile_coord(
+    vector2 coord,
+    float rotation,
+    vector2 rotation_range,
+    float scale,
+    vector2 scale_range,
+    float offset,
+    vector2 offset_range)
+{
+    float sqrt3_2 = sqrt(3.0) * 2.0;
+
+    // Scale coord to maintain the original fit
+    vector2 st = coord * sqrt3_2;
+
+    // Skew input space into simplex triangle grid
+    // (1, 0, -tan(30), 2*tan(30))
+    vector2 st_skewed = vector2(st.x + st.y * -0.57735027, st.y * 1.15470054);
+
+    // Barycentric weights
+    vector2 st_frac = fmod(st_skewed, 1.0);
+    // Handle negative coordinates properly
+    if (st_frac.x < 0.0) st_frac.x += 1.0;
+    if (st_frac.y < 0.0) st_frac.y += 1.0;
+
+    vector temp = vector(st_frac.x, st_frac.y, 0.0);
+    temp[2] = 1.0 - temp[0] - temp[1];
+
+    float s = step(0.0, -temp[2]);
+    float s2 = 2.0 * s - 1.0;
+
+    float w1 = -temp[2] * s2;
+    float w2 = s - temp[1] * s2;
+    float w3 = s - temp[0] * s2;
+
+    // Vertex IDs
+    int base_id_x = int(floor(st_skewed.x));
+    int base_id_y = int(floor(st_skewed.y));
+    int si = int(s);
+    int id1_x = base_id_x + si;
+    int id1_y = base_id_y + si;
+    int id2_x = base_id_x + si;
+    int id2_y = base_id_y + 1 - si;
+    int id3_x = base_id_x + 1 - si;
+    int id3_y = base_id_y + si;
+
+    // Vertex IDs as array for looping
+    int id_x[3] = { id1_x, id2_x, id3_x };
+    int id_y[3] = { id1_y, id2_y, id3_y };
+
+    // Tile centers, random values, and offsets
+    vector2 ctr[3], rand[3], offsets[3];
+    vector2 seed_offset = vector2(0.12345, 0.12345);  // To avoid some zeros
+    vector2 rr = vector2(radians(rotation_range.x), radians(rotation_range.y));
+    vector rotations, scales;
+
+    for (int i = 0; i < 3; i++)
+    {
+        ctr[i] = vector2(float(id_x[i]) + float(id_y[i]) * 0.5, float(id_y[i]) / 1.15470054) / sqrt3_2;
+        rand[i] = mx_hextile_hash(vector2(float(id_x[i]), float(id_y[i])) + seed_offset);
+        rotations[i] = mix(rr.x, rr.y, rand[i].x * rotation);
+        scales[i] = mix(1.0, mix(scale_range.x, scale_range.y, rand[i].y), scale);
+        offsets[i] = vector2(
+            mix(offset_range.x, offset_range.y, rand[i].x * offset),
+            mix(offset_range.x, offset_range.y, rand[i].y * offset));
+    }
+
+    vector sin_r = sin(rotations);
+    vector cos_r = cos(rotations);
+
+    HextileData tile_data;
+    tile_data.weights = vector(w1, w2, w3);
+    tile_data.rotations = rotations;
+
+    // Transform coordinates for each tile
+    for (int i = 0; i < 3; i++)
+    {
+        vector2 d = coord - ctr[i];
+        float c = cos_r[i];
+        float s = sin_r[i];
+        float sc = scales[i];
+
+        tile_data.coords[i] = vector2(
+            (d.x * c - d.y * s) / sc + ctr[i].x + offsets[i].x,
+            (d.x * s + d.y * c) / sc + ctr[i].y + offsets[i].y);
+    }
+
+    return tile_data;
+}
+
+// Blend 3 normals by blending the gradients
+// Morten S. Mikkelsen, Surface Gradient-Based Bump Mapping Framework, Journal of
+// Computer Graphics Techniques (JCGT), vol. 9, no. 3, 60-90, 2020
+// http://jcgt.org/published/0009/03/04/
+vector mx_normals_to_gradient(vector N, vector Np)
+{
+    float d = dot(N, Np);
+    vector g = (d * N - Np) / max(1e-8, abs(d));
+    return g;
+}
+
+vector mx_gradient_blend_3_normals(vector N, vector N1, float N1_weight, vector N2, float N2_weight, vector N3, float N3_weight)
+{
+    float w1 = clamp(N1_weight, 0.0, 1.0);
+    float w2 = clamp(N2_weight, 0.0, 1.0);
+    float w3 = clamp(N3_weight, 0.0, 1.0);
+
+    vector g1 = mx_normals_to_gradient(N, N1);
+    vector g2 = mx_normals_to_gradient(N, N2);
+    vector g3 = mx_normals_to_gradient(N, N3);
+
+    // Blend
+    vector gg = w1 * g1 + w2 * g2 + w3 * g3;
+
+    // Gradient to normal
+    return normalize(N - gg);
+}
+
+// Axis-angle rotation matrix
+matrix mx_axis_rotation_matrix(vector axis, float angle)
+{
+    float s = sin(angle);
+    float c = cos(angle);
+    float omc = 1.0 - c;
+    vector a = axis;
+
+    return matrix(
+        a[0]*a[0]*omc + c,       a[0]*a[1]*omc - a[2]*s,  a[0]*a[2]*omc + a[1]*s,  0.0,
+        a[1]*a[0]*omc + a[2]*s,  a[1]*a[1]*omc + c,       a[1]*a[2]*omc - a[0]*s,  0.0,
+        a[2]*a[0]*omc - a[1]*s,  a[2]*a[1]*omc + a[0]*s,  a[2]*a[2]*omc + c,       0.0,
+        0.0,                     0.0,                     0.0,                     1.0
+    );
+}
--- a/libraries/stdlib/genosl/mx_hextiledimage_color3.osl
+++ b/libraries/stdlib/genosl/mx_hextiledimage_color3.osl
@@ -1,0 +1,55 @@
+#include "lib/$fileTransformUv"
+#include "lib/mx_hextile.osl"
+
+// Morten S. Mikkelsen, Practical Real-Time Hex-Tiling, Journal of Computer Graphics
+// Techniques (JCGT), vol. 11, no. 2, 77-94, 2022
+// http://jcgt.org/published/0011/03/05/
+void mx_hextiledimage_color3(
+    textureresource file,
+    color default_value,
+    vector2 texcoord,
+    vector2 tiling,
+    float rotation,
+    vector2 rotationrange,
+    float scale,
+    vector2 scalerange,
+    float offset,
+    vector2 offsetrange,
+    float falloff,
+    float falloffcontrast,
+    color lumacoeffs,
+    output color result)
+{
+    if (file.filename == "")
+    {
+        result = default_value;
+        return;
+    }
+
+    vector2 coord = mx_transform_uv(texcoord) * tiling;
+
+    HextileData tile_data = mx_hextile_coord(coord, rotation, rotationrange, scale, scalerange, offset, offsetrange);
+
+    // Sample textures for each tile
+    color c[3];
+    vector cw;
+
+    for (int i = 0; i < 3; i++)
+    {
+        c[i] = texture(file.filename, tile_data.coords[i].x, tile_data.coords[i].y,
+                       "missingcolor", default_value,
+                       "swrap", "periodic", "twrap", "periodic"
+#if (OSL_VERSION_MAJOR == 1 && OSL_VERSION_MINOR >= 14) || (OSL_VERSION_MAJOR > 1)
+                       , "colorspace", file.colorspace
+#endif
+                       );
+        cw[i] = dot(vector(c[i]), vector(lumacoeffs));
+    }
+
+    // Luminance-based blend weights
+    cw = mix(vector(1.0), cw, vector(falloffcontrast));
+    vector w = mx_hextile_compute_blend_weights(cw, tile_data.weights, falloff);
+
+    // Blend colors
+    result = w[0] * c[0] + w[1] * c[1] + w[2] * c[2];
+}
--- a/libraries/stdlib/genosl/mx_hextiledimage_color4.osl
+++ b/libraries/stdlib/genosl/mx_hextiledimage_color4.osl
@@ -1,0 +1,65 @@
+#include "lib/$fileTransformUv"
+#include "lib/mx_hextile.osl"
+
+// Morten S. Mikkelsen, Practical Real-Time Hex-Tiling, Journal of Computer Graphics
+// Techniques (JCGT), vol. 11, no. 2, 77-94, 2022
+// http://jcgt.org/published/0011/03/05/
+void mx_hextiledimage_color4(
+    textureresource file,
+    color4 default_value,
+    vector2 texcoord,
+    vector2 tiling,
+    float rotation,
+    vector2 rotationrange,
+    float scale,
+    vector2 scalerange,
+    float offset,
+    vector2 offsetrange,
+    float falloff,
+    float falloffcontrast,
+    color lumacoeffs,
+    output color4 result)
+{
+    if (file.filename == "")
+    {
+        result = default_value;
+        return;
+    }
+
+    vector2 coord = mx_transform_uv(texcoord) * tiling;
+
+    HextileData tile_data = mx_hextile_coord(coord, rotation, rotationrange, scale, scalerange, offset, offsetrange);
+
+    // Sample textures for each tile
+    color c[3];
+    float alpha[3];
+    vector cw;
+
+    for (int i = 0; i < 3; i++)
+    {
+        c[i] = texture(file.filename, tile_data.coords[i].x, tile_data.coords[i].y,
+                       "alpha", alpha[i],
+                       "missingcolor", default_value.rgb, "missingalpha", default_value.a,
+                       "swrap", "periodic", "twrap", "periodic"
+#if (OSL_VERSION_MAJOR == 1 && OSL_VERSION_MINOR >= 14) || (OSL_VERSION_MAJOR > 1)
+                       , "colorspace", file.colorspace
+#endif
+                       );
+        cw[i] = dot(vector(c[i]), vector(lumacoeffs));
+    }
+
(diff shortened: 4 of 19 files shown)


source/MaterialXCore/PerfettoSink.cpp
```diff
@@ -0,0 +1,173 @@
+//
+// Copyright Contributors to the MaterialX Project
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <MaterialXCore/PerfettoSink.h>
```
jstone-lucasfilm: Usually we think of `MaterialXCore` as the minimal, most lightweight core library for MaterialX clients, and I wonder if the proposed Perfetto tracing functionality would be better suited for `MaterialXTest`, which currently contains the early beginnings of a `Benchmark Test` framework:

https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/source/MaterialXTest#benchmark-tests

I'm additionally CC'ing @ashwinbhat, who worked on the early benchmark system for MaterialX, to see if there might be a useful harmonization across these different generations of ideas for effectively profiling MaterialX.

ppenenko: While the new tracing mechanism is only used in tests in this PR, it's applicable to to all MaterialX libraries. `MaterialXView` and `MaterialXGraphEditor` could definitely benefit from this for analyzing the interactive performance. The design is open to adding a USD tracing integration, which would make it possible to see MaterialX events together with USD events together in one capture. For comparison, tracing is implemented in `base/trace`, so I would argue that `MaterialXCode` is the closest match.

ppenenko: As for the existing Catch2-based benchmarking, it's limited compared to tracing, and definitely not sufficient for profiling per-frame and per-material performance.

Both Catch2 benchmarking and the frame time stat in `MaterialXView` do some averaging of the measurements at run time, but the tracing approach allows us to record what actually happened at run time at an arbitrary granularity, which enables flexible analysis offline, e.g. visually or with SQL in the Perfetto UI, or with a Python script calling into https://perfetto.dev/docs/analysis/trace-processor

jstone-lucasfilm:
@ppenenko I have a few additional questions, as I'd like to better understand the advantages of the approach that you're proposing.

Historically, we've used _sampling profilers_ for performance analysis of the MaterialX project, and here's an example profile of the MaterialX Viewer using the built-in sampling profiler in Visual Studio:

MaterialXView_SamplingProfiler

Can you give a sense of the advantages provided by the _instrumented profiling_ in your PR, in order to better understand the benefits that might be associated with this additional infrastructure in MaterialX?

And when profiling the results of shader generation in hardware shading languages (e.g. GLSL, MSL, Slang), I would naturally think of GPU profilers such as RenderDoc as the natural approach for recording and comparing performance data.

Is your proposed approach in this PR intended as a replacement for this form of traditional GPU profiling, or is it purely focused on CPU profiling?  If it supports GPU profiling, what are its benefits over RenderDoc?

ashwinbhat:
The current catch2 based benchmark test in MaterialX is very simple and having a more robust tracing system that integrates well in libraries/applications that MaterialX is used in will be valuable. For example when MaterialX is integrated into USD we can enable USDTrace and MaterialX tracing to get captures with more detailed view. 

@jstone-lucasfilm would this PR be more acceptable if we create a new MaterialXTrace module that brings in Perfetto. So this would be similar to USD tracing but more simpler. [Here](https://openusd.org/dev/api/trace_page_front.html) is some additional info about USD Trace https://github.com/PixarAnimationStudios/OpenUSD/tree/dev/pxr/base/trace

jstone-lucasfilm:
@ashwinbhat For adding an instrumented trace concept to MaterialX, I agree a `MaterialXTrace` module would be better than adding new source code to `MaterialXCore`, and I'm open to this approach.

I'd still like to learn more, though, about the advantages offered by _instrumented profilers_ such as Perfetto over the existing _sampled profilers_ we've used previously in MaterialX performance evaluation, which don't require additional code to be inserted into applications at all.  So far, I haven't encountered cases where the precision of sampled profilers was insufficient to fully understand the impact of a proposed optimization or codebase change, and I'd love to learn more about the cases that other teams have run into where sampling profilers were not up to the task.

Also, I'd like to better understand the relationship of Perfetto to _GPU profiling_, which is just as important (or perhaps more important) in evaluating the impact of a ShaderGen change on the real-time rendering performance of MaterialX content.

ppenenko:
@jstone-lucasfilm I agree that sampling profilers are of course useful tools in their own right, but instrumentation adds a different useful perspective. Some advantages:

1. Most importantly, we can instrument with arbitrary strings that matter to us, e.g. material and shader names, not just C++ symbols.
2. Cross-platform, while sampling profilers are typically not. Same trace format recorded on any platform.
3. Sampling profilers can miss, under-represent or over-represent events due to temporal aliasing.

ppenenko:
Some benefits of Perfetto specifically:

1. Captures can be analyzed in a web browser any platform, without the need to install any specialized tools. Thus, easy to share with anyone.
2. Rich analysis functionality with SQL and a Python API.
3. Potential for IPC backends.
4. Overall, feature-rich and battle-tested by Chrome.

ppenenko:
Good point about CPU vs. GPU traces @jstone-lucasfilm . AFAIK Perfetto doesn't offer facilities for GPU profiling out of the box, but the application itself can perform GPU perf queries and report the results via Perfetto. In fact, this is what I'm working on in another branch. At the same time, CPU and GPU events can be analyzed in the same UI in Perfetto, which is convenient. Besides, MaterialX has no CPU-side tracing system anyway, so integrating Perfetto will help many more CPU performance experiments in the future.

Of course, Renderdoc would offer a much more detailed picture of a GPU frame, but it's designed for profiling at a lower level. Its captures would contain GPU resources, which would have a much higher overhead (capture file size) than the per-material measurements I'm adding with Perfetto. So, I would suggest identifying the materials that are slow to render or benefit from a certain optimization by analyzing the existing render test across 180+ materials with Perfetto (I can already confirm that that is working well) and then, if necessary, profiling specific materials in Renderdoc, Nsight etc.

By the way, if we wanted to see material and shader names in Renderdoc captures, we'd have to instrument our code either way. So the instrumentation macros I'm proposing could have a Renderdoc implementation in the future.

ppenenko:
Great idea about `MaterialXTrace` @ashwinbhat - it does feel better encapsulated. It would be a very small library though. Anyway, I'll refactor this aspect.

jstone-lucasfilm:
@ppenenko @ashwinbhat Pivoting for a moment, let's focus on the "steelman" arguments for why a `MaterialXTrace` module providing instrumented build support might be a really _good_ idea.

Would `MaterialXTrace` allow us to integrate an instrumented run of `MaterialXTest` into the GitHub CI for MaterialX, with the instrumented build support providing consistent profile results from run to run, independent of the actual time each run of MaterialXTest took on the GitHub CI virtual machine?

That would be a really compelling argument, if true, and it's something that a sampling profiler couldn't hope to provide, as the distribution of time samples are profoundly dependent on the performance of the virtual machines on which our GitHub CI workflows are run.

ppenenko:
@jstone-lucasfilm Perfetto tracing is completely orthogonal to how a VM's configuration and load affect timings - it just records wall-clock times, so unfortunately it wouldn't improve on that. When testing on the same machine and avoiding heavy parallel loads (e.g. running Maya in parallel), the timings should be reliable. E.g. I can measure the expected speedups of @ld-kerley 's #2499 reliably.

But of course, we could record traces in CI/CD. At the very least, these would give us a kind of a rich log, with meaningful events stacked and grouped by thread. We can be sure, for example, that this or that shader graph optimization pass has been applied to the given material. And optimization passes #2499 should show similar _relative_ improvements across different configurations.

A related use case would be if a user experiences performance issues with MaterialX - they can easily record a capture and share it with the maintainers. And, as I mentioned, it could one day be a USD trace including MaterialX events.

jstone-lucasfilm:
It does sound interesting to integrate MaterialXTrace into our GitHub CI in the future, even if it doesn't produce perfectly consistent results between runs.

It seems slightly surprising that you would want to use Perfetto to measure _GPU timing improvements_ from shader generation optimizations, e.g. the hardware shader optimizations found in #2499.

Am I wrong to think that this category of optimization would normally be evaluated with either RenderDoc or a real-time, smoothed frame timer like the one we use in MaterialXView?

ppenenko:
@jstone-lucasfilm could we please separate the concerns in this code review a bit?

First of all, can we all agree that _a_ tracing mechanism would be an essential addition to MaterialX? I stated the motivations above, and the most immediate and straightforward one is to to measure how long a particular operation took for a particular MaterialX entity - material, node, shader - which is not identified uniquely by a C++ symbol, and therefore can't be captured by a sampling profiler. Let's set GPU timings aside for a moment, because there are a few types of CPU events that we care about in the context of shader graph optimizations: shader codegen timings and shader compilation timings. So even if we disagree on GPU profiling, I would argue that such a mechanism is a must-have. Chrome and USD corroborate this argument.

If we do agree on the above, let's discuss if Perfetto is the right implementation choice. I would argue that it is, due to its exceptional maturity and widespread adoption.

As for GPU instrumentation - my main argument is that if we have a tracing mechanism anyway, for all the other reasons, recording GPU timings there is simple and convenient. Taking the frame duration with OpenGL queries is about a dozen lines of C++, and then the result can be recorded via Perfetto.

As for Renderdoc - it's not well-suited for recording the durations of hundreds of frames because:
1. Its number one use case is debugging and profiling a __single__ frame.
2. It records complete graphics API (e.g. OpenGL) call traces and GPU resources, for every frame - something we don't need for our kind of test and something that would amount to huge amounts of data across hundreds of frames.
3. Its main use case if frame debugging, not frame profiling. For profiling, GPU hardware-specific tools (Nsight etc.) are more useful but they're even less cross-platform.
4. GPU frame debuggers/profilers in general are known for compatibility issues across platforms, graphics APIs and application usage patterns.
5. For measuring the duration of a frame or a draw call, Renderdoc would use the same graphics API calls that we can use - it doesn't have special hooks into the driver, unlike HW vendor tools. So our measurements can be as precise.

Now, regarding MaterialXView.
1. It can only measure one material at a time.
2. For some reason, the reported frame duration is truncated to integer milliseconds. E.g. 1.9f would be truncated to 1.
3. It doesn't measure GPU durations. It also doesn't seem to do `glFinish` or `glFlush`, so even CPU frame durations are questionable. The only CPU/GPU synchronization it seems to have is spamming the video driver with frames until it starts blocking.
4. It does do averaging, but we don't have to do it at run time. With a recorded trace, we can examine multiple events per material after the fact, either in a GUI or with Python. Perfetto UI can do all sorts of statistic analyses, and a Python script can bring the data into Pandas and Matplotlib. Having the complete data offline makes it possible to analyze distributions as well - e.g. when the first frame per material is much longer due to shader compilation.
5. Running an interactive app and jotting down a single milliseconds value is less reproducible and shareable than running a test suite which records timings in nanosecond precision. The file that can be shared with others, the test suite can be run by others.

jstone-lucasfilm:
@ppenenko I'm definitely interested in the benefits offered by an instrumented tracing system like Perfetto, and you've made a good case for how it might complement the range of non-instrumented CPU and GPU profiling systems that we regularly use on the MaterialX project.

In my very early days writing real-time rendering systems for games, I can recall situations where tracing a very small component of a large scene render was critical, and as USD and MaterialX become more common components of real-time engines, I can see the value in making it straightforward for both USD and MaterialX to report this category of profiling data to clients.

For your current PR, in addition to moving the new functionality into a `MaterialXTrace` module, do you think that it would be possible to upload a Perfetto trace of `MaterialXTest` in one of our GitHub CI jobs, following the pattern that we currently use to generate code coverage and render test results in our GitHub CI?  If build times would be negatively affected by this change, then it should be fine to run the Perfetto trace in our nightly build, in the same fashion that we currently do for MDL shader testing.

ppenenko:
Makes sense, @jstone-lucasfilm, after all Perfetto should be tested in at least some CI/CD configurations. It's not even built by default. Ideally, we should have CI/CD build both with and without it. Do you think it would be reasonable to build with it disabled in most CI/CD configurations but to build with it enabled and actually collect some traces in nightly builds?

jstone-lucasfilm:
@ppenenko Yes, that sounds exactly right to me.  Perfetto should ideally be disabled by default in MaterialX, and we can manually enable it in a single job within the nightly build of our GitHub CI.  This should have a minimal impact on build times, while giving us our first running history of instrumented profiles of `MaterialXTest`, which can be compared to new profiles when we're analyzing a new optimization or performance regression.

ppenenko:
@jstone-lucasfilm I have a question: how best to enable the run-time switch for tracing in nightly CI/CD? I added a tracing option for tests in [resources\Materials\TestSuite\_options.mtlx](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2742/changes#diff-a2d63b2fbd0bbd70019003edf2b5e1d9020a6e1ca09354f5a02e91a79c337e78) - that's a source-controlled file which is installed as a part of the MaterialX build. But how can we edit one of these settings programmatically, i.e. from GHA?

jstone-lucasfilm:
@ppenenko We don't yet have functionality to override properties of the TestSuite options dynamically, and although that's a great idea for the future, I wouldn't want to burden your PR with adding that functionality.

For now, would it be reasonable to write the Perfetto output to a subfolder of the current working directory when no location has been set in the TestSuite options, and we can upload the profile data from that subfolder in GitHub CI?

ppenenko:
@jstone-lucasfilm sure, the directory isn't an issue - currently, if it's not specified, the capture would just be saved side-by-side with the logs (in CWD). I was rather thinking about the [run-time on/off switch](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2742/changes#diff-a2d63b2fbd0bbd70019003edf2b5e1d9020a6e1ca09354f5a02e91a79c337e78R86) It's also off by default. However, considering that the __build-time__ switch is __off__ by default, maybe we should set make the __run-time__ switch __on__ by default?

Going forward, maybe we could make the path to `_options.mtlx` configurable via an env var or CLI argument? Right now, it's inconvenient to make changes to that file because its location is fixed and it's source-controlled - so there's a risk of accidentally losing or committing local options.

jstone-lucasfilm:
That approach would be fine for now, @ppenenko, or alternatively you could make the test suite profiling switch a CMake build option rather than a property of the options text file.

.github/workflows/main.yml
```diff
@@ -206,6 +206,8 @@ jobs:
           if [ "${{ matrix.extended_build_mdl_sdk }}" == "ON" -a "${{ runner.os }}" == "Windows" ]; then
             EXTENDED_BUILD_CONFIG="$EXTENDED_BUILD_CONFIG -DMATERIALX_MDL_SDK_DIR=C:/vcpkg/installed/x64-windows"
           fi
+          # Enable Perfetto tracing
+          EXTENDED_BUILD_CONFIG="$EXTENDED_BUILD_CONFIG -DMATERIALX_BUILD_TRACING=ON"
```
ppenenko: FYI @jstone-lucasfilm : enabling for all extended builds.

ppenenko:
Triggered an extended build at https://github.com/autodesk-forks/MaterialX/actions/runs/21048543911 - let's see if we can find the trace in its artifacts.

ppenenko:
Extended build fails to compile, I'll investigate.

.github/workflows/main.yml
```diff
@@ -277,7 +279,7 @@ jobs:
         else
           brew install cppcheck
         fi
-        cppcheck --project=build/compile_commands.json --error-exitcode=1 --suppress=normalCheckLevelMaxBranches --suppress=*:*/External/* --suppress=*:*/NanoGUI/*
+        cppcheck --project=build/compile_commands.json --error-exitcode=1 --suppress=normalCheckLevelMaxBranches --suppress=*:*/External/* --suppress=*:*/NanoGUI/* --suppress=*:*perfetto*
```
ppenenko, line 284: Excluding Perfetto from cppcheck because it's 3rd-party code.

source/MaterialXTrace/Tracing.h
```diff
@@ -0,0 +1,255 @@
...
+/// @param outputPath Path to write the trace file when the sink is destroyed
+/// @param bufferSizeKb Size of the trace buffer in KB (default 32MB)
+/// @return A unique_ptr to the Perfetto sink
+///
+/// Usage:
+///   Dispatcher::getInstance().setSink(createPerfettoSink("trace.perfetto-trace"));
+///   Dispatcher::ShutdownGuard guard;
+///   // ... traced work ...
+///   // guard destructor writes the trace file
+MX_TRACE_API std::unique_ptr<Sink> createPerfettoSink(
```
ppenenko, line 203: Not exposing Perfetto across DLL boundaries.

ppenenko:
Refactoring done - `MaterialXTrace` is a separate module. GLSL render tests aren't enabled in CI/CD, so I added instrumentation in codegen tests. An extended build with trace artifacts can be found at https://github.com/autodesk-forks/MaterialX/actions/runs/21080475833/job/60632726674 A Windows step is still running, taking a while to install the MDL SDK, but otherwise it looks good.

ppenenko:
FYI @jstone-lucasfilm https://github.com/autodesk-forks/MaterialX/actions/runs/21080475833/ is green except for JavaScript which always fails on forks, and the traces are downloadable as artifacts.

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.h
```diff
@@ -118,6 +118,27 @@ class TestSuiteOptions
     // Enable reference quality rendering. Default is false.
     bool enableReferenceQuality;
 
+    // Base directory for all test output artifacts (shaders, images, logs).
+    // If empty, use default locations. If set, all artifacts go to this directory.
+    mx::FilePath outputDirectory;
+
+    // Enable Perfetto tracing during render tests (requires MATERIALX_BUILD_TRACING).
+    // Default is false to avoid overhead when not profiling.
+    bool enableTracing = false;
```
jstone-lucasfilm, line 127: Although I like the idea of adopting in-class member initializers for `TestSuiteOptions` and other MaterialX classes in the future, I'd recommend following the existing pattern of in-constructor member initializers for now, so that we don't use a mixed approach that confuses readers.

source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
```diff
@@ -12,13 +12,15 @@
...
 #include <MaterialXRender/StbImageLoader.h>
+#include <MaterialXTrace/Tracing.h>
 #if defined(MATERIALX_BUILD_OIIO)
 #include <MaterialXRender/OiioImageLoader.h>
 #endif
 
 #include <MaterialXFormat/Util.h>
 
 namespace mx = MaterialX;
+using Cat = mx::Tracing::Category;
```
jstone-lucasfilm: Although this is a reasonable abbreviation, I'd recommend using the fully qualified `mx::Tracing::Category` names for now, especially in the phase where the `Tracing` library is brand new.

CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ option(MATERIALX_BUILD_OCIO "Build OpenColorIO support for shader generators." O
 option(MATERIALX_BUILD_TESTS "Build unit tests." OFF)
 option(MATERIALX_BUILD_BENCHMARK_TESTS "Build benchmark tests." OFF)
 option(MATERIALX_BUILD_OSOS "Build OSL .oso's of standard library shaders for the OSL Network generator" OFF)
+option(MATERIALX_BUILD_TRACING "Build with Perfetto tracing support for performance analysis." OFF)
```
jstone-lucasfilm: Let's call this flag `MATERIALX_BUILD_PERFETTO_TRACING`, so that we have a clear path for future extensions to Tracy and other instrumented profiling systems.

.github/workflows/main.yml
```diff
@@ -206,6 +206,8 @@ jobs:
           if [ "${{ matrix.extended_build_mdl_sdk }}" == "ON" -a "${{ runner.os }}" == "Windows" ]; then
             EXTENDED_BUILD_CONFIG="$EXTENDED_BUILD_CONFIG -DVCPKG_TARGET_TRIPLET=x64-windows-release -DMATERIALX_MDL_SDK_DIR=C:/vcpkg/installed/x64-windows-release"
           fi
+          # Enable Perfetto tracing
+          EXTENDED_BUILD_CONFIG="$EXTENDED_BUILD_CONFIG -DMATERIALX_BUILD_TRACING=ON"
```
jstone-lucasfilm: It's a huge benefit that we can include Perfetto in any GitHub CI build, but for now I'd recommend that we just select a _single_ build in which to include it, so that users are clearly directed to read a single `Tracing` artifact in each extended run of our CI.

To focus this on one build, I'd recommend adding a `matrix.extended_build_perfetto` flag, and feel free to choose which of our builds sets this to true.  Perhaps it could be a non-Windows build, since OIIO and MDL lean heavily on extended Windows builds.

jstone-lucasfilm (commented):
Following up on our last MaterialX TSC meeting, it's clear that we should move forward with this new work, so I've started writing up more detailed recommendations for improving it.

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.h
ppenenko: There are existing in-class member initializers in this class above, e.g. https://github.com/AcademySoftwareFoundation/MaterialX/pull/2742/changes#diff-c4ebf60508ed32c63965ee7be263bccb0bf0c669da8faf58999809867f787225R80

I agree that having some members initialized and some uninitialized here is messy, but that's the status quo.

The real problem with this class though is that it doesn't have a non-default constructor defined (nowhere to put in-constructor member initializers), so some of the Boolean members remain uninitialized after construction, which is UB-prone.

Should I instead make sure that all the POD members are initialized in the class definition, just like the new addition?

CMakeLists.txt
ppenenko: Good point.

source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
ppenenko: Will do, I was just trying to make the markers less verbose.

.github/workflows/main.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           compiler: gcc
           compiler_version: "14"
           python: 3.13
+          extended_build_perfetto: ON
```
ppenenko, line 34: The new switch.

.github/workflows/main.yml
ppenenko: Done, as far as I can tell.

CMakeLists.txt
ppenenko: Done.

source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp
ppenenko: Done.

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.h
jstone-lucasfilm: Ah, you're right, @ppenenko.  Since this conflict already exists in the class, I'll retract my suggestion above, and let's consider this work for the future.

jstone-lucasfilm (approved):
This looks great to me, @ppenenko, and thanks for the contribution to MaterialX!  Once we start seeing the new tracing artifacts in nightly builds, we can augment this with additional improvements.

PR #2743, opened 2026-01-09
Add support for df_vulkan as MDL test renderer

jreichel-nvidia:
Wrapper code based on that for the OSL test renderer and intermediate code from #2254.

The new CMake variable MATERIALX_MDL_BINARY_TESTRENDER contains the path of the df_vulkan binary, and is derived from MATERIALX_MDL_SDK_DIR.

Use x64-windows-release triplet for the MDL SDK to save disk space on the runners (and build time).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -468,6 +468,10 @@ jobs:
         name: MaterialX_Python_SDist
         path: sdist
 
+    - name: Install Doxygen (Windows)
+      uses: ssciwr/doxygen-install@v1
+      if: runner.os == 'Windows'
+
     - name: Build Wheel
       uses: pypa/cibuildwheel@v2.23.2
       with:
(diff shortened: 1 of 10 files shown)


jreichel-nvidia:
Extended build pipeline results:
https://github.com/jreichel-nvidia/MaterialX/actions/runs/20829640388

Does MaterialX use the free Github runners? Because using x64-windows as triplet hit a disk space limit (the free runners have a documented limit of 14 GB). Apparently, we have already been close to that limit before, and adding the df_vulkan example plus the Vulkan SDK as dependency crossed that limit. I've now switched the build_mdk_sdk feature to x64-windows-release to save disk space.

I guess build_oiio can't use x64-windows-release because it links to the OIIO library and one cannot mix debug/release on Windows? That means build_oiio and build_mdl_sdk can't be enabled in the same job ...

jstone-lucasfilm:
@jreichel-nvidia Yes, as of today we only use the free GitHub Actions runners on the MaterialX project, as this allows all developers to run identical tests in their fork of MaterialX.  In the future, though, we're interested in integrating the larger, GPU-enabled runners that the ASWF has access to, though these wouldn't be accessible in MaterialX forks.

jstone-lucasfilm:
The new extended build looks good in my local tests, and although I'd like to try testing renders through `MaterialXRenderMdl`, I haven't yet had the chance to try this.

Just following up on your question above, I like the idea of installing OIIO using `x64-windows-release` as well, as this would simplify our build process and remove unneeded overhead.  Did you already try this experiment and run into a specific technical hurdle, or is testing OIIO as release-only still future work?

jreichel-nvidia:
I didn't try since I remember difficulties mixing debug/release on Windows. From an older experiment I know that always using the static release libraries for OIIO does not work. Maybe it works for DLLs.

PR #2747, opened 2026-01-14
Use heuristic to preserve volume IOR in MDL

jreichel-nvidia:


jstone-lucasfilm (approved):
This looks good to me, thanks @jreichel-nvidia!

Issue #2421, opened 2025-05-28
MaterialX's cmake config should find_dependency its own OpenGL and X11 dependencies


spiffmon:
Hi, please have a look at [OpenUSD Issue 3648](https://github.com/PixarAnimationStudios/OpenUSD/pull/3648), which is a proposed fix to USD's cmake to work around the lack in MaterialX, when building USD with MaterialX.  It would be nice if we didn't need to intuit the transitive dependencies ourselves, and wonder if other MaterialX users haven't run into this?

Thanks!
--spiff

qbisi:
materialX should propagate opengl/x11 conditionally when target MaterialXRenderGlsl is linked against. I would suggest the following patch to materialX cmake config file:
```
# Always-available components
set(_MaterialX_known_components
  Core
  ...
  RenderGlsl
)

# Validate components
foreach(comp IN LISTS MaterialX_FIND_COMPONENTS)
  if(NOT comp IN_LIST _MaterialX_known_components)
    set(MaterialX_FOUND FALSE)
    set(MaterialX_NOT_FOUND_MESSAGE
      "Unknown MaterialX component: ${comp}")
    return()
  endif()
endforeach()

# Only GLSL requires OpenGL/X11
if("RenderGlsl" IN_LIST MaterialX_FIND_COMPONENTS)
  find_dependency(OpenGL)
  if(UNIX AND NOT APPLE)
    find_dependency(X11)
  endif()
endif()
```

Then downstream openusd can use find_dependency(MaterialX REQUIRED COMPONENTS RenderGlsl) in their cmake config file.

PR #2752, opened 2026-01-19
Propagate CMake dependency for opengl/x11

qbisi:
Hi, this pr aims to fix cmake configure error of vtk-9.6.0 when building with openusd with materialx.
In nixpkgs, the cmake configure system will complain
```
CMake Error at /nix/store/0b2sw8gp8qa0pqznnhirlm2gqx8h919b-python3.13-materialx-1.39.4/lib/cmake/MaterialX/MaterialXTargets.cmake:270 (set_target_properties):
  The link interface of target "MaterialXRenderHw" contains:

    X11::X11

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```
also the target OpenGL::OpenGL X11::Xt is required to link with other MaterialX target.

So, i believe MaterialX need to propagate dependency OpenGL/libX11/libXt  for downstream project to link with.

Note, in the last commit, i propagate OpenGL dependency conditionally as only the target MaterialXRenderGlsl link publicly with OpenGL, so i propagate OpenaGl optionally when downstream project require the component RenderGlsl explicitly (e.g. openusd might need this), tell me if you want this to be unconditionally propagated.

resolve https://github.com/AcademySoftwareFoundation/MaterialX/issues/2421.
cc @spiffmon

### DIFF
--- a/cmake/modules/MaterialXConfig.cmake.in
+++ b/cmake/modules/MaterialXConfig.cmake.in
@@ -41,6 +41,10 @@ endif()
 
 if(UNIX AND NOT APPLE)
     find_dependency(X11 REQUIRED COMPONENTS Xt)
+    if("RenderGlsl" IN_LIST MaterialX_FIND_COMPONENTS)
+        find_dependency(OpenGL REQUIRED)
+        set(MaterialX_RenderGlsl_FOUND TRUE)
+    endif()
 endif()
 
 check_required_components(@CMAKE_PROJECT_NAME@)


jstone-lucasfilm:
This looks very promising, thanks @qbisi, and I'd love to get feedback from @spiffmon, @JGamache-autodesk, and other domain experts before merging it.

qbisi:
Thanks for quick responding, updated the last commit (find opengl should be in scope unix and not macos).

qbisi:
Hi, any plan to make a new release in 2026. I would like to make a pr to downstream openusd that make use of the finding MaterialXRenderHw component.

jstone-lucasfilm:
@qbisi Yes, we're in the process of organizing a MaterialX v1.39.5 release, and we'll send out announcements once the first release candidate is available.

PR #2755, opened 2026-01-23
Fix FILENAME inputs in SHADER_INTERFACE_REDUCED mode

ppenenko:
## Summary

In `SHADER_INTERFACE_REDUCED` mode, unconnected `FILENAME` inputs on internal nodes were being inlined as literal path strings in the generated shader code. This caused GLSL compilation errors because real-time shading languages cannot handle file paths as string literals.

For example, a compound node call was being generated as:
```glsl
NG_network_surface(resources/Images/grid.png, N_surfaceshader_out);  // Invalid GLSL
```

Instead of the correct form with a uniform:
```glsl
uniform sampler2D N_surfaceshader_base_color_filename;
NG_network_surface(N_surfaceshader_base_color_filename, N_surfaceshader_out);  // Valid
```

## Solution

The fix modifies `ShaderGraph::finalize()` to publish `FILENAME` node inputs to graph input sockets even in `REDUCED` mode, ensuring they become uniforms with proper variable names. This mirrors the behavior of `COMPLETE` mode but only for `FILENAME` types, which cannot be inlined as literals.

## Test Plan

- [x] Existing GLSL render test suite passes with `shaderInterfaces=1` (REDUCED mode)
- [x] `network_surfaceshader.mtlx` material now compiles successfully in REDUCED mode
- [x] No regressions in COMPLETE mode

---

Assisted-by: Claude (Anthropic) via Cursor IDE
Signed-off-by: Pavlo Penenko <pavlo.penenko@autodesk.com>

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -916,40 +916,57 @@ void ShaderGraph::finalize(GenContext& context)
     // Sort the nodes in topological order.
     topologicalSort();
 
-    if (context.getOptions().shaderInterfaceType == SHADER_INTERFACE_COMPLETE)
+    // Publish node inputs that should be exposed as graph input sockets.
+    // In COMPLETE mode, all unconnected editable inputs are published.
+    // In REDUCED mode, only filename inputs are published since file paths
+    // can't be inlined as literals in real-time shading languages.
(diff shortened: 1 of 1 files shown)


ppenenko:
This is more complicated than I thought. Should probably just log an issue.

jstone-lucasfilm:
Thanks @ppenenko, and that sounds like the right approach here.  I'll close out this PR, and feel free to write up an Issue that captures your findings.

Issue #1261, opened 2023-02-25
MaterialXGraphEditor : renaming inputs / outputs does not work


kwokcb:
The node editor seems to get into some confused state when you rename any input or output
in a node graph. e.g. try this with the default marble graph that is loaded.

Seems that only the local node is renamed but nothing else is. This seems result in side-effects such as connections
being lost, output and inputs disappearing when going back to the document level, and inconsistency in
linkage (mtllx graph and imgui graph have different links set).

crystalqdm:
Hello! I would like to pick this issue up for ASWF Dev Days 2024 🙂

crystalqdm:
Sorry! Didn't mean to unassign myself 🙂 Still keen on picking up this issue.

ekim-01:
Hello!  I am new to ASWF Dev Days this year.  I have already asked if I could be assigned to issue#1749.  If that is not possible, I would gladly accept the responsibility of investigating this issue.  Thank you for your consideration!

ld-kerley:
I assigned you issue #1749 - but If you want to switch over to this one just let me know.

PR #2756, opened 2026-01-25
Unify element storage in Graph Editor classes

jstone-lucasfilm:
This changelist unifies element storage and replaces direct data references with accessor methods in the `UiNode`, `UiPin`, and `UiEdge` classes, providing better encapsulation, and addressing edge cases where data copies within these classes could fall out of sync.

Fixes https://github.com/AcademySoftwareFoundation/MaterialX/issues/1261 and related issues after user edits.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -319,13 +319,13 @@ ed::PinId Graph::getOutputPin(UiNodePtr node, UiNodePtr upNode, UiPinPtr input)
     {
         // For nodegraph need to get the correct output pin according to the names of the output nodes
         mx::OutputPtr output;
-        if (input->getNode()->getNode())
+        if (input->getUiNode()->getNode())
         {
-            output = input->getNode()->getNode()->getConnectedOutput(input->getName());
+            output = input->getUiNode()->getNode()->getConnectedOutput(input->getName());
         }
-        else if (input->getNode()->getNodeGraph())
+        else if (input->getUiNode()->getNodeGraph())
         {
-            output = input->getNode()->getNodeGraph()->getConnectedOutput(input->getName());
+            output = input->getUiNode()->getNodeGraph()->getConnectedOutput(input->getName());
         }
 
         if (output)
@@ -2085,7 +2085,7 @@ int Graph::getNodeId(ed::PinId pinId)
     {
         if (pin->getPinId() == pinId)
         {
-            return findNode(pin->getNode()->getId());
+            return findNode(pin->getUiNode()->getId());
         }
     }
     return -1;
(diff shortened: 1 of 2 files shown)


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2971,79 +2981,77 @@ void Graph::deleteNode(UiNodePtr node)
...
         for (UiPinPtr pin : outputPin.get()->getConnections())
         {
             mx::ValuePtr val;
-            if (pin->_pinNode->getNode())
+            if (pin->getNode()->getNode())
             {
-                mx::NodeDefPtr nodeDef = pin->_pinNode->getNode()->getNodeDef(pin->_pinNode->getNode()->getName());
-                val = nodeDef->getActiveInput(pin->_input->getName())->getValue();
-                if (pin->_pinNode->getNode()->getType() == mx::SURFACE_SHADER_TYPE_STRING)
+                mx::NodeDefPtr nodeDef = pin->getNode()->getNode()->getNodeDef(pin->getNode()->getNode()->getName());
```
lfl-eholthouser: Could the new getNode() function for the pins be called something like getUiNode() instead of just getNode() since that already exists? I think having pin->getNode()->getNode() is a bit hard to read.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2857,41 +2869,40 @@ void Graph::deleteLinkInfo(int startAttr, int endAttr)
...
                 {
-                    pin->_input->setConnectedNode(nullptr);
+                    pin->getInput()->setConnectedNode(nullptr);
                 }
                 if (_graphNodes[upNode]->getInput())
                 {
                     // Remove interface value in order to set the default of the input
-                    pin->_input->setConnectedInterfaceName(mx::EMPTY_STRING);
-                    setDefaults(pin->_input);
-                    setDefaults(_graphNodes[upNode]->getInput());
```
lfl-eholthouser, line 2879: I'm curious why the node input no longer needs to set to the default as well?

source/MaterialXGraphEditor/Graph.cpp
jstone-lucasfilm: Good catch, and I missed that!  I've now renamed `UiPin::getNode` to `UiPin::getUiNode` for clarity.

source/MaterialXGraphEditor/Graph.cpp
jstone-lucasfilm: Ah, good question.  That was an opportunistic fix that I found during testing of the graph editor, where we were unintentionally clearing the defaults of upstream inputs.  I believe this new version of the logic is slightly better, and in my local testing I get more consistent behavior when disconnecting and reconnecting pins.

lfl-eholthouser: Ah copy that! These changes look great. Thanks @jstone-lucasfilm

PR #2757, opened 2026-01-25
ci: Skip JavaScript job on forks

ppenenko:
The JavaScript job in the main CI workflow attempts to deploy to GitHub Pages, which fails on forks since they don't have the same Pages configuration as upstream.

E.g. https://github.com/autodesk-forks/MaterialX/actions/runs/21080475833/

This adds a condition to skip the JavaScript job on forks, matching the existing pattern used by the `sdist` and `wheels` jobs.


jstone-lucasfilm:
This is a very reasonable proposal, @ppenenko, though we'll need to provide an alternative mechanism for developers to _request_ a deployment of the MaterialX Web Viewer in their MaterialX fork.

Maybe we could link this to _extended_ CI builds, so that a manual trigger of the main MaterialX workflow would include a deployment of the MaterialX Web Viewer?

Other ideas here are welcome, and the goal is to provide developers with easy access to testing their proposed changes to the MaterialX Web Viewer, without impacting teams that might be restricted from enabling web deployment for all MaterialX builds.

ppenenko:
Good point @jstone-lucasfilm but I'm not sure if extended builds would succeed for most forks. Let me dig deeper into the failures.

.github/workflows/main.yml
```diff
@@ -481,7 +481,10 @@ jobs:
       working-directory: javascript/MaterialXView
 
     - name: Deploy Web Viewer
-      if: github.event_name != 'pull_request'
+      if: >-
+        github.event_name != 'pull_request' &&
+        (github.repository == 'AcademySoftwareFoundation/MaterialX' ||
+         github.event_name == 'workflow_dispatch')
```
ppenenko, line 487: FYI @jstone-lucasfilm - does this provide the necessary control?

PR #2758, opened 2026-01-25
Remove `choco` dependency from CI

jstone-lucasfilm:
This changelist removes a dependency on `choco` from Python wheels builds in our GitHub CI, so that common disruptions to this package manager don't block MaterialX builds.


jstone-lucasfilm:
Let's merge this administratively to unblock MaterialX CI, and we can discuss additional ideas once we're up and running.

PR #2759, opened 2026-01-27
Connect `geometry_thin_walled` to `surface` in OpenPBR

jreichel-nvidia:
And make it uniform.
See also https://github.com/AcademySoftwareFoundation/OpenPBR/pull/250.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -48,6 +48,8 @@ The current Voting Members of the MaterialX TSC are:
 - Doug Smythe - Industrial Light & Magic
 - Niklas Harrysson - Lumiere Software
 - Orn Gunnarsson - Autodesk
+- Lee Kerley - Apple
+- Frankie Liu - NVIDIA
 - Andréa Machizaud - Adobe
 
 ### Stakeholders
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This is definitely the right path forward, @jreichel-nvidia, but I wonder whether this should wait on the formal release of OpenPBR 1.2, which is the version of OpenPBR that will contain this improvement:

https://github.com/AcademySoftwareFoundation/OpenPBR/tree/dev_1.2

jreichel-nvidia:
Hmm, I see. Do you know by chance when this release is planned to happen?

jreichel-nvidia:
Since several parties are now using MaterialX with this patch, I wonder whether we shouldn't merge it even though OpenPBR 1.2 is not yet released ...

jstone-lucasfilm:
I'm inclined to agree, @jreichel-nvidia, and perhaps we should think of this change as a _bug fix_ to the MaterialX graph for OpenPBR Surface, rather than a design change that requires versioning of the shading model.

I'm CC'ing @AdrienHerubel and @theblackunknown for their thoughts, and let's see if there's general agreement on this idea.

AdrienHerubel:
I agree, it is a bugfix of the graph and in sync with future versions of the specification. It is fine to get it in.

theblackunknown:
Post-mortem review: like Adrien all those changes look good to me and will be appreciated even before OpenPBR v1.2 release

PR #2764, opened 2026-01-31
Graph Editor improvements for edges and links

jstone-lucasfilm:
This changelist improves the management of edges and links in the MaterialX Graph Editor, including the following specific changes:

- Consolidate duplicated edge creation and erasure into `Graph::createEdge` and `UiNode::eraseEdge` methods.
- Synchronize `Graph::removeEdge` and `Graph::deleteNode` with the global edge list, fixing cases where stale edges could persist.
- Replace the static ID counter in `Link` with explicit construction through `GraphState::nextUiId`.
- Simplify `Graph::edgeExists` and `Graph::linkExists` implementations for clarity, taking arguments by const reference.
- Add bounds and null checks across edge and link operations to prevent undefined behavior.


lfl-eholthouser (approved):
This looks great, thanks @jstone-lucasfilm !

PR #2767, opened 2026-02-05
AOUSD formatting proposals for NPR Nodes

jstone-lucasfilm:
This changelist contains an initial set of formatting proposals for the NPR Nodes specification, organized by the AOUSD Materials Working Group.

The node signatures for the NPR Nodes have been aligned with the latest signatures for the Standard and PBR Nodes, and a few minor clarifications have been made to node descriptions.

### DIFF
--- a/documents/Specification/MaterialX.NPRSpec.md
+++ b/documents/Specification/MaterialX.NPRSpec.md
@@ -53,7 +53,7 @@ The view direction is a normalized vector from the viewer position to the curren
 <a id="node-facingratio"> </a>
 
 ### `facingratio`
-The geometric facing ratio of surface normals and the view direction.
+The geometric facing ratio of the view direction and surface normal.
 
 Facing ratio is computed as the dot product between the view direction and surface normal.
 


documents/Specification/MaterialX.NPRSpec.md
```diff
@@ -36,35 +36,55 @@ This document describes a number of MaterialX nodes primarily applicable to non-
...
 
 
 
 ## NPR Utility Nodes
 
 <a id="node-facingratio"> </a>
 
-* **`facingratio`**: returns the geometric facing ratio, computed as the dot product between the view direction and geometric normal.  Output is a float between 0.0 and 1.0.
+### `facingratio`
+The geometric facing ratio of surface normals and the view direction.
```
dbsmythe: Perhaps should be "of the surface normal" rather than "of surface normals"?

documents/Specification/MaterialX.NPRSpec.md
```diff
@@ -36,35 +36,55 @@ This document describes a number of MaterialX nodes primarily applicable to non-
...
-    * `light_direction` (vector3): the incoming predominant lighting direction in world space, defaults to (1.0, -0.5, -0.5).
+### `gooch_shade`
+Computes the color from single-pass shading portion of the Gooch[^Gooch1998] lighting model.
+
+Gooch shade provides an illustrative shading effect by blending colors based on the angle between the surface normal and the light direction. It also provides a simple Phong specular highlight, on top of the warm and cool colors.
+
+|Port                |Description                            |Type   |Default      |
+|--------------------|---------------------------------------|-------|-------------|
+|`warm_color`        |The color facing toward the light      |color3 |0.8, 0.8, 0.7|
+|`cool_color`        |The color facing away from the light   |color3 |0.3, 0.3, 0.8|
```
dbsmythe, line 83: The original description mentioned that these colors were specified in the "lin_rec709" colorspace; perhaps we should find a place to keep this important detail in the node description?

documents/Specification/MaterialX.NPRSpec.md
jstone-lucasfilm: Good point, @dbsmythe, and I've updated this phrasing for better consistency with the data tables.

documents/Specification/MaterialX.NPRSpec.md
jstone-lucasfilm: That's a great question, and I think the removal of this detail was _intentional_, as these colors are still governed by MaterialX color management, so they can be specified in any source color space, and will be transformed to the linear color space of the renderer.

In the future, we may want to specify that the _default_ values for these two inputs are in `lin_rec709`, but that's not yet true in the data definition of the node, so we shouldn't state that here.

PR #2772, opened 2026-02-13
Update TSC members

jstone-lucasfilm:
This changelist updates Lee Kerley and Frankie Liu to voting members of the MaterialX TSC, as confirmed in our recent TSC vote.


dbsmythe (approved):
Looks good to me!

PR #2773, opened 2026-02-13
Add `enableDirectLight` option to MaterialX Viewer

hybridherbst:
This PR adds a commandline flag to disable direct lighting, so scenes can be rendered purely with image-based lighting. This makes it easier to compare to other environments where direct lighting from environment maps is not available.

For example, comparison with Unity:

| Unity<br>just IBL | MaterialXView<br>`--enableDirectLight false` | MaterialXView<br>default settings | 
| -- | -- | -- |
| image | image | image |

### DIFF
--- a/source/MaterialXView/Main.cpp
+++ b/source/MaterialXView/Main.cpp
@@ -30,7 +30,7 @@ const std::string options =
     "    --envSampleCount [INTEGER]     Specify the environment sample count (defaults to 16)\n"
     "    --envLightIntensity [FLOAT]    Specify the environment light intensity (defaults to 1)\n"
     "    --lightRotation [FLOAT]        Specify the rotation in degrees of the lighting environment about the Y axis (defaults to 0)\n"
-    "    --directLight [BOOLEAN]        Specify whether direct lighting is enabled (defaults to true)\n"
+    "    --enableDirectLight [BOOLEAN]  Specify whether direct lighting is enabled (defaults to true)\n"
     "    --shadowMap [BOOLEAN]          Specify whether shadow mapping is enabled (defaults to true)\n"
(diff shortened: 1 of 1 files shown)


source/MaterialXView/Main.cpp
```diff
@@ -30,6 +30,7 @@ const std::string options =
     "    --envSampleCount [INTEGER]     Specify the environment sample count (defaults to 16)\n"
     "    --envLightIntensity [FLOAT]    Specify the environment light intensity (defaults to 1)\n"
     "    --lightRotation [FLOAT]        Specify the rotation in degrees of the lighting environment about the Y axis (defaults to 0)\n"
+    "    --directLight [BOOLEAN]        Specify whether direct lighting is enabled (defaults to true)\n"
```
jstone-lucasfilm: This new feature looks good, though I'd recommend "--enableDirectLight" as the command-line option, so that we don't give the impression that this option controls _which_ direct lighting rig should be used.

source/MaterialXView/Main.cpp
```diff
@@ -91,6 +92,7 @@ int main(int argc, char* const argv[])
     int envSampleCount = mx::DEFAULT_ENV_SAMPLE_COUNT;
     float envLightIntensity = 1.0f;
     float lightRotation = 0.0f;
+    bool directLight = true;
```
jstone-lucasfilm: As above, I'd recommend "enableDirectLight" here for clarity.

jstone-lucasfilm (commented):
This looks like a good proposal to me, @hybridherbst, and I had just one recommendation on the name of the command-line option.

jstone-lucasfilm (approved):
This looks good to me, thanks @hybridherbst!

PR #2779, opened 2026-02-15
Add GLSL functions for PDFs

jstone-lucasfilm:
This changelist adds explicit GLSL functions for PDFs (probability density functions) to MaterialX, making its physically based shading code clearer and more self-documenting, and providing a stepping stone towards future use of generated GLSL/ESSL/MSL in path tracers as well as rasterization renderers.

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_environment_fis.glsl
@@ -33,7 +33,7 @@ vec3 mx_environment_radiance(vec3 N, vec3 V, vec3 X, vec2 alpha, int distributio
 
         // Sample the environment light from the given direction.
         vec3 Lw = mx_matrix_mul(tangentToWorld, L);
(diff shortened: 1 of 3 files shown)


libraries/pbrlib/genglsl/lib/mx_microfacet_sheen.glsl
```diff
@@ -70,9 +70,8 @@ float mx_imageworks_sheen_dir_albedo_monte_carlo(float NdotV, float roughness)
         float reflectance = mx_imageworks_sheen_brdf(NdotL, NdotV, NdotH, roughness);
 
         // Add the radiance contribution of this sample.
-        //   uniform_pdf = 1 / (2 * PI)
         //   radiance = reflectance * NdotL / uniform_pdf;
-        radiance += reflectance * NdotL * 2.0 * M_PI;
+        radiance += reflectance * NdotL / mx_uniform_hemisphere_PDF();
```
niklasharrysson, line 74: I guess this new division is easy for the compiler to remove since the PDF function returns a constant.

PR #2780, opened 2026-02-18
Add missing OCIO include in RenderView.cpp

jstone-lucasfilm:
This changelist adds a missing include of `OcioColorManagementSystem.h` in `RenderView.cpp`, which is required when building the Graph Editor with OCIO support.


Martinfx:
> This changelist adds a missing include of `OcioColorManagementSystem.h` in `RenderView.cpp`, which is required when building the Graph Editor with OCIO support.

I tested on FreeBSD 15.0. It is ok.

jstone-lucasfilm:
Thanks for the validation, @Martinfx.  I'm CC'ing @JGamache-autodesk for visibility, though I believe he's currently on sabbatical.

I'll keep this PR open for a bit longer for the community to weigh in, and if there are no objections, let's plan to merge it for MaterialX v1.39.5.

Issue #2769, opened 2026-02-09
Direct lighting missing in MaterialX Viewer on MacOS


jstone-lucasfilm:
I've recently been testing the MaterialX Viewer on MacOS, and I'm noticing that the _direct lighting_ term is missing on this platform, while it renders correctly in Windows and Linux builds of the MaterialX Viewer.

The missing direct lighting term appears as an unexpectedly dark render of the environment lighting in the MaterialX Viewer on MacOS, as the sunlight in the San Giuseppe Bridge environment strongly relies upon the direct lighting term.

For reference, here are recent renders from the GitHub CI for MaterialX on MacOS, where the same artifact is visible in the MaterialX Viewer renders that are performed for each commit to the project:

<img width="128" height="128" alt="Image" src="https://github.com/user-attachments/assets/498d4629-b7b4-40cc-9e7e-3d9ee2c3bd4d" />
<img width="128" height="128" alt="Image" src="https://github.com/user-attachments/assets/698876a7-a4a2-4333-bd36-a65f4d7ff839" />

jstone-lucasfilm:
This has been fixed in #2781.

PR #2781, opened 2026-02-18
Fix direct lighting in `MslProgram::bindLighting`

jstone-lucasfilm:
This changelist fixes a uniform name mismatch in `MslProgram::bindLighting`, which was causing the direct lighting term to be missing on MacOS, as reported in https://github.com/AcademySoftwareFoundation/MaterialX/issues/2769.

In the MSL rendering path, uniforms are stored with a `GlobalContext` struct prefix, and the `hasUniform` and `bindUniform` methods correctly resolve these prefixed names through the global uniform name map.


niklasharrysson (approved):
LGTM!

PR #2525, opened 2025-08-17
Restore modern Python versions to CI

jstone-lucasfilm:
This changelist restores support for modern Python versions to the GitHub CI for MaterialX, replacing a CMake call to `setup.py` with `pip install`.

Additionally, the minimum Python version in our root `pyproject.toml` file has been raised from 3.6 to 3.9, aligning with the current minimum version of Python required by MaterialX.

Issue #2414, opened 2025-05-26
Modernize `setup.py` system for MaterialX Python


jstone-lucasfilm:
This task covers additional work to modernize the `setup.py` system used in local builds of MaterialX Python, as described in this document:

https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

MaterialX already has a `pyproject.toml` file for its PyPI generation step, which can be leveraged in the new work for local MaterialX Python builds:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/pyproject.toml

jstone-lucasfilm:
Marking this as resolved by the combination of #2525 and #2782.

PR #2782, opened 2026-02-20
Migrate Python packaging to `pyproject.toml`

jstone-lucasfilm:
This changelist migrates the MaterialX Python packaging configuration from the legacy `setup.py` script to the modern `pyproject.toml` standard, as defined in PEP 517 and PEP 621.


jstone-lucasfilm:
Once this latest improvement is reviewed and merged, I believe we can mark issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/2414 as resolved.

JeanChristopheMorinPerso (approved):
I don't remember exactly why this is still needed, but it looks good on the surface.

jstone-lucasfilm:
Thanks @JeanChristopheMorinPerso, this is mainly the completion of work that we started in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2525, fully moving over from `setup.py` to `pyproject.toml`.

My expectation is that the main benefit will be improved maintainability going forward, and that Python developers won't notice any change at all.

PR #2783, opened 2026-02-20
Add retroreflective to specular BSDFs

msuzuki-nvidia:
Add a `retroreflective` boolean input to the specular BSDF nodes (`dielectric_bsdf`, `conductor_bsdf`, `generalized_schlick_bsdf`).

* Update the PBR specification with the new input
* Add `retroreflective` input to node definitions in `pbrlib_defs.mtlx`
* Add GLSL implementation. OSL and MDL have interface-only changes (the parameter is ignored)

Also includes minor test infrastructure fixes:

* Make `FilePath::createDirectory()` recursive to support nested output directories
* Resolve relative paths to absolute in `OslRenderer` before working directory changes
* Skip `blur.mtlx` in MDL render tests due to material name mismatch

### DIFF
--- a/source/MaterialXFormat/File.cpp
+++ b/source/MaterialXFormat/File.cpp
@@ -278,17 +278,20 @@ FilePathVec FilePath::getSubDirectories() const
     return dirs;
 }
 
-void FilePath::createDirectory() const
+void FilePath::createDirectory(bool recursive) const
 {
-    if (isEmpty() || exists())
+    if (recursive)
     {
-        return;
-    }
+        if (isEmpty() || exists())
+        {
+            return;
+        }
 
-    FilePath parent = getParentPath();
-    if (!parent.isEmpty() && !parent.exists())
-    {
-        parent.createDirectory();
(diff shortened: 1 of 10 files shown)


source/MaterialXFormat/File.cpp
```diff
@@ -280,6 +280,17 @@ FilePathVec FilePath::getSubDirectories() const
...
 {
+    if (isEmpty() || exists())
+    {
+        return;
+    }
+
+    FilePath parent = getParentPath();
+    if (!parent.isEmpty() && !parent.exists())
+    {
+        parent.createDirectory();
```
jstone-lucasfilm: @msuzuki-nvidia This looks like a good extension of `FilePath::createDirectory`, but we should probably gate this behavior with a new `recursive` argument to the method, where the default value of the argument would be `false` to maintain the same behavior and performance for existing callers.

msuzuki-nvidia: sounds good, updated

Issue #2763, opened 2026-01-29
Document inputs don't work when connected to NodeGraph inputs


frohnej-adsk:
I don't think this is mentioned explicitly in the specification but judging by PR #1453, documents are allowed to contain `<input>` elements.
- Document inputs work if they feed into an input on a `Node` ✅ 
- Document inputs don't work if they feed into an input of a `NodeGraph` ❌ 

## ✅ `Node` input referencing a document input
This renders correctly:
<img width="553" height="200" alt="Image" src="https://github.com/user-attachments/assets/598c17f3-4029-4492-ad94-4851675acb4e" />
```XML
<?xml version="1.0"?>
<materialx version="1.39">
  <!-- Document input -->
  <input name="A" type="color3" value="0, 1, 0" />

  <add name="add1" type="color3" nodedef="ND_add_color3">
    <!-- Node input connected to document input is resolved correctly. -->
    <input name="in1" type="color3" interfacename="A" />
  </add>

  <convert name="convert1" type="surfaceshader" nodedef="ND_convert_color3_surfaceshader">
    <input name="in" type="color3" nodename="add1" />
  </convert>
  <surfacematerial name="surfacematerial1" type="material" nodedef="ND_surfacematerial">
    <input name="surfaceshader" type="surfaceshader" nodename="convert1" />
  </surfacematerial>
</materialx>
```

## ❌ `NodeGraph` input referencing a document input
This renders black:
<img width="553" height="200" alt="Image" src="https://github.com/user-attachments/assets/79c0bc90-ab43-4458-a586-707fc6f30cd5" />
```XML
<?xml version="1.0"?>
<materialx version="1.39">
  <!-- Document input -->
  <input name="A" type="color3" value="0, 1, 0" />

  <nodegraph name="compound1">
    <!-- NodeGraph input connected to document input. -->
    <!-- This isn't resolved correctly by `Input::getInterfaceInput()` -->
    <input name="B" type="color3" interfacename="A" />

    <add name="add1" type="color3" nodedef="ND_add_color3">
      <!-- Calling `Input::getInterfaceInput()` on this input returns input B. -->
      <!-- This isn't wrong, but it might be desirable to follow interfacenames -->
      <!-- recursively and return input A. -->
      <input name="in1" type="color3" interfacename="B" />
    </add>
    <output name="out" type="color3" nodename="add1" />
  </nodegraph>

[...]

Issue #1832, opened 2024-05-22
Validation incorrectly categorizes top level inputs connected to graphs as invalid


kwokcb:
## Issue

Validation considers a top level `input` conneted to an interface input on a `nodegraph` as invalid even though this
should be valid and is accepted by code generation.

## Test data
```xml
<?xml version="1.0"?>
<materialx version="1.39" colorspace="lin_rec709">
  <input name="input_color4" type="color3" value="0,0,0" xpos="10.094203" ypos="0.439655" />
  <input name="input_float1" type="float" value="0" xpos="9.862319" ypos="1.818966" />
  <nodegraph name="custom_graph_1">
    <input name="matrixIn" type="matrix33" value="1,0.2,0.2,0,1,0,0,0,1" />
    <input name="floatIn" type="float" interfacename="input_float1" />
    <input name="color3In" type="color3" interfacename="input_color4" />
    <output name="output_matrix34" type="matrix33" nodename="dot_matrix34" />
    <output name="floatOut" type="float" nodename="dot_float1" />
    <output name="color3Out" type="color3" nodename="dot_color4" />
    <output name="color3Gamma22Out" type="color3" nodename="colorcorrect_color4" />
    <dot name="dot_float1" type="float">
      <input name="in" type="float" interfacename="floatIn" />
      <input name="note" type="string" value="" />
    </dot>
    <dot name="dot_matrix34" type="matrix33">
      <input name="in" type="matrix33" interfacename="matrixIn" />
      <input name="note" type="string" value="" />
    </dot>
    <dot name="dot_color4" type="color3">
      <input name="in" type="color3" interfacename="color3In" />
      <input name="note" type="string" value="" />
    </dot>
    <colorcorrect name="colorcorrect_color4" type="color3">
      <input name="in" type="color3" interfacename="color3In" />
      <input name="hue" type="float" value="0" />
      <input name="saturation" type="float" value="1" />
      <input name="gamma" type="float" value="2.2" />
      <input name="lift" type="float" value="0" />
      <input name="gain" type="float" value="1" />
      <input name="contrast" type="float" value="1" />
      <input name="contrastpivot" type="float" value="0.5" />
      <input name="exposure" type="float" value="0" />
    </colorcorrect>
  </nodegraph>
  <nodegraph name="ramp_graph">
    <input name="input_vector3" type="vector2" value="0,0" />
    <input name="Left_color" type="color3" value="0, 0, 0" />
[...]

PR #2785, opened 2026-02-20
Improvements to interface name validation

jstone-lucasfilm:
This changelist improves the validation of interface name attributes in `ValueElement::validate`, including the following specific changes:

- Replace `getAncestorOfType<NodeGraph>` with `getAncestorOfType<GraphElement>`, allowing interface names to be validated within any graph scope, including the document scope.
- When a value element is a direct child of a graph element, resolve its interface name against the parent graph rather than the containing graph, fixing validation of nodegraph inputs that reference document-level inputs.
- When no declaration is found for a graph element, fall back to the graph element itself as the declaration scope.

Collectively, these improvements address issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/1832.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,56 @@
 
 ## [1.39.5] - Development
 
+### Added
+- Added support for the [Slang shading language](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2548) in MaterialX shader generation and rendering, with [language bindings](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2679) for Python and JavaScript.
+- Added [OSL implementations](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2734) for hex-tiled images, completing cross-language coverage for hextiling nodes.
+- Added an [OSL command string code generator](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2603), enabling the mixing of external OSL shaders with MaterialX-generated nodes.
+- Added support for [GLSL probability density functions](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2779), providing a foundation for path-tracer use of generated GLSL, ESSL, and MSL shaders.
+- Added an [enableDirectLight option](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2773) to the MaterialX Viewer, enabling pure image-based lighting without direct light contributions.
+- Added support for [upstream and downstream node traversal](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2225) in the property editor of the MaterialX Graph Editor.
+- Added shader generation options for [Airy Fresnel iterations](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2569) and [constant node eliding](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2521), supporting quality and performance trade-offs on iridescent materials and interactive editing workflows.
+- Added [Python operator bindings](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2594) for Vector and Matrix types.
+- Added support for [Doxygen-generated docstrings](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2657) in MaterialX Python bindings, enabling IDE support and interactive help.
+- Added a [nodedef comparison script](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2723) for validating alignment between data library definitions and specification documents.
+- Added [Perfetto tracing](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2742) as an optional performance instrumentation framework.
+- Added a [nightly build workflow](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2621) to GitHub Actions, with [test suite rendering on MacOS](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2675) and [MDL compiler validation](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2254).
+- Added [developer guidelines](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2632) to the MaterialX contributing guide.
+
+### Changed
+- Improved specifications for the [Standard Nodes](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2520), [PBR Nodes](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2732), and [NPR Nodes](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2767), converting prose descriptions to structured tables.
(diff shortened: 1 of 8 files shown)


source/MaterialXCore/Element.cpp
```diff
@@ -793,10 +793,21 @@ bool ValueElement::validate(string* message) const
...
+            // interface name references a value element in the parent scope.
+            ConstElementPtr graphParent = graph->getParent();
+            graph = graphParent ? graphParent->getAncestorOfType<GraphElement>() : nullptr;
+        }
+        if (graph)
         {
+            ConstInterfaceElementPtr decl = graph->getDeclaration();
+            if (!decl)
+            {
+                decl = graph;
```
kwokcb, line 809: I don't think this is correct if the graph is not the grandparent. I believe you will look at siblings of the input or token which should not be allowed.

source/MaterialXTest/MaterialXCore/Node.cpp
```diff
@@ -81,6 +81,23 @@ TEST_CASE("Interface Input Validation", "[node]")
...
+    mx::InputPtr docInput = doc->addInput("doc_color", "color3");
+    mx::InputPtr graphInput = nodegraph->addInput("graph_color", "color3");
+    graphInput->setInterfaceName(docInput->getName());
+    bool valid = doc->validate(&validationErrors);
+    if (!valid)
+    {
+        INFO(validationErrors);
+    }
+    REQUIRE(valid);
+
```
kwokcb, line 96: As noted above, a test should be added to try and specify an interfacename which is a sibling input / token -- which should fail.

source/MaterialXCore/Element.cpp
```diff
@@ -793,10 +793,21 @@ bool ValueElement::validate(string* message) const
...
         validateRequire(isA<Input>() || isA<Token>(), res, message, "Only input and token elements support interface names");
-        ConstNodeGraphPtr nodeGraph = getAncestorOfType<NodeGraph>();
-        ConstInterfaceElementPtr decl = nodeGraph ? nodeGraph->getDeclaration() : nullptr;
-        if (decl)
+        ConstGraphElementPtr graph = getAncestorOfType<GraphElement>();
+        if (graph && graph == getParent())
+        {
+            // This element is a direct child of a graph element, so its
+            // interface name references a value element in the parent scope.
+            ConstElementPtr graphParent = graph->getParent();
```
kwokcb, line 801: If this is a token, then the spec says document level tokens have no effect so I think the previous search for NodeGraph needs to be preserved. (see [this test](https://github.com/AcademySoftwareFoundation/MaterialX/blob/0d1731a8832f5f919d30414a0d4ebaafaf4cd6f6/resources/Materials/TestSuite/stdlib/texture/token_graph.mtlx#L24))

source/MaterialXTest/MaterialXCore/Node.cpp
jstone-lucasfilm: That's a great suggestion, thanks @kwokcb, and I've added this unit test to the PR.  No other code changes were needed, and the validation result correctly matches expectations (validation fails, and the unit test succeeds).

source/MaterialXCore/Element.cpp
jstone-lucasfilm: I've traced through the current code, and I believe it handles this case correctly, as verified by the new unit test.  Let me know if there's another case that you're thinking about here, and we can add a corresponding unit test for that case as well.

source/MaterialXCore/Element.cpp
jstone-lucasfilm: I'd gently push back on this idea, as I don't see any specification language that prevents tokens from existing at document scope, and there's nothing to prevent an interface name from referencing such a token.

The note in the example that you cite is about runtime token _substitution_ behavior, not about the validity of document-scope tokens in general, and I think it should be a separate discussion if we want to change that behavior in MaterialX.

source/MaterialXCore/Element.cpp
```diff
@@ -793,10 +793,21 @@ bool ValueElement::validate(string* message) const
     if (hasInterfaceName())
     {
         validateRequire(isA<Input>() || isA<Token>(), res, message, "Only input and token elements support interface names");
-        ConstNodeGraphPtr nodeGraph = getAncestorOfType<NodeGraph>();
-        ConstInterfaceElementPtr decl = nodeGraph ? nodeGraph->getDeclaration() : nullptr;
-        if (decl)
+        ConstGraphElementPtr graph = getAncestorOfType<GraphElement>();
```
frohnej-adsk, line 796: [Suggestion] I feel like the use of `getAncestorOfType` is misleading. It suggests that we don't know how many levels up the referenced interface is.

But I think we do know: It should always be two levels up (i.e. the grandparent or its node definition). Should this be made explicit? It might make the code easier to read.

Perhaps something like this:
```C++
        validateRequire(isA<Input>() || isA<Token>(), res, message, "Only input and token elements support interface names");
        ConstInterfaceElementPtr referencedInterface;
        if (getParent() && getParent()->getParent())
        {
            referencedInterface = getParent()->getParent()->asA<InterfaceElement>();
        }
        validateRequire(referencedInterface != nullptr, res, message, "Referenced declaration not found");
        if (referencedInterface)
        {
            if (ConstInterfaceElementPtr decl = referencedInterface->getDeclaration())
            {
                referencedInterface = decl;
            }
            ValueElementPtr valueElem = referencedInterface->getActiveValueElement(getInterfaceName());
            validateRequire(valueElem != nullptr, res, message, "Interface name not found in referenced declaration");
            if (valueElem)
            {
                validateRequire(getType() == valueElem->getType(), res, message, "Interface name refers to value element of a different type");
            }
        }
```

jstone-lucasfilm: That's a completely valid suggestion, @frohnej-adsk.

I agree that `getAncestorOfType` feels a bit open-ended since the two possible hierarchies are well-known, but my thinking is that the ancestor-based approach is a better fit, as our semantic intent is to "find the nearest containing graph scope", rather than to go up exactly two levels.

I'm imagining that the supported set of graph hierarchies in MaterialX will evolve over time, and the "nearest containing graph scope" can adapt naturally to new features, while a hardcoded level count has the potential to be more fragile.

source/MaterialXCore/Element.cpp
kwokcb: I wasn't sure. If it passes tests then it sounds solid to me.

source/MaterialXTest/MaterialXCore/Node.cpp
kwokcb: Thanks for adding this in.

source/MaterialXCore/Element.cpp
kwokcb: That's a good point about substitution logic vs validity. I have no strong feelings about changing validity so let's go with your logic on this.

source/MaterialXCore/Element.cpp
kwokcb: I would agree with using a semantic vs hard-coding 2 levels just happens to be the current element structure. There are other use cases where the starting point may differ and the traversal is not necessarily 2 levels.

frohnej-adsk: Makes sense, thanks! Although the semantic intent is to "find the nearest containing graph scope, that's not the direct parent" now, isn't it?

It wouldn't really make sense for an element reference itself, so we always need to go at lest one level up, and this PR adds a special case for the direct parent, so we always go at lest two levels up.

Perhaps `getParent()->getParent()->getAncestorOfType<GraphElement>()` would make sense.

I'm wondering because something similar will need to be implemented in `Input::getInterfaceInput()` to resolve issue #2763.

PR #2789, opened 2026-02-23
Improve vertical refinement in Graph Editor layout

jstone-lucasfilm:
This changelist improves the `refineLayerY` method in the Graph Editor layout engine, addressing a downward drift bias in vertical node placement that was caused by the use of a single top-to-bottom sweep for overlap resolution.

Overlaps are now resolved with a symmetric bidirectional technique, where top-to-bottom and bottom-to-top passes are run independently and their results are averaged, producing balanced placement without an explicit re-centering step.

Additionally, the median calculation now properly averages the two middle values for even-sized neighbor sets.


jstone-lucasfilm:
Standard Surface Graph Layout (Before):

StandardSurface_Old

Standard Surface Graph Layout (After):

StandardSurface_New

PR #2791, opened 2026-02-24
MaterialXGraphEditor: UI/UX improvements

msuzuki-nvidia:
Improve the MaterialXGraphEditor's usability and visual polish:

- Fix node search popup not closing when selecting via Enter key
- Make node search case-insensitive
- Auto-select newly created nodes
- Bundle Roboto as the default font (with fallback to ImGui default)
- Tune node styling and layout

Current:
MaterialX Graph Editor_current

New:
MaterialX Graph Editor_new

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -12,7 +12,6 @@
 #include <imgui_node_editor_internal.h>
 #include <widgets.h>
 
-#include <algorithm>
 #include <cctype>
 #include <iostream>
 


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -12,6 +12,8 @@
 #include <imgui_node_editor_internal.h>
 #include <widgets.h>
 
+#include <algorithm>
```
jstone-lucasfilm: Minor note:  The standard `algorithm` header is already brought into this source file through [Library.h](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Library.h), so we don't need to repeat it here.

jstone-lucasfilm (commented):
This looks really promising, thanks @msuzuki-nvidia!  I had just one minor note about header inclusion, and I'm very interested in @lfl-eholthouser's feedback on the visual results of the UI/UX improvements.

source/MaterialXGraphEditor/Graph.cpp
msuzuki-nvidia: removed

Issue #2655, opened 2025-11-03
Black rendering for standard surface on Android as of 1.39.4


kwokcb:
## Issue

Standard surface renders completely black on Android devices.
This appears to be a regression from 1.39.3 to 1.39.4. 

Have not invested as to why.

kwokcb:
@jstone-lucasfilm Is it possible to get this fixed for 1.39.5 ?

The official web site still shows this issue:

<img width="2340" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b5ec58c5-8bb5-41f4-84e3-e7881fb17cca" />

jstone-lucasfilm:
@kwokcb Alas, I don't have any access to Android phones or their development environment, so I'll need to leave this improvement to members of the MaterialX community who have access to this ecosystem.

I can confirm, though, that the MaterialX Web Viewer still works smoothly in iOS 26:

![Image](https://github.com/user-attachments/assets/76019235-cd93-4afb-9abf-cf08b07677b9)

kwokcb:
I'll trying pinging on the Slack channel. I'd really like to see the shaders on my phone :).

AntonPalmqvist:
Works in Firefox and Chromium on my Pixel 8a running Android 16. I do however get this black artifact on transparent materials.

<img width="3201" height="1433" alt="Image" src="https://github.com/user-attachments/assets/23775eaf-1133-418e-8be5-c92a2e543085" />

kwokcb:
That's a separate logged issue. There appears to be division by 0 for low values of specular roughness. I think due to lower precision math.

This is for iOS and Android.

kwokcb:
It seems the root of the problem is precision . With mediuimp (medium) precision there may be different side-effects on different device's renderers. 

As `ESSL 3.0` is required, one overall fix is to use `highp` ([which is core (required) in 3.0+](https://docs.imgtec.com/reference-manuals/open-gl-es-extensions/html/topics/GL_OES/fragment-precision-high.html)). ESSL 3.0 / highp [is supported on Android 4.3+](https://developer.android.com/develop/ui/views/graphics/opengl/about-opengl).

[...]

Issue #2654, opened 2025-11-03
Corrupted rendering of specular roughness for shading modules on Android


kwokcb:
## Issue

On Android devices if you set the specular roughness low enough, you appear to get solid white pixels in the areas facing the primary light direction which I believe is computed based on the light environment map.

This occurs for all shading models so appears to be a common math issue.

This does not occur on Apple devices.

## Triage

- Around 0.07 for glTF and 0.08 for OpenPBR and standard surface causes this with all other input values the default.
- Code is ESSL with mediump precision setting.

After hunting around a guess as to areas with issues are the following:

```glsl
// FIX 1: Add safety to GGX Smith G1 function
float mx_ggx_smith_G1(float cosTheta, float alpha)
{
    float cosTheta2 = mx_square(cosTheta);
    float tanTheta2 = (1.0 - cosTheta2) / max(cosTheta2, 1e-6); // PREVENT DIVISION BY ZERO
    float alpha2 = mx_square(max(alpha, 0.001)); // ENFORCE MINIMUM ALPHA
    return 2.0 / (1.0 + sqrt(1.0 + alpha2 * tanTheta2));
}

// FIX 2: Add safety to GGX Smith G2 function  
float mx_ggx_smith_G2(float NdotL, float NdotV, float alpha)
{
    float alpha2 = mx_square(max(alpha, 0.001)); // ENFORCE MINIMUM ALPHA
    
    // ADD SAFETY CHECKS FOR NEAR-ZERO DOT PRODUCTS
    NdotL = max(NdotL, 1e-6);
    NdotV = max(NdotV, 1e-6);
    
    float lambdaL = sqrt(alpha2 + (1.0 - alpha2) * mx_square(NdotL));
    float lambdaV = sqrt(alpha2 + (1.0 - alpha2) * mx_square(NdotV));
    
    // PREVENT DIVISION BY ZERO
    float denominator = lambdaL * NdotV + lambdaV * NdotL;
    return 2.0 * NdotL * NdotV / max(denominator, 1e-8);
}

// FIX 3: Add safety to the main BSDF response calculation
// In mx_dielectric_bsdf and mx_generalized_schlick_bsdf, find this line:
bsdf.response = D * F * G * comp * safeTint * closureData.occlusion * weight / (4.0 * NdotV);

// CHANGE TO:
float denominator = 4.0 * max(NdotV, 1e-6);
[...]

PR #2792, opened 2026-02-27
Use high precision in generated ESSL

kwokcb:
## Change

- Specify `highp` to use 32-bit precision with ESSL

Fixes: #2655, #2654.

### Details

High Precision Change

The minimum ESSL version required is 3.0 , where `highp` precision is mandatory in fragment shaders, so the switch is safe.
Instead of IEEE 754 half‑precision (16‑bit) we now use single‑precision (32‑bit). This avoids cases of underflow/overflow and reduces accumulated math errors (for Std surface, OpenPBR and glTF).

### Results

I have pushed locally to: https://kwokcb.github.io/MaterialX/

| Viewer on Samsung | Viewer on Desktop |
| :--: | :--: |
|  image  image |   image |

Specific testing for specular roughness shows that this is now working:

| Viewer on Samsung | Test | 
| :--: | :--: |
| image | Metalness up and specular roughness to 0 |
|  screenshot_20260228_095800_duckduckgo_720 | Same test glTF|
|  screenshot_20260228_095644_duckduckgo_720 | Same test OpenPBR |
 |  screenshot_20260228_095603_duckduckgo_720 | Same test USDPreview |

### DIFF
--- a/libraries/stdlib/genglsl/lib/mx_math.glsl
+++ b/libraries/stdlib/genglsl/lib/mx_math.glsl
@@ -1,8 +1,4 @@
-#ifdef GL_ES
-    #define M_FLOAT_EPS 0.0001
-#else
-    #define M_FLOAT_EPS 1e-8
-#endif
+#define M_FLOAT_EPS 1e-8
 
 #define mx_mod mod
 #define mx_inverse inverse


libraries/stdlib/genglsl/lib/mx_math.glsl
```diff
@@ -1,4 +1,8 @@
-#define M_FLOAT_EPS 1e-8
+#ifdef GL_ES
```
kwokcb: This does not affect Metal / iOS as there is a separate mx_math.metal file. On iOS it seems it's always 32-bit precision so the issues do not show up.

jstone-lucasfilm:
@kwokcb I think you're right on the mark with your change from `mediump` to `highp` in ESSL, which aligns with modern expectation in ESSL 3.0, and I'd have no objections to merging that part of your proposal.

I'm less convinced by the universal raising of M_FLOAT_EPS from `1e-8` to `1e-4`, which would affect all of our Fresnel and energy compensation computations at grazing angles, in addition to the roughness clamping that originally motivated this adjustment.

What are your thoughts on the [earlier suggestion](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2654#issuecomment-3481234513) to provide a shader generator option for `MIN_SAFE_ALPHA`?  That still seems like a safer approach to me, if it addresses the roughness errors that you're seeing on Android platforms.

kwokcb:
The issue with the approach is that there is no way to detect when your generating the code where it's going to be deployed to, and I'm not sure if these are the only places that can be affected. 

So I propose to either:

1. Just remove this second change or

2. Add in this:
 ```glsl
#ifdef GL_ES
    #ifdef GL_FRAGMENT_PRECISION_HIGH
        #define MX_FLOAT_EPS 1e-8
    #else
       #define MX_FLOAT_EPS 1e-4
   #endif
#else
    #define M_FLOAT_EPS 1e-8
#endif
```
This would not affect any device which guarantees `highp` (like iOS), but also
provides a safe-guard for Android variants.

I can remove the second change and see how it goes. If it's still not good, I can put up a separate PR for a better tolerance approach, or add 2. into this PR if your okay with it. How does that sound ?

kwokcb:
The latest deployment seems good with just the `highp` fix. Added new snapshot which explicitly tests specular roughness value issue.

Issue #2794, opened 2026-03-02
GraphEditor : Crashes on document open after nodegraph change


kwokcb:
## Symptoms

I can't reliably reproduce this but if I open a file and go into nodegraph and change layout and back out
and load in another file it can sometimes crash. 

Seems something is not being cleaned up properly on file new as a guess.

This is with SHA `bc1175de637bf663f70f9c3085071310122d1b68` on local Windows build.

### Crash Info

Seems to all hinge on getting document root.

One possible stack
```cmd
>	MaterialXGraphEditor.exe!MaterialX_v1_39_5::Element::getRoot() Line 331	C++
 	MaterialXGraphEditor.exe!MaterialX_v1_39_5::Element::getDocument() Line 344	C++
 	MaterialXGraphEditor.exe!MaterialX_v1_39_5::Node::getNodeDef(const std::string & target, bool allowRoughMatch) Line 88	C++
 	MaterialXGraphEditor.exe!Graph::propertyEditor() Line 3368	C++
 	MaterialXGraphEditor.exe!Graph::graphButtons() Line 3093	C++
 	MaterialXGraphEditor.exe!Graph::drawGraph(ImVec2 mousePos) Line 4033	C++
 	MaterialXGraphEditor.exe!main(int argc, char * const * argv) Line 291	C++
```

Another possible stack:
```cmd
 	MaterialXGraphEditor.exe!MaterialX_v1_39_5::Element::getRoot() Line 331	C++
>	MaterialXGraphEditor.exe!MaterialX_v1_39_5::Element::getDocument() Line 344	C++
 	MaterialXGraphEditor.exe!MaterialX_v1_39_5::ValueElement::getValue() Line 666	C++
 	MaterialXGraphEditor.exe!Graph::showPropertyEditorValue(std::shared_ptr<UiNode> node, std::shared_ptr<MaterialX_v1_39_5::Input> input, const MaterialX_v1_39_5::UIProperties & uiProperties) Line 760	C++
 	MaterialXGraphEditor.exe!Graph::propertyEditor() Line 3496	C++
 	MaterialXGraphEditor.exe!Graph::graphButtons() Line 3093	C++
 	MaterialXGraphEditor.exe!Graph::drawGraph(ImVec2 mousePos) Line 4033	C++
 	MaterialXGraphEditor.exe!main(int argc, char * const * argv) Line 291	C++
```

```c++
ConstElementPtr Element::getRoot() const
{
[...]

PR #2795, opened 2026-03-02
Add missing initialization on document load in Graph Editor

kwokcb:
## Fixes

Fixes: #2794.
There are references to the previous MaterialX documents on file load

### Change

- Clear `_currUiNode` : Would hold onto to last selection if any. This seems the root cause of the property editor crash.
- Clear `_currRenderNode` : Should also be cleared to prevent trying to render last node. This seems to also fix having extra renders when load and select nodegraph.


kwokcb:
@jstone-lucasfilm . Maybe this is due to recent refactoring ? Anyways this should make it safe in general now as no data members should hold on to previous MTLX data.

PR #2796, opened 2026-03-03
Align `UsdPreviewSurface` opacity with upcoming specification

jstone-lucasfilm:
This changelist updates the opacity behavior in the MaterialX graph definition of `UsdPreviewSurface`, aligning it with upcoming revisions to the `UsdPreviewSurface` specification.

Our Pixar colleagues have shared the following language from the next release of the `UsdPreviewSurface` specification, providing additional clarity on the edge case behavior of opacity:

- `opacity`: When opacity is 1.0 then the prim is fully opaque, if it is smaller than 1.0 then the prim is translucent, when it is 0 the prim is fully transparent. Fully transparent objects will still receive a lighting response when the opacityMode is transparent. When the opacityMode is set to presence the lighting response will be scaled by the opacity.
- `opacityMode`: This input dictates how materials with zero opacity are interpreted and takes one of two inputs: transparent or presence. When the value is transparent materials with zero opacity still receive a lighting response as, for example, perfectly clear glass still has a specular response. However, when set to presence these materials will not receive any lighting and thus they always serve as cutouts. When using presence and semi-opaque materials the entire lighting response, including specular, will be scaled by opacity to allow for a smooth transition between materials with zero and non-zero opacity.
- `opacityThreshold`: The opacityThreshold input is useful for creating geometric cut-outs based on the opacity input. A value of 0.0 indicates that no masking is applied to the opacity input, while a value greater than 0.0 indicates that rendering of the surface is limited to the areas where the opacity is greater or equal to that value. A classic use of opacityThreshold is to create a leaf from an opacity input texture, in that case the threshold determines the parts of the opacity texture that will be fully transparent and not receive lighting. Note that when opacityThreshold is greater than zero, the opacity values less than the opacityThreshold will not be rendered, and the opacity values greater than or equal to the opacityThreshold will be fully visible. Thus, the opacityThreshold serves as a switch for how the opacity input is interpreted; this “translucent or masked” behavior is common in engines and renderers, and makes the UsdPreviewSurface easier to interchange. Recreating a translucent material that also provides a zero opacity-based mask, can be achieved by setting opacityThreshold to zero and opacityMode to presence.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,9 @@ on:
     - cron: '30 10 * * *'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
 
   build:
@@ -130,7 +133,7 @@ jobs:
 
     steps:
     - name: Sync Repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       with:
         submodules: recursive
 
@@ -180,7 +183,7 @@ jobs:
 
     - name: Install Python ${{ matrix.python }}
       if: matrix.python != 'None'
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
       with:
         python-version: ${{ matrix.python }}
         architecture: ${{ matrix.architecture }}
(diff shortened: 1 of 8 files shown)


sherholz:
@jstone-lucasfilm, how would the `opacityMode` influence the behavior of the `opacityThreshold`?
- 'presence':  it acts like a clear presence/opacity cut off as described. 
- 'transparent': above the threshold opaticity descriptes the transmittance, but what is happening below the threshold? I then the optacity value suddenly a cutoff mask?

I think adding formulas or diagrams would help make this definition much clearer.

jstone-lucasfilm:
@sherholz I'd have no objections to updating the specification language for additional clarity, but I should note that the language above is copied directly from Pixar's work-in-progress specification, and we're not actually authoring new language in this PR.

For this PR, our main goal is to make sure we're correctly _interpreting_ Pixar's specification language, without making any changes to it.

klucknav:
Due to feedback we are planing to update the wording as follows:

- `opacity`: When opacity is 1.0 then the prim is fully opaque, if it is smaller than 1.0 then the prim is translucent, when it is 0 the prim is fully transparent. Fully transparent objects will still receive a lighting response when the opacityMode is transparent. When the opacityMode is set to presence the lighting response will be scaled by the opacity.
- `opacityMode`: This input dictates how materials with opacity less than one are interpreted and accepts one of two possible values: transparent or presence. When the value is transparent, opacity reduces the contribution of the non-specular lighting components in favor of transparency, while the specular response remains at full weight. This models materials such as glass, where even perfectly clear glass still produces a specular reflection. When set to presence, opacity scales the entire lighting response, including specular, so the material response fades proportionally as opacity decreases. In this mode the surface transitions toward full transparency as opacity approaches zero, behaving as if the object itself is disappearing rather than remaining optically present through specular reflection.
- `opacityThreshold`: The opacityThreshold input is useful for creating geometric cut-outs based on the opacity input. A value of 0.0 indicates that no masking is applied to the opacity input, while a value greater than 0.0 indicates that rendering of the surface is limited to the areas where the opacity is greater or equal to that value. A classic use of opacityThreshold is to create a leaf from an opacity input texture, in that case the threshold determines the parts of the opacity texture that will be fully transparent and not receive lighting. Note that when opacityThreshold is greater than zero, regardless of the opacityMode, the opacity values less than the opacityThreshold will not be rendered, and the opacity values greater than or equal to the opacityThreshold will be fully visible. Thus, the opacityThreshold serves as a switch for how the opacity input is interpreted; this “translucent or masked” behavior is common in engines and renderers, and makes the UsdPreviewSurface easier to interchange. Recreating a translucent material that also provides a zero opacity-based mask, can be achieved by setting opacityThreshold to zero and opacityMode to presence.

This is purely a wording change to hopefully make the behavior clearer, and does not reflect the need to make any further changes to this PR.

PR #2798, opened 2026-03-04
Web Viewer UI Improvements

kwokcb:
## Changes

- Add FPS timer. Toggles via T / t(imer). ( SHIFT-f / F already used for frame-grab so don't want to mix things up ).
- Make UI consistent for font, and color
- Make menus not be file names but try and parse out useful words. Remove extensions.
- Prevent UI from overlapping and wrapping when page is narrow (mobile and desktop).

## Results

| Material labels | Geometry labels |
| :--: | :--: |
| image  |  image |

| With background | Without background 
| :--: | :--: |
| image | image |

- UI overlap guards

| Mobile | Desktop | 
| :--: | :--: |
| Avoid labels + menu items from wrapping and overwriting each other when narrow. Isn't really responsive but don't want to change current fixed formatting. (e.g. iPhone 12 Pro Sim) | image |
| Prevent property editor / menu overlapping as much as possible | image |

### DIFF
--- a/javascript/MaterialXView/index.ejs
+++ b/javascript/MaterialXView/index.ejs
@@ -8,8 +8,42 @@
     <style>
       body {
         margin: 0;
-        font-family: Arial
+        font-family: "Segoe UI", Roboto, Arial, sans-serif;
+        font-size: 12px;
       }
+
+      /* Responsive select menus */
+      select {
+        max-width: 40vw;
+        width: 100%;
+        box-sizing: border-box;
+        white-space: normal;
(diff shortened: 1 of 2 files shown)


kwokcb:
@jstone-lucasfilm , this is mostly style sheets and text formatting. The FPS is a small change to give some rudimentary performance number. Some shaders are very slow (separate issue) but relative perf can be given as a number at least.

jstone-lucasfilm (approved):
This looks like a worthwhile improvement, thanks @kwokcb!

PR #2800, opened 2026-03-05
Fix inconsistent background color in MaterialXView on Metal

### LINKED ISSUE #2606: Inconsistent background color in MaterialXViewer

aabhinavvvvvvv:
## Summary

Fixes #2606

The `MaterialXView` background color appeared lighter on Mac (Metal) than on Windows (OpenGL) due to a color space mismatch in the Metal rendering pipeline.

## Root Cause

The Metal renderer uses an `MTLPixelFormatRGBA16Float` (linear) framebuffer. macOS applies gamma correction when displaying linear float values. The background color stored in `m_background` comes from `DEFAULT_SCREEN_COLOR_SRGB = (0.3, 0.3, 0.32)` — sRGB-encoded values. Passing these directly to `setClearColor` on a linear framebuffer causes macOS to gamma-correct them on display (`sRGB(0.3) ≈ 0.58`), producing a noticeably lighter background.

On Windows/OpenGL, `glClearColor` writes values into a non-linear framebuffer which the display interprets as sRGB — so the appearance is correct.

The previous code also had a `linearToSrgb` conversion applied in the **wrong direction**, converting already-sRGB values further into sRGB and making the discrepancy worse.

## Fix

Replace `linearToSrgb` with `srgbToLinear` in `RenderPipelineMetal.mm`. This converts the sRGB background values to their linear equivalents before storing in the float framebuffer. macOS then gamma-corrects them back on display, producing the intended `(0.3, 0.3, 0.32)` gray — matching the Windows/OpenGL output.

This aligns with the finding in the issue that using `DEFAULT_SCREEN_COLOR_LIN_REC709` (which is `DEFAULT_SCREEN_COLOR_SRGB.srgbToLinear()`) resolves the inconsistency.

### DIFF
--- a/source/MaterialXView/RenderPipelineMetal.mm
+++ b/source/MaterialXView/RenderPipelineMetal.mm
@@ -472,17 +472,14 @@ void MetalRenderPipeline::renderFrame(void* color_texture, int shadowMapSize, co
     // background color must be converted to linear before being set as the clear color.
     // On OpenGL (Windows), the background is stored directly in a non-linear framebuffer
     // and no conversion is needed.
-    auto srgbToLinear = [](float srgb) -> float {
-        if (srgb <= 0.04045f)
-            return srgb / 12.92f;
-        else
-            return std::pow((srgb + 0.055f) / 1.055f, 2.4f);
-    };
+    mx::Color3 bgLinear = mx::Color3(_viewer->m_background[0],
+                                     _viewer->m_background[1],
+                                     _viewer->m_background[2]).srgbToLinear();
 
     [renderpassDesc.colorAttachments[0] setClearColor:MTLClearColorMake(
-                                        srgbToLinear(_viewer->m_background[0]),
-                                        srgbToLinear(_viewer->m_background[1]),
-                                        srgbToLinear(_viewer->m_background[2]),
+                                        bgLinear[0],
+                                        bgLinear[1],
+                                        bgLinear[2],
                                         _viewer->m_background[3])];
     [renderpassDesc.colorAttachments[0] setLoadAction:MTLLoadActionClear];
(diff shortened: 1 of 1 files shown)


source/MaterialXView/RenderPipelineMetal.mm
```diff
@@ -467,10 +467,22 @@
     {
         [renderpassDesc.colorAttachments[0] setTexture:MTL(currentFramebuffer())->getColorTexture()];
     }
+    // The Metal framebuffer stores linear color values (MTLPixelFormatRGBA16Float).
+    // macOS applies gamma correction when displaying linear float values, so the sRGB
+    // background color must be converted to linear before being set as the clear color.
+    // On OpenGL (Windows), the background is stored directly in a non-linear framebuffer
+    // and no conversion is needed.
+    auto srgbToLinear = [](float srgb) -> float {
```
jstone-lucasfilm: Instead of this custom lambda, could we use the existing [Color3::srgbToLinear](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXCore/Types.cpp#L63) in MaterialXCore, so that this math remains centralized?

jstone-lucasfilm (commented):
This looks like a valuable improvement, @aabhinavvvvvvv, and I had just one recommendation about using the existing Color3::srgbToLinear instead of a custom lambda.

jstone-lucasfilm (approved):
Thanks for this improvement, @aabhinavvvvvvv!

PR #2803, opened 2026-03-08
AOUSD formatting for Specification Proposals

jstone-lucasfilm:
This changelist applies the formatting conventions of the AOUSD Materials Working Group to `MaterialX.Proposals.md`, aligning its node signatures with the latest conventions of the Standard Nodes, PBR Nodes, and NPR Nodes specification documents.

### DIFF
--- a/documents/Specification/MaterialX.Proposals.md
+++ b/documents/Specification/MaterialX.Proposals.md
@@ -169,16 +169,6 @@ Inheritance of material-type custom nodes is also allowed, so that new or change
 ### `tokenvalue`
 A constant "interface token" value, may only be connected to &lt;token>s in nodes, not to &lt;input>s.
 
-|Port   |Description                         |Type                                    |Default |
-|-------|------------------------------------|----------------------------------------|--------|
-|`value`|The value that will be sent to `out`|float, colorN, vectorN, boolean, integer|__zero__|
-|`out`  |Output: `value`                     |Same as `value`                         |__zero__|
-
-|Port   |Description                         |Type           |Default|
-|-------|------------------------------------|---------------|-------|
-|`value`|The value that will be sent to `out`|matrixNN       |__one__|
-|`out`  |Output: `value`                     |Same as `value`|__one__|
-
 |Port   |Description                         |Type            |Default  |
 |-------|------------------------------------|----------------|---------|
 |`value`|The value that will be sent to `out`|string, filename|__empty__|
@@ -386,17 +376,6 @@ Output the value of one of two input streams, according to whether the value of 
 
 
 ### Channel Nodes
-
-<a id="node-extractrowvector"> </a>
-
-### `extractrowvector`
-Extract the specified row vector number from a matrixN stream.
-
-|Port   |Description                                                                       |Type   |Default|
-|-------|----------------------------------------------------------------------------------|-------|-------|
-|`in`   |The input matrix                                                                  |matrixN|       |
-|`index`|The row number to extract; should be 0-2 for matrix33 streams, or 0-3 for matrix44|integer|0      |
-|`out`  |Output: the extracted row vector                                                  |vectorN|       |
 
 <a id="node-separatecolor4"> </a>
 


documents/Specification/MaterialX.Proposals.md
```diff
@@ -208,89 +233,131 @@ In #1201 it was decided that separate periodic versions of all of the noises is
 
 ### Geometric Nodes
 
-<a id="node-bump"> </a>
-
-* **`bump`**: Existing node, proposal to add a vector3 `bitangent` input
-
-Note: when &lt;geompropvalueuniform> is added, the text in the first paragraph of the Specification about Node Inputs should be revised to include "&lt;geompropvalueuniform>" as an example of "or any other node whose output is explicitly declared to be uniform".
```
dbsmythe, line 215: This section is removed with no replacement- do we feel the information here is no longer relevant?

documents/Specification/MaterialX.Proposals.md
```diff
@@ -208,89 +233,131 @@ In #1201 it was decided that separate periodic versions of all of the noises is
...
-    * `coneangle` (float): the half-angle of a cone about the surface normal, within which geometric surface features are considered as potential occluders.  The unit for this input is degrees, and its default value is 90.0 (full hemisphere).
-    * `maxdistance` (float): the maximum distance from the surface point at which geometric surface features are considered as potential occluders.  Defaults to 1e38, e.g. "unlimited".
+### `ambientocclusion`
+Compute the ambient occlusion at the current surface point, returning a scalar value between 0 and 1.  Ambient occlusion represents the accessibility of each surface point to ambient lighting, with larger values representing greater accessibility to light.
+
+|Port         |Description                                      |Type |Default|
+|-------------|-------------------------------------------------|-----|-------|
+|`coneangle`  |The half-angle of the occlusion cone, in degrees |float|90.0   |
```
dbsmythe, line 239: The original text clarified that the default value of 90.0 represented the "full hemisphere"; while that may be obvious to those who read carefully, it's easy to see this and wonder why the default isn't 180 to cover the full hemisphere.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -208,89 +233,131 @@ In #1201 it was decided that separate periodic versions of all of the noises is
...
-    * `inz` (float or colorN): the image to be projected in the direction from the +Z axis back toward the origin.  Default is 0 in all channels.
-    * `position` (vector3): a spatially-varying input specifying the 3D position at which the projection is evaluated.  Default is to use the current 3D object-space coordinate.
-    * `normal` (vector3): a spatially-varying input specifying the 3D normal vector used for blending.  Default is to use the current object-space surface normal.
-    * `blend` (float): a 0-1 weighting factor for blending the three axis samples using the geometric normal, with higher values giving softer blending.  Default is 1.0.
-    * `filtertype` (uniform string): the type of texture filtering to use; standard values include "closest" (nearest-neighbor single-sample), "linear", and "cubic".  If not specified, an application may use its own default texture filtering method.
+### `triplanarblend`
+Samples data from three inputs, and projects a tiled representation of the images along each of the three respective coordinate axes, computing a weighted blend of the three samples using the geometric normal.
```
dbsmythe, line 276: It may be worth noting somewhere in here that if this is implemented, it's intended that the existing triplanarprojection node would be reimplemented to call triplanarblend with its three resolved images.

documents/Specification/MaterialX.Proposals.md
jstone-lucasfilm: Good point, @dbsmythe, and I've added a sentence to clarify this.

documents/Specification/MaterialX.Proposals.md
jstone-lucasfilm: That's a good point as well, and although this language didn't exist in the previous version, I've now added it to this PR.

documents/Specification/MaterialX.Proposals.md
jstone-lucasfilm: Oh, I see that there was an independent sentence about `geompropvalueuniform` above, and I should check whether that has been fulfilled as well, or whether it is still outstanding work.

jstone-lucasfilm: As it turns out, this proposal has already been fulfilled in the Standard Nodes specification and codebase, where there's now a bitangent input to bump. That was the motivation for removing it entirely, though I likely should have called this out in the PR description body.

Similarly, the `geompropvalueuniform` node is now supported in the Standard Nodes specification, so the extra sentence above was likely just a leftover from that earlier work.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -166,36 +166,61 @@ Inheritance of material-type custom nodes is also allowed, so that new or change
...
 
+|Port   |Description                         |Type                                    |Default |
+|-------|------------------------------------|----------------------------------------|--------|
+|`value`|The value that will be sent to `out`|float, colorN, vectorN, boolean, integer|__zero__|
+|`out`  |Output: `value`                     |Same as `value`                         |__zero__|
+
+|Port   |Description                         |Type           |Default|
+|-------|------------------------------------|---------------|-------|
+|`value`|The value that will be sent to `out`|matrixNN       |__one__|
+|`out`  |Output: `value`                     |Same as `value`|__one__|
```
dbsmythe: Thinking about this- because Tokens are used exclusively for Image Filename Substitutions, I can't think of any situation where a matrix value (or vector or color or boolean for that matter) would be useful.

documents/Specification/MaterialX.Proposals.md
dbsmythe: Sounds good, thanks

documents/Specification/MaterialX.Proposals.md
```diff
@@ -302,26 +373,41 @@ Note: when &lt;geompropvalueuniform> is added, the text in the first paragraph o
...
-    * `in` (matrix<em>N</em>): the input value or nodename
-    * `index` (integer): the row number to extract, should be 0-2 for matrix33 streams, or 0-3 for matrix44 streams.
+### `extractrowvector`
+Extract the specified row vector number from a matrixN stream.
+
+|Port   |Description                                                                       |Type   |Default|
+|-------|----------------------------------------------------------------------------------|-------|-------|
+|`in`   |The input matrix                                                                  |matrixN|       |
+|`index`|The row number to extract; should be 0-2 for matrix33 streams, or 0-3 for matrix44|integer|0      |
+|`out`  |Output: the extracted row vector                                                  |vectorN|       |
```
dbsmythe: With the new addition of matrix types to the existing &lt;extract> node to extract a row vector, we can now completely remove this from the Proposals document.

dbsmythe:
Made a couple more comments, otherwise this is looking good.

documents/Specification/MaterialX.Proposals.md
jstone-lucasfilm: Good point, and I've removed these two tables.

documents/Specification/MaterialX.Proposals.md
jstone-lucasfilm: Good point, and I've removed this entry from the document.

PR #2804, opened 2026-03-09
Improve material name handling in shader generation

jreichel-nvidia:
Convert names of shader graph nodes early to identifiers to increase the chance of a name without suffixes. This allows to render all materials from the testsuite and examples with MDL without the need to skip some of them.


jreichel-nvidia:
I believe the new version inside the constructor is cleaner. Let me know if I should restore the old version though.

PR #2806, opened 2026-03-09
Fix missing origin annotations for MDL

jreichel-nvidia:
These are based on the node paths and were missing for the second input of a unit transform node and input sockets of node graphs.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -51,6 +51,7 @@ void ShaderGraph::addInputSockets(const InterfaceElement& elem, GenContext& cont
         {
             inputSocket->setUniform();
         }
+        inputSocket->setPath(input->getNamePath());
         GeomPropDefPtr geomprop = input->getDefaultGeomProp();
         if (geomprop)
         {
@@ -384,6 +385,12 @@ void ShaderGraph::addUnitTransformNode(ShaderInput* input, const UnitTransform& 
         shaderInput->setPath(input->getPath());
         shaderInput->setUnit(input->getUnit());
         shaderInput->setColorSpace(input->getColorSpace());
+
+        ShaderInput* shaderInput2 = unitTransformNode->getInput(1);
+        if (shaderInput2)
+        {
+            shaderInput2->setPath(input->getPath());
+        }
 
         if (input->isBindInput())
         {
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -386,12 +386,6 @@ void ShaderGraph::addUnitTransformNode(ShaderInput* input, const UnitTransform& 
         shaderInput->setUnit(input->getUnit());
         shaderInput->setColorSpace(input->getColorSpace());
 
-        ShaderInput* shaderInput2 = unitTransformNode->getInput(1);
-        if (shaderInput2)
-        {
-            shaderInput2->setPath(input->getPath());
-        }
-
         if (input->isBindInput())
         {
             ShaderOutput* oldConnection = input->getConnection();


jreichel-nvidia:
On second thought I'm not that sure anymore about the change for the unit transform node. Do we want multiple nodes with the same path? I don't know where paths are used elsewhere and would leave that decision to others.

If the decision is that multiple nodes with the same path is ok/the correct thing to do, then we might want to avoid for the MDL backend multiple parameters with the same origin annotations. These are used by integrations to map arguments for the MaterialX material to parameters of the generated MDL code. The 2nd input of the unit transform node does not have a direct counterpart in the original material. Mapping the value from that path would be logically wrong (and doesn't work anyway because of type mismatches). Instead of asking all integrations to add code for this special case it is probably better not to emit the origin annotation in this case (as before my change).

jstone-lucasfilm:
@jreichel-nvidia This single-line change looks fine to me, assuming it addresses the issue you're encountering in MDL.  Can you provide a quick description of the case that it addresses?

jreichel-nvidia:
Actually I have difficulties finding the commonality of the affected cases. In most cases the path is set somewhere later, but in a few cases these locations are not hit. Looks like it has to do with nodedef nodes, but I suppose there is something special/different with those cases that I don't see.

Here is the list of the generated MDL materials with the affected paths:
```
resources/Materials/TestSuite/libraries/metal/libraries/metal_definition/NG_TestMetal_out.genmdl.mdl
- ND_TestMetal/MetalF0
- ND_TestMetal/SurfaceRoughness
- ND_TestMetal/SurfaceAlbedo
- ND_TestMetal/SurfaceAnisotropy
- ND_TestMetal/SurfaceNormal
- ND_TestMetal/SurfaceRotation
- ND_TestMetal/SurfaceCutout
resources/Materials/TestSuite/pbrlib/bsdf/mix_bsdf/IMP_substrateshader_out.genmdl.mdl
- ND_substrateshader/albedo
- ND_substrateshader/subsurface
- ND_substrateshader/subsurface_weight
- ND_substrateshader/emission
- ND_substrateshader/opacity
resources/Materials/TestSuite/pbrlib/multioutput/multishaderoutput/NG_multi_burley_out.genmdl.mdl
- ND_multi/diffuse_color
resources/Materials/TestSuite/pbrlib/multioutput/multishaderoutput/NG_multi_dielectric_out.genmdl.mdl
- ND_multi/diffuse_color
resources/Materials/TestSuite/pbrlib/surfaceshader/shader_ops/NG_checker_float_out.genmdl.mdl
- ND_checker_float/scale
resources/Materials/TestSuite/stdlib/nodegraphs/nodegraph_nodegraph/NG_upstream_graph_nd_graph_out_image.genmdl.mdl
- ND_upstream_graph/nd_file
resources/Materials/TestSuite/stdlib/nodegraphs/nodegraph_nodegraph/NG_upstream_graph_nd_graph_out_image2.genmdl.mdl
- ND_upstream_graph/nd_file2
resources/Materials/TestSuite/stdlib/nodegraphs/nodegraph_nodegraph/upstream_graph_graph_out_image.genmdl.mdl
- upstream_graph/file
resources/Materials/TestSuite/stdlib/nodegraphs/nodegraph_nodegraph/upstream_graph_graph_out_image2.genmdl.mdl
- upstream_graph/file2
resources/Materials/TestSuite/stdlib/texture/token_graph/Parent_Token_Graph_out.genmdl.mdl
- Parent_Token_Graph/Image_Filename
resources/Materials/TestSuite/stdlib/version/multiple_version_test/NG_myimage_color3_v1_out.genmdl.mdl
- ND_myimage_color3_v1/file
resources/Materials/TestSuite/stdlib/version/multiple_version_test/NG_myimage_color3_v2_out.genmdl.mdl
- ND_myimage_color3_v2/file```

jstone-lucasfilm (approved):
This looks fine to me, thanks @jreichel-nvidia!

PR #2808, opened 2026-03-09
Use directional albedo for prefiltered environment lighting

shill-lucasfilm:
While working on other changes, I noticed that the GLSL prefiltered environment lighting code path was only performing a single evaluation of the `F` and `G` terms, leading to innaccurate results for rougher materials. To get an approximation of the correct behaviour, we need to account for the particular Fresnel function in use. To that end, I've introduced a new variation of `mx_ggx_dir_albedo()`.

Here are some roughness wedges for the different cases:

**Thin Film** (resources/Materials/TestSuite/pbrlib/bsdf/thin_film_bsdf.mtlx)
thin_film_prefilter_before
Before
thin_film_prefilter_after
After
thin_film_fis
FIS (reference)

**Dielectric** (resources/Materials/TestSuite/pbrlib/bsdf/dielectric.mtlx)
dielectric_prefilter_before
Before
dielectric_prefilter_after
After
dielectric_fis
FIS (reference)

**Conductor** (resources/Materials/TestSuite/pbrlib/bsdf/conductor.mtlx)
conductor_prefilter_before
Before
conductor_prefilter_after
After
conductor_fis
FIS (reference)

**Schlick** (resources/Materials/TestSuite/pbrlib/bsdf/generalized_schlick.mtlx)
generalized_shlick_prefilter_before
Before
generalized_shlick_prefilter_after
After
generalized_shlick_fis
FIS (reference)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,9 @@ on:
     - cron: '30 10 * * *'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
 
   build:
@@ -130,7 +133,7 @@ jobs:
 
     steps:
     - name: Sync Repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       with:
         submodules: recursive
 
@@ -180,7 +183,7 @@ jobs:
 
     - name: Install Python ${{ matrix.python }}
       if: matrix.python != 'None'
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
       with:
         python-version: ${{ matrix.python }}
         architecture: ${{ matrix.architecture }}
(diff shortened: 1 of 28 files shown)


libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
```diff
@@ -482,6 +482,35 @@ vec3 mx_compute_fresnel(float cosTheta, FresnelData fd)
     }
 }
 
+// Directional albedo accounting for different Fresnel functions.
```
shill-lucasfilm, line 485: It seemed natural to put this after `mx_compute_fresnel` for maintenance reasons (a new model would need to be added to both), but it could also go next to the other `mx_ggx_dir_albedo` implementations. That said, if we add dedicated `mx_ggx_dir_albedo` functions in the future for each model then they could go next to the existing `mx_ggx_dir_albedo` function and this one could stay here.

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
```diff
@@ -482,6 +482,35 @@ vec3 mx_compute_fresnel(float cosTheta, FresnelData fd)
...
+    {
+        float F0 = mx_ior_to_f0(fd.ior.x);
+        return mx_ggx_dir_albedo(NdotV, alpha, F0, 1.0);
+    }
+    else if (fd.model == FRESNEL_MODEL_CONDUCTOR)
+    {
+        vec3 F0 = mx_fresnel_conductor(1.0, fd.ior, fd.extinction);
+        return mx_ggx_dir_albedo(NdotV, alpha, F0, vec3(1.0));
+    }
+    else
```
shill-lucasfilm, line 508: We'd benefit from dedicated directional albedo tables / fits for each of these models but for now I think this is a reasonable compromise.

shill-lucasfilm:
@jstone-lucasfilm: thanks! The results can be slightly less accurate at low roughness when the shape of `mx_fresnel_dielectric` or `mx_fresnel_conductor` isn't precisely captured by the Schlick approximation used by the existing `mx_ggx_dir_albedo` function, so that's where dedicated tables/fits for these cases would be most beneficial – something I plan to look at in the future.

PR #2809, opened 2026-03-09
Initial VDF support in hardware shading

jstone-lucasfilm:
This changelist adds initial support for Volume Distribution Functions (VDFs) in hardware shader generation, giving the VDF type its own struct, and approximating volumetric light absorption for the transmissive term.  The following specific changes are included:

- Add a dedicated VDF struct to GLSL, MSL, and Slang, replacing the previous use of the BSDF struct as an alias for the VDF type.
- Add a simple approximation of volumetric absorption to `mx_anisotropic_vdf`, computing throughput using Beer's law in the transmissive lighting term.
- Fix `mx_layer_vdf` to multiply the surface response and throughput by the volume throughput, replacing the previous placeholder math.

### DIFF
--- a/libraries/pbrlib/genglsl/mx_anisotropic_vdf.glsl
+++ b/libraries/pbrlib/genglsl/mx_anisotropic_vdf.glsl
@@ -1,6 +1,11 @@
 #include "lib/mx_closure_type.glsl"
 
-void mx_anisotropic_vdf(ClosureData closureData, vec3 absorption, vec3 scattering, float anisotropy, inout BSDF bsdf)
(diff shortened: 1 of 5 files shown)


jstone-lucasfilm:
To provide visual intuition for the impact of this change, here are before and after renders for a simple OpenPBR Tinted Glass material in the MaterialX Viewer:

OpenPBR Tinted Glass (Before):
OpenPBR_GlassTinted_Old

OpenPBR Tinted Glass (After):
OpenPBR_GlassTinted_New

shill-lucasfilm (approved):
LGTM!

PR #2812, opened 2026-03-10
Add extract node support for matrix33 and matrix44 types

JHUbisoft:
Add extract node definitions for matrix33 and matrix44 types, enabling
  row vector extraction by index from matrices.

  - matrix33 extract returns vector3 (row index 0-2)
  - matrix44 extract returns vector4 (row index 0-3)

  Includes node definitions, implementations for GLSL, MDL, and OSL
  backends, and corresponding unit tests.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -2458,6 +2458,7 @@ void Graph::addLink(ed::PinId startPinId, ed::PinId endPinId)
             {
                 // Get or create the actual node's input (not the NodeDef's input)
                 mx::InputPtr nodeInput = addNodeInput(uiDownNode, pin->getInput());
+                pin->setElement(nodeInput);
 
                 // Update value to be empty
                 if (uiDownNode->getNode() && uiDownNode->getNode()->getType() == mx::SURFACE_SHADER_TYPE_STRING)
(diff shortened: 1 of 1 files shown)


dbsmythe:
In the MaterialX Proposals document https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Proposals.md#propose-stdlib-nodes something essentially identical to this is proposed for addition to the spec, although there it was called &lt;extractrowvector> which is marginally clearer.  Is there a strong feeling from the community which node naming would be preferred for this?

dbsmythe:
Also noting that if we add a new node or new signatures for an existing node, the MaterialX.Specification.md document must also be updated to keep them in sync.

JHUbisoft:
About the documentation : it would mean that extract purpose instead of extracting float from vectors, it would be more : extract a n-1 dimension subpart  from a n dimension aggregate. What do you think ? 
By the way, I have updated the standardnodes documentation consequently.

dbsmythe:
The changes to MaterialX.Specification.md look good to me; I'll let other reviewers approve the code changes.

jstone-lucasfilm:
This looks really promising, @JHUbisoft, and thanks to @dbsmythe for the specification language review.

Let's plan to take a close look at the implementations soon, so that we have the option of including this work in v1.39.5 if it's ready to merge.

jstone-lucasfilm:
@JHUbisoft The GLSL and MDL implementations look correct to me, but am I right in my observation that the `mx_extract` overloads for matrix types are missing in OSL?  Without those, I don't believe the generated OSL code will compile after this PR is merged, but let me know if I'm missing an important detail.

JHUbisoft:
> @JHUbisoft The GLSL and MDL implementations look correct to me, but am I right in my observation that the `mx_extract` overloads for matrix types are missing in OSL? Without those, I don't believe the generated OSL code will compile after this PR is merged, but let me know if I'm missing an important detail.

Oups, you are right. I have added OSL code.

JHUbisoft:
Thank you

PR #2814, opened 2026-03-10
Proposed Spec Clarification : Input Nodes / Interface References

kwokcb:
## Change

This is some additional wording to clarify the definition of "root" level input nodes
This includes clarifying usage of `interfacename` from nodegraph input interfaces.

### DIFF
--- a/documents/Specification/MaterialX.Specification.md
+++ b/documents/Specification/MaterialX.Specification.md
@@ -612,7 +612,33 @@ Unless specified otherwise, all inputs default to a value of 0 in all channels f
 
 Standard MaterialX nodes have exactly one output, while custom nodes may have any number of outputs; please see the [Custom Nodes](#custom-nodes) section for details.
 
+### Input Nodes
 
+Input nodes (&lt;input>) may be defined at the root level of a MaterialX document. The input values of these nodes are considered to be part of the interface of the document.
+
+Input ports on nodes or nodegraphs may referenced input nodes at the same scope as the node or nodegraph by specifying the name of the input node as the value of the `interfacename` attribute of the port.  
+
+```xml
+  <nodegraph name="nodegraph">
+    <input name="graph_input" type="color3" interfacename="root_input" />
+  </nodegraph>
+  <add name="add" type="color3">
+    <input name="in1" type="color3" interfacename="root_input" />
+    <input name="in2" type="color3" value="0, 0, 0" />
+  </add>
+  <input name="root_input" type="color3" value="0, 0, 0" />
+```
+
+Input nodes may not be connected to the output of any node or nodegraph.
+
+```xml
+  <add name="node_reference" type="color3" />
+  <nodegraph name="nodegraph_reference" />
+  <input name="input_node_reference" type="color3" value="0, 0, 0" />
+  <input name="input_node_1" type="color3" interfacename="input_node_reference" />
+  <input name="input_node_2" type="color3" nodename="node_reference" />
+  <input name="input_node_3" type="color3" nodegraph="nodegraph_reference" />
+```
 
 ## Node Graph Elements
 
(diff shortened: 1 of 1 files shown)


documents/Specification/MaterialX.Specification.md
```diff
@@ -1193,7 +1219,11 @@ A **compound &lt;nodegraph>** element may specify one or more child &lt;input> a
...
+A compound nodegraph provides a set of named input and output connection ports which may be referenced by its contained nodes using `interfacename` attributes, and interface token names whose values may be substituted into filenames used within the nodegraph; nodes within this &lt;nodegraph> adopt the context of that nodegraph.
+
+The &lt;input> ports of a compound graph may reference &lt;input> nodes outside the &lt;nodegraph> at the same scope as the &lt;nodegraph> itself using `interfacename` attributes, but may not reference sibling &lt;input> ports within the nodegraph.
```
kwokcb, line 1224: Only this part is new.

documents/Specification/MaterialX.Specification.md
```diff
@@ -612,7 +612,33 @@ Unless specified otherwise, all inputs default to a value of 0 in all channels f
 
 Standard MaterialX nodes have exactly one output, while custom nodes may have any number of outputs; please see the [Custom Nodes](#custom-nodes) section for details.
 
+### Input Nodes
 
+Input nodes (&lt;input>) may be defined at the root level of a MaterialX document. The input values of these nodes are considered to be part of the interface of the document.
```
kwokcb, line 617: Unsure what the "proper" name is for this so have called them "input nodes".

documents/Specification/MaterialX.Specification.md
```diff
@@ -612,7 +612,33 @@ Unless specified otherwise, all inputs default to a value of 0 in all channels f
...
+Input nodes may not be connected to the output of any node or nodegraph.
+
+```xml
+  <add name="node_reference" type="color3" />
+  <nodegraph name="nodegraph_reference" />
+  <input name="input_node_reference" type="color3" value="0, 0, 0" />
+  <input name="input_node_1" type="color3" interfacename="input_node_reference" />
+  <input name="input_node_2" type="color3" nodename="node_reference" />
+  <input name="input_node_3" type="color3" nodegraph="nodegraph_reference" />
+```
```
dbsmythe, line 641: I'm assuming that by "Input nodes", you mean "Root-level input nodes", as we're only adding Input "nodes" at the root level, other &lt;input> semantics within nodes and nodegraphs are unchanged from previous syntax.
Given that, I am confused about this example: the text says that input nodes may not be connected to the output of any node or nodegraph (which sounds correct to me), but then the example shows input nodes with exactly that kind of prohibited connection.

documents/Specification/MaterialX.Specification.md
dbsmythe: To me, a "Node" is something that would be shown as a standalone block in a graph editor; earlier in the document (within the "Nodes" section), they are referred to as "Input elements", as they are children of the node element.
If we want to support &lt;input>s as a root-level node, maybe we should clarify that "Input nodes may only be placed at the root level of a document", although this then immediately causes confusion about &lt;input> elements placed immediately within a node (as has always been allowed).
Further thought on wording is required.

documents/Specification/MaterialX.Specification.md
```diff
@@ -612,7 +612,33 @@ Unless specified otherwise, all inputs default to a value of 0 in all channels f
 
 Standard MaterialX nodes have exactly one output, while custom nodes may have any number of outputs; please see the [Custom Nodes](#custom-nodes) section for details.
 
+### Input Nodes
 
+Input nodes (&lt;input>) may be defined at the root level of a MaterialX document. The input values of these nodes are considered to be part of the interface of the document.
+
+Input ports on nodes or nodegraphs may referenced input nodes at the same scope as the node or nodegraph by specifying the name of the input node as the value of the `interfacename` attribute of the port.
```
dbsmythe, line 619: Typo: "referenced" should be "reference".

documents/Specification/MaterialX.Specification.md
```diff
@@ -612,7 +612,33 @@ Unless specified otherwise, all inputs default to a value of 0 in all channels f
...
+```xml
+  <nodegraph name="nodegraph">
+    <input name="graph_input" type="color3" interfacename="root_input" />
+  </nodegraph>
+  <add name="add" type="color3">
+    <input name="in1" type="color3" interfacename="root_input" />
+    <input name="in2" type="color3" value="0, 0, 0" />
+  </add>
+  <input name="root_input" type="color3" value="0, 0, 0" />
+```
```
dbsmythe, line 630: The ordering of elements in this example is confusing- for clarity, the "root_input" &lt;input> should be declared first, then the &lt;add>, and then the &lt;nodegraph>.

documents/Specification/MaterialX.Specification.md
dbsmythe: We may want to consider a specification-document-wide review of the usage of the terms "element" vs. "port", as sometimes &lt;input>s are referred to as "elements", sometimes (now) as "nodes", and sometimes as "ports".  I think this only applies to &lt;input>s and &lt;output>s (where the term "port" makes sense to use), not globally.

PR #2817, opened 2026-03-11
Fix MDL generation for empty materials

jreichel-nvidia:
Repro case: M_EmptySurface in surfacematerial_with_graph.mtlx

The test driver skips such materials, but they can be explicitly requested via the API and caused invalid MDL code to be generated.


jstone-lucasfilm (approved):

PR #2823, opened 2026-03-12
Implement frame node for MDL via scene data

pablode:
As discussed in Slack.


jstone-lucasfilm:
Thanks for putting together this PR, @pablode, and I'll defer to @jreichel-nvidia on reviewing this specific implementation.

jreichel-nvidia:
LGTM, except that I think 0.0 would be better default.

jstone-lucasfilm:
@jreichel-nvidia That might be a broader change than the MDL implementation in this PR, since the `frame` node declares its default return value as one:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.StandardNodes.md#frame

jreichel-nvidia:
Ok, fine. I didn't realize that 1.0 is the explicit default in the spec, just thought of zero as a general default value.

Issue #1902, opened 2024-06-24
Allow support for custom nodes that return surfacematerial data


mkuo-lucasfilm:
I came across shader compilation errors when trying to implement a new LamaSurface node (Renderman definition: https://rmanwiki.pixar.com/display/REN26/LamaSurface). The LamaSurface node has two inputs of BSDF type and returns a surfacematerial.

We saw that the MaterialX Viewer skips the LamaSurface example because of this function (https://github.com/AcademySoftwareFoundation/MaterialX/blob/328fde8c5f293c1e6d6b14df0877c533be059855/source/MaterialXGenShader/Util.cpp#L368), which requires the top-level surfacematerial to be connected to a top-level surfaceshader. We believe that this check is in place as MaterialX does not yet have logic that handles custom nodes that return surfacematerial data, as this pattern is not used in any of the examples in the MaterialX repository.

jstone-lucasfilm:
Thanks for this note, @mkuo-lucasfilm, and I'm CC'ing @niklasharrysson so that he's in the loop.

niklasharrysson:
Thanks for the report @mkuo-lucasfilm. This is a case we have not run into before, where the concept of front and back shader assignment needs to be wrapped up in a custom node. We have considered the `surfacematerial` node to be a terminal node that just handles shader assignment to geometry.

But I can see how this is needed to wrap up the `LamaSurface` node as a self-contained node. So this is something we need to address in shader generation, to support having custom nodes with outputs of `material` type.

Issue #2811, opened 2026-03-10
Lama Surface errors in MaterialX Graph Editor


HardCoreCodin:
Using MaterialX Graph Editor, when connecting any BSDF to any of Lama surface node's BSDF inputs, the viewport goes blank and the following error is logged:
```
Error in compiling fragment shader:
0(1104) : error C1503: undefined variable "closureData"
```

Later selecting something else logs:
`OpenGL error after program bind: 1281`

jstone-lucasfilm:
Thanks for the report, @HardCoreCodin, and this looks like a bug in our hardware shader generation that would be valuable to fix.

jstone-lucasfilm:
I believe this is closely related to earlier issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/1902.

jstone-lucasfilm:
Fixed in #2829.

PR #2829, opened 2026-03-17
Add HW generation support for compound material nodes

jstone-lucasfilm:
This changelist adds hardware shader generation support for compound material nodes such as LamaSurface, which are defined as node graphs with material-type outputs.  The following specific changes are included:

- Add an `HwMaterialCompoundNode` class that extends `HwSurfaceNode` for nodegraphs with material-type outputs, discovering BSDF, EDF, and opacity input name mappings by analyzing the nodegraph's internal surface node connections.
- Refactor `HwSurfaceNode` to resolve input names through virtual methods rather than hardcoded strings, and add null-safety for inputs that may not be present on compound material nodes.
- Fix the fallback condition in `getShaderNodes` to check whether shader nodes were found, rather than whether the material node has inputs.
- Fix `isTransparentSurface` to identify material nodes by output type rather than node category, so that compound material nodes are handled correctly.

Fixes https://github.com/AcademySoftwareFoundation/MaterialX/issues/1902 and https://github.com/AcademySoftwareFoundation/MaterialX/issues/2811.

### DIFF
--- a/source/MaterialXGenHw/Nodes/HwMaterialCompoundNode.cpp
+++ b/source/MaterialXGenHw/Nodes/HwMaterialCompoundNode.cpp
@@ -15,6 +15,7 @@ MATERIALX_NAMESPACE_BEGIN
 namespace
 {
 
+const string MATERIAL_TYPE = "material";
 const string SURFACE_CATEGORY = "surface";
 
 } // anonymous namespace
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
To provide a visual sense of the results, here's a MaterialX graph of Lama nodes in the Graph Editor, using a LamaSurface node for the final render:

LamaGraphExample

PR #2824, opened 2026-03-13
GPU async timing instrumentation via MX_TRACE_ASYNC

ppenenko:
## Summary

Extends the `MaterialXTrace` API with support for **async events with explicit timing**, enabling GPU render-time-per-material analysis on a dedicated Perfetto track.

- **`MX_TRACE_ASYNC` macro**: Records events with explicit start time and duration, rendered on a separate Perfetto track from the CPU thread slices
- **`AsyncTrack` enum**: Identifies async tracks (currently `GPU`, extensible for future use cases like shader compilation)
- **`Sink::asyncEvent()`**: New virtual method on the tracing sink interface, with a Perfetto implementation using custom track descriptors
- **`GpuTimerQuery`**: Helper class wrapping GL_TIME_ELAPSED queries for measuring actual GPU execution time
- **Single-frame GPU timing**: Wraps `_renderer->render()` with a timer query and emits the result as an async event

This enables visualizing GPU render time per material alongside CPU codegen/compilation traces in the Perfetto UI.

image

## Test plan

- [x] Verified C++ build with `MATERIALX_BUILD_PERFETTO_TRACING=ON` on Windows (MSVC)
- [x] Verified C++ build with `MATERIALX_BUILD_PERFETTO_TRACING=OFF` (no-op macros)
- [x] Verified GPU async track events appear in Perfetto traces alongside CPU slices

## Related

- #2742 — Perfetto tracing infrastructure (merged)
- #2820 — CPU trace markers for codegen hot paths (companion PR)
- #2821 — Test suite options: `envSampleCount`, `framesPerMaterial` (companion PR, adds multi-frame rendering)
- #2822 — Python comparison scripts (companion PR, consumes GPU track events)

---

```
Assisted-by: Claude (Anthropic) via Cursor IDE
Signed-off-by: Pavlo Penenko <pavlo.penenko@autodesk.com>
```

PR #2822, opened 2026-03-12
Python comparison scripts for test suite output (diff_test_runs)

ppenenko:
## Summary

A new Python package under `python/MaterialXTest/diff_test_runs/` provides three scripts for comparing pairs of MaterialX test suite output directories (baseline vs. optimized). The trace comparison script (`diff_traces.py`) consumes the Perfetto traces and GPU track events produced by the tracing instrumentation from #2742.

| Script | Purpose |
|--------|---------|
| `diff_images.py` | Perceptual image comparison using [NVIDIA FLIP](https://github.com/NVlabs/flip), with HTML side-by-side reports. Below is an example diff between two test runs with 1024 and 1 environment samples, respectively: <br>image |
| `diff_traces.py` | Perfetto trace comparison with per-material CPU slice and GPU render-time analysis, multiple `--slice` filters, `--warmup-frames` burn-in period, inline SVG charts in HTML reports, e.g. image |
| `diff_shaders.py` | Offline shader analysis: line count, SPIR-V binary size, `glslangValidator` compile time, `spirv-opt` optimization time; auto-discovers Vulkan SDK tools in `PATH`, e.g. image |
| `_report.py` | Shared utilities for comparison tables, SVG chart generation, and HTML report building |

Key features:
- **HTML reports** with interactive, searchable inline SVG bar charts
- **Sorted comparison tables** showing per-material deltas (absolute and percentage)
- **Summary statistics** (mean, median, best/worst) for quick assessment
- **Configurable filtering** (`--min-delta-ms`, `--min-delta-pct`) to focus on significant changes
- **Warm-up frame support** (`--warmup-frames`) to discard initial GPU frames for more stable averages

## Motivation

This tooling was developed to measure and validate shader generation optimizations (early lobe pruning, dead code elimination) in a companion branch. The infrastructure is generally useful for:

- **Regression testing**: Detect visual regressions from codegen changes via FLIP
- **Performance profiling**: Compare shader generation time, compilation time, and GPU render time across branches
- **Shader quality analysis**: Quantify code size and SPIR-V improvements from optimizations

## Test plan

- [x] Verified all three comparison scripts produce correct HTML reports when comparing two test suite runs
- [x] Verified `diff_traces.py` correctly processes Perfetto traces with CPU slices and GPU async track events
- [x] Verified `diff_images.py` produces FLIP-based perceptual difference reports
- [x] Verified `diff_shaders.py` compiles dumped GLSL to SPIR-V and reports LOC/size/timing deltas

## Related

[...]

PR #2820, opened 2026-03-12
Extended tracing instrumentation for shader codegen and rendering

ppenenko:
## Summary

Building on the Perfetto tracing from #2742, this PR adds CPU trace markers to additional codegen and rendering hot paths. This also required conditionally linking `MaterialXTrace` to `MaterialXGenShader`, which was not needed by #2742 (tracing was only in test code) but is now required for the new codegen instrumentation.

- **Shader codegen**: `ShaderGraph` traversal (`createConnectedNodes`, `addUpstreamDependencies`, `createNode`), `ShaderGenerator` emit functions (`emitFunctionDefinitions`, `emitFunctionCalls`, `getImplementation`), and `CompoundNode` operations

  image

- **GLSL build**: `GlslProgram::build()`, `DumpGeneratedCode`, `DumpUniformsAndAttributes`, `CaptureAndSaveImage`

## Motivation

This instrumentation was developed to measure and validate shader generation optimizations (early lobe pruning, dead code elimination) in a companion branch. The infrastructure is generally useful for identifying bottlenecks in shader generation and compilation.

## Test plan

- [x] Verified C++ build with `MATERIALX_BUILD_PERFETTO_TRACING=ON` on Windows (MSVC)
- [x] Verified C++ build with `MATERIALX_BUILD_PERFETTO_TRACING=OFF` (no-op macros, no link dependencies)
- [x] Verified Perfetto traces contain the new CPU slices

## Related

- #2742 — Perfetto tracing infrastructure (merged)
- #2774 — Original combined PR (superseded by this and follow-up PRs)
- Companion PRs for GPU async instrumentation, test suite options, and Python comparison scripts

---

```
Assisted-by: Claude (Anthropic) via Cursor IDE
Signed-off-by: Pavlo Penenko <pavlo.penenko@autodesk.com>
```

PR #2832, opened 2026-03-19
Propose an initial graph refactoring framework for shader generation

jstone-lucasfilm:
This changelist proposes an initial graph refactoring framework for MaterialX shader generation, replacing hardcoded graph optimizations with a composable set of named refactoring passes.  The following specific changes are included:

- Add a `ShaderGraphRefactor` base class, with three concrete subclasses for node elision, premultiplied BSDF add (valuable in hardware languages), and layer-over-mix distribution (valuable in MDL).
- Add a virtual `setDefaultOptions` method to `ShaderGenerator`, enabling each subclass to set target-appropriate defaults at `GenContext` construction time.
- Add a `MIX` classification bit to ShaderNode for efficient mix node identification.
- Remove hand-authored GLSL and MDL graph overrides for `open_pbr_surface` and `standard_surface`, as these refactoring steps are now generated automatically.


jstone-lucasfilm:
This proposal will require significant discussion from the MaterialX community, and I'm initially CC'ing @niklasharrysson, @ld-kerley, @ppenenko, and @jreichel-nvidia for their expertise and review.

This changelist directly addresses issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/2480, but it goes significantly further than this, providing a composable system of graph refactoring passes that handles constant/dot node elision, BSDF mix pre-multiplication for hardware shading languages, layer-over-mix distribution for MDL, and potentially far more in the future.

Let me know your thoughts, and I'd prefer that we take our time with this proposal, rather than rushing it into a release of MaterialX.

jstone-lucasfilm:
Here I'll share renders and performance numbers for common OpenPBR Surface materials, and the general pattern is that the current refactoring PR actually provides _better_ real-time shading performance than the original manually-optimized graphs, which was not the case for earlier automated optimization proposals:

OpenPBR Surface Carpaint, Manual Optimization: 8 ms/frame
OpenPbrSurface_Carpaint_Manual

OpenPBR Surface Carpaint, Current PR: 7ms/frame
OpenPbrSurface_Carpaint_Automated

OpenPBR Surface Glass, Manual Optimization: 12ms/frame
OpenPbrSurface_Glass_Manual

OpenPBR Surface Glass, Current PR: 11ms/frame
OpenPbrSurface_Glass_Automted

OpenPBR Surface Pearl, Manual Optimization: 13ms/frame
OpenPbrSurface_Pearl_Manual

OpenPBR Surface Pearl, Current PR: 12ms/frame
OpenPbrSurface_Pearl_Automated

ppenenko:
Looks great @jstone-lucasfilm! I think it's important to clarify that these changes would not conflict with my current WIP on #2680 - even though I initially experimented at the shader graph optimization/rewriting stage as well, my code ended up moving to the shader graph construction stage. So, the two optimizations should be complimentary in the end.

ppenenko:
I would also suggest the infrastructure we discussed recently as an automated way to measure the performance impact of these changes with more precision and test against correctness regressions, namely:

- #2820 and #2824: nanosecond-precision instrumentation instead of truncated milliseconds in MaterialXView; GPU timer queries with proper CPU/GPU synchronization; checking the impact on codegen and compilation times;
- #2821: `framesPerMaterial` to get more frame duration samples from MaterialXTest for averaging; `envSampleCount` for conditions more representative of interactive rendering and not as dominated by texture sampling;
- #2822: automated perceptual image comparison.

jreichel-nvidia:
Nice! That should be much more flexible than the old approach.

I verified that the generated MDL code is identical for all materials (i.e. `renderTestPaths` set to `resources/Materials`). A manual test shows that the graph transformation also kicks in when OpenPBR itself is translated (good), while the old mechanism only handled uses of OpenPBR, but not OpenPBR itself.

`setDefaultOptions()`: Isn't that rather a getter (with out parameter instead of return value) than a setter?

Looks like all passes are applied in a fixed order, potentially skipped depending on the options? I wonder whether it is worthwile to support a configurable order and repeated runs, like "run pass A first, then pass B, then another run of A". But that can probably be done later when there is an actual use case for that.

jstone-lucasfilm:
Good points, @ppenenko and @jreichel-nvidia.

For clarity, I've renamed `ShaderGenerator::setDefaultOptions` to `ShaderGenerator::applyDefaultOptions`, since the action of the function is to apply the preferences of a generator to the given `GenOptions` object, with support for layering this action through a hierarchy of generator subclasses.

niklasharrysson:
This looks great! 👍 👍

niklasharrysson (approved):
LGTM

jstone-lucasfilm:
Thanks so much for the positive feedback, @niklasharrysson, @ppenenko, and @jreichel-nvidia, and I believe this proposal has landed in a good place.

One additional motivation for this work, which wasn't apparent when I originally wrote up https://github.com/AcademySoftwareFoundation/MaterialX/issues/2480, is that the OpenPBR project is rapidly iterating on the graph for [OpenPBR Surface 1.2](https://github.com/AcademySoftwareFoundation/OpenPBR/blob/dev_1.2/reference/open_pbr_surface.mtlx), and it will be increasingly important to validate this evolving graph without encountering MDL errors in Omniverse or performance limitations in the MaterialX Web Viewer.

Based on the great feedback in this thread and these upcoming needs, I'd say that we should move forward with this PR, and we can continue to build upon this functionality going forward.

PR #2834, opened 2026-03-19
Re-enable rendermdl tests

jreichel-nvidia:
Attempt to re-enable the rendermdl tests for the nightly test runs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,6 @@ jobs:
   build:
     name: Build
     runs-on: ${{ matrix.os }}
-    continue-on-error: true
     env:
       IS_EXTENDED_BUILD: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
     strategy:
@@ -385,6 +384,7 @@ jobs:
 
     - name: Upload Log Files
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      if: always()
       with:
         name: Log_files_${{ matrix.name }}
         path: build/**/*.txt


.github/workflows/main.yml
```diff
@@ -216,8 +217,9 @@ jobs:
         fi
         TEST_RENDER_CONFIG="-DMATERIALX_TEST_RENDER=OFF"
         if [ "${{ matrix.test_render }}" == "ON" ]; then
+          TEST_RENDER_CONFIG="-DMATERIALX_TEST_RENDER=ON"
           if [ "${{ runner.os }}" == "macOS" ]; then
-            TEST_RENDER_CONFIG="-DMATERIALX_TEST_RENDER=ON -DMATERIALX_RENDER_MSL_ONLY=ON"
+            TEST_RENDER_CONFIG="$TEST_RENDER_CONFIG -DMATERIALX_RENDER_MSL_ONLY=ON"
```
jstone-lucasfilm, line 222: Looking at our existing approach for platform-specific render tests, I'd probably recommend following the existing pattern of `MATERIALX_RENDER_MSL_ONLY`, which was added by @kwokcb to support render tests in MSL without activating all shading languages.

Taking this approach, we could simplify this new MDL-only render test in GitHub CI with a corresponding `MATERIALX_RENDER_MDL_ONLY` flag, which would be enabled in our nightly builds.

jreichel-nvidia: When I started I still had the impression that they ran originally (my MR back then also installed the required dependencies) and got disabled by accident, but with more and more CI runs and subsequent changes I also got to the conclusion that his was never the case. I'll look into your suggestion with MATERIALX_RENDER_MSL_ONLY.

jreichel-nvidia:
Closing this one. Running df_vulkan in the CI on Windows needs a Vulkan driver with a hardware or software device. Removing the unused vcpkg dependencies and keeping log artifacts happens in new MRs.

Issue #2819, opened 2026-03-11
Update flake nodes to OSL-compatible hash functions and test performance


jstone-lucasfilm:
Following up on https://github.com/AcademySoftwareFoundation/MaterialX/pull/2676, the MaterialX rendering community has noted that it would be valuable to use OSL-compatible hash functions such as `mx_cell_noise_float` and `mx_hash_int` instead of custom functions, making it more straightforward for teams to add support for `flake` nodes in their own implementations.

For this task, we need a developer to make these changes in a PR and validate that the performance of `flake` nodes is not significantly impacted, as performance was the original motivation for the custom functions in the current implementations.

See additional details in the discussion at https://github.com/AcademySoftwareFoundation/MaterialX/pull/2676#discussion_r2525242488.

PR #2842, opened 2026-03-30
Update flake nodes to standard cellnoise functions

jstone-lucasfilm:
This changelist updates the implementations of `flake2d` and `flake3d` to use standard cellnoise functions across all shading languages, including a new implementation in OSL.  The following specific changes are included:

- Replace the custom random number generator with standard cellnoise functions in all languages, using scalar cellnoise for density and priority rejection and vector cellnoise for flake rotation.
- Optimize `mx_cell_noise_vec3` in GLSL and `mx_cell_noise_float3` in MDL to share a single bjmix across all three output channels, matching the existing optimization in the native OSL version.
- Add OSL implementations of the `flake2d` and `flake3d` nodes, removing corresponding exclusions in the render test suite.
- Fix the GLSL rotation matrix to use correct sine and cosine terms in the Arvo fast rotation construction, aligning with the MDL implementation.
- Tighten the bounding sphere early-rejection test to the exact circumsphere of the flake cube.

Fixes issue https://github.com/AcademySoftwareFoundation/MaterialX/issues/2819.


jstone-lucasfilm:
Here's a comparison of a Standard Surface Flake3D material in GLSL, where you can see that the overall distribution of flakes and performance profile are maintained, while the individual positions and rotations of flakes have been updated to match the hash behavior of OSL `cellnoise`:

Standard Surface Flake3D Material (Old):

StandardSurface_Flake3D_Old

Standard Surface Flake3D Material (New):

StandardSurface_Flake3D_New

msuzuki-nvidia:
Do you have a comparison with larger flake size, as we can see a flake shape?

jstone-lucasfilm:
@msuzuki-nvidia Sure, here's the same before-and-after comparison with the flake size set to 0.1:

Standard Surface Flake3D Material (Old):

StandardSurfaceFlake3D_BigFlakes_Old

Standard Surface Flake3D Material (New):

StandardSurfaceFlake3D_BigFlakes_New

msuzuki-nvidia:
@jstone-lucasfilm Thank you, looks good to me!

Issue #2833, opened 2026-03-19
OSL implementation of `transformmatrix` vector3M4 variant ignores translation


jstone-lucasfilm:
The OSL implementation of the vector3M4 variant of `transformmatrix` treats the input as a direction (w=0), ignoring the translation component of the matrix:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx#L501

The GLSL and MDL implementations correctly treat it as a point (w=1), which includes translation:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/genglsl/mx_transformmatrix_vector3M4.glsl
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenMdl/mdl/materialx/stdlib_1_6.mdl#L2302

The OSL implementation is also inconsistent with its corresponding vector2M3 variant, which correctly casts to point to include translation:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/genosl/mx_transformmatrix_vector2M3.osl

One proposed fix would be the following:

```
sourcecode="transform({{mat}}, point({{in}}))"
```

aviralgarg05:
Hi, I would like to solve this issue if someone could assign me

jstone-lucasfilm:
Thanks @aviralgarg05, and it's now assigned to you!

jstone-lucasfilm:
This issue has been addressed by @aviralgarg05 in #2843.

PR #2843, opened 2026-03-31
Fix OSL transformmatrix vector3M4 translation

aviralgarg05:
## Title

Fix OSL `transformmatrix` translation for `vector3` + `matrix44`. #2833

**Description**

## Summary

This fixes an OSL backend bug in `transformmatrix` for the `vector3` + `matrix44` variant.

In OSL, `ND_transformmatrix_vector3M4` was using:

```osl
transform({{mat}}, {{in}})
```

That treats the input as a direction-like vector, so matrix translation is ignored.

The fix changes the OSL implementation to:

```osl
transform({{mat}}, point({{in}}))
```

which preserves translation and brings the behavior in line with the existing GLSL and MDL implementations, as well as the OSL `vector2M3` path.

## What changed

- Updated the OSL stdlib implementation for `IM_transformmatrix_vector3M4_genosl`
- Added a focused regression test in the OSL test suite to verify:
  - the implementation mapping uses `point({{in}})`
  - generated OSL source includes point semantics for this case

## Why this is needed

The issue was that a `matrix44` transform with translation would not move a `vector3` input in generated OSL, even though the same MaterialX node behaves correctly in other backends.

That made OSL inconsistent and caused translation to be silently dropped.

## Files changed

- [`libraries/stdlib/genosl/stdlib_genosl_impl.mtlx`](/Users/aviralgarg/Everything/MaterialX/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx)
- [`source/MaterialXTest/MaterialXGenOsl/GenOsl.cpp`](/Users/aviralgarg/Everything/MaterialX/source/MaterialXTest/MaterialXGenOsl/GenOsl.cpp)

## Validation

Ran the new regression test:

```bash
build/bin/MaterialXTest "GenShader: OSL TransformMatrix Point Semantics"
```

Ran targeted OSL checks:

```bash
build/bin/MaterialXTest "GenShader: OSL Implementation Check"
build/bin/MaterialXTest "GenShader: OSL Shader Generation"
build/bin/MaterialXTest '[genosl]'
build/bin/MaterialXTest '[genosl]'
```

Ran the full locally configured test suite:

```bash
ctest --output-on-failure
```

Result:

- `79/79` tests passed
- OSL test suite passed repeatedly
- `git diff --check` passed

## Notes

This change is intentionally narrow and only touches the OSL implementation path that was dropping translation, plus the regression coverage needed to keep it from coming back.


source/MaterialXTest/MaterialXGenOsl/GenOsl.cpp
```diff
@@ -179,6 +179,45 @@ TEST_CASE("GenShader: OSL Metadata", "[genosl]")
     REQUIRE(shader != nullptr);
 }
 
+TEST_CASE("GenShader: OSL TransformMatrix Point Semantics", "[genosl]")
+{
+    const std::string expectedSourceCode = "transform({{mat}}, point({{in}}))";
+    const std::string testDocumentString = R"(<materialx version="1.39">
```
jstone-lucasfilm: One downside of _inlining_ an example file is that it won't be included when we perform syntax upgrades and reformatting on our example data, and a better approach would be to leverage our standard test suite, which includes MaterialX documents that leverage `transformmatrix`:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/resources/Materials/TestSuite/stdlib/math/transform.mtlx

If there's already an appropriate example there, feel free to load and test it here; if not, then you can add another example to the same document in the test suite.

jstone-lucasfilm (commented):
This looks very promising, @aviralgarg05, and I had just one suggested improvement for the new test that you've added.

jstone-lucasfilm (approved):
Thanks for the fix, @aviralgarg05!

PR #2846, opened 2026-04-01
Fix sRGB exponent in Metal rendering

exavi:
This tiny change should 'fix' the sRGB encoding used on the metal backend.

I noticed this while using the MaterialXViewer in a CI setup when generating images where no single pixel matched against reference values (not even the background color :D)


jstone-lucasfilm:
Thanks for this PR, @exavi, and since this code goes all the way back to the original integration of Metal support in 2023, I'm CC'ing @Morteeza, @lkerley, and @dgovil for their domain expertise.

doug-walker:
The proposed change seems correct, the exponent for the sRGB formula should use 2.4 rather than 2.2.

dgovil:
I agree this looks like the change is correct, as far as I understand gamma approximations anyway. I will ask Lee to check me on that but I think its a good catch

ld-kerley:
Looks good to me too - I trust Doug for color stuff more than myself :). 

I am a little surprised we're not just handling this conversion natively in Metal, but thats a possible improvement for the future.

ld-kerley (approved):

jstone-lucasfilm:
Great, thanks to all of our reviewers!

@exavi Before we can merge this work, we'll need you to address the EasyCLA errors, which you can do by following the steps listed here:

https://github.com/AcademySoftwareFoundation/MaterialX/pull/2846#issuecomment-4169324639

exavi:
Awesome! I’ll sort it out when I get back to the office!

PR #2853, opened 2026-04-03
Run pip install at end of MaterialX install

ashwinbhat:
Relocating the pip install at the end. This allows any customizations to be included in the install.

In our fork of MaterialX we install additional libraries 
https://github.com/autodesk-forks/MaterialX/blob/4c20cc3426812b1a196176f0a13ade2f34a36751/CMakeLists.txt#L132
https://github.com/autodesk-forks/MaterialX/blob/4c20cc3426812b1a196176f0a13ade2f34a36751/CMakeLists.txt#L620
This allows us to keep our fork in-line with main. 
I would to like to install additional libraries from our fork before pip install is run. 

Relocating the pip install at the end, avoids the amount of patches we have to keep against main.


jstone-lucasfilm (approved):
This looks like a good improvement to me, thanks @ashwinbhat.

PR #2855, opened 2026-04-04
Add Python wrapper for Metal image handler creation

kwokcb:
## Change

There is no way to create an image handler in Python for MSL as there is no API exposed wrapper. This change adds in the wrapper on MslRenderer to allow image support.

## Tests

Test using Python renderer of standard surface using wood textures. GLSL has always been possible. MSL equivalent is also possible now with this change.

| GLSL Python render | MSL Python render | 
| :--: | :--: |
| Tiled_Wood_genglsl | Tiled_Wood_genmsl |


jstone-lucasfilm (approved):
Thanks for the improvement, @kwokcb!

PR #2858, opened 2026-04-08
Upgrade three.js from r152 to r183

jstone-lucasfilm:
This changelist upgrades the three.js dependency in the MaterialX Web Viewer from r152 to r183, adapting to API changes across the intervening releases.


javascript/package-lock.json
```diff
@@ -26,8 +26,9 @@
       "version": "1.0.0",
       "license": "ISC",
       "dependencies": {
+        "fflate": "^0.8.2",
         "lil-gui": "^0.19.2",
-        "three": "^0.152.2"
+        "three": "^0.183.0"
```
kwokcb, line 31: Looks like the desired and resolved versions differ. Maybe best to just use "^0.183.2" (including patch version) in  package.json (like it was done for the previous version.)

source/MaterialXGenGlsl/GlslShaderGenerator.cpp
```diff
@@ -243,10 +243,13 @@ void GlslShaderGenerator::emitTransmissionRender(GenContext& context, ShaderStag
     emitLineBreak(stage);
 }
 
-void GlslShaderGenerator::emitDirectives(GenContext&, ShaderStage& stage) const
+void GlslShaderGenerator::emitDirectives(GenContext& context, ShaderStage& stage) const
 {
-    emitLine("#version " + getVersion(), stage, false);
-    emitLineBreak(stage);
+    if (context.getOptions().hwEmitVersionDirective)
```
kwokcb: I'm not so sure about skipping version number as it's supposed to be mandatory afaik (https://registry.khronos.org/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.html#version-declaration) . With out it is incomplete as a stand-alone shader (outside the context of three). 

Also I don't think you want to allow this for general GLSL. Not sure what it does for Vulkan as well.

Even in ThreeJS, I'm not sure how it knows which particular version of v3 it is using -- it just seems to say v1 or v3.

kwokcb (commented):
Looks good in general. I think we want to keep explicit versioning for shaders and not add in the new shader gen option.

javascript/package-lock.json
jstone-lucasfilm: Good point, @kwokcb, and done.

source/MaterialXGenGlsl/GlslShaderGenerator.cpp
jstone-lucasfilm: This is a trickier case, as Three.js will actually _fail_ to compile shaders that already contain a version directive, as it automatically adds its own version information, and two version directives in a single shader are invalid.

This leaves us with two options for generating correct shaders in Three.js:

1. Using string processing operations in JavaScript to _manually remove_ the version directive from our generated shader before passing it along to Three.js.
2. Adding a shader generator option that controls the _addition_ of version directives.

Of these two options, (2) seemed more robust to me, and that was the motivation for the current solution.

Making this a hardware shader option seemed the most targeted and consistent with existing features in shader generation, though it's true that we don't yet have a use case for skipping version directives in GLSL or MSL.

What are your thoughts?

ashwinbhat: I would prefer not adding new attributes to GenOptions. Is this a new limitation/issue in three.js, perhaps we can get some input from three.js folks.

jstone-lucasfilm: Ok, I've spoken offline with @ashwinbhat and @kwokcb, and I think they're right that the approach I've taken here is _too_ general, when all we need is a local workaround for the expectations of Three.js.

I'll make the change!

ashwinbhat (approved):
The change to locally remove `version` directive to support three.js RawShaderMaterial looks good to me.

kwokcb (approved):

PR #2861, opened 2026-04-11
Improve build configuration in GitHub CI

jstone-lucasfilm:
This changelist improves the build configuration in GitHub CI, reducing duplication across the build matrix and adjusting parallelism per platform.  The following specific changes are included:

- Consolidate extended build configuration as a local `extended_cmake_config` field on each matrix row, reducing the CMake Generate step to a single call.
- Adjust build parallelism per platform through the CMAKE_BUILD_PARALLEL_LEVEL environment variable, defaulting to 4 and reduced to 1 on Windows, as the MSVC `/MP` flag already handles file-level parallelism within each project.
- Simplify the MacOS rendering setup to setting a debug flag in extended builds, as the `MTL_HARDWARE_RENDERING` and `LIBGL_ALWAYS_SOFTWARE` overrides are not actually supported in Metal rendering.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,10 @@ on:
 permissions:
   contents: read
 
+env:
+  CMAKE_BUILD_PARALLEL_LEVEL: 4
+  CTEST_PARALLEL_LEVEL: 4
+
 jobs:
 
   build:
@@ -35,6 +39,7 @@ jobs:
           compiler_version: "14"
           python: 3.13
           extended_build_perfetto: ON
+          extended_cmake_config: -DMATERIALX_BUILD_PERFETTO_TRACING=ON
 
         - name: Linux_GCC_14_Python314
           os: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


kwokcb (commented):
This is much nicer to have the localized extended build options per config vs a big if statement. 

The counter comment I have is if you want more than one config to run the same extended options then it seems you need to repeat them (e.g. if you want to run MDL  on > 1 platform/config).  This could be error prone.

I guess you could do something like make a resuable option (e.g. `extended_cmake_config_mdl` or structure the matrix to extend `extended_cmake_config` another way ?

jstone-lucasfilm:
@kwokcb I think your note about reusable extended commands is a good one for the future, and we'll have many options available to share commands across jobs if this ends up being required.

At the current scale of our extended builds, I feel the approach proposed here strikes a better balance in clarity and readability than its predecessor, and it should be straightforward to evolve based on future needs.

kwokcb (approved):
I was mostly checking that this won't cause issues with any future reuse. Sounds like this can be extended when needed. Looks good to  check in. Thanks.

PR #2863, opened 2026-04-14
Omit `specular_weight` from metal edge tint in OpenPBR

jstone-lucasfilm:
This changelist removes `specular_weight` from the computation of metal edge tint in OpenPBR Surface, aligning the behavior of this shading model with a [proposed fix](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/240) for OpenPBR 1.2.

As described in the proposed fix, the edge-tint color for the `generalized_schlick_bsdf` should not be multiplied by `specular_weight`, as the whole lobe is multiplied by `specular_weight` as well.


portsmouth (commented):
LGTM

portsmouth (approved):

PR #2869, opened 2026-04-20
Refactor runRenderer to use structured types

ashwinbhat:
Replace the 10-parameter runRenderer signature with RenderSession, RenderItem, and RenderProfileResult value types.

Each call now returns isolated profiling data that the caller accumulates. Test run is now decoupled, so in future, we would be able to introduce some parallelism in these tests if needed.

Minor fixes:
- correct ScopedTimer scoping in validate(),
- fix a copy-paste timer name in the Slang renderer.

### DIFF
--- a/source/MaterialXTest/MaterialXRender/RenderUtil.h
+++ b/source/MaterialXTest/MaterialXRender/RenderUtil.h
@@ -186,17 +186,13 @@ struct RenderItem
 {
     RenderItem(mx::TypedElementPtr elem,
                mx::FileSearchPath searchPath,
-               mx::FilePath outPath,
-               mx::ImageVec* images = nullptr);
+               mx::FilePath outPath);
 
-    mx::TypedElementPtr element;
-    mx::DocumentPtr document;
-    mx::FileSearchPath imageSearchPath;
-    mx::FilePath outputPath;
-    mx::ImageVec* imageVec = nullptr;
-    std::string shaderName;
-
-    mx::DocumentPtr doc() const { return document; }
+    const mx::TypedElementPtr element;
+    const mx::DocumentPtr document;
+    const mx::FileSearchPath imageSearchPath;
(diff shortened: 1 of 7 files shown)


source/MaterialXTest/MaterialXRender/RenderUtil.h
```diff
@@ -161,6 +173,41 @@ struct TestRunState
...
+    RenderItem(mx::TypedElementPtr elem,
+               mx::FileSearchPath searchPath,
+               mx::FilePath outPath,
+               mx::ImageVec* images = nullptr);
+
+    mx::TypedElementPtr element;
+    mx::DocumentPtr document;
+    mx::FileSearchPath imageSearchPath;
+    mx::FilePath outputPath;
+    mx::ImageVec* imageVec = nullptr;
```
jstone-lucasfilm: Would this variable be better as an `ImageVec images` member of `RenderProfileResult`, rather than a pointer-to-vector member of `RenderItem`?

That would allow function arguments of type `const RenderItem&` to remain truly constant, without making this vector a silently-mutable data member of `RenderItem`.

source/MaterialXTest/MaterialXRender/RenderUtil.h
```diff
@@ -161,6 +173,41 @@ struct TestRunState
...
+    std::string shaderName;
+
+    mx::DocumentPtr doc() const { return document; }
```
jstone-lucasfilm: Now that `document` is a public member of `RenderItem`, do we still need this additional `doc` accessor?

jstone-lucasfilm (commented):
I really like this proposal overall, @ashwinbhat, and I've written up two notes on potential improvements before we merge this work.

jstone-lucasfilm (approved):
Thanks for this improvement, @ashwinbhat, and this looks good to me!

Issue #2851, opened 2026-04-03
'Indirect Lighting' MaterialXView option broken on macOS


pablode:
It does not seem to do anything, while on Windows it works:

https://github.com/user-attachments/assets/7f762469-9120-4d59-a86a-236b4e39903a

jstone-lucasfilm:
Good catch, @pablode, and my initial guess is that we're referencing a hardcoded uniform name that doesn't match the behavior of the MSL shader generator.

jstone-lucasfilm:
Thanks for posting this issue, @pablode, and I believe it has been addressed by @warrenm in #2870.

PR #2870, opened 2026-04-20
Unify environment light texture handling in MaterialXRenderMsl

warrenm:
This PR seeks to address #2851.

The core issue is that the Metal rendering code wasn't honoring the indirect lighting property of the light handler. However, changing the logic to match that of the implementation in `bindLighting()`  in [GlslProgram](https://github.com/AcademySoftwareFoundation/MaterialX/blob/d69fcb10283f0bbc5f17415c33a6f2e6b63622ad/source/MaterialXRenderGlsl/GlslProgram.cpp#L585-L600) isn't quite enough. There were actually two independent paths for binding IBL textures in the Metal implementation, one in `bindLighting()` and one in the lower-level `bindTextures()`. This change unifies those two paths into a single path relying on the _explicit bound texture_ mechanism currently used for shadow mapping.

This change seems fairly natural to me, but I would appreciate any other Metal renderer implementation experts weighing in.

Tested in MaterialXView on macOS Tahoe 26.4:

https://github.com/user-attachments/assets/f3c48288-2eb8-420b-b700-b9ca6d9e436b

### DIFF
--- a/source/MaterialXRenderMsl/MslPipelineStateObject.mm
+++ b/source/MaterialXRenderMsl/MslPipelineStateObject.mm
@@ -580,7 +580,7 @@ MaterialX::ConstValuePtr MslProgram::findUniformValue(const string& uniformName,
 }
 
 void MslProgram::bindTextures(id<MTLRenderCommandEncoder> renderCmdEncoder,
-                              LightHandlerPtr lightHandler,
+                              LightHandlerPtr /*lightHandler*/,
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
This looks like a great improvement to me, @warrenm, and I'm CC'ing @lkerley and @dgovil for their thoughts and feedback.

jstone-lucasfilm (approved):
Thanks for this fix, @warrenm!

Issue #1136, opened 2022-11-08
Gltf attribute attenuation_distance does not have a default value


JGamache-autodesk:
As [seen here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/gltf_pbr.mtlx#L27):
```xml
    <input name="attenuation_distance" uniform="true" type="float" uimin="0" uiname="Attenuation Distance" uifolder="Volume" />
```
And if I read the [specs](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_materials_volume/README.md#attenuation) correctly, the default should be `+infinity`. Should we represent that using FLOAT_MAX?

JGamache-autodesk:
It is going to be interesting to build a UI to manipulate that value.

jstone-lucasfilm:
Good point, @JGamache-autodesk, and I wonder if we should select a suitably-large finite value as the canonical default, for the benefit of UI systems that don't support infinity.

PR #2872, opened 2026-04-22
Connect gltf pbr attenuation distance

bhouston:
Another fix resulting from my use of the MaterialX Fidelity tool here: https://materialx-fidelity.ben3d.ca

I was finding that materials with attention colours where not rendering with any tint in the MaterialXView, like this material:

https://materialx-fidelity.ben3d.ca/api/asset/gltf_pbr/default_subset_transmission.mtlx.zip

Was rendering like this:

materialxview-4

I fixed it to render like this via this PR:

materialxview-3

The fix seems to be just modifying this one mtlx file since its attention sub-graph was disconnected.

### DIFF
--- a/libraries/bxdf/gltf_pbr.mtlx
+++ b/libraries/bxdf/gltf_pbr.mtlx
@@ -45,9 +45,16 @@
       <input name="in" type="vector3" nodename="attenuation_color_vec" />
     </ln>
 
+    <ifgreater name="safe_attenuation_distance" type="float">
+      <input name="value1" type="float" interfacename="attenuation_distance" />
+      <input name="value2" type="float" value="0.0" />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks for this fix, @bhouston, and the VDF layering you've added here is clearly a valuable improvement.

For the attenuation distance fallback, I wonder if it would actually be better to address this by stating a concrete default value for this input, as proposed in this earlier issue from @JGamache-autodesk:

https://github.com/AcademySoftwareFoundation/MaterialX/issues/1136

That would remove the need to choose an arbitrary fallback such as 1 in the shading graph, and the stated default value would be visible to users of the shading model, rather than a hidden implementation detail.

PR #2877, opened 2026-05-02
Stabilize flake `id` across floating-point precisions

jstone-lucasfilm:
This changelist reduces the scale factor for flake `id` from `INT32_MAX` to `2^24 - 1`, providing a stronger guarantee of consistency between 64-bit and 32-bit shading environments.

### DIFF
--- a/libraries/stdlib/genglsl/lib/mx_flake.glsl
+++ b/libraries/stdlib/genglsl/lib/mx_flake.glsl
@@ -112,7 +112,8 @@ void mx_flake(
     float xi1 = flake_noise.y;
 
     rand = flake_noise.z;
-    id = int(rand * 2147483647.0);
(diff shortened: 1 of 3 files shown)


code-monkey (approved):
This will be much more stable for systems with different float precision.  But as you note in the comment the range for `id` is now less.  I don't _think_ this is a concern, but it might be unexpected.  I expect we're just trying to get a random integer for a seed to be used elsewhere.
Some quick internal testing shows the integer value seems more stable with 32 and 64 bit precision.

msuzuki-nvidia:
LGTM. I agree that keeping IDs consistent across precisions is more important than preserving a wider ID range.

msuzuki-nvidia (approved):

PR #2880, opened 2026-05-04
Single File MaterialX Container Proposal

bhouston:
Per discussions on Slack, I've added a formal proposal for a mtlz file.  Details in PR.


documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,128 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
 
+## Single-File MaterialX Container
+
+A number of users and developers have discussed the usefulness of a single-file MaterialX container format, along with features that would make such a format practical for online material libraries, publishing workflows, and interchange between applications.  This proposal suggests formalizing the `.mtlx.zip` convention already used by the [AMD Material Library](https://matlib.gpuopen.com/main/materials/all) and [Poly Haven](https://polyhaven.com/materials) as a dedicated `.mtlz` format.
+
+The proposed `.mtlz` format is a zip archive containing one root MaterialX document and its referenced resources.  It is intended as a package format for sharing and delivery, not as a replacement for ordinary `.mtlx` files or directory-based production libraries.
+
+### Motivation
+
+Downloading a single file for a complete MaterialX material, including textures and other referenced files, is convenient for web delivery and content exchange.  Online MaterialX libraries already use `.mtlx.zip` archives for this purpose, since downloading and tracking multiple related files is awkward for browsers, APIs, and asset management systems.
```
dgovil, line 81: Note that this is convenient for distribution, but delivery has some complexities.
ie when bundled into a zip, you lose the ability to cache individual assets (two materials sharing a texture) and the ability to download piece meal (maybe you don't need all the textures but you can't choose ahead of time).

dgovil, line 81: >  since downloading and tracking multiple related files is awkward for browsers
This is actually something that browsers are very, very good at and vastly prefer. It's more an issue for engine runtimes

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,128 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+marble_cliff.mtlz
+|-- marble_cliff.mtlx
+`-- textures/
+    |-- marble_cliff_diff.jpg
+    |-- marble_cliff_nor_gl.exr
+    |-- marble_cliff_rough.exr
+    `-- marble_cliff_disp.png
+```
+
+The rule that exactly one `.mtlx` file appears at the root lets readers identify the package entry point without requiring a manifest.  Other `.mtlx` files may be included in subdirectories when referenced by the root document.
```
dgovil, line 121: Note that USDZ takes the convention of the first file in the zip. I think that's worth having as a requirement, even if you do only include a single file in the root as a secondary requirement. 

It lets you take some shortcuts with reading the file. I see you have this in the next section, but I'll leave some notes on that separately

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,128 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+
+### Archive Layout Recommendations
+
+For efficient streaming, the root `.mtlx` file should be stored first in the zip archive.  This lets a reader inspect the MaterialX graph and its referenced resources before the rest of the archive has been downloaded.
```
dgovil: While I agree with this paragraph as a whole, I should mention that a zip file cannot be streamed easily.  You can mandate that the file has to be first in the binary structure of the zip, but you will spend many http requests streaming byte chunks over + requires the file server to support streamed chunks (which most but not all do)

Unfortunately one trick of the zip file is that the ToC is at the end of the file. But it is possible to do a straight ahead zip parser if you do mandate that it has to be physically first in the zip, and not just ordered first in the ToC

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,128 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+
+For efficient direct access, package writers should consider storing files without zip compression.  This follows the same general rationale as USDZ, where uncompressed zip storage can support direct memory access to package contents.  Many texture formats are already internally compressed, so zip compression may provide limited benefit while adding decoding overhead.
+
+These layout recommendations are not required for a package to be structurally valid, but they provide useful guidance for tools that create `.mtlz` files intended for web delivery, rendering, or large material libraries.
```
dgovil: IMHO we should explicitly require that it be uncompressed as a zip package. I don't think we should leave this as just a recommendation.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,128 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+    `-- marble_cliff_disp.png
+```
+
+The rule that exactly one `.mtlx` file appears at the root lets readers identify the package entry point without requiring a manifest.  Other `.mtlx` files may be included in subdirectories when referenced by the root document.
+
+### Archive Layout Recommendations
+
+For efficient streaming, the root `.mtlx` file should be stored first in the zip archive.  This lets a reader inspect the MaterialX graph and its referenced resources before the rest of the archive has been downloaded.
+
+For efficient direct access, package writers should consider storing files without zip compression.  This follows the same general rationale as USDZ, where uncompressed zip storage can support direct memory access to package contents.  Many texture formats are already internally compressed, so zip compression may provide limited benefit while adding decoding overhead.
```
dgovil: Another recommendation I would strongly suggest is that a container format include adequate inter asset padding to align with common memory mapping access patterns. This is less critical for the mtlx file itself, but greatly helps with texture reads without needing to copy the full byte range of textures into memory.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,128 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+`-- textures/
+    |-- marble_cliff_diff.jpg
+    |-- marble_cliff_nor_gl.exr
+    |-- marble_cliff_rough.exr
+    `-- marble_cliff_disp.png
+```
+
+The rule that exactly one `.mtlx` file appears at the root lets readers identify the package entry point without requiring a manifest.  Other `.mtlx` files may be included in subdirectories when referenced by the root document.
+
+### Archive Layout Recommendations
```
dgovil: Some missing aspects to this:

1. Should the format be zip32 or zip64? zip32 is much simpler to parse, but zip64 isn't much more of a headache. One issue of note is that USDZ is zip32 and I think GLB also expects 32bit offsets (but I could be wrong)

2. Encryption like AES or other zip extensions should probably be called out as explicitly out of scope

dgovil (commented):
I've left some notes. 

Personally I believe we should leave packaging up to the individual scene formats to handle instead, but if we were to do this my recommendation would be to follow the USDZ spec 1:1 as Pixar have put a lot of thought into the performance characteristics.

This does not require any dependency on the USD library but has some benefits:

1. The USDZ format is well documented and is very easy to read, it also takes into account several performance characteristics

2. This would make MTLZ play very transparently with USDZ, while still not requiring USD itself when used by itself or with other formats

jstone-lucasfilm:
Thanks for putting together this proposal, @bhouston, and it's my sense that your high-level approach here is correct, and this feature would provide significant benefits to the MaterialX/USD/glTF ecosystem.  @dgovil brings up some important points on specific details, though, and we should resolve these before proceeding.

I'll bring up one discussion point of my own, which is that I'd recommend that we incorporate properties such as `materialx_license`, `materialx_description`, and `materialx_authors` as _standard_ attributes in the MaterialX specification, rather than custom attributes that are only used in MTLZ files.

In the current MaterialX specification, we support the definition of custom attributes via `attributedef` for properties that are too application-specific or architecture-specific to standardize in the specification, but most of the new attributes that you suggest here seem truly universal.

documents/Specification/MaterialX.Proposals.md
bhouston: I'll modify it to suggest 64 byte alignment.

documents/Specification/MaterialX.Proposals.md
bhouston: Should we say all files are uncompressed in the archive or just the resource files that may be memory mapped?

documents/Specification/MaterialX.Proposals.md
bhouston: I've make that clear.  Thank you!

documents/Specification/MaterialX.Proposals.md
bhouston: I've added clarification to make it clear it is also the first file in the archive.

documents/Specification/MaterialX.Proposals.md
bhouston: I've added an explanation of the pain points to the doc to be more clear.

bhouston: > ie when bundled into a zip, you lose the ability to cache individual assets (two materials sharing a texture) 

Could you use the CRC32 + filesize in the zip file to deduplicate the textures without loading and re-hashing them?

documents/Specification/MaterialX.Proposals.md
bhouston: I've added to the proposal that it is only ZIP32 (to be consistent with USDZ and glB.). We can always add ZIP64 support in the future.

I've said no encryption as well.

I am not sure what you mean by "zip extensions" so I didn't yet put it into the proposal.

bhouston:
> Thanks for putting together this proposal, @bhouston, and it's my sense that your high-level approach here is correct, and this feature would provide significant benefits to the MaterialX/USD/glTF ecosystem. @dgovil brings up some important points on specific details, though, and we should resolve these before proceeding.

I think I've address his points in the latest version.
 
> I'll bring up one discussion point of my own, which is that I'd recommend that we incorporate properties such as `materialx_license`, `materialx_description`, and `materialx_authors` as _standard_ attributes in the MaterialX specification, rather than custom attributes that are only used in MTLZ files.
> 
> In the current MaterialX specification, we support the definition of custom attributes via `attributedef` for properties that are too application-specific or architecture-specific to standardize in the specification, but most of the new attributes that you suggest here seem truly universal.

Okay, I've moved it to its own proposal separate from the single file container and switched it to optional attributes, rather than using the custom attributes.

documents/Specification/MaterialX.Proposals.md
spiffmon: @dgovil , we _might_ be able to address the caching/sharing problem in a usdMtlx implementation by virtue of how the "package identifiers" are returned, i.e. the two materials would both return to USD and the renderer the same path that looks like `@materialBundle.mtlz[textures/shared_tex.avif@`

spiffmon: Oh, I think the problem you're identifying is having two materials defined in two _separate_ `.mtlz` packages, both consumed by the same asset, @dgovil ?  That would indeed create a sharing problem for any `usdMtlx` implementation, as well.  @bhouston , the problem with any zip-specific deduplication is that you're piercing an abstraction that spans several architectural layers in OpenUSD, at least.  Without passing any judgments, just noting it would be a big lift to communicate that information "in time" from low-level file-format reader all the way up to the imaging systems.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,131 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+
+* The issue with multiple downloads for consumption by a website (e.g 3D viewer), is that if you have multiple files with dependencies between, you have to download the first mtlx file, parse it, and then make subsequent requests to the server for the textures/resources.  Those subsequent requests best case are around 100ms later because of this sequentialization.  Downloading the file zip at once, especially if you are incrementally parsing, means you avoid that additional roundtrip request.
+
+### Technical Format
+
+An `.mtlz` package is a zip archive with the following baseline rules:
+
+* The archive contains exactly one `.mtlx` file at the root level AND it is also the first file stored in the zip archive.  This file is the root MaterialX document to load.
+* All other content is stored in subdirectories, including textures, included node definitions, and other referenced resources.
```
spiffmon, line 100: If you are going to make that requirement (subdirectories), does it add value to go further and stipulate known subdirectories for `libraries`, `textures`, etc -- calling out that in usdMtlx, "libraries" that wanted to inject shader definitions into the runtime would require different handling than "simple" material definitions.

spiffmon: Additionally, is there a strong motivation for disallowing multiple "root level" .mtlx files?  It seems to me that if this were allowed, it would provide a symmetric solution to the texture sharing problem @dgovil called out above.  Drawing inspiration from USDZ and how usdMtlx might leverage any .mtlz packages, every .mtlx file in a package can be unambiguously addressed from outside (e.g. the asset that binds to the materials) by referring to `@bananaMaterials.mtlz[skin.mtlx]@` and `bananaMaterials.mtlz[flesh.mtlx]@` individually.  

You'd invoke your packager logic to add all of an asset's materials into the package, with all (shared, at least) textures going into the `textures` folder within the package, and then "asset-identifier based" sharing/deduplication just works.

MaterialX does allow for multiple materials to be defined in a single .mtlx, which would be another solution, but if authoring tools don't make that easy to do, then allowing for aggregation in the package format provides versatility.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,131 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+* Relative file references in the root `.mtlx` document are resolved within the package.
```
spiffmon, line 101: This should be stated of all mtlx files in the package, no?

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,131 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+* The file extension is `.mtlz`.
+* The media type is `model/materialx+zip`.
+* Archive must be in ZIP32 format (not ZIP64).
+* Archive must not be encrypted.
```
spiffmon, line 105: The archive itself must also not be compressed.  This restriction stems from the same place as no-encryption: direct consumption of embedded data via mmapping, which is invaluable (and a requirement of USDZ).  The savings from allowing mtlx files to be compressed are almost always going to be dwarfed by allowing compressed textures.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,131 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+* Resource files in the archive must not be compressed.
+* Resource files in the archive must be aligned at 64 byte boundaries.
```
spiffmon, line 107: I think use of "resource" is inconsistent here, if we are trying to match USDZ for maximal compatibility and integration?

- **Every** file contained in the package must be 64-byte aligned
- Any resource type whose native reader does not automatically support compression must not be compressed. So AVIF is OK, but mtlx.zip is not.

documents/Specification/MaterialX.Proposals.md
```diff
@@ -37,6 +37,131 @@ As such, those forward-looking proposals have been moved from the formal Specifi
...
+
```
spiffmon, line 108: OK, although this might be contentious, I'd like to make the argument that we go further here and incorporate USDZ's image format restrictions, which means AVIF is the sanctioned super-compressed format, not ktx2/basis.  Arguments:

1. It is unbelievably fantastic that MaterialX is being adopted in both glTf and USD ecosystems, which serve sometimes subtly, sometimes grossly different needs with different fidelity and performance tradeoffs.  But MaterialX, like USD, evolved out of the needs of film studios, and when weighing choices, should lean towards best-serving those needs.  From that perspective, AVIF is a better choice than ktx2/basisU because it provides significantly greater image fidelity, even for _smaller_ file-sizes than basis, and small size is great not just for USDZ delivery, but also for renderfarm network traffic (for sure .exr is going to be used for hero assets, but for those cases where we would reach for compression, AVIF gets us better pixels, acknowledging the greater GPU footprint.
2. Why not support both? Ktx/basis is not supported natively by Hydra (exr and AVIF are), and we have the same challenge in making them useful that VFX-standard OpenImageIO has (which also lacks support, which also means the OSL runtime lacks support): modifying the image architectures to deliver compressed data to the GPU would be a non-trivial effort.  And every file format we add to the spec comes with a cost for implementers (I'm told the basisu library is around 100K loc, which is not at all trivial or palatable to (e.g.) inter into OpenUSD).
3. Anecdotally, we also have not heard of desire for ktx support from our gaming colleagues.
4. Whereas VFX does not support ktx2 well/much, every browser supports AVIF, as does three.js and Babylon, and though it looks still early-ish, Khronos _does_ have an extension for AVIF.  So in theory, glTf/glb _could_ support a `mtlz` that contains AVIF, even though they won't be as GPU-friendly.
5. Alternatively, building on the already-understood premise that USD content generally needs to be "distilled" to be representable in glTf, we can say the same for MaterialX, without changing anything fundamental about the glTF "asset packaging" process.  Just as already, .exr files would need to be converted to ktx2, we'd stipulate that glTf does not natively consume mtlz files, and instead, as part of the packaging process, extracts the .mtlx files, and converts all the textures to ktx2, and puts all of it into the .glb

If this makes sense, we should stipulate the standard-compliant allowable texture formats here, also.

spiffmon (commented):
Hi all, left several comments with an eye towards keeping this effort, if it proceeds, aligned with the goal of optimal integration of MaterialX content into OpenUSD.  That's an ongoing goal for us, and is distinct from the comments I made on slack about the challenges of adding support for `.mtlx` into the USD/USDZ normative specifications.

spiffmon:
Building on @dgovil 's initial comment, and inspired by a question from @meshula, the most concise, least-perturbance to the ecosystem way of formulating this spec would be by formal reference to the USDZ spec, with enumeration of the differences, which is primarily replacing allowable contained filetypes (`usd, usda, usdc`) with `mtlx`, and removing the audio filetypes.  Not only would that make the goal of compatibility clear, but means adding support in OpenUSD would be straightforward, and the USD toolset would actually be usable on mtlz packages.

jstone-lucasfilm:
> Building on @dgovil 's initial comment, and inspired by a question from @meshula, the most concise, least-perturbance to the ecosystem way of formulating this spec would be by formal reference to the USDZ spec, with enumeration of the differences, which is primarily replacing allowable contained filetypes (`usd, usda, usdc`) with `mtlx`, and removing the audio filetypes. Not only would that make the goal of compatibility clear, but means adding support in OpenUSD would be straightforward, and the USD toolset would actually be usable on mtlz packages.

I really like your suggestion to share definitions between MTLZ and USDZ, but perhaps the reference should be in the reverse direction, in the same spirit of our current referencing of the MaterialX specification from the USD specification in the AOUSD?

As long as our two teams closely _collaborate_ on the definition of MTLZ, making sure it fully meets the needs of both projects, it seems natural to reference the MTLZ specification from the USDZ specification, making the latter a superset of the former.  This approach would work smoothly in glTF as well, where it can reference the standalone definition of MTLZ without bringing in concepts from an independent scene graph framework.

spiffmon:
> I really like your suggestion to share definitions between MTLZ and USDZ, but perhaps the reference should be in the reverse direction, in the same spirit of our current referencing of the MaterialX specification from the USD specification in the AOUSD?
> 
> As long as our two teams closely _collaborate_ on the definition of MTLZ, making sure it fully meets the needs of both projects, it seems natural to reference the MTLZ specification from the USDZ specification, making the latter a superset of the former. This approach would work smoothly in glTF as well, where it can reference the standalone definition of MTLZ without bringing in concepts from an independent scene graph framework.

There's a difference between referencing the MaterialX spec in the UsdShade specification, where we are defining what the opt-in domain of materials and shading means, vs the **core spec**'s task of simply defining a container type for generic scenegraph data, and it would be a bit of a disconnect to reference a materials standard in that context (I believe the only external specifications the core references are unicode, zip, and lz4?).

But my main concern/request is that we be really clear about the problems that need solving, here.  Nowhere in this proposal (apologies I have not read the entirety of the originating slack thread) is it obvious that `.glb` packages would need or want to directly embed any MaterialX container in them, and therefore there need be no reference to them in the glTf specification.  Glb is already a perfectly competent container format, and unlike USD, which is strongly concerned with being able to represent and preserve **source asset structure and fidelity**, it is already incumbent on glTf packagers to deal with the entire complexity of baking, optimizing, and distilling a multitude of source scenegraph and texture formats down into the highly gpu-optimized representations at the core of the glTf spec.  So I conclude (and I could be entirely missing something important!) that any MaterialX container would be a concern only of the DCC's and services that already deal with USD, openExr, and all the other "source fidelity" encodings.

Web delivery and manipulation of materials is mentioned as a use-case for this container, but if you were selling or allowing the creation of materials in a web app, wouldn't you expect to be delivering source-quality textures that could then be consumed and quanitized/processed/compressed independently for many different needs, which would then be converted to a super-compressed format of a quality level selected independently for each endpoint/use-case?

meshula:
> but perhaps the reference should be in the reverse direction

Note that core spec has discussed an intent to publish the container definitions separately, for a clear separation of concerns. So I'd suggest "where" the container specs "live" is of a broader interest. It would probably be a good idea to bring to the TAC to see where the consensus currently lies on factoring the containers from the core specification.

I don't think it would make sense for the specification to live in MaterialX, but it would make sense for MaterialX and USD to refer to the same specification; where it lives now is USD core spec. So any factoring of that specification has consequences and would need consensus building not to mention review, voting, ratification, and new core spec publication. So not an inconsequential or quick undertaking.

jstone-lucasfilm:
@spiffmon This is an area where I still have more to learn from the USD/GLTF communities, but I would only expect the proposed MTLZ file to be referenced from _pure text_ scene files such as USDA and GLTF, not from binary scene files such as USDZ and GLB.  Since the latter formats have their own binary packaging conventions, I would expect them to extract the raw MTLX data from an MTLZ, and then to encode it using their own standards.  In the case of a USDZ, I wouldn't see any problem with the data being encoded as pure UsdShade, and in the case of GLB, I wouldn't see any problem with the data being encoded using their [MaterialX-aligned procedural textures extension](https://github.com/KhronosGroup/glTF/tree/KHR_texture_procedurals/extensions/2.0/Vendor/EXT_texture_procedurals_mx_1_39).

@meshula I see your point about the problems of hosting these conventions entirely within MaterialX, and I wonder if the best analogy here would be NanoColor, where the ASWF now hosts a set of conventions that can be leveraged by MaterialX, USD, and others in the future.  Going forward, would it make sense for the ASWF to host a specification for packaging conventions that could similarly be leveraged by MaterialX, USD, and others in the future?

I wouldn't necessarily hold up @bhouston's proposal on the establishment of this shared set of conventions, but it seems like a path forward that would allow the USD/MaterialX ecosystem to harmonize on a standard without excluding groups such as Khronos, which may find it easier to align with an ASWF-hosted specification than a USD-hosted specification.

jstone-lucasfilm:
@meshula I like your suggestion to have a shared specification for efficient ZIP packaging conventions, and I wonder if the best analogy here would be NanoColor, where the ASWF hosts an independent standard for color spaces that is currently leveraged by MaterialX and USD, and can easily be used by other material and scene formats in the future.  By analogy, would it make sense for the ASWF to host a specification for _packaging_ conventions that could similarly be leveraged by MaterialX, USD, OTIO, and others?  Hosting this as an independent standard under the ASWF would make it maximally accessible, and wouldn't require cross dependencies between Khronos and Pixar standards.

@spiffmon This is an area where others in the USD/glTF communities have more knowledge than I do, but I would only expect the proposed MTLZ file to be referenced from _pure text_ scene files such as USDA and GLTF, not from binary scene files such as USDZ and GLB.  Since the latter formats have their own binary packaging conventions, I would expect them to extract the raw MTLX data from an MTLZ, and then to encode it using their own standards.  In the case of a USDZ, it seems logical for the material data to be encoded as UsdShade, and in the case of GLB, it seems logical for the material data to be encoded using their [MaterialX-aligned procedural textures extension](https://github.com/KhronosGroup/glTF/tree/KHR_texture_procedurals/extensions/2.0/Vendor/EXT_texture_procedurals_mx_1_39).

To my mind, @bhouston's MTLZ proposal is most valuable in the context of _material libraries_, as a more principled replacement for the ad hoc MaterialX libraries found today in Poly Haven, AMD GPUOpen, and other sources.  With a well-defined standard for the contents of a MTLZ file, I would expect these open material libraries to become even more popular in the future, providing a common base of material data for the USD and glTF ecosystems.

Issue #2881, opened 2026-05-05
Is MaterialXView's auto flipping on load of the V texcoord of glTF models a bug?


bhouston:
This piece of code here is run MaterialXView by default upon glTF load (I'll save you the stack trace but it is the case):

```cpp
// Perform v-flip
if (isTexCoordStream && v == 1)
{
    if (!texcoordVerticalFlip)
        {
            floatValue = 1.0f - floatValue;
        }
    }
```

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXRender/CgltfLoader.cpp#L402

I think this is a bug.  I think this is trying to compensate for MaterialX assuming the UV space is flipped (e.g V'= 1-V) as compared to OpenGL/WebGL, etc.  But this compensation on load only appears to works but is completely wrong if any math done on the V coordinate prior to texture access.

Proof:
```
prefliped V with math != post-flipped V with math

flipV(V) + 0.1 != flipV( V + 0.1 )
( 1 - V ) + 0.1 != 1 - ( V + 0.1 )
1.1 - V != 0.9 - V
```

The correct fix I believe is to flip V (e.g. V' = 1 - V) when fetching the texel from the texture or there are options to have OpenGL/WebGL flip the textures automatically upon load.

This was caught by [Brecht Van Lommel](https://projects.blender.org/brecht) from blender here: https://projects.blender.org/blender/blender/issues/158138#issuecomment-1920144

BTW it appears that MaterialXView also flips V on texture access - so it is just straight out wrong to do the pre-flip as well:

```GLSL
vec2 mx_transform_uv(vec2 uv, vec2 uv_scale, vec2 uv_offset)
{
    uv = uv * uv_scale + uv_offset;
    return vec2(uv.x, 1.0 - uv.y);
}
```
https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/genglsl/lib/mx_transform_uv_vflip.glsl

bhouston:
Further investigate reveals the original cause of the bug.  There are two obj files that have inverted UV coordinates compared to what is desired, thus there is special case to enable flipping in those cases:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/e8c12a2baae2419d5e66676a5aac197505e3dc1a/source/MaterialXTest/MaterialXRenderGlsl/RenderGlsl.cpp#L270

[...]

PR #2882, opened 2026-05-05
Remove multiple flips of the V texcoord in MaterialXView (upon glTF load and when accessing textures)

bhouston:
Fixes the issue identified here: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2881

The MaterialXViewer flips V texture coordinates of obis and glTF files upon load.  It also flips the V texture coordinate when accessing textures.  I believe both of these flips are incorrect.  Neither should exist.  This PR removes both of them.

It leads to simpler code but also now the UV space of a mesh does not have to be modified in order to render it.

Lastly, math operations in a flip UV space are not the same as in a non-flipped UV space.  Thus this flipped UV space leads to incorrect math results:

```
prefliped V with math != post-flipped V with math

flipV(V) + 0.1 != flipV( V + 0.1 )
( 1 - V ) + 0.1 != 1 - ( V + 0.1 )
1.1 - V != 0.9 - V
```

And also pre flipped + post-flipped V doesn't equal no flipping:
```
flipV( flipV(V) + 0.1 ) !== V + 0.1
1 - ( 1 - V + 0.1 ) != V + 0.1
1 - 1 + V - 0.1 != V + 0.1
V - 0.1 != V + 0.1
```

### DIFF
--- a/source/MaterialXGraphEditor/RenderView.cpp
+++ b/source/MaterialXGraphEditor/RenderView.cpp
@@ -160,7 +160,6 @@ RenderView::RenderView(mx::DocumentPtr doc,
 
     // Set default Glsl generator options.
     _genContext.getOptions().targetColorSpaceOverride = "lin_rec709";
-    _genContext.getOptions().fileTextureVerticalFlip = true;
     _genContext.getOptions().hwShadowMap = true;
     // Make sure all uniforms are added so value updates can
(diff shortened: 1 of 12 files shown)


bhouston:
This was an incorrect PR.  glTF's UV space is flipped compared to MaterialX/Blender.

Issue #1518, opened 2023-09-07
Build and runtime error on MacOS Catalina


pablode:
When trying to build MaterialXView, I encounter following error with Xcode 12.4 on Catalina 10.15.7 (sorry for the outdated setup!):
```
[100%] Linking CXX executable ../../bin/MaterialXView
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_MTLTileRenderPipelineDescriptor", referenced from:
      objc-class-ref in libMaterialXRenderMsl.a(MetalState.mm.o)
ld: symbol(s) not found for architecture x86_64
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
make[3]: *** [bin/MaterialXView] Error 1
make[2]: *** [source/MaterialXView/CMakeFiles/MaterialXView.dir/all] Error 2
make[1]: *** [source/MaterialXView/CMakeFiles/MaterialXView.dir/rule] Error 2
make: *** [MaterialXView] Error 2
```
The issue can be resolved as follows
```diff
--- a/source/MaterialXGenMsl/CMakeLists.txt
+++ b/source/MaterialXGenMsl/CMakeLists.txt
@@ -21,15 +21,18 @@ set_target_properties(
     OUTPUT_NAME ${MATERIALX_MODULE_NAME}${MATERIALX_LIBNAME_SUFFIX}
     COMPILE_FLAGS "${EXTERNAL_COMPILE_FLAGS}"
     LINK_FLAGS "${EXTERNAL_LINK_FLAGS}"
+    INTERFACE_LINK_LIBRARIES "-undefined dynamic_lookup"
     INSTALL_RPATH "${MATERIALX_SAME_DIR_RPATH}"
     VERSION "${MATERIALX_LIBRARY_VERSION}"
     SOVERSION "${MATERIALX_MAJOR_VERSION}")
```
which, however, leads to a number of linker warnings.

Starting MaterialXView causes a sanitizer error
```
/Users/pablode/tmp/MaterialX/source/MaterialXRenderMsl/MetalState.mm:60:29: runtime error: load of value 190, which is not a valid value for type 'bool'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/pablode/tmp/MaterialX/source/MaterialXRenderMsl/MetalState.mm:60:29 in
[1]    48884 abort      ./bin/MaterialXView
```
which can be fixed using following patch:
```diff
[...]

PR #2884, opened 2026-05-05
Fix uninitialized member variable in MetalState

warrenm:
This PR seeks to address issue #1518.

I believe the first part of this issue to be obsolete. The CMake targets have changed substantially since the issue was reported, and I was able to build with Xcode 12.4 on macOS Catalina 10.15.8 without errors.

The second part of the issue is still relevant. When compiled on a system with a macOS 11+ SDK but running on an OS version prior to macOS 11, the `supportsTiledPipeline` member variable is not initialized by the constructor or the `initialize()` method, and can hold an invalid value. This doesn't cause a crash because of further runtime checks in `initLinearToSRGBKernel()`, but it does raise a runtime issue when compiled with UBSAN.

These changes ensure that the `supportsTiledPipeline` variable is always initialized, regardless of SDK version or OS version.

Tested on macOS Catalina 10.15.8 with Xcode 12.4 and macOS Tahoe 26.4.1 with Xcode 26.4.1.

### DIFF
--- a/source/MaterialXRenderMsl/MetalState.h
+++ b/source/MaterialXRenderMsl/MetalState.h
@@ -56,7 +56,7 @@ struct MX_RENDERMSL_API MetalState
     id<MTLRenderCommandEncoder> renderCmdEncoder = nil;
     std::stack<MaterialX::MetalFramebufferPtr> framebufferStack;
 
-    bool supportsTiledPipeline;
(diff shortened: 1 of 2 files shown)


jstone-lucasfilm (approved):
This looks like a correct fix to me, and the missing initializer in some MacOS environments was a valid bug.  Thanks @warrenm!

Issue #2885, opened 2026-05-06
Hextile discrepancy between OSL and Metal/GLSL


bhouston:
I recently added OSL and Metal rendering to the Material Fidelity Test Suite: https://material-fidelity.ben3d.ca

I discussed there is ordering discrepancy between applying scale and tiling with the Hextile now.  OSL does it in the opposite order than Metal/GLSL.  This wasn't caught because all existing hextile examples, such as oynx, either do tiling or scale, but not both.

If you have scaling and tiling you get problems like this:

<img width="632" height="280" alt="Image" src="https://github.com/user-attachments/assets/5c9f2d9a-9e99-4239-b41a-3e42905c3450" />

If I were to propose a PRs to fix these issue, should I correct Metal/GLSL or OSL?  I ask because if people are using MaterialX OSL code in production fixing this bugs may be sensitive.

bhouston:
@jstone-lucasfilm has replied off Github to say fix the OSL to match the GLSL/Metal code base.  I will do that next.

jstone-lucasfilm:
Thanks for this report, @bhouston, as well as for the fix in #2888!

PR #2888, opened 2026-05-06
Bug fix: tiling/scaling order in OSL hextiling to match GLSL

bhouston:
This fix addresses the bug outlined here: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2885


jstone-lucasfilm (approved):
These fixes look good to me, thanks @bhouston!

Issue #2887, opened 2026-05-06
OSL Transformmatrix implementation uses column-vector, rather than MaterialX's row-vector convention


bhouston:
I recently added OSL and Metal rendering to the Material Fidelity Test Suite: https://material-fidelity.ben3d.ca

In GLSL/Metal it does v' = vM where as in OSL it does v' = Mv.

In the MaterialX docs, it seems to suggest that row-vector, e.g v' = vM, is the correct way.

If you have scaling and tiling you get problems like this:

<img width="1380" height="653" alt="Image" src="https://github.com/user-attachments/assets/a1eeb37d-acbb-4c22-aed2-5a56154058b4" />

If I were to propose a PRs to fix these issue, should I correct Metal/GLSL or OSL?  I ask because if people are using MaterialX OSL code in production fixing this bugs may be sensitive.

bhouston:
@jstone-lucasfilm has replied off Github to say fix the OSL to match the GLSL/Metal code base.  I will do that next.

jstone-lucasfilm:
Thanks for posting this, @bhouston, though I'm realizing it duplicates the Good First Issue in https://github.com/AcademySoftwareFoundation/MaterialX/issues/2833, which is the subject of a PR for Dev Days.  Let's mark this as a duplicate, so that our Dev Days participant can finish their PR, borrowing any additional ideas from this issue as needed.

PR #2889, opened 2026-05-06
Bug fix: OSL transformmatrix for matrix44 to match GLSL/Metal

bhouston:
This fixes issue #2887

The original issue was not row versus column, it was just that the vector-matrix multiply was not treated as a point being multiplied by a homogeneous matrix.  Thus its translation was lost.


jstone-lucasfilm:
Thanks for posting this, @bhouston, though I'm realizing it duplicates the Dev Days PR at https://github.com/AcademySoftwareFoundation/MaterialX/pull/2843.  Let's mark this as a duplicate, so that our Dev Days participant can finish their PR, borrowing any additional ideas from your implementation as needed.

PR #2890, opened 2026-05-07
Optimize GLSL/OSL `rotate3d` using Rodrigues' rotation formula

bhouston:
While fixing some issues with the rotate3d for my blender importer, I noticed that we can replace the matrix44 construction for just rotating a vector with Rodrigues' formula:

https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula

This PR makes that change.  Fidelity test passes with successful results so there was no degradation.

### DIFF
--- a/libraries/stdlib/genglsl/mx_rotate_vector3.glsl
+++ b/libraries/stdlib/genglsl/mx_rotate_vector3.glsl
@@ -1,6 +1,9 @@
 void mx_rotate_vector3(vec3 _in, float amount, vec3 axis, out vec3 result)
 {
-    // https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula
+    // based on https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula
+    // but the code in the Wikipedia article is for v' = M * v, where as 
+    // MaterialX is v' = v * M, thus the order of parameters into the first cross are reversed
+
     axis = normalize(axis);
     float rotationRadians = mx_radians(amount);
     float s = mx_sin(rotationRadians);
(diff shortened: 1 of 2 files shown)


libraries/stdlib/genglsl/mx_rotate_vector3.glsl
```diff
@@ -1,19 +1,10 @@
...
 {
+    // https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula
+    axis = normalize(axis);
     float rotationRadians = mx_radians(amount);
-    mat4 m = mx_rotationMatrix(axis, rotationRadians);
-    result = mx_matrix_mul(m, vec4(_in, 1.0)).xyz;
+    float s = mx_sin(rotationRadians);
+    float c = mx_cos(rotationRadians);
+    float oc = 1.0 - c;
+    result = _in * c + cross(_in, axis) * s + axis * dot(axis, _in) * oc;
```
jstone-lucasfilm, line 12: Are the two terms in the cross product above _reversed_, with respect to the Rodrigues formula at https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula?

I would have expected the cross product to read `cross(axis, _in`), but it's very possible that I'm misreading.

jstone-lucasfilm (commented):
Thanks @bhouston, I think this optimization is generally a good idea, though I had a question about one detail in the math.

libraries/stdlib/genglsl/mx_rotate_vector3.glsl
bhouston: The wikipedia article's code is for v\' = M\*v, but the convention in MaterialX is v\' = v\*M.  Thus the sign change / flipped parameters to the cross product.

I have re-confirmed that the replacement of the existing code with this new code produces identical results via the fidelity test suite.

bhouston: I will update this PR to add a comment that clarifies this to avoid confusion in the future.

jstone-lucasfilm (approved):
Thanks for the explanation and updated comments, @bhouston, and this looks good to me.

PR #2893, opened 2026-05-07
Align OSL `normalmap` tangent frame with GLSL/MSL

bhouston:
## Problem
The `color3_vec3_cm` standard surface sample rendered incorrectly with `materialx-osl` compared with `materialx-glsl`, `materialx-metal`, and other renderers. Investigation showed the color-management and `color3 -> vector3` conversion path was correct: the generated shader applied `srgb_texture_to_lin_rec709`, then converted the resulting `color3` to `vector3` before feeding `normalmap`.

The mismatch came from the OSL normal-map implementation and tangent-frame setup. OSL used an OSL-specific Gram-Schmidt path and `dPdv`-based bitangents, while the hardware backends use a tangent plus `cross(N, T)` bitangent convention.

Screenshot 2026-05-07 at 2 27 38 PM

This also affected "tangent" and "binormal" nodes, oil results were a mismatch compared to glsl/metal:

Screenshot 2026-05-07 at 2 26 16 PM

## Solution
Align the OSL normal-map path with the GLSL/MSL convention:

- Update `mx_normalmap.osl` to use the same normal-map formula as the hardware implementations.
- Normalize OSL tangent geomprops and derive bitangents as `cross(N, T)`.
- Update the local `materialx-osl` and OSL render-test tangent/bitangent overrides to use the same convention.

## How It Was Realized
The generated OSL for the failing sample was inspected to confirm the color transform and `convert_color3_vector3` graph were emitted correctly. The divergence was isolated to the tangent-space normal-map path. Since OSL `testrender` does not currently expose mesh tangent attributes through the OBJ path, the fix uses the available OSL derivatives consistently and derives the bitangent from `N` and normalized `dPdu`, matching the hardware backend convention more closely.

Screenshot 2026-05-07 at 2 27 47 PM

And tangent and binormal nodes in oil now match glsl/metal:

Screenshot 2026-05-07 at 2 26 32 PM

### DIFF
--- a/libraries/stdlib/genosl/mx_normalmap.osl
+++ b/libraries/stdlib/genosl/mx_normalmap.osl
@@ -1,21 +1,7 @@
 void mx_normalmap_vector2(vector value, vector2 normal_scale, vector N, vector T, vector B, output vector result)
 {
-    if (value == vector(0.0))
-    {
-        result = N;
-    }
-    else
-    {
-        // The OSL backend uses dPdu and dPdv for tangents and bitangents, but these vectors are not
-        // guaranteed to be orthonormal.
-        //
-        // Orthogonalize the tangent frame using Gram-Schmidt, unlike in the other backends.
-        //
-        vector v = value * 2.0 - 1.0;
-        vector Tn = normalize(T - dot(T, N) * N);
-        vector Bn = normalize(B - dot(B, N) * N - dot(B, Tn) * Tn);
-        result = normalize(Tn * v[0] * normal_scale.x + Bn * v[1] * normal_scale.y + N * v[2]);
-    }
+    vector v = (dot(value, value) == 0.0) ? vector(0.0, 0.0, 1.0) : value * 2.0 - 1.0;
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm (approved):
Thanks for this fix, @bhouston, and it's a clear parity improvement that we should merge.  Aligning the OSL normal-map math and cross(N, T) bitangent convention with the GLSL/MSL generators is the right approach (the hardware path uses implicit bitangents by default, so dropping the dPdv path is exactly correct), and the before/after comparisons are convincing.

Just as a reference for future readers, there are two details that are worth calling out explicitly, in case we find that they're areas for future improvements:

1. This intentionally removes the Gram-Schmidt orthonormalization to match GLSL/MSL, which don't re-orthogonalize the tangent frame.  The trade-off is that the OSL frame is no longer guaranteed orthonormal.  This is consistent with the other generators, but worth knowing if we ever require universal orthonormalization.
2. This uses dPdu as a stand-in for a real mesh tangent because testrender doesn't expose tangents through the OBJ path.  We should think about whether/how to surface genuine mesh tangents in OSL, and how handedness behaves for mirrored/flipped UVs on production OSL renderers.

PR #2897, opened 2026-05-07
HLSL backend (codegen + {D3D11 & D3D12) renderer)

soufianekhiat:
| HLSL | GLSL | Diff (x8) |
|---|---|---|
| hlsl | glsl | diff |

Adds `MaterialXGenHlsl` and `MaterialXRenderHlsl`, + Python and JS bindings, plus the matching libraries under `libraries/{stdlib,pbrlib,nprlib,lights}/genhlsl/`. Existing GLSL / MSL / Slang code is untouched.

The codegen reuses the GLSL `.glsl` impl files via `file="../genglsl/..."` and runs a small post-emit pass for the GLSL→HLSL deltas (`mix→lerp`, `dFdx→ddx`, vector-splat C-cast, `texture()→mx_texture_sample`, etc.). Two HLSL-native helpers (`mx_math.hlsl`, `mx_texture.hlsl`) fill in the rest.

The renderer is D3D11. FXC for SM 5.x, DXC loaded dynamically for SM 6+. `HlslMaterial` owns per-stage cbuffers with a CPU mirror so partial uniform writes don't clobber neighbours — D3D11 cbuffers are stateful, no `glProgramUniform` analogue. `HlslTextureHandler` is an `ImageHandler` subclass. `HlslRenderer` auto-binds camera, lights, env, file textures and multi-mesh geometry from reflection.

Tests: 7 cases / 3369 assertions in `[genhlsl]`, 19 / 213 in `[renderhlsl]`. Headless CI safe - every D3D-touching case skips
cleanly when `tryCreateContext()` returns null. The full FXC compile sweep (~22 min) is tagged `[!slow]` so quick CI runs can skip it.

Validation: 31 materials rendered side-by-side against GLSL on the shaderball (StandardSurface, OpenPBR, glTF PBR, Disney, SimpleHair), plus the 15-material chess_set scene. All visually identical; 30/31 single-material RMSE under 2 on the 0-255 scale. See [Compare](https://github.com/soufianekhiat/MaterialX/issues/1).

Build: `MaterialXGenHlsl` cross-platform, `MaterialXRenderHlsl` gated on `WIN32`. Stages `dxcompiler.dll` + `dxil.dll` from the Windows SDK.

## Why not "just emit HLSL with Slang"?

Slang would give us HLSL source, not a D3D11 renderer - we'd still need most of `MaterialXRenderHlsl`. And FXC/DXC reflection round-trips MaterialX uniform names directly with native emit; through Slang they get mangled, breaking the per-uniform `setVariable` API that GLSL/MSL already provide.

Disclosure: This PR was created assicted with Claude Opus 4.7.


jstone-lucasfilm:
This proposal looks very promising, thanks @soufianekhiat, and I'd be interested in thoughts from @ashwinbhat, who has given a good deal of thought to strategies for HLSL integration in MaterialX.

Back in 2022, Ashwin presented an overview of different HLSL approaches to the MaterialX TSC, and I'll share the Google Doc from that presentation:

https://docs.google.com/document/d/1UamANfIivGFcWAB3URlJ6CRf_KEsuAGLSRbXkfzcsxU/edit?tab=t.0#heading=h.lyr31crlybwd

ZapAndersson:
Heh, I was looking at having a Robot Friend do the same thing, but you beat me to it!

For my use, I have a very specific requirement, though, because I need to glue my HLSL into some special constraint.

I need the the per-node genhlsl/*.hlsl library files are guaranteed to be valid standalone HLSL _functions_ (no entry-point glue, callable in isolation). If yes, that's worth documenting because it opens the door to host integrations that consume MaterialX nodes as fragments rather than full shaders. 

Will your implementation be able to do that?

ZapAndersson:
Ok I downloaded built and tested this, and it seems to do everything I need already - awesome!

libraries/stdlib/genhlsl/lib/mx_math.hlsl
```diff
@@ -0,0 +1,179 @@
...
+float4 mx_ddx(float4 p) { return ddx(p); }
+float  mx_ddy(float  p) { return ddy(p); }
+float2 mx_ddy(float2 p) { return ddy(p); }
+float3 mx_ddy(float3 p) { return ddy(p); }
+float4 mx_ddy(float4 p) { return ddy(p); }
+
+/// The GLSL we are piggybacking on has all its matrices transposed compared to HLSL and the MaterialX spec.
+/// (The matrices are defined like mat3(1, 2, 3, 4, 5, 6, 7, 8, 9) where the spec says it should be row-major order, but GLSL creates it as col-major)
+/// So when GLSL code says "mul(M, v)" it means "v * transpose(M)", and since in HLSL the matrices are stored
+/// in row-major order (when declared without a layout qualifier), we need to reverse the order of multiplication to get the same result.
```
ZapAndersson, line 30: OkayOkay, super-nitpicky detail here; (and I hope I'm not misunderstanding the comment here) we need to very clearly state and document that your code is expecting the host to upload matrices in the "GLSL order" because that is what MaterialX currently does. However, in my use case I may want to put this into a completely different, HLSL native host, which will then upload matrices in the original row-major order.

So my suggestion is do one of the following:

a) Just very clearly upfront state with blinking arrows and warnings signs this matrix upload contract
or
b) Add a GenOption where I can set what format *my* particular host code happens to upload matrices in.

Either works for me, it just stood out as a minor detail.

ZapAndersson: Or maybe I am confusing myself? :)

ashwinbhat:
> This proposal looks very promising, thanks @soufianekhiat, and I'd be interested in thoughts from @ashwinbhat, who has given a good deal of thought to strategies for HLSL integration in MaterialX.
> 
> Back in 2022, Ashwin presented an overview of different HLSL approaches to the MaterialX TSC, and I'll share the Google Doc from that presentation:
> 
> https://docs.google.com/document/d/1UamANfIivGFcWAB3URlJ6CRf_KEsuAGLSRbXkfzcsxU/edit?tab=t.0#heading=h.lyr31crlybwd

+1 on the proposal. Great work and I'm glad that this proposal includes MaterialXRenderHlsl for testing purposes to ensure future maintainability. 

The pipeline I had proposed in 2022 focused on ease of maintainability and leveraged existing transpiling tools so that we would have single shadergen (GLSL), but generate additional targets for Metal, Vulkan and DX.

Some downsides of the use of transpiling tools are
- pipeline can be cumbersome to setup for realtime workflows 
- generated shaders are hard to read and debug especially the code comments are lost due to SPIRV representations. 

Since we have direct implementation for other HW target, I would support a HLSL backend that can be refined to better use new HLSL semantics and feature levels.

ashwinbhat:
@soufianekhiat Would it be a lot of effort to support DX12? Do you think we might need some utilities to support extraction of shader metadata for root signature generation?

soufianekhiat:
> @soufianekhiat Would it be a lot of effort to support DX12? Do you think we might need some utilities to support extraction of shader metadata for root signature generation?

It depends if it's for production readyness I would add a [D3D12MemoryAllocator](https://github.com/GPUOpen-LibrariesAndSDKs/D3D12MemoryAllocator) as a dependency. Otherwise I can build a simpler version only for the viewer.

ashwinbhat:
> > @soufianekhiat Would it be a lot of effort to support DX12? Do you think we might need some utilities to support extraction of shader metadata for root signature generation?
> 
> It depends if it's for production readyness I would add a [D3D12MemoryAllocator](https://github.com/GPUOpen-LibrariesAndSDKs/D3D12MemoryAllocator) as a dependency. Otherwise I can build a simpler version only for the viewer.

For production, I think the hlsl shader should suffice. I was suggesting DX12 for the testrenderer and viewer.

soufianekhiat:
Not forgot the PR, just need few time to add the dx12 path.

libraries/stdlib/genhlsl/lib/mx_math.hlsl
soufianekhiat: Fixed comment.

PR #2904, opened 2026-05-08
Bug fix: incorrect transpose in Metal inverse world matrices

bhouston:
The world inverse matrix uniforms in metal are incorrectly transposed.  This removes this incorrect transposes

You can tell pretty clearly that this is incorrect because when it sets the WORLD_MATRIX just a few lines above, it doesn't transpose it, so why transpose the inverse?

https://github.com/bhouston/MaterialX/blob/fe92050e77f6f64984958abf0461747369c713b4/source/MaterialXRenderMsl/MslPipelineStateObject.mm#L1163

This was found via the fidelity test suite.


jstone-lucasfilm (approved):
Great catch, @bhouston, and your fix looks correct to me!

PR #2906, opened 2026-05-08
Align OSL implementation of `cellnoise3d` with GLSL

bhouston:
I will be honest, I am not sure why this fixes the issue but it does.

Prior to this fix cellnoise3d looked different for OSL as compared to Metal/GLSL:

Screenshot 2026-05-08 at 4 56 37 PM

After this fix they match:

Screenshot 2026-05-08 at 4 57 53 PM

### DIFF
--- a/libraries/stdlib/genosl/mx_cellnoise3d_float.osl
+++ b/libraries/stdlib/genosl/mx_cellnoise3d_float.osl
@@ -1,4 +1,6 @@
 void mx_cellnoise3d_float(vector position, output float result)
 {
-    result = cellnoise(position);
+    // Use the explicit point overload for deterministic 3D cell lookup.
+    // Avoid relying on implicit vector conversion in the OSL builtin.
+    result = cellnoise(point(position));
 }


jstone-lucasfilm (approved):
This change looks good to me, @bhouston, as there's no `cellnoise(vector)` signature in OSL, so the previous code relied upon an implicit conversion that could resolve in an unexpected (or implementation dependent) way.

Additionally, your updated approach aligns with how we use `cellnoise` in OSL flake node implementations.

Let's move forward with this fix, and thanks for tracking this down!

fpsunflower:
Would it be possible to get a copy of the generated osl code for this shader? We would be interested in investigating exactly what OSL was doing here, the overload resolution should have been treating treating `vector` and `point` the same way.

jstone-lucasfilm:
Good question, @fpsunflower, and I'm CC'ing @bhouston to see if he has access to the OSL output for his examples above.

lgritz:
Any follow-up on this?

PR #2907, opened 2026-05-08
Use `windows-2022` for VS2022 in GitHub CI

jstone-lucasfilm:
This changelist updates GitHub CI to use `windows-2022` consistently for VS2022 builds, as an upcoming change to GitHub Actions will replace `windows-2025` with a VS2026-focused environment.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,7 +116,7 @@ jobs:
           cmake_config: -G "Visual Studio 17 2022" -A "Win32"
 
         - name: Windows_VS2022_x64_Python313
-          os: windows-2025
+          os: windows-2022
           architecture: x64
           python: 3.13
           cmake_config: -G "Visual Studio 17 2022" -A "x64"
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Let's merge this as an administrative update, as the GitHub change noted above will occur next week, and it's valuable to have consistent CI for Dev Days.

lgritz:
Do you have a reference for this change?

jstone-lucasfilm:
@lgritz This week, we've been seeing the following notices for `windows-2025` jobs in our GitHub CI:

```
NOTICE: windows-2025 requests are being redirected to windows-2025-vs2026 by May 12, 2026
```

That would be a problem for our previous CI configuration, as we were assuming `windows-2025` has access to VS2022, and this isn't present in `windows-2025-vs2026`.

Usually I'd expect to see formal documentation of this change in the [runner-images](https://github.com/actions/runner-images) project, but I don't see any documentation beyond the notices themselves, so go figure!

PR #2908, opened 2026-05-10
Initial render comparison in GitHub CI

jstone-lucasfilm:
This changelist takes a step toward broader render validation in our GitHub CI, adding an extended job that generates a downloadable HTML/PDF report comparing rasterized (MSL) and ray-traced (OSL) renders of the test suite side-by-side with per-pixel differences, allowing contributors to validate shading updates in their fork without building OSL locally.  The following specific changes are included:

- Add an extended CI job that builds OSL v1.14.10.0 from source, runs MSL and OSL render tests at reference quality, and uploads an HTML/PDF comparison report as an artifact.
- Add the `OpenPBR` and `UsdPreviewSurface` example folders to the render test suite, replacing the less-critical `pbrlib/surfaceshader` folder, and adding a `renderTestExcludeFiles` option to skip individual files.
- Tune reference-quality sample counts to fit within the render capabilities and time budget of GitHub CI: rasterization supersample factor 8 -> 2, OSL lit -aa 32 -> 16 (1024 -> 256 spp), OSL unlit -aa 8 -> 2 (64 -> 4 spp).
- Centralize reference quality parameters in `getRasterEnvSampleCount` and `getRasterSupersampleFactor` helpers, removing duplicated constants across hardware shading targets.
- Replace the `MATERIALX_RENDER_MSL_ONLY` flag with `MATERIALX_TEST_RENDER_GLSL` for clarity, and replace the `enableReferenceQuality` test flag with a `MATERIALX_TEST_REFERENCE_QUALITY` build option.

As a logical next step, we might consider publishing the HTML artifact as a MaterialX Fidelity Suite (in the spirit of https://github.com/bhouston/material-fidelity), allowing any member of the community to view the current state of MaterialX render parity at a glance.


.github/workflows/main.yml
```diff
@@ -191,6 +193,33 @@ jobs:
         C:/vcpkg/vcpkg install mdl-sdk --triplet=x64-windows-release
         Add-Content $env:GITHUB_PATH "C:/vcpkg/installed/x64-windows-release/bin"
 
+    - name: Install OSL
```
kwokcb, line 196: I'm curious if any caching can be done here to not have to rebuild on every check-in ?  I assume it's a noticeable amount of time and could gate the time for getting back CI results. 

Another point of note is could this be put into a script file for maintenance and reuse. e.g. if a local dev wants to build OSL, this would be a useful script.

.github/workflows/main.yml
```diff
@@ -200,7 +229,7 @@ jobs:
 
     - name: Install Python Dependencies
       if: matrix.python != 'None'
-      run: pip install setuptools
+      run: pip install setuptools ${{ env.IS_EXTENDED_BUILD == 'true' && matrix.extended_build_osl == 'ON' && 'Pillow' || '' }}
```
kwokcb: Might be better to create a `requirements.txt` file and install that. Then it's flexible and reusable.
e.g.  Switching to use OpenImageIO instead of Pillow would only require changing the requirements file.

.github/workflows/main.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           compiler: gcc
           compiler_version: "14"
           python: 3.14
-          test_render: ON
+          render_captures: ON
```
kwokcb: No big deal. This just sounds a bit strange to me. 
Maybe something like `render_reference_images` instead ?

.github/workflows/main.yml
```diff
@@ -281,37 +312,58 @@ jobs:
...
 
+    - name: Generate Render Comparison Report
+      if: env.IS_EXTENDED_BUILD == 'true' && matrix.extended_build_osl == 'ON'
+      run: |
+        python "$GITHUB_WORKSPACE/python/MaterialXTest/tests_to_html.py" \
+          -i1 Materials \
+          -l1 msl -l2 osl -d \
+          --order-from "$GITHUB_WORKSPACE/resources/Materials/TestSuite/_options.mtlx" \
+          -o MaterialX_RenderComparison.html
+        CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
```
kwokcb, line 358: Seems a bit strange to use Chrome print to convert.  Suggest to use something like `Playright` CI action or install something like `Pandoc`.

kwokcb: Also I seems for ability to view easily it could just be left as html.

python/MaterialXTest/tests_to_html.py
```diff
@@ -53,12 +78,16 @@ def main(args=None):
...
     parser.add_argument('-l3', '--lang3', dest='lang3', action='store', help='Third target language for comparison. Default is empty', default="")
     parser.add_argument('-e', '--error', dest='error', action='store', help='Filter out results with RMS less than this. Negative means all results are kept.', default=-1, type=float)
+    parser.add_argument('-of', '--order-from', dest='order_from', action='store', help='Path to a MaterialX _options.mtlx file. When provided, output sections are ordered to match its renderTestPaths input.', default="")
 
     args = parser.parse_args(args)
 
     fh = open(args.outputfile,"w+")
     fh.write("<html>\n")
     fh.write("<style>\n")
+    # Preserve inline background-color when printing to PDF, so transparent renders
```
kwokcb, line 88: Is the background color for the image not set when rendering ? Or is it set to be transparent ?

kwokcb (commented):
This looks like a great step to getting reference compares up.
There are a few points left in the code, with the most noteworthy is the time to build OSL which looks like it will occur for every check-in.
Also just wanted to check that a regular Mac render test run will still by default render out both GLSL and Metal but the safeguards are in place for MaterialXView.

.github/workflows/main.yml
jstone-lucasfilm: Although caching is a valid option for the future, the OSL build step only occurs in nightly/manual builds of MaterialX, rather than per-commit builds.  Additionally, the OSL install/build step is remarkably fast on MacOS, taking just 3 minutes and 16 seconds in the latest run!

If we wanted to add caching in the future, we should be careful about the increased _cache-invalidation surface area_ of this approach, where subtle changes in dependencies can lead to caching errors that are hard to track down.

.github/workflows/main.yml
jstone-lucasfilm: For now, since this is just a single line, and the Pillow dependency isn't repeated in other steps or scripts, I'd recommend staying with the explicit statement here.  If our set of render dependencies grows beyond just Pillow, though, I'd say you're right, and that an independent list of requirements would be worthwhile.

.github/workflows/main.yml
jstone-lucasfilm: I'm very open to ideas on improving this name, though `render_captures` is the flag associated with our Viewer/Editor frame captures, not with the render test suite.  Naming it `render_reference_images` might blur this distinction, making it sound like this flag is associated with the render test suite.

.github/workflows/main.yml
jstone-lucasfilm: I tried a few different approaches before settling on Chrome, and here are the advantages that I found with Chrome:

- Chrome comes _pre-installed_ on MacOS environments, while Playwright/Pandoc have additional dependencies that require separate installation.
- Chrome handles _colors_ more accurately than the other approaches I tried, and honors the `print-color-adjust: exact` setting in CSS.
- Some alternatives (e.g. `weasyprint`) failed to handle the formatting in our HTML, while Chrome handles them well.

On your note about providing HTML directly, this workflow uploads the full HTML and source images as well, giving developers the choice of which they prefer to view.

python/MaterialXTest/tests_to_html.py
jstone-lucasfilm: Yes, the PNGs have alpha and the table cells supply a black backdrop.

jstone-lucasfilm:
Thanks for your notes above, @kwokcb!

To answer your question about MacOS default behavior, there are two distinct cases worth separating out:

- Render tests: `MATERIALX_TEST_RENDER_GLSL` defaults to ON, so a default Mac build still runs both the GLSL and MSL render-test paths.  Only the extended MacOS CI job opts out by passing -DMATERIALX_TEST_RENDER_GLSL=OFF.  It's the same effective behavior as the old MATERIALX_RENDER_MSL_ONLY=ON, just renamed.
- MaterialXView: Removing the MSL_ONLY branch is a no-op for the standard Mac build, as `USE_OPENGL_BACKEND_ON_APPLE_PLATFORM` defaults to OFF on macOS ≥ 10.14, which already selects the Metal backend.  Anyone who hasn't explicitly opted into OpenGL continues to get Metal.

resources/Materials/TestSuite/_options.mtlx
```diff
@@ -62,14 +62,10 @@
...
+    <input name="renderTestPaths" type="string" value="resources/Materials/Examples/StandardSurface,resources/Materials/Examples/OpenPbr,resources/Materials/Examples/UsdPreviewSurface,resources/Materials/TestSuite/pbrlib/bsdf,resources/Materials/TestSuite/stdlib/procedural,resources/Materials/TestSuite/stdlib/color_management,resources/Materials/TestSuite/stdlib/convolution,resources/Materials/TestSuite/nprlib" />
 
-    <!-- Enable reference quality rendering.
-         This option enables higher sample counts and supersampling in render tests,
-         allowing for visual comparisons and differencing across shading languages,
-         but requiring a more powerful GPU and longer CPU render times.
-    -->
-    <input name="enableReferenceQuality" type="boolean" value="false" />
+    <!-- List of document base names to exclude from render tests -->
+    <input name="renderTestExcludeFiles" type="string" value="standard_surface_look_brass_tiled.mtlx,standard_surface_look_wood_tiled.mtlx,standard_surface_chess_set.mtlx" />
```
ld-kerley, line 71: why are these files being excluded - would be good to include a comment explaining why - so we don't change this in the future without understanding the thought process here.

.github/workflows/main.yml
ld-kerley: relying on chrome being delivered from the MacOS GitHub image feels like we're exposing ourselves to changing changes in the future - as opposed to installing some explicitly versioned package - as Bernard requests. The GitHub CI images seem to change a worryingly frequent amount.

.github/workflows/main.yml
ld-kerley: having an external `requirements.txt` would make it a little more straight forward to replicate the setup of these tests locally? 

I think it would be good to move in a direction where the environment curation and launching of the CI tests are managed by scripts that can also be run locally - that would improve the developer experience.

.github/workflows/main.yml
ld-kerley: I'm with Bernard here - when I read `captures` it wasn't immediately obvious to me what these were. Perhaps `render_refernce_tool_images`?

resources/Materials/TestSuite/_options.mtlx
jstone-lucasfilm: Good point, and done!

.github/workflows/main.yml
jstone-lucasfilm: @ld-kerley I'm open to adding an explicit `brew install --cask google-chrome` here if that sounds valuable, but it's worth noting that a missing Chrome is not fatal to the workflow, and the GitHub CI would simply generate HTML only and emit a warning.  Let me know if you'd like me to add a command along these lines.

.github/workflows/main.yml
jstone-lucasfilm: @ld-kerley I'm very open to the local reproducibility argument that you make here, though it seems slightly premature to introduce a `requirements.txt` file for the single install of `Pillow` for render testing.  Perhaps we should wait until there's a second requirement and then take this step?

.github/workflows/main.yml
jstone-lucasfilm: Thanks @ld-kerley, and I'm open to coming up with a better name in this thread.

My hesitation with `render_reference_tool_images` is the use of the word "reference", which overlaps our usage of `MATERIALX_TEST_REFERENCE_QUALITY` and other terminology in the render test suite.

This flag actually gates a mix of render captures across Python and C++ tools -- `baketextures.py`, `translateshader.py`, `MaterialXView`, and `MaterialXGraphEditor` -- so perhaps `tool_captures` would strike the right balance?

.github/workflows/main.yml
kwokcb: That’s what I was wondering. Could we leave a comment for clarity, otherwise looks good.

.github/workflows/main.yml
kwokcb: I think it’s still better to setup this in a file, even if it’s one dependency. I doubt anyone would look at ci for dependicies she running locally.

.github/workflows/main.yml
kwokcb: `capture_tool_renders` ? Btw, won’t hold up the or for the naming :).

.github/workflows/main.yml
kwokcb: Perhaps we can iterate on this. As said, I think it’s fine to just leave as html.

kwokcb (approved):
Talked with Jonathan offline about requirements.txt. Rest is good enough to go for this initial change

hybridherbst:
Is there a place where I can look at the rendered images already? (I couldn't find it)

jstone-lucasfilm:
@hybridherbst GitHub has a size limit for files posted in comments, so I'll post it in the Slack thread for today's TSC session.  Going forward, I'd love to post the HTML directly as an ASWF-hosted site, which would work around this issue in a compelling way.

kwokcb:
The metal images in resources are packaged as part of ci. I think you'll also get the osl images now. It can be downloaded, though I think with this pr you don't get the comparison html output. Guess when an aswf repo is available the results can automated to be put there along with preview html or markdown.

jstone-lucasfilm:
@kwokcb The generated artifact includes _both_ HTML and PDF versions, so you can choose which you'd like to view.  Once this PR is merged, we should see these new artifacts generated in every nightly build, but for now they're only in the fork of MaterialX associated with the in-progress PR.

kwokcb:
That's great. Should be pretty trivial to automate grabbing artifacts.

.github/workflows/main.yml
jstone-lucasfilm: Got it, and I've made the change.

.github/workflows/main.yml
jstone-lucasfilm: Got it, thanks @kwokcb, and I've proposed `tool_render_captures` in the latest update to the PR.

PR #2876, opened 2026-04-29
Add specification proposal for maxcomponent and mincomponent nodes

chinmaychahar:
## Summary

Add specification proposal for `maxcomponent` and `mincomponent` nodes to the Math Nodes section of the Specification Proposals document.

Closes #2802

## Description

Per the discussion in #2802 with @jstone-lucasfilm :

- These nodes support **all vectorN and colorN types** (`vector2`, `vector3`, `vector4`, `color3`,
`color4`)
- Output type is **float**

The proposal format follows existing nodes such as `magnitude` and `dotproduct` in the Standard Nodes document.

### Use Cases

As mentioned in the issue, two places in the MaterialX data libraries currently use verbose 5-node patterns (3 `extract` + 2 `max`/`min`) that these nodes would simplify to a single node:

1. **Sheen intensity for glTF PBR** (`libraries/bxdf/gltf_pbr.mtlx`) — uses `maxcomponent` pattern
2. **Absorption coefficient for OpenPBR Surface** (`libraries/bxdf/open_pbr_surface.mtlx`) — uses `mincomponent` pattern

### Scope

This PR addresses the specification proposal only.  I can work on the follow-up implementation via graph definitions as a
separate PR once the proposal is reviewed and merged.

Issue #2802, opened 2026-03-08
Add specification proposal for `maxcomponent` and `mincomponent` nodes


jstone-lucasfilm:
The MaterialX Standard Nodes would benefit from the addition of `maxcomponent` and `mincomponent` nodes, allowing convenient computation of the maximum and minimum components of `vectorN` and `colorN` type values.

Two examples where this node would be used in the MaterialX data libraries are in the computation of [sheen intensity for glTF PBR](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/gltf_pbr.mtlx#L240) and the[ absorption coefficient for OpenPBR Surface](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/gltf_pbr.mtlx#L240).

For this GitHub Issue, we'd like a developer to write a proposal for these nodes to the [Math Nodes section of the Specification Proposals document](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Proposals.md#math-nodes), which can then be reviewed by the community and merged to main.

Once the proposal has been merged to main, the same developer can optionally tackle the work of _implementing_ these nodes using graph definitions, but the central goal of this GitHub Issue is to write the specification proposal itself.

chinmaychahar:
Hi @jstone-lucasfilm ! I’d like to work on this.

chinmaychahar:
I had a couple of clarifying questions:
1. Should these nodes support all vector and color types?
2. Would the output type be a float for max/min component?

Also, is there an existing spec proposal I should follow as a template for structure? Or can I get started by modifying [Math Nodes section of the Specification Proposals document](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Proposals.md#math-nodes)?

jstone-lucasfilm:
Thanks for volunteering, @chinmaychahar, and I'll assign it to you!

On your questions:

1. Yes, they should support all `vectorN` and `colorN` types in MaterialX.
2. That's exactly right, the output type for these nodes would be `float`.

We don't yet have a specification proposal, and providing that language in the Math Nodes section of our Specification Proposals document is the right place to start in addressing this GitHub Issue.

chinmaychahar:
Hi @jstone-lucasfilm , I opened PR #2876 to add the specification proposals for both the nodes, to the Proposals document.

[...]

PR #2909, opened 2026-05-11
Add `maxcomponent` and `mincomponent` nodes

chinmaychahar:
## Summary

Implement `maxcomponent` and `mincomponent` nodes using graph definitions, as proposed in #2802.

The specification proposal was merged in #2876

## Description

This PR adds node definitions, nodegraph implementations, and test suite entries for `maxcomponent` and `mincomponent`. These nodes compute the maximum and minimum component of vectorN and colorN values, outputting a float.


chinmaychahar:
Hey @jstone-lucasfilm as discussed [here](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2876#issuecomment-4347215682), this code change implements both the max and min node components. 

This PR's scope is only to implement the nodes and for now no use-cases have been updated.

Lemme know your thoughts! Thanks

libraries/stdlib/stdlib_ng.mtlx
```diff
@@ -3153,6 +3153,262 @@
     <output name="out" type="float" nodename="N_mtlxmagnitude" />
   </nodegraph>
 
+  <!--
+    Node: <maxcomponent>
+    Output the maximum component of the incoming vector or color stream.
+  -->
+  <nodegraph name="NG_maxcomponent_color3" nodedef="ND_maxcomponent_color3">
+    <extract name="N_extract_0" type="float">
```
kwokcb: You should be able to use `separate` nodes instead of multiple extracts. This also avoids storing and testing intermediate values in case they become precision sensitive.

resources/Materials/TestSuite/stdlib/math/vector_math.mtlx
```diff
@@ -139,4 +139,52 @@
     </transpose>
     <output name="out" type="vector4" nodename="transformvector1" />
   </nodegraph>
+  <nodegraph name="maxcomponent_color3">
```
kwokcb, line 142: For visual testing, I'd suggest to make the max value be the same for all tests. You might also want to add a test where 2 or more components differ only by a very small delta to test precision.

libraries/stdlib/stdlib_ng.mtlx
jstone-lucasfilm: @chinmaychahar I think @kwokcb's suggestion above is a good one, and using `separateN` nodes would be better aligned with existing graph definitions than `extract`.  Does this seem like a reasonable improvement to you as well?

jstone-lucasfilm:
This looks like a great start, @chinmaychahar, and my one addition to @kwokcb's suggestions above would be to move your specification definition of these nodes from the [Proposals](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Proposals.md#maxcomponent) document to the [Standard Nodes](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.StandardNodes.md#math-nodes) document.

This will allow us to merge the specification and data library changes as a single PR, keeping them in sync.

chinmaychahar:
Sure, lemme take a look and address the comments today

libraries/stdlib/stdlib_ng.mtlx
chinmaychahar: Yes, let me fix this

jstone-lucasfilm (approved):
This looks great to me, thanks @chinmaychahar!

PR #2910, opened 2026-05-11
Fix edge case behavior of `dielectric_bsdf` in MDL

krohmerNV:
Address an issue that was discovered with OpenPBR.
The specular lobe was still visible with specular_weight=0 and specular_ior=1.

Improved the comment that explains the reasoning for chosing custom_curve over fresnel in the dielectric_bsdf.


jstone-lucasfilm (approved):
This looks like an improvement to me, thanks @krohmerNV!

krohmerNV:
Thank you!

PR #2911, opened 2026-05-11
Bug fix: align osl range with glsl range via clamp-> max/min

bhouston:
I noticed via the Fidelity Suite (https://materials-fidelity.ben3d.ca) that the range renders did not align between GLSL and OSL.  I have narrowed down the issue to that OSL uses a Clamp operation, where as GLSL uses nested Min/Max.  These can, when the range is inverted, produce different results.

For the parameters:
* in = 0.50
* low = 0.80
* high = 0.20 (note low > high)

Nested min/max produces: 0.20

```
min(max(in, low), high)
= min(max(0.50, 0.80), 0.20)
= min(0.80, 0.20)
= 0.20
```

Where as OSL's clamp produces: 0.8

```
if (in < low) return low;
else if (in > high) return high;
else return in;

0.50 < 0.80 is true
result = 0.80
```

Thus PR thus replaces OSL's native clamp with the equivalent to GLSL's nested min/max so that it produces identical results across the parameter range.

An alternative solution, would be to replace GLSL and other renderer's code with OSL equivalents.  I can see the argument for either solution.

### DIFF
--- a/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
+++ b/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
@@ -436,17 +436,17 @@
   <implementation name="IM_sign_vector4_genosl" nodedef="ND_sign_vector4" target="genosl" sourcecode="sign({{in}})" />
 
   <!-- <clamp> -->
-  <implementation name="IM_clamp_float_genosl" nodedef="ND_clamp_float" target="genosl" sourcecode="clamp({{in}}, {{low}}, {{high}})" />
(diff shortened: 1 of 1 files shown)


ld-kerley:
Rather than picking one language over the other. What about resolving this already broken case more agnostically.

We could say that if `low` is greater than `high` then we just swap `low` and `high` so they are correctly ordered again.  Or choose to set `low` equal to `high` or `high` equal to `low` - but that choice feels a little arbitrary too

jstone-lucasfilm:

PR #2913, opened 2026-05-11
Bug fix: guard asin/acos inputs outside [-1, 1] in osl

bhouston:
I noticed via the https://material-fidelity.ben3d.ca test suite that out of range asin/acos values render different in MaterialX OSL as compared to the GLSL / Metal versions.

Before fix render comparisons:
Screenshot 2026-05-11 at 12 17 56 PM

This updates the `genosl` stdlib implementations for `asin` and `acos` to explicitly guard out-of-domain inputs so that the resulting value matches that produced by materialx glsl.

- `ND_asin_float`: `asin(in)` -> `mx_ternary(abs(in) <= 1.0, asin(in), 0.0)`
- `ND_acos_float`: `acos(in)` -> `mx_ternary(abs(in) <= 1.0, acos(in), 0.0)`

## Why

In degenerate-but-valid authoring cases, some graphs intentionally feed values outside `[-1, 1]` into inverse trig nodes. Backend behavior for out-of-domain inputs can differ (NaN propagation, clamping, or undefined results), which causes renderer divergence.

This change makes `genosl` behavior explicit and deterministic for those edge cases, improving cross-backend consistency.

## Validation

After fix render (I also fixed the blender-materialx-importer as well) comparisons:
Screenshot 2026-05-11 at 12 18 36 PM

## Notes

This PR is intentionally minimal and only changes `genosl` source expressions for float `asin`/`acos`.

### DIFF
--- a/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
+++ b/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
@@ -387,8 +387,8 @@
   <implementation name="IM_sin_float_genosl" nodedef="ND_sin_float" target="genosl" sourcecode="sin({{in}})" />
   <implementation name="IM_cos_float_genosl" nodedef="ND_cos_float" target="genosl" sourcecode="cos({{in}})" />
   <implementation name="IM_tan_float_genosl" nodedef="ND_tan_float" target="genosl" sourcecode="tan({{in}})" />
-  <implementation name="IM_asin_float_genosl" nodedef="ND_asin_float" target="genosl" sourcecode="asin({{in}})" />
-  <implementation name="IM_acos_float_genosl" nodedef="ND_acos_float" target="genosl" sourcecode="acos({{in}})" />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
@bhouston This is an important issue to address in the cross-language parity for MaterialX, but I don't think the solution is quite as simple as adding new guards in OSL.

See https://github.com/AcademySoftwareFoundation/MaterialX/pull/2914#issuecomment-4432346378 for thoughts on how we might proceed with this category of improvement.

fpsunflower:
OSL clamps the input to `[-1,1]` to avoid common bugs like `acos(dot(a,b))` returning `NaN` when the dot product of two unit vectors is slightly outside `[-1,1]`.

I feel like returning 0 in such cases is not a very intuitive behavior, and would require additional clamping in common cases, bringing the total number of clamps in the OSL case up to 3 which doesn't seem great.

Maybe we should consider aligning other shader languages with the OSL spec instead?

jstone-lucasfilm:

PR #2914, opened 2026-05-11
Bug fix: guard ln against nonpositive inputs in OSL

bhouston:
I noticed OSL rendered negative ln node inputs different than GLSL and Metal.  It is because it lacks a guard against negative values.

The visual results were this:

Screenshot 2026-05-11 at 12 33 42 PM

So I upated the `genosl` float `ln` implementation to handle nonpositive inputs explicitly:

- `ND_ln_float`: `log(in)` -> `mx_ternary(in > 0.0, log(in), 0.0)`

## Why

Degenerate-but-valid MaterialX graphs may feed zero or negative values into `ln`. In those cases, backend behavior can diverge due to undefined/non-finite math handling. This produced visible mismatch in `materialx-osl` versus the reference renderer behavior.

Making the fallback explicit in `genosl` improves deterministic behavior and cross-renderer consistency for these edge cases.

## Validation

The visual results across the renderers now match after this fix:

Screenshot 2026-05-11 at 12 33 51 PM

### DIFF
--- a/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
+++ b/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
@@ -416,7 +416,7 @@
   <implementation name="IM_sqrt_vector4_genosl" nodedef="ND_sqrt_vector4" target="genosl" sourcecode="sqrt({{in}})" />
 
   <!-- <ln> -->
-  <implementation name="IM_ln_float_genosl" nodedef="ND_ln_float" target="genosl" sourcecode="log({{in}})" />
+  <implementation name="IM_ln_float_genosl" nodedef="ND_ln_float" target="genosl" sourcecode="mx_ternary({{in}} > 0.0, log({{in}}), 0.0)" />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
@bhouston This is an important issue to address in the cross-language parity for MaterialX, but I don't think the solution is quite as simple as adding new guards in OSL.

In GLSL, for example, computing `log` of a negative number is officially [undefined](https://registry.khronos.org/OpenGL-Refpages/gl4/html/log.xhtml), and the driver implementation makes the choice about whether to return zero, return NaN, or any other behavior.

As of today, we similarly declare that computing `ln` of a negative number is [undefined](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.StandardNodes.md#ln) in MaterialX, but there have been robust discussions of the idea that we should formally _specify_ a behavior for negative numbers instead.

I'm very open to that idea, and the best place to propose improvements to the formal specification of `ln` would be in the [Math Nodes](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Proposals.md#math-nodes) section of our Specification Proposals document.

jstone-lucasfilm:

PR #2915, opened 2026-05-11
Bug fix: smoothstep should emulate glsl behavior in degenerate cases

bhouston:
Like most of my PRs I noticed via https://material-fidelity.ben3d.ca that smooth step when the bounds are inverted does not replicate glsl behaviour.  This PR fixes that.

## Summary

Update the float `smoothstep` implementation in `genosl` to handle degenerate/inverted bounds explicitly.

- `ND_smoothstep_float` now uses:
  - normal `smoothstep(low, high, in)` when `high > low`
  - a deterministic fallback when `high <= low`: hard step at `high` (`in < high ? 0.0 : 1.0`)

## Why

For inverted bounds, `smoothstep` behavior can diverge across backends due to undefined/implementation-dependent handling. This caused visible mismatch in `materialx-osl` for degenerate node tests.

Making the fallback explicit in `genosl` improves determinism and cross-renderer consistency for these edge cases.

## Validation

Renders from OSL, Metal and GLSL now match:

Screenshot 2026-05-11 at 12 38 10 PM

### DIFF
--- a/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
+++ b/libraries/stdlib/genosl/stdlib_genosl_impl.mtlx
@@ -544,9 +544,9 @@
 
   <!-- <smoothstep> -->
   <implementation name="IM_smoothstep_float_genosl" nodedef="ND_smoothstep_float" target="genosl" sourcecode="mx_ternary({{high}} > {{low}}, smoothstep({{low}}, {{high}}, {{in}}), mx_ternary({{in}} < {{high}}, 0.0, 1.0))" />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
See https://github.com/AcademySoftwareFoundation/MaterialX/pull/2914#issuecomment-4432346378 for recommendations on resolving this category of undefined behaviors through new specification proposals, so that our per-language implementations have an unambiguous reference to match.

tdavidovicNV:
fwiw, there is already a doc change proposal for this, but old by now: https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1851

jstone-lucasfilm:

PR #2916, opened 2026-05-11
Bug fix: make remap deterministic for degenerate input ranges

bhouston:
Found via https://material-fidelity.ben3d.ca.  remap is not deterministic for metal, glsl or osl with them producing different results:

Screenshot 2026-05-11 at 12 49 41 PM

Define deterministic behavior for float `remap` when the input range is degenerate (`inhigh == inlow`).

- In GLSL codegen (`IM_remap_float_genglsl`), add a guard:
  - if `abs(inhigh - inlow) <= 1e-8`, return `outlow`
  - else use the standard remap expression
- In OSL support code (`mx_remap(float, ...)`), add the same guard before division.

## Why

`remap` with a zero-width input range is undefined under the raw formula due to division by zero. Different backends were producing different results for `nodes/remap_degenerate` because of NaN/Inf handling differences.

This change establishes a stable, cross-backend default:
- degenerate input range -> `outlow`

## Scope

- Limited to float remap behavior in `genglsl` and `genosl`.
- Non-degenerate remap behavior is unchanged.

## Validation

Re-rendered `nodes/remap_degenerate` with MaterialX renderers and they all match:

Screenshot 2026-05-11 at 12 55 01 PM


libraries/stdlib/genosl/include/mx_funcs.h
```diff
@@ -149,7 +149,12 @@ vector4 mx_extract_matrix44(matrix in, int index)
 
 float mx_remap(float in, float inLow, float inHigh, float outLow, float outHigh, int doClamp)
 {
-      float x = (in - inLow)/(inHigh-inLow);
+      float denom = inHigh - inLow;
```
ld-kerley, line 152: It's worth noting that in OSL "divide-by-zero" is not degenerate.  It's defined to return 0.

jstone-lucasfilm:
See https://github.com/AcademySoftwareFoundation/MaterialX/pull/2914#issuecomment-4432346378 for recommendations on resolving this category of undefined behaviors through new specification proposals, so that our per-language implementations have an unambiguous reference to match.

jstone-lucasfilm:

Issue #2922, opened 2026-05-13
`createValidChildName()` can create invalid documents


ld-kerley:
The `createValidChildName()` method is often called when adding a new node to a MaterialX document to ensure the newly created node has a unique name within the document.  If the document has referenced the standard data library in using `setDataLibrary()` as opposed to importing the contents with `importLibrary()`, then we can end up with a document that doesn't pass validation.

What can happen is that if a node is created with a name that matches the name of a child in the data library, then during validation we end up discovering the wrong child.  I'm including a python snippet below as an example of how this can go wrong where a node is named, matching an existing node definition name.  The `createValidChildName()` function does not create a unique name.

This happens because the names of the elements in the data library are not populated in to the documents `Element::_childMap`.  So the existing code 
```
    string createValidChildName(string name) const
    {
        name = name.empty() ? "_" : createValidName(name);
        while (_childMap.count(name))
        {
            name = incrementName(name);
        }
        return name;
    }
```
is not able to consider the names in the referenced data library.

Simple test to demonstrate.
```
import MaterialX as mx


def main():
    stdlib = mx.createDocument()
    search_path = mx.getDefaultDataSearchPath()
    library_folders = mx.getDefaultDataLibraryFolders()
    mx.loadLibraries(library_folders, search_path, stdlib)

    doc = mx.createDocument()
    doc.setDataLibrary(stdlib)

    shader_def = stdlib.getNodeDef("ND_standard_surface_surfaceshader")
    shader = doc.addNodeInstance(shader_def, "my_surface_shader")

    material_def = stdlib.getNodeDef("ND_surfacematerial")
    material = doc.addNodeInstance(material_def, "my_material")

    surface_input = material.addInput("surfaceshader", "surfaceshader")
    surface_input.setNodeName(shader.getName())

    valid, message = doc.validate()
    print(f"Validation #1 (after creation): valid={valid}")
    if message:
        print(f"  message: {message}")

    shader.setName("ND_standard_surface_surfaceshader")

    valid, message = doc.validate()
    print(f"Validation #2 (after rename): valid={valid}")
    if message:
        print(f"  message: {message}")


if __name__ == "__main__":
    main()
```

result
```
>python mtlx_validation_test.py
Validation #1 (after creation): valid=True
Validation #2 (after rename): valid=False
  message: Invalid port connection: <input name="surfaceshader" type="surfaceshader" nodename="my_surface_shader">
```

kwokcb:
[...]

PR #2923, opened 2026-05-13
Fix createValidChildName for Documents with referenced data libraries

ld-kerley:
This should address #2922.

### DIFF
--- a/source/MaterialXCore/Document.h
+++ b/source/MaterialXCore/Document.h
@@ -78,6 +78,18 @@ class MX_CORE_API Document : public GraphElement
         return _dataLibrary;
     }
 
+    /// Using the input name as a starting point, modify it to create a valid,
+    /// unique name for a child element.
+    string createValidChildName(string name) const override
+    {
+        name = name.empty() ? "_" : createValidName(name);
+        while (_childMap.count(name) || (_dataLibrary && _dataLibrary->getChild(name)))
+        {
+            name = incrementName(name);
+        }
+        return name;
+    }
+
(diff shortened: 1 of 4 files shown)


source/MaterialXCore/Document.h
```diff
@@ -78,6 +78,18 @@ class MX_CORE_API Document : public GraphElement
         return _dataLibrary;
     }
 
+    /// Using the input name as a starting point, modify it to create a valid,
+    /// unique name for a child element.
+    string createValidChildName(string name) const override
```
kwokcb: I don't know if this is better but would it be better to improve `getChild()` to take into account data libraries and use that as a central logic for query ? As it is now there is very similar code between the base + derived implementations that if the name map logic changes then they could deviate.

Note that data libraries are considered for `Element::getChildOfType() / getChildrenOfType()` so it it seems there is also an API consistency issue -- i.e. `getChild()` will not find the data library element but this method and the 2 others will.

kwokcb (commented):
Nice to see your looking at this. There have been a few very long Slack threads on this with no resolution.

I have left a API logic consistency comment. What do you think ?

source/MaterialXCore/Document.h
ld-kerley: I think that makes sense to me - I think I reflected your idea in the updated commit here - let me know if I missed the mark.

kwokcb (approved):
Thanks for the update. Looks great!

jstone-lucasfilm (approved):

Issue #2901, opened 2026-05-08
MaterialX UsdTransform2D rotation direction does not align with USD specification


ld-kerley:
The USD specification states explicitly that the rotation component in `UsdTransform2D` is counter-clockwise and specified in degrees around the origin. The USD specification also has an explicit callout that its transforming the texture coordinates, not the texture.

> NOTE : The UsdTransform2d node transforms texture coordinates, not the textures themselves, so to get the effect of moving, resizing, or rotating an image being applied to a surface, one must apply the inverse transformation of how you expect the image to move.

It appears that the MaterialX implementation doesn't follow this convention.

[rotate_test_file.zip](https://github.com/user-attachments/files/27533293/rotate_test_file.zip)

The test files zip contain both a USD version where you can see the rotation occurs in opposing directions between USD and MaterialX. 

### USD version
<img width="1257" height="1142" alt="Image" src="https://github.com/user-attachments/assets/3efba947-56b1-4682-a8e7-48ff531275a4" />

### MaterialX version
<img width="1257" height="1142" alt="Image" src="https://github.com/user-attachments/assets/be36dccc-9652-4fd9-a49c-2edb916b61f1" />

Caveat - the MaterialX version does have some modifications to work around other shortcomings in the MaterialX implementation of the UsdPreviewSurface node.

ld-kerley:
Actually turns out that on closer inspection all of the transforms are being applied inverted - likely due to a mis-understanding that is clarified in the _NOTE_ from the UsdPreviewSurface spec referenced above.

jstone-lucasfilm:
This sounds like a good observation to me, @ld-kerley, and I'm CC'ing @kwokcb for his thoughts, as he was the original author of the USD texturing and placement nodes in MaterialX.

kwokcb:
Review looks good and should be faster to compute as well.

PR #2924, opened 2026-05-13
Fix ND_UsdTransform2D implementation

ld-kerley:
Addresses #2901 

Reversed the direction of the transformations applied - aligning the MaterialX implementation with the USD implementation.

### DIFF
--- a/libraries/bxdf/usd_preview_surface.mtlx
+++ b/libraries/bxdf/usd_preview_surface.mtlx
@@ -430,14 +430,28 @@
   </nodegraph>
 
   <!-- Node: UsdTransform2d -->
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Great to see this fix, @ld-kerley!  I'm CC'ing @kwokcb, who should be the one to formally review this, but it seems like the right path forward.

kwokcb (approved):
This looks good and matches exactly https://openusd.org/dev/spec_usdpreviewsurface.html.
Nice to avoid the extra math for pivots and TRS ordering support.

jstone-lucasfilm (approved):
Ship it!

PR #2926, opened 2026-05-14
Fix alpha handling in HSV/RGB conversion in GLSL/OSL

### LINKED ISSUE #2765: Incorrect GLSL implementation of color4 typed RgbToHsv, HsvToRgb and HsvAdjust
Spec says alpha channel should be forwarded unchanged, implementation always sets to 1.
HsvAdjust is a node graph implementation, but by virtue of it using the RgbToHsv and HsvToRgb nodes, it inherits these bugs so is also bugged.

paytonGilbertson:
`mx_rgbtohsv` and `xm_hsvtorgb` had hardcoded output alpha set to `1.0`. The alpha channel should forward the preexisting alpha value `_in.a` unchanged.

Changes made in
- `/libraries/stdlib/genglsl/mx_hsvtorgb_color4.glsl`
- `/libraries/stdlib/genglsl/mx_rgbtohsv_color4.glsl`
- `/libraries/stdlib/genosl/mx_hsvtorgb_color4.osl`
- `/libraries/stdlib/genosl/mx_rgbtohsv_color4.osl`

This fixes #2765

### DIFF
--- a/libraries/stdlib/genglsl/mx_hsvtorgb_color4.glsl
+++ b/libraries/stdlib/genglsl/mx_hsvtorgb_color4.glsl
@@ -2,5 +2,5 @@
 
 void mx_hsvtorgb_color4(vec4 _in, out vec4 result)
 {
-    result = vec4(mx_hsvtorgb(_in.rgb), 1.0);
+    result = vec4(mx_hsvtorgb(_in.rgb), _in.a);
 }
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm (approved):
Thanks for the fix, @paytonGilbertson, and this looks correct to me!

PR #2927, opened 2026-05-14
Fix nullptr dereference in MaterialNode

### LINKED ISSUE #2918: nullptr dereference possible in MaterialNode::emitFunctionCall
An MTLX document that contains a `<surfacematerial>` inside of a `<nodegraph>` whose `surfaceshader` input is bound through the graphinterface with no upstream shader node causes `ShaderNode::getImplementation()` to dereference a null `shared_ptr`

```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <nodegraph name="green_material_graph">
    <input name="input_shader" type="surfaceshader" />
    <surfacematerial name="mat" type="material">
      <input name="surfaceshader" type="surfaceshader" interfacename="input_shader" />
    </surfacematerial>
    <output name="out" type="material" nodename="mat" />
  </nodegraph>
</materialx>
```

This can be picked up by the UBSan in Clang.

There is an example test case recreating the issue here https://github.com/seang-ilm/MaterialX/commit/389d0916cfe0b968b8e54ece5619342635e8cd73

```
$ctest -R MaterialXTest_GenShader_Surface_Material_In_Nodegraph_With_Interface_Input -VV
...
shared_ptr_base.h:980: runtime error: reference binding to null pointer of type
    'element_type' (aka 'MaterialX_v1_39_5::ShaderNodeImpl')
  #0  ShaderNode::getImplementation() const
  #1  ShaderStage::addFunctionCall(ShaderNode const&, GenContext&, bool)
  #2  ShaderGenerator::emitFunctionCall(ShaderNode const&, GenContext&, ShaderStage&) const
  #3  MaterialNode::emitFunctionCall(ShaderNode const&, GenContext&, ShaderStage&) const
```

This issue is likely that `MaterialNode::emitFunctionCall` is missing a call to `ShaderNode::isAGraph()` before calling `emitFunctionCall` on the connected node

https://github.com/AcademySoftwareFoundation/MaterialX/blob/acbd32ccfcabb88b10dcd0ad9d21a949d987c291/source/MaterialXGenShader/Nodes/MaterialNode.cpp#L47-L49

The same issue likely exists in `MaterialNode::addClassification`

codesavory:
Fixes #2918

## Summary
- Add `isAGraph()` guard in `MaterialNode::emitFunctionCall` to return early with default output variables when the `surfaceshader` input is bound through a graph interface with no upstream shader node, preventing a null dereference on `ShaderNode::getImplementation()`
- Add `isAGraph()` guard in `MaterialNode::addClassification` to skip classification inheritance in the same case

## Test
Regression test added by @seang-ilm in the linked branch covering the case where a `<surfacematerial>` is placed inside a `<nodegraph>` whose `surfaceshader` input is bound via `interfacename` to a graph interface input with no upstream node.


codesavory:
@jstone-lucasfilm this PR is ready for review!
@seang-ilm CC'd

I think the python wheel failure is a sporadic issue, let me know if I should re-run the builds.

jstone-lucasfilm:
Thanks @codesavory, and you're right that the Python Wheel error is just a fluke of GitHub CI, so I've requested a rebuild of that specific job.

jstone-lucasfilm (commented):
This fix looks excellent to me, thanks @codesavory.

My one recommendation would be to move your test example from an inlined string to a MTLX file in our test suite, e.g. `resources/Materials/TestSuite/stdlib/materials/surfacematerial_interface_input.mtlx`.  Once it's there, you likely don't need this additional test logic at all, as the new test suite example will automatically be run through our shader generation suite across all of the languages that MaterialX supports.

codesavory:
> This fix looks excellent to me, thanks @codesavory.
> 
> My one recommendation would be to move your test example from an inlined string to a MTLX file in our test suite, e.g. `resources/Materials/TestSuite/stdlib/materials/surfacematerial_interface_input.mtlx`. Once it's there, you likely don't need this additional test logic at all, as the new test suite example will automatically be run through our shader generation suite across all of the languages that MaterialX supports.

That was a test written by @seang-ilm. I took your feedback and updated the code. Good to know this architecture of files auto-executed as tests!

jstone-lucasfilm (approved):
Ship it!

Issue #2919, opened 2026-05-12
Add utility function to fix connection type mismatches


kwokcb:
### Context

It is possible for documents to contain connections between mismatched types.
This is not considered to valid based on the MaterialX specification and is caught by validation.

### Premise

MaterialX should provide core support to "patch" these mismatched type connections.
The proposal is to add in explicit `convert` nodes as appropriate noting some conversions
do no make sense since they cannot be type converted. 

#### Part Ia : Core Logic

- Add a new utility function to find type mismatches. e.g. `addExplicitTypeConversions()` or 
something better :).
- Consider the existing `convert` nodes as the set of available supported types. Fail / skip
if not supported.
- Otherwise insert the appropriate `convert` node between the existing connection. e.g
if there is a `vector3` output connected to a `color3` input, then create a convert node which
handles `vector3` to `colore` conversion and insert it between the existing connection.
- It would be useful as feedback to return the list of nodes added w.r.t. the connection, and
a list of connections which cannot be handled.

Note: There is overlap with the existing `validate()` logic which already performs detection.
It would be desirable to reuse calling into that logic, or refactor to create a separate reusable detection
function.

#### Part Ib: Test 
 
- Add in some test cases that perform validation, fix, and re-validation to make sure it 
works. This can be in C++ as part of unit tests.
- This is a composite of possible connection configurations to test:
   - node -> node connections
   - input interface -> node connections -> + output interface 
   - upstream nodes with single / multiple connections ("multioutput")
   - node -> input interface connections
   - output interface -> node connections
   - nesting of `GraphElement`s : Currently only `NodeGraph`s under a `Document`
   - fan out from upstream outputs to more than  1 input downstream input.

<img width="100%" alt="Image" src="https://github.com/user-attachments/assets/7e2dffdf-89fc-450f-8e1b-d0c02ff89751" />

 
#### Part II: API
 
- Add Python and Javascript bindings. 
 
#### Part III : API command

- Add in a sample command to perform type connection fixes.

---

If this is a single day dev-days project then Part I could be considered an initial task.

christopher-jacobs:
I would like to take a look at this issue for the upcoming Dev Days! If it's alright, I might start working on it today so that I can have a better chance of submitting a PR Thursday.

jstone-lucasfilm:
Thanks @christopher-jacobs, and I have assigned it to you!

kwokcb:
Additional clarification points:

[...]

PR #2929, opened 2026-05-15
Implemented Part I of addExplicitTypeConversions()

christopher-jacobs:
In regards to #2919 I implemented addExplicitTypeConversions() and added two tests to show a valid conversion and an invalid conversion. The valid test connects a vector3 output to a color3 output, and the invalid test attempts to connect a vector3 output to a float output. This currently loops through the entire NodeGraph to see if any conversions are possible, but I believe that this could be more easily implemented to be integrated with validate() (possibly taking the outputted message to find the questionable node and then check this connection directly).

This is my first time contributing to an open source project so let me know what can be improved and also any feedback. Looking at Parts II and III this is a very interesting and cool idea that I would love to continue if possible! Thank you!


kwokcb (commented):
This looks like a good start on this.

However, there are additional complications and scope.... 

1. You can have "interface" inputs (input nodes) connected to internal node inputs.
2. You have have internal node outputs connected to interface "outputs" (output nodes).  

The "get" / "set" connected nodes will probably not work in these cases as they are not stricty node-node connections and also they will traverse to the parent.

3. You need to consider not just `NodeGraphs` but "graphs" in general which is the parent class `GraphElement`. This will allow include of `Documents` which inherit from `GraphElement`. 

I will also add this to the issue, as this should be clarified there more succinctly. Apologies for the general language used.

christopher-jacobs:
Thank you for that feedback! I'll reiterate on this and make sure that I take these into consideration!

kwokcb:
As mentioned in the issue, I think it might be better to use some form of the validation logic to find the problem connections and then fix them afterwards.

source/MaterialXCore/Node.h
```diff
@@ -399,6 +399,11 @@ class MX_CORE_API NodeGraph : public GraphElement
     /// @param interfaceName The new interface name.
     void modifyInterfaceName(const string& inputPath, const string& interfaceName);
 
+    /// Modify the nodes that can be converted explicitly within the associated NodeGraph.
+    /// @param addedNodes List of added nodes that were successfully integrated (w.r.t. the connection).
+    /// @param invalidConnections List of connections that could not be successfully converted.
+    void addExplicitTypeConversions(vector<NodePtr>& addedNodes, vector<NodePtr>& invalidConnections);
```
ld-kerley: @kwokcb I'm curious stylistically if you think its important that this method exist on the NodeGraph class itself, or if instead we should consider it a true utility - and make it a stand-alone function.

In the back of my head I worry we start expanding the API surface of these objects, things may get harder in the future to manage API/ABI compatibility.

Also in any future world where we have the ability to feed ShaderGen with some other data model, if these "utility" functions existed stand-alone - we might be able to serve the same logic through the abstract interface, and say allow USD materials to also have conversion nodes inserted on demand.

kwokcb: I don't see any reason that it needs to be on the core classes (*). It does need an input object (`GraphElement`) but that could be wrapped up into some opaque wrapper. 

I'd prefer if these we're not global methods but instead add a "operation stack" base class which takes one or more "operations" (wrappers) to perform. You could create a MTLX version here and USD could create their own. 

Is this kind of what your thinking of?  ( Std disclaimer is that I haven't followed any of the ShaderGen refactor stuff of late.)

(*) @christopher-jacobs : This is not meant to overwhelm you with longer term complications. I think for this task it can just be on the `GraphElement` class.

christopher-jacobs: No worries, if anything I like observing this conversation and learning more about the different ways this could be implemented. I'll go ahead and implement the changes within `GraphElement` as we discussed earlier. I'm very interested though in this conversation and to explore and learn more! Thank you both!

jstone-lucasfilm:
Thanks for taking this on, @christopher-jacobs!

This task turned out to involve considerably more connection cases than the original issue suggested, and you've made real progress on a challenging problem for a first contribution.  The move to `GraphElement` and the recursion into child graphs are the right structure, and I'd like to give you the chance to carry this forward if you have the bandwidth.  If your availability has changed since Dev Days, that's completely understandable, and just let us know so that we can arrange for someone to build on your work.

From a fresh read of the current PR, here are the items I'd consider important before we move into detailed review:

1. The final loop over node inputs calls `resolveNameReference<NodeGraph>` with the input's `nodegraph` string unconditionally, and then calls `getOutput` on the result.  For the common case of an input with no `nodegraph` attribute, the lookup returns null and the next line dereferences it, so the function will crash on most documents.  Guarding that branch with `hasNodeGraphString()` should resolve it, and the same treatment applies to `getDeclaration`, which returns null for a `Document`, and to `getActiveValueElement`, which returns null for an unresolved interface name.

2. The new convert nodes are named with the nodedef name, or with fixed strings such as `convert_node`, and since `addChild` throws on a duplicate name, the second conversion of the same type pair in one graph will throw.  I'd suggest `addNodeInstance(convertDef, createValidChildName("convert"))`, which also sets the `nodedef` attribute, following the pattern of the existing convert insertion in `Version.cpp`.

3. `Input::getConnectedNode` follows interface names and nodegraph outputs into other scopes, which is the concern @kwokcb raised in May.  When it returns a node from the parent document or from inside a nodegraph, `setConnectedNode` on the new convert node writes a `nodename` that doesn't resolve in the convert node's own graph.  I'd suggest working from the raw `nodename`, `nodegraph`, `output`, and `interfacename` attributes of each port, so that each connection type is handled in the scope where it lives.  The nodegraph-output branch also needs a `continue`, since it currently falls through to the node-name branch and inserts a second convert node for the same input.

4. Types are compared against the upstream node's type rather than the upstream output's type, so a `multioutput` upstream node would look for a nonexistent `ND_convert_multioutput_*` nodedef and be reported as invalid.  Comparing against the specific output and carrying the `output` attribute onto the convert node's input should handle the multioutput case from the issue.

5. The test sections for the other connection cases are currently commented out, along with the call under test in the remaining section, so the new function isn't exercised by the suite.  Once the items above are in place, restoring those sections would give us the validate, fix, and re-validate coverage described in the issue.  It's worth noting that `validate` doesn't currently flag a type mismatch across an `interfacename` reference, so the interface tests will need to assert on the resulting graph structure rather than on a validation failure, and that gap is worth a note on #2919.

A few smaller notes:

- The `std::cout` calls and `<iostream>` includes in `Node.cpp` and the test should be removed, along with the commented-out code blocks.
- A pass with our `.clang-format` will bring the braces and spacing in line with the rest of MaterialXCore, and I'd suggest `string` over `std::string` and camelCase for local variables, matching the surrounding code.
- For `invalidConnections`, a vector of the downstream ports would identify each connection unambiguously, rather than pushing pairs of elements consecutively into a single vector.

Let us know how you'd like to proceed, and thanks again for your work on this!

kwokcb:
I think different cases can be handled in an incremental manner. Perhaps this could be a useful way to stage things:

1. Handle connections between nodes with same immediate parent. Start with document level.
2. Handle itme 1. at nodegraph level.
3. Validate these cases which can be thought of as connections between nodes that do not deal with graph interfaces.
4. Handle connections between nodegraphs and nodes.
5. Handle connections between nodegraphs and nodegraphs.
6. Handle connections between input interfaces and node inputs
7. Handle connections between node outputs and output interfaces.
8. Handle connections between input interfaces on nodedefs and inputs on nodes in functional graphs. This one is a bit more involved, and maybe should not be automated?

I am wondering if the approach taken by the graph editor might be easiest to conceptualize as it remaps all of the materialx cases into a more uniform data structure. Then work on that structure instead.

That is why I was thinking a validation preprocess can record incompatible links vs just errors. I haven't looked at this in a while so just revisiting with ideas.

christopher-jacobs:
Thank you Johnathan and Bernard for all your assistance on my first issue with MaterialX. I apologize for the amount of time it has been since I have revisited this issue. I have done some more work on this, but have not submitted a review with the edits because I want to refine it to be more suitable as your notes have suggested (thank you for those!).

I would like to continue working on this issue if at all possible in the meantime and implement your comments while using Bernard's incremental approach to chip away at this once again. It is very daunting of course but I would like to challenge myself so that I can continue to improve.

I will resume work on this issue so that I can submit another PR with the notes implemented using the incremental approach given. Thank you both once again!

PR #2931, opened 2026-05-15
Fix material switching in the MaterialX Web Viewer

jstone-lucasfilm:
This changelist fixes a regression in the MaterialX Web Viewer, where switching to any example material other than Standard Surface Default would fail to update the rendered material.

Since the update to `Element::getChild` in #2923, calling `Element::getDescendant` on a Document with a referenced data library can resolve uniform paths _into_ the library, returning elements such as NodeDef inputs whose parent is not a Node.  The Web Viewer's property editor then throws a `TypeError` when calling `getNodeDef` on the NodeDef parent, preventing materials other than Standard Surface Default (which declares all inputs explicitly) from displaying.

The fix is to skip descendants whose active source URI differs from that of the working document, ensuring the property editor only processes elements from the content document.

### DIFF
--- a/javascript/MaterialXView/source/viewer.js
+++ b/javascript/MaterialXView/source/viewer.js
@@ -1111,8 +1111,14 @@ export class Material
                     {
                         continue;
                     }
-                    
-                    // Skip non-input types and anything > 2 levels deep 
+
+                    // Skip elements from the referenced data library
+                    if (currentElem.getActiveSourceUri() !== elem.getDocument().getSourceUri())
+                    {
+                        continue;
+                    }
+
+                    // Skip non-input types and anything > 2 levels deep
                     if (!currentElem.asAInput() || currentElem.getNamePath().split('/').length > 2)
                     {
                         continue;


javascript/MaterialXView/source/viewer.js
```diff
@@ -1111,8 +1111,14 @@ export class Material
                     {
                         continue;
                     }
-                    
-                    // Skip non-input types and anything > 2 levels deep 
+
+                    // Skip elements from the referenced data library
+                    if (currentElem.getActiveSourceUri() !== elem.getDocument().getSourceUri())
```
kwokcb, line 1116: Would this work and be a bit more robust ?
```
if (currentElem.getDocument() != elem.getDocument())
```
As the working and data `Document` should always be different ?

jstone-lucasfilm: @kwokcb Although that approach seems promising at first glance, it would make this test a _pointer comparison_ instead of a string comparison, which is much less robust in the JavaScript/Emscripten context.  If we switched over to pointer comparisons, it seems likely that the `!=` comparison would _always_ return true, because of the nature of pointer bindings.

Another note is that we use the `getActiveSourceUri` string comparison pattern elsewhere in MaterialX when we're testing whether a given element belongs to the content document, as in [this example](https://github.com/AcademySoftwareFoundation/MaterialX/blob/b661a4bf653d4720c6eda678dd4676208ee53649/source/MaterialXGenShader/Util.cpp#L417).

kwokcb: I had wandered back into "C++ thinking" for a bit. This all makes sense.
It would be nice if it was a reusable utility as a follow-up at some point for maintenance.

kwokcb (approved):

javascript/MaterialXView/source/viewer.js
jstone-lucasfilm: That's a great suggestion, @kwokcb, and I agree an `Element::belongsToContentDocument` method would be much more elegant, and could be applied across our C++/Python/JavaScript codebase.

That sounds like a worthwhile Dev Days task of its own, if you'd like to add a GitHub Issue for this.

kwokcb: Added: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2934

PR #2932, opened 2026-05-15
Make mxvalidate autoload std libraries

zychen8:
issue2506-Make mxvalidate autoload std libraries

### DIFF
--- a/python/Scripts/mxvalidate.py
+++ b/python/Scripts/mxvalidate.py
@@ -12,7 +12,6 @@ def main():
     parser = argparse.ArgumentParser(description="Verify that the given file is a valid MaterialX document.")
     parser.add_argument("--resolve", dest="resolve", action="store_true", help="Resolve inheritance and string substitutions.")
     parser.add_argument("--verbose", dest="verbose", action="store_true", help="Print summary of elements found in the document.")
-    parser.add_argument("--stdlib", dest="stdlib", action="store_true", help="Deprecated: Standard MaterialX libraries are imported automatically.")
     parser.add_argument(dest="inputFilename", help="Filename of the input document.")
     opts = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


python/Scripts/mxvalidate.py
```diff
@@ -12,26 +12,23 @@ def main():
     parser = argparse.ArgumentParser(description="Verify that the given file is a valid MaterialX document.")
     parser.add_argument("--resolve", dest="resolve", action="store_true", help="Resolve inheritance and string substitutions.")
     parser.add_argument("--verbose", dest="verbose", action="store_true", help="Print summary of elements found in the document.")
-    parser.add_argument("--stdlib", dest="stdlib", action="store_true", help="Import standard MaterialX libraries into the document.")
+    parser.add_argument("--stdlib", dest="stdlib", action="store_true", help="Deprecated: Standard MaterialX libraries are imported automatically.")
```
jstone-lucasfilm: I'd take this one step further and fully _remove_ the `--stdlib` option, rather than silently accepting the input.  Although it's uncommon for this script to be used directly in studio pipelines, if teams were calling this script with `--stdlib`, it seems better for them to see the change right away and remove the flag from their own calls.

jstone-lucasfilm (commented):
This looks great to me, @zychen8, and I had just one suggestion for a simplification.

python/Scripts/mxvalidate.py
kwokcb: Agreed. This was what was the proposal in the issue outline.

zychen8:
Thank you for your suggestion! Just removed the flag.

kwokcb (approved):
Thanks for the updates. Looks good to me.

jstone-lucasfilm (approved):
Great work, thanks @zychen8!

PR #2933, opened 2026-05-15
Improving NodeGraph::getNodeDef performance

seang-ilm:
Fuzz testing was consistently findings some very slow inputs. Looking at this inputs through `perf`.

I've added a performance regression case that builds example trees that have this issue.

```
test -R MaterialXTest_NodeGraph_getNodeDef_performance -VV
...
27: -------------------------------------------------------------------------------
27: NodeGraph getNodeDef performance
27: -------------------------------------------------------------------------------
27: /home/sean/MaterialX_Clean/source/MaterialXTest/MaterialXCore/Performance.cpp:53
27: ...............................................................................
27:
27: benchmark name                       samples       iterations    estimated
27:                                      mean          low mean      high mean
27:                                      std dev       low std dev   high std dev
27: -------------------------------------------------------------------------------
27: validate, stubs=200 refs=100                   100             1     2.91805 m
27:                                         4.20239 ms    4.00369 ms    4.54723 ms
27:                                         1.30152 ms    875.207 us    2.10073 ms
27:
27: validate, stubs=1000 refs=100                  100             1      1.5165 s
27:                                         13.4589 ms    12.2686 ms    15.7144 ms
27:                                          8.1092 ms    5.15263 ms      13.99 ms
27:
27: validate, stubs=10000 refs=100                 100             1     23.5368 s
27:                                         154.384 ms    141.568 ms    192.638 ms
27:                                         103.788 ms    40.5831 ms    224.837 ms
27:
27:
27: 29.193 s: NodeGraph getNodeDef performance
27: ===============================================================================
27: test cases: 1 | 1 passed
27: assertions: - none -
27:
1/1 Test #27: MaterialXTest_NodeGraph_getNodeDef_performance ...   Passed   30.73 sec
```

I think I've found a couple of issues that come together to cause the processing time for loading .mtlx files grow linearly.

- `Syntax::makeValidName` uses `std::find` on `_reservedWords`. This is a `std::set<string>` and supports a faster O(log N) `.find()`. Give's a minor improvement (5-6%)

https://github.com/AcademySoftwareFoundation/MaterialX/blob/9ebe5f119adde1562bddf0e107d521eb733fd9a0/source/MaterialXGenShader/Syntax.cpp#L152-L162

- `NodeGraph::getNodeDef` calls `Document::getImplementations()` on every invocation. This rebuilds a fresh vector by scanning every document-level child with a dynamic_pointer_cast on each, with no result reuse. `Document::Cache` already maintains hash-map lookups for ports, nodedefs, and nodedDefImplementation, but the reverse direction (nodegraph-name -> implementation) was missing. I've added a _nodeGraphImplMap to the cache, populated during rebuild() alongside the existing maps, and replaced the linear scan with an O(1) lookup.

https://github.com/AcademySoftwareFoundation/MaterialX/blob/b661a4bf653d4720c6eda678dd4676208ee53649/source/MaterialXCore/Node.cpp#L742-L756

If we look at benchmark time I get a significant improvement

|Baseline|Time(s)|
|----|----|
|Unmodified|44.8|
|Syntax::makeValidName, fix _reservedWords check|41.94|
|Use Document::Cache lookups|12.34|

### DIFF
--- a/source/MaterialXCore/Document.cpp
+++ b/source/MaterialXCore/Document.cpp
@@ -69,6 +69,13 @@ class Document::Cache
         return (it != _implementationMap.end()) ? it->second : vector<InterfaceElementPtr>();
     }
 
+    ImplementationPtr getImplementationForNodeGraph(const string& nodeGraphName)
+    {
+        auto lock = refreshWithLock();
+        auto it = _nodeGraphImplMap.find(nodeGraphName);
+        return (it != _nodeGraphImplMap.end()) ? it->second : ImplementationPtr();
+    }
+
   private:
     std::shared_lock<std::shared_mutex> refreshWithLock()
     {
@@ -103,6 +110,7 @@ class Document::Cache
         _portElementMap.clear();
         _nodeDefMap.clear();
         _implementationMap.clear();
+        _nodeGraphImplMap.clear();
 
         // Traverse the document to build a new cache.
         for (ElementPtr elem : doc->traverseTree())
@@ -119,6 +127,14 @@ class Document::Cache
                 if (portElem)
                 {
                     _portElementMap[portElem->getQualifiedName(portKey)].push_back(portElem);
+                }
+            }
+            if (!nodeGraphName.empty())
+            {
+                ImplementationPtr impl = elem->asA<Implementation>();
+                if (impl)
+                {
+                    _nodeGraphImplMap[impl->getQualifiedName(nodeGraphName)] = impl;
                 }
             }
             if (!nodeString.empty())
@@ -152,6 +168,7 @@ class Document::Cache
     std::unordered_map<string, std::vector<PortElementPtr>> _portElementMap;
     std::unordered_map<string, std::vector<NodeDefPtr>> _nodeDefMap;
     std::unordered_map<string, std::vector<InterfaceElementPtr>> _implementationMap;
+    std::unordered_map<string, ImplementationPtr> _nodeGraphImplMap;
 };
 
 //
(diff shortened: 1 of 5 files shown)


source/MaterialXCore/Document.cpp
```diff
@@ -152,6 +168,7 @@ class Document::Cache
     std::unordered_map<string, std::vector<PortElementPtr>> _portElementMap;
     std::unordered_map<string, std::vector<NodeDefPtr>> _nodeDefMap;
     std::unordered_map<string, std::vector<InterfaceElementPtr>> _implementationMap;
+    std::unordered_map<string, ImplementationPtr> _nodeGraphImplMap;
```
ld-kerley, line 171: I think its possible for more than one Implementation element to use the same NodeGraph - so this might need to be

std::unordered_map<string, std::vector<ImplementationPtr>>

For standard surface ([here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/standard_surface.mtlx#L107)) we have two different Implementation elements that have the same NodeGraph

source/MaterialXGenShader/Syntax.cpp
```diff
@@ -153,7 +153,7 @@ void Syntax::makeValidName(string& name) const
 {
     std::replace_if(name.begin(), name.end(), isInvalidChar, '_');
     name = replaceSubstrings(name, _invalidTokens);
-    if (std::find(_reservedWords.begin(), _reservedWords.end(), name) != _reservedWords.end())
+    if (_reservedWords.find(name) != _reservedWords.end())
```
ld-kerley, line 156: This seems like a nice improvement :) good catch

source/MaterialXCore/Node.cpp
```diff
@@ -745,12 +745,10 @@ NodeDefPtr NodeGraph::getNodeDef() const
     // If not directly defined look for an implementation which has a nodedef association
     if (!nodedef)
     {
-        for (auto impl : getDocument()->getImplementations())
+        ImplementationPtr impl = getDocument()->getImplementationForNodeGraph(getQualifiedName(getName()));
```
kwokcb, line 748: Would it make sense that instead  of caching the implementation reference and then getting the nodedef here, just store the nodedef reference as the lookup  ?

source/MaterialXCore/Document.cpp
kwokcb: Curious if there are any metrics for the additional update cost especially if the  cache updates frequently ? At most currently 1 or a few items will ever be found and it's mostly static during the lifetime of a document.

PR #2936, opened 2026-05-15
Add Element helper for content document membership

### LINKED ISSUE #2934: Add method to determine if an element belongs to content (main) Document
## Issue

There is some repeated code that is used to test if a `Element` is part of the content (main) Document or part of a dependent
data library which is set using `setDataLibrary()` on the main `Document`.

This is used in cases when you want to search for an associated node but that node needs to be in the main Document.

The passing logic (C++) is:
```c++
DocumentPtr doc;                // Content document
ElementPtr elementToTest;   // Element to test
bool inContentDoc = elementToTest->getActiveSourceUri()` == doc->getActiveSourceUri()); // Is in the content document
```

A suggested name and location are: `Element::belongsToContentDocument()`.

API wrappers for Python + JS are required.

## Existing Usage Examples

- [C++ example](https://github.com/AcademySoftwareFoundation/MaterialX/blob/b661a4bf653d4720c6eda678dd4676208ee53649/source/MaterialXGenShader/Util.cpp#L417)
- [JS example](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/javascript/MaterialXView/source/viewer.js#L1116)

popsiclelmlm:
## Summary
- add Element::belongsToContentDocument() for filtering elements imported from libraries
- expose the helper through Python and JavaScript bindings
- update existing C++ and JavaScript source-URI checks to use the new helper
- add C++, Python, and JavaScript test coverage

Fixes #2934

## Validation
- cmake --build build --target MaterialXTest PyMaterialXCore --parallel 4
- ./build/bin/MaterialXTest "[element]"
- PYTHONPATH=build/lib /opt/homebrew/bin/python3 - <<'PY'
import PyMaterialXCore as mx
content_doc = mx.createDocument()
content_doc.setSourceUri('content.mtlx')
content_elem = content_doc.addChildOfCategory('generic', 'contentElem')
assert content_elem.belongsToContentDocument()
lib_doc = mx.createDocument()
lib_doc.setSourceUri('library.mtlx')
lib_doc.addChildOfCategory('generic', 'libElem')
content_doc.importLibrary(lib_doc)
lib_elem = content_doc._getChild('libElem')
assert lib_elem
assert not lib_elem.belongsToContentDocument()
print('python binding ok')
PY
- git diff --check


jstone-lucasfilm:
Thanks for this proposal, @popsiclelmlm, and you should be able to resolve the EasyCLA warning above by following the steps in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2936#issuecomment-4462958154.

popsiclelmlm:
review

source/MaterialXGenShader/Util.cpp
kwokcb, line 398: If you don't mind adding this in, this method is missing the library check for materialNode.

source/MaterialXGenShader/Util.cpp
```diff
@@ -414,15 +414,15 @@ vector<TypedElementPtr> findRenderableElements(ConstDocumentPtr doc)
         {
             for (OutputPtr output : graph->getOutputs())
             {
-                if (output->getActiveSourceUri() == doc->getActiveSourceUri())
+                if (output->belongsToContentDocument())
```
kwokcb: This looks like it's in the wrong scope. It should be enough to test `graph->belongsToContentDocument()`. Individual outputs will never have uris, but the nodegraph may. Could you move it up a level. Thanks.

source/MaterialXTest/MaterialXCore/Element.cpp
```diff
@@ -60,6 +60,19 @@ TEST_CASE("Element", "[element]")
...
 
+    // Check content document membership.
+    mx::DocumentPtr contentDoc = mx::createDocument();
+    contentDoc->setSourceUri("content.mtlx");
+    mx::ElementPtr contentElem = contentDoc->addChildOfCategory("generic", "contentElem");
+    REQUIRE(contentElem->belongsToContentDocument());
+    mx::DocumentPtr lib = mx::createDocument();
+    lib->setSourceUri("library.mtlx");
+    lib->addChildOfCategory("generic", "libElem");
+    contentDoc->importLibrary(lib);
```
kwokcb, line 72: Could you also check the other variant of this which is using `setDataLibrary()` instead of `importLibrary()` as that's the main case we want to do checks against.

source/MaterialXCore/Element.h
```diff
@@ -749,6 +749,13 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
         return EMPTY_STRING;
     }
 
+    /// Return true if this element belongs to its content document, and not
+    /// to an imported library with a distinct source URI.
```
kwokcb, line 753: Could you add in a @returns line for doxygen documentation. Thanks.

kwokcb (commented):
Thanks for doing this addition. Left a  few minor points including that test should
really test the `setDataLibrary()` library dependency as that's the main reason
for this safeguard logic. Maybe the easiest would be to load in the standard libraries, then look for the existing items in the library. None should pass the test when using `setDataLibrary()` while all should be found for `importLibrary`.

Also, I'm not sure if this is all the places this check exists  so if you don't mind could
you do a quick double check on this. Much appreciated.

popsiclelmlm:
Updated in 5797d42:

- C++ and Python tests now load the standard libraries and check `ND_image_color3` through both `setDataLibrary()` and `importLibrary()`.
- The JS binding test now covers `setDataLibrary()` with a source URI-marked library element in its Core-only fixture.
- I double-checked current `belongsToContentDocument()` call sites; outside bindings and tests, the active uses are in `source/MaterialXGenShader/Util.cpp` and `javascript/MaterialXView/source/viewer.js`.

Local checks:

- `cmake --build build --target MaterialXTest -- -j2`
- `./build/bin/MaterialXTest-1.39.5 "[element]"`
- `cmake --build build --target PyMaterialXCore PyMaterialXFormat -- -j2`
- `PYTHONPATH=/tmp/materialx-pytest:. python3.14 -m unittest python.MaterialXTest.main.TestMaterialX.test_ElementContentDocument`
- `node --check javascript/MaterialXTest/element.spec.js`

popsiclelmlm:
Pushed a follow-up in 765856b for the JavaScript CI failure. The previous JS local fixture did not model `setDataLibrary()` the same way as a loaded library document, so the JS test now loads the standard library files and checks `ND_image_color3` through both `setDataLibrary()` and `importLibrary()`, matching the C++ and Python coverage.

Local checks for the follow-up:

- `node --check javascript/MaterialXTest/element.spec.js`
- `git diff --check`

I do not have the local Emscripten JS build artifacts in this worktree, so the full JS test is left to the PR CI rerun.

jstone-lucasfilm:
This looks really promising, @popsiclelmlm!

The remaining recommendations from @kwokcb above are good ones (e.g. fixing the scope of calls, adding a Doxygen return value), and I'll just note that a few additional cases of the same pattern exist in the MaterialX codebase, which could be updated to use your new method:

```
source/MaterialXTest/MaterialXFormat/XmlIo.cpp, lines 75 and 86
source/MaterialXRender/ImageHandler.cpp, line 182
source/MaterialXView/Viewer.cpp, line 1381
source/MaterialXGraphEditor/Graph.cpp, line 1915
```

popsiclelmlm:
Pushed a follow-up in d7e9e8d addressing the latest review suggestions:

- Added a Doxygen return value for `belongsToContentDocument()`.
- Updated the additional source URI checks in `XmlIo.cpp`, `ImageHandler.cpp`, `Viewer.cpp`, and `Graph.cpp` to use the new helper.

Local checks:

- `git diff --check`
- `cmake --build build --target MaterialXTest --parallel 4`
- `./build/bin/MaterialXTest "[element]"`
- `./build/bin/MaterialXTest "[xmlio]"`
- `cmake -S . -B build-verify-render -DMATERIALX_BUILD_RENDER=ON -DMATERIALX_BUILD_RENDER_PLATFORMS=OFF -DMATERIALX_BUILD_TESTS=OFF -DMATERIALX_BUILD_VIEWER=OFF -DMATERIALX_BUILD_GRAPH_EDITOR=OFF -DMATERIALX_BUILD_PYTHON=OFF -DMATERIALX_BUILD_JS=OFF`
- `cmake --build build-verify-render --target MaterialXRender --parallel 4`

I also tried configuring a Viewer/GraphEditor verification build, but this checkout does not have the NanoGUI submodule initialized, so CMake stopped before generating those targets.

source/MaterialXCore/Element.h
```diff
@@ -749,6 +749,14 @@ class MX_CORE_API Element : public std::enable_shared_from_this<Element>
...
+    /// @return True if the active source URI matches the root document URI.
+    bool belongsToContentDocument() const
+    {
+        return getActiveSourceUri() == getRoot()->getSourceUri();
```
jstone-lucasfilm: There's a slight difference in meaning between `getRoot` and `getDocument`, though the two methods resolve to the same object today.

In order to make this logic more future-proof, and to better align with existing code in Element.cpp, I'd recommend changing this line to the following:

```
return getActiveSourceUri() == getDocument()->getSourceUri();
```

source/MaterialXGenShader/Util.cpp
jstone-lucasfilm: I like @kwokcb's recommendation here, and I think it would be valuable to move this check up one level, so that we're not calling the new method repeatedly for each output.

jstone-lucasfilm (commented):
This looks very close to the mark, thanks @popsiclelmlm, and I had just two additional notes, one of which is a reminder about @kwokcb's earlier suggestion.

popsiclelmlm:
Pushed a follow-up in 5aecabd for the two latest review notes:

- Moved `belongsToContentDocument()` out of the header so it can compare against `getDocument()->getSourceUri()` as suggested.
- Moved the graph output source-URI check up to the node graph level in `findRenderableElements()`.

Local checks:

- `git diff --check`
- `cmake --build build --target MaterialXTest --parallel 4`
- `./build/bin/MaterialXTest "[element]"`
- `./build/bin/MaterialXTest "[genshader]"`

jstone-lucasfilm (approved):
Great work, @popsiclelmlm, and thanks for working through all of the details with us!

PR #2938, opened 2026-05-18
Support multiple glTF UV sets

hybridherbst:
This PR adds support for multiple UV sets throughout the apps in this repository. So MaterialXView, MaterialXGraphEditor and so on can now work with materials and objects that have multiple UV sets.

### DIFF
--- a/source/MaterialXRenderGlsl/GlslProgram.cpp
+++ b/source/MaterialXRenderGlsl/GlslProgram.cpp
@@ -261,7 +261,15 @@ void GlslProgram::bindAttribute(const GlslProgram::InputMap& inputs, MeshPtr mes
         MeshStreamPtr stream = mesh->getStream(input.first);
         if (!stream)
         {
-            throw ExceptionRenderError("Geometry buffer could not be retrieved for binding: " + input.first + ". Index: " + std::to_string(index));
+            // Fall back to the first texcoord stream when the mesh has fewer UV sets than the material.
+            if (input.first.find("i_texcoord_") != std::string::npos)
+            {
+                stream = mesh->getStream(MeshStream::TEXCOORD_ATTRIBUTE, 0);
+            }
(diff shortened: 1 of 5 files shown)


source/MaterialXRenderMsl/MslPipelineStateObject.mm
```diff
@@ -369,6 +369,11 @@ int GetStrideOfMetalType(MTLDataType type)
 
         unsigned int stride = 0;
         MeshStreamPtr stream = mesh->getStream(input.first);
+        // Fall back to the first texcoord stream when the mesh has fewer UV sets than the material.
```
jstone-lucasfilm, line 372: These two updates to GLSL and MSL look great, but I believe we need a corresponding update for Slang as well, in `SlangProgram.cpp`.

source/MaterialXRenderGlsl/GlslProgram.cpp
```diff
@@ -261,7 +261,15 @@ void GlslProgram::bindAttribute(const GlslProgram::InputMap& inputs, MeshPtr mes
         MeshStreamPtr stream = mesh->getStream(input.first);
         if (!stream)
         {
-            throw ExceptionRenderError("Geometry buffer could not be retrieved for binding: " + input.first + ". Index: " + std::to_string(index));
+            // Fall back to the first texcoord stream when the mesh has fewer UV sets than the material.
+            if (input.first.find("i_texcoord_") != std::string::npos)
```
jstone-lucasfilm, line 265: Perhaps we should use the existing constant `HW::IN_TEXCOORD`, rather than adding this hardcoded string?  The same approach could be used in MSL and Slang as well.

jstone-lucasfilm (commented):
Thanks for this contribution, @hybridherbst!  This looks close to the mark, and I had just two recommendations for improvements.

PR #2943, opened 2026-05-20
Support token editing within GraphEditor UI

mialana:
Screenshot 2026-05-20 130151

> Preview of new detailed Tokens Table feature.

Hello! This is a PR for Dev Days 2026. It features:

1. Token editing directly within the graph editor UI
2. Display of other token data within `Tokens` table to improve user experience. Namely:
  - A `Source Element` column, which contains the graph element where the token is declared / was discovered during upwards traversal.
  - A `Affected Inputs` column, which lists which of the node's inputs reference the token.
3. Tooltip features for both the overall tokens table and each of the individual columns (see preview GIF below).
materialx_token_tooltips_preview

> Preview of new supplementary tooltip features.

Feel free to point out any inefficiencies or inconsistencies I've made in implementation. Thank you!


source/MaterialXGraphEditor/UiNode.cpp
```diff
@@ -95,6 +95,72 @@ mx::NodeGraphPtr UiNode::getNodeGraph() const
...
+
+    mx::ElementPtr currElem = getNode();
+    while (currElem)
+    {
+        if (mx::ConstInterfaceElementPtr interfaceElem = currElem->asA<mx::InterfaceElement>())
+        {
+            handleTokenMapping(interfaceElem, currElem);
+
+            // If the node is a nodegraph, also check for tokens on corresponding nodedef
+            if (mx::ConstNodeGraphPtr nodegraph = currElem->asA<mx::NodeGraph>())
```
kwokcb: I believe this logic will need to also handle node instances as the corresponding nodedef could also have exposed tokens. I don't think there is an example such as nodedef so may need to make one. Perhaps a simple example is a nodedef implemented as a image node + a token node, where the image node's filename and the token input are exposed as interface inputs.

source/MaterialXGraphEditor/UiNode.cpp
```diff
@@ -95,6 +95,72 @@ mx::NodeGraphPtr UiNode::getNodeGraph() const
...
+{
+    // Helper inline lambda function to avoid repeating code for mapping of tokens declared on element itself vs on element's corresponding nodedef
+    auto handleTokenMapping = [&](const mx::ConstInterfaceElementPtr& interfaceElem, mx::ElementPtr sourceElem)
+    {
+        std::vector<mx::TokenPtr> tokens = interfaceElem->getActiveTokens();
+        for (auto token : tokens)
+        {
+            std::string key = token->getName();
+            
+            // Insert into map, but do not allow parent values to override child values
```
kwokcb: Thanks for handling token hierarchies. 

I believe you will get the correct ordering if `getActiveTokens()` maintains child->parent ordering (which it should). Another option would be to use full paths and check if the current token is a parent.

kwokcb (commented):
Hi @mialana,
This looks really good! There are some minor comments on variants to check.

I see you find all parent tokens for the active child node. 
You could simplify logic by just getting the tokens for the current (selected) node/nodegraph. The user interaction and display would be simplified as well as a user knows exactly what scope the token change is occurring at. I leave it up to you -- either way is fine.

I didn't run this but have you checked that a token change can indeed cause the image's referenced to update, or is this a second step after the core + edit logic here is added ?

Thanks.

source/MaterialXGraphEditor/UiNode.cpp
mialana: Sure! Do these [new additions](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2943/changes/33fe3a72dc45b50b25d07ceee70a513a581e921e#diff-c9914ed9938c8bc1a7e6093544177600745c75705684c3ede56ea2be4db5748bR129-R134) cover the cases you are thinking of?

mialana:
> Hi @mialana, This looks really good! There are some minor comments on variants to check.
> 
> I see you find all parent tokens for the active child node. You could simplify logic by just getting the tokens for the current (selected) node/nodegraph. The user interaction and display would be simplified as well as a user knows exactly what scope the token change is occurring at. I leave it up to you -- either way is fine.
> 
> I didn't run this but have you checked that a token change can indeed cause the image's referenced to update, or is this a second step after the core + edit logic here is added ?
> 
> Thanks.

Great! Yes, here is a demo where I edit a token in the table, and the material preview is updated thereafter. 

materialx_token_edit_demo

For the design decision you mention, my initial thought process was that displaying parent tokens is useful as then the table exposes all tokens that are available for use in `filename`-type inputs of the current node -- i.e. all the tokens that are currently in scope.

Does that make sense to you as well? Thank you for the code review!

source/MaterialXGraphEditor/UiNode.cpp
mialana: Right, I did go down the implementation code of `getActiveTokens()`. If I understand it correctly, it traverses inheritance, calls `getChildrenOfType<Token>()`, and appends the return tokens to a result vector. 

But of course I can also look into the other solution you suggested! Is there a helper function anywhere in the code base for detecting if path strings have parent-child relationships?

kwokcb: Thanks for checking the logic. If this is ordering than it's fine.

source/MaterialXGraphEditor/UiNode.cpp
kwokcb: Looks good. Thanks.

kwokcb (commented):
This looks good to go.

Just leaving a final test check that the editing values works for the cases in 
the `token_graph` and `token_graph_material`. A snapshot or short gif/video would be good to show the result. There is a nodedef with token instance so that will check the latest logic you added in to cover that case. Thanks!

mialana:
[MaterialX Token Editing - 'token_graph_material.mtlx' Functionality Demo](https://youtu.be/ep8SFb5hfqA)

[MaterialX Token Editing - 'token_graph.mtlx' Functionality Demo](https://youtu.be/lLDoa-ey3M4)

@kwokcb Sorry for the delay in getting these videos posted. I will say that the videos demonstrate that the editing process is not yet completely bug-free. Namely:

- In `token_graph_material.mtlx` the third test scenario with `token_nodedef_graph` will invoke an error message: "Could not find a matching implementation for node 'token_image' matching target 'genglsl'". From debugging, this occurs from calling `updateMaterials()` after editing. It seems that during this re-render the graph node traversal is different / incomplete, causing the implementation `NG_token` of `ND_token` to not be visited at all.
- In `token_graph.mtlx`, only the `Parent_Token_Graph` test case has a material preview in the viewport. Furthermore, the mini 2D preview of the filename input in the sibling test does not show up (and never did). 

Given these videos, I am curious whether you'd like to fix these bugs before this PR is approved? I believe that the problems demonstrated should have already existed before my changes, and if possible could be addressed separately. 

I can also provide some more details on anything if needed! Thank you, hope my update make sense and let me know preferred next steps.

kwokcb:
> > Hi @mialana, This looks really good! There are some minor comments on variants to check.
> > I see you find all parent tokens for the active child node. You could simplify logic by just getting the tokens for the current (selected) node/nodegraph. The user interaction and display would be simplified as well as a user knows exactly what scope the token change is occurring at. I leave it up to you -- either way is fine.
> > I didn't run this but have you checked that a token change can indeed cause the image's referenced to update, or is this a second step after the core + edit logic here is added ?
> > Thanks.
> 
> Great! Yes, here is a demo where I edit a token in the table, and the material preview is updated thereafter.
> 
> [ ](https://private-user-images.githubusercontent.com/96770662/595735113-39ff7e22-487c-47d1-8a71-8c74a5020f46.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzk5NzMxNzcsIm5iZiI6MTc3OTk3Mjg3NywicGF0aCI6Ii85Njc3MDY2Mi81OTU3MzUxMTMtMzlmZjdlMjItNDg3Yy00N2QxLThhNzEtOGM3NGE1MDIwZjQ2LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MjglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTI4VDEyNTQzN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWY4ZTc1MmIxYzRhMTdjYmJlMmE4YzVhOWMzYzg0YTgwMTE1MWQ1OTgzMzZiMTcyNjc0Mzg2NWU4NmFhY2VlOTEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.Mso3MHKPe8RPjepBrPpirScGtiPzPlZwMjtxlHtYaSM)materialx_token_edit_demo
> 
> [ ](https://private-user-images.githubusercontent.com/96770662/595735113-39ff7e22-487c-47d1-8a71-8c74a5020f46.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzk5NzMxNzcsIm5iZiI6MTc3OTk3Mjg3NywicGF0aCI6Ii85Njc3MDY2Mi81OTU3MzUxMTMtMzlmZjdlMjItNDg3Yy00N2QxLThhNzEtOGM3NGE1MDIwZjQ2LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MjglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTI4VDEyNTQzN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWY4ZTc1MmIxYzRhMTdjYmJlMmE4YzVhOWMzYzg0YTgwMTE1MWQ1OTgzMzZiMTcyNjc0Mzg2NWU4NmFhY2VlOTEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.Mso3MHKPe8RPjepBrPpirScGtiPzPlZwMjtxlHtYaSM)
> For the design decision you mention, my initial thought process was that displaying parent tokens is useful as then the table exposes all tokens that are available for use in `filename`-type inputs of the current node -- i.e. all the tokens that are currently in scope.
> 
> Does that make sense to you as well? Thank you for the code review!

- Thanks for the update preview. Looks great.
- For exposing all the tokens. I think the only concern I have is that parent tokens can affect more than the  currently selected node. e.g. a parent token could affect 2 or more child images. Seems good to leave it as is and gather
user feedback on this.

kwokcb:
> [MaterialX Token Editing - 'token_graph_material.mtlx' Functionality Demo](https://youtu.be/ep8SFb5hfqA)
> 
> [MaterialX Token Editing - 'token_graph.mtlx' Functionality Demo](https://youtu.be/lLDoa-ey3M4)
> 
> @kwokcb Sorry for the delay in getting these videos posted. I will say that the videos demonstrate that the editing process is not yet completely bug-free. Namely:
> 
> * In `token_graph_material.mtlx` the third test scenario with `token_nodedef_graph` will invoke an error message: "Could not find a matching implementation for node 'token_image' matching target 'genglsl'". From debugging, this occurs from calling `updateMaterials()` after editing. It seems that during this re-render the graph node traversal is different / incomplete, causing the implementation `NG_token` of `ND_token` to not be visited at all.
> * In `token_graph.mtlx`, only the `Parent_Token_Graph` test case has a material preview in the viewport. Furthermore, the mini 2D preview of the filename input in the sibling test does not show up (and never did).
> 
> Given these videos, I am curious whether you'd like to fix these bugs before this PR is approved? I believe that the problems demonstrated should have already existed before my changes, and if possible could be addressed separately.
> 
> I can also provide some more details on anything if needed! Thank you, hope my update make sense and let me know preferred next steps.

- Hmm. The first seems like a bug in the editor update logic not picking up the local nodedef properly but is picked up on initial file load. . Can you leave a new issue for this. Thanks.
- `token_graph.mtlx` is unrelated and is relate to discovery of what is considered "renderable". Should log another issue for this, but you don't need to do this :).

kwokcb (approved):
Thanks for your work on this Amy. This looks good to go!

If you can log the update issue you found at your convenience that would be much
appreciated. Thx.

source/MaterialXGraphEditor/UiNode.h
```diff
@@ -160,6 +165,76 @@ class UiPin
...
+
+  private:
+    const mx::TokenPtr _tokenPtr;
+    const mx::ElementPtr _sourceElement;
+
+    std::vector<mx::InputPtr> _affectedInputs{};
+    std::ostringstream _affectedInputsStream{};
+
+    // Track whether changes were made to inputs in order to re-build stream accordingly
+    std::atomic<bool> _isAffectedInputsDirty{ true };
```
jstone-lucasfilm: Perhaps the introduction of `std::atomic` to this code is not needed?  Assuming this code will only be called from the single-threaded UI of the Graph Editor, I wouldn't imagine we need to bring new thread-safe constructs into this logic.

jstone-lucasfilm (commented):
Thanks @mialana and @kwokcb!  I'm looking forward to reading through this code in more depth, and on my first read I had one thought on the introduction of `std::atomic` to the Graph Editor.

source/MaterialXGraphEditor/UiNode.h
mialana: Hi @jstone-lucasfilm.
Sure that makes sense! I have removed the usage of `std::atomic` in my latest commit.

jstone-lucasfilm:
@mialana Thanks for the `atomic` fix!  When you have a chance, would you mind resolving the merge conflicts between your PR and recent Graph Editor work from other Dev Days contributors?

mialana:
> @mialana Thanks for the `atomic` fix! When you have a chance, would you mind resolving the merge conflicts between your PR and recent Graph Editor work from other Dev Days contributors?

Hi @jstone-lucasfilm sorry for such the slow turnaround times! I've just resolved the conflict. Please let me know if anything else is needed!

source/MaterialXGraphEditor/UiNode.h
```diff
@@ -160,6 +164,76 @@ class UiPin
...
+    std::ostringstream _affectedInputsStream{};
+
+    // Track whether changes were made to inputs in order to re-build stream accordingly
+    bool _isAffectedInputsDirty{ true };
+
+    void buildAffectedInputsStream()
+    {
+        if (!_isAffectedInputsDirty)
+            return;
+        _affectedInputsStream.clear();
```
jstone-lucasfilm: Since `clear` just resets the error/state flags for an `ostringstream`, maybe you meant to call `str("")` here?  That would clear the _contents_ of the stream, which seems to match your intent.

source/MaterialXGraphEditor/UiNode.cpp
```diff
@@ -95,6 +95,65 @@ mx::NodeGraphPtr UiNode::getNodeGraph() const
...
+        std::string inputValue = input->getValueString();
+        if (inputValue.empty() && input->hasInterfaceName())
+            inputValue = input->getInterfaceInput()->getValueString(); // Get value from referenced interface
+
+        for (const auto& entry : inputTokens)
+        {
+            if (inputValue.find(entry.first) != std::string::npos)
+            {
+                // Append to affected inputs of corresponding entry in token map
+                _uiTokenMap[inputTokensRenormalized[entry.first]]->addAffectedInput(input);
```
jstone-lucasfilm: To make this more robust, I'd recommend checking that the value in the map isn't null before dereferencing it, as in the following:

```
auto it = _uiTokenMap.find(inputTokensRenormalized[entry.first]);
if (it != _uiTokenMap.end())
    it->second->addAffectedInput(input);
```

jstone-lucasfilm (commented):
This looks nearly ready to merge, thanks @mialana, and I had just two additional suggestions for improvements.

source/MaterialXGraphEditor/UiNode.h
mialana: Ah yes, I was not aware I was using `clear()` incorrectly. Thank you!

source/MaterialXGraphEditor/UiNode.cpp
mialana: Sure that makes senses. Fixed in latest commit

mialana:
Hi @jstone-lucasfilm your last two suggestions have both been applied. Let me know how it looks now. Thank you so much for taking the time to give me pointers on my code quality!

source/MaterialXGraphEditor/UiNode.cpp
```diff
@@ -95,6 +95,68 @@ mx::NodeGraphPtr UiNode::getNodeGraph() const
...
+            {
+                if (mx::NodeDefPtr nodedef = node->getNodeDef())
+                    UiToken::applyTokenMapping(&_uiTokenMap, nodedef, nodedef);
+            }
+        }
+        currElem = currElem->getParent();
+    }
+
+    // Traverse through inputs and determine which tokens their value depends on
+    for (const auto& input : getNode()->getActiveInputs())
```
jstone-lucasfilm: Apologies, as I should have caught this in my earlier review, but I believe we need a null check for this call to `getNode`, as it's very possible that the current element is not a node.

Since you're storing the return value of `getNode` in the `currElem` variable above, you might consider storing this value as `node` and then returning early if its value is null, as there is effectively no further work to do in this case.

For example, this might look like the following:

```
void UiNode::buildUiTokenMap()
{
    _uiTokenMap.clear();

    mx::NodePtr node = getNode();
    if (!node)
        return;

    mx::ElementPtr currElem = node;
    while (currElem)
    {
        ...
    }

    for (const auto& input : node->getActiveInputs())
    {
        ...
    }
}
```

jstone-lucasfilm (commented):
Just one more suggested fix, and then I think this is good to go!

mialana:
@jstone-lucasfilm I addressed the latest suggestion just now -- added the null check and renamed some variables within the while-loop for consistency. Let me know if this matches what you were thinking!

jstone-lucasfilm (approved):
Great work, thanks @mialana!

PR #2944, opened 2026-05-21
Add UsdUVTexture v2.3 variant with vector4 scale and bias

### LINKED ISSUE #2895: UsdUVTexture `bias` and `scale` port types differ
In MaterialX `ND_UsdUVTexture_23` specifies the `bias` and `scale` inputs as type `color4`.
```
<input name="scale" type="color4" value="1, 1, 1, 1" uniform="true" />
<input name="bias" type="color4" value="0, 0, 0, 0" uniform="true" />
```

The UsdPreviewSurface specification defines them to be `float4` (which corresponds to the MaterialX type `vector4` most closely).
```
    float4 inputs:scale = (1.0, 1.0, 1.0, 1.0) (
        connectability = "interfaceOnly"
        doc = """Scale to be applied to all components of the texture.
                 value * scale + bias"""
    )

    float4 inputs:bias = (0.0, 0.0, 0.0, 0.0) (
        connectability = "interfaceOnly"
        doc = """Bias to be applied to all components of the texture.
                 value * scale + bias"""
    )
```

We need these types to be aligned, but also importantly the types here infer different handling when color management is considered. MaterialX considers these to be color-managed values, but USD does not. I believe USD is correct here, when using `bias` and `scale` we are performing mathematical operations on the color value that has been retrieved from the texture file.

ld-kerley:
Resolves #2895 

Introduces `ND_UsdUVTexture_23_vector4` nodedef and I`MP_UsdUVTexture_23_vector4` nodegraph implementation, where the `scale` and `bias` inputs are typed as vector4 instead of the original color4-derived behavior. The prior variant is retained for now for compatibility but could be removed in a future MaterialX release via an upgrade function.


kwokcb (approved):
Looks good to me. Maybe add an instance of this variant to the existing usd or image sample test files.so can be used for mtlx and usd testing.

ld-kerley:
added unit test - now we have a test that tests UsdUVTexture and also the new typed  scale and bias.

resources/Materials/Examples/UsdPreviewSurface/usd_preview_surface_brass.mtlx
kwokcb, line 1: Just a couple of small items:

- Both of these and the  existing `usd_preview_surface_brass_tiled.mtlx` in the same folder use the same `surfacematerial` name `USD_Tiled_Brass`. Could you give unique names to these new ones so they can be generate uniquely named output folders + image file names. I think they may overwrite each other.
- For efficiency, could you combine them into 1 file as well, then any further variants will only require a single file update.
Thanks.

ld-kerley: Done.

Although I wonder if we can make the test image generation more robust than needing unique names across multiple files?
* use the filename?
* have the test system merge all the unit test files together an validate? 

Merging these three files in to one also feels a little more fragile - its much easier to just copy the entire file and modify the material - move these to all be in the same file required more care with node naming etc - easier to get wrong?

PR #2946, opened 2026-05-22
Fix index count for unindexed primitives in `CgltfLoader`

jstone-lucasfilm:
This changelist fixes an indexing bug in `CgltfLoader` for glTF primitives that omit an index buffer.

When no index buffer is present, the loader synthesizes a sequential index buffer covering the position stream.  The previous code assigned the raw size of the position float buffer to `indexCount`, which is three times the vertex count for 3D positions, producing an inflated face count and out-of-range indices in the MaterialX Viewer.

This fix divides by `MeshStream::STRIDE_3D` to obtain the actual vertex count, matching the convention already used a few lines below when computing `setVertexCount`.

### DIFF
--- a/source/MaterialXRender/CgltfLoader.cpp
+++ b/source/MaterialXRender/CgltfLoader.cpp
@@ -426,7 +426,7 @@ bool CgltfLoader::load(const FilePath& filePath, MeshList& meshList, bool texcoo
                 }
                 else
                 {
-                    indexCount = positionStream->getData().size();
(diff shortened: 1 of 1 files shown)


kwokcb (approved):
Looks good to me.

jstone-lucasfilm:
Thanks @kwokcb, and let's go ahead and include this in v1.39.5, since it fixes a crash that can occur in production workflows (albeit a very rare crash).

Issue #2526, opened 2025-08-17
Example graph for interior mapping


jstone-lucasfilm:
We'd benefit from having a simple MaterialX example of the [interior mapping](https://godotshaders.com/shader/interior-mapping-shader/) technique in our repository, demonstrating how this shading effect can be built from the existing `viewdirection` node.

Initially, this graph might be contributed to the [nprlib](https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/resources/Materials/TestSuite/nprlib) folder of our test suite, and further in the future it might be upgraded to a standard node with a graph definition.

As inspiration, here are two existing examples of MaterialX graphs for interior mapping:

- https://github.com/pablode/materialx-interior-mapping (@pablode)
- https://github.com/jakethorn/ShadingLanguageX/blob/main/examples/interiormapping.mxsl (@jakethorn)

juandejoya:
@jstone-lucasfilm would you be okay if I took a look at this for next week's DevDays?

jstone-lucasfilm:
Welcome, @juandejoya, and I'll go ahead and assign this to you!

jakethorn:
What should the NodeDef signature to look like? 
Possible inputs could be:
1. Image filepaths for the walls/ceiling/floor textures. In my implementation, I used 5 separate textures, but @pablode used a single environment map. There's also an alternative approach where the user doesn't input the textures, but the node outputs the UVs which the user can then pass onto image nodes later in the network. This is more flexible if the user wants to use more than just basic albedo textures, or they need to adjust the UVs. Perhaps there could be two implementations, one which takes the filepaths and another which returns the UVs.
2. Texture coordinates (vec2).
3. Space: world/object/model (string).
4. Dimensions of the room (vec3). At the moment, my (and I think @pablode, sorry if I'm wrong) implementations assume a room with equal width and height. It would be good to extend this to rooms of different dimensions.

Outputs would then be:
1. Output color for the node that takes in filepaths or UVs for the one that doesn't. For the second approach, an int output might be needed, to indicate which wall the ray hit.

juandejoya:
Thanks for the frame of reference @jakethorn, and apologies for the slow response. I was waiting until the Dev Days started (and finishing up my work priorities end of day) before I took a closer look.

I've starting testing some of the recommendations for the node signature and will chime in on here or Slack to follow up.

juandejoya:
[...]

PR #2951, opened 2026-05-25
added interior mapping example with equirectangular samples

codesavory:
Adds a basic example for the issue: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2526

## Reference Renders

`room.png` (brick room):
screenshot_room

`cornell.png` (Cornell box):
screenshot_cornell

You can read more about the design, architecture and reasoning here:
[interior-mapping-design.md](https://github.com/user-attachments/files/28204509/interior-mapping-design.md)
<details>
<summary>Design Notes</summary>

# Interior Mapping for MaterialX — Design Notes

**Issue**: [#2526 — Example graph for interior mapping](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2526)
**Branch**: `issue-2526-interior-mapping`
**Files**: `resources/Materials/TestSuite/nprlib/interior_mapping.mtlx`, `room.png`, `cornell.png`

---

## Background

Interior mapping is a shading technique that fakes 3D rooms behind flat window surfaces — no room geometry required. Each UV tile becomes a separate room instance. The core idea: cast a ray from the camera through the surface into a virtual box, find which interior wall the ray hits, sample a texture at that point.

The ask (from jstone-lucasfilm) is simple: add a working `.mtlx` example to the `nprlib` test suite. Tagged `good first issue`. No new node definition required for a first contribution — a bare `nodegraph` like `starfield.mtlx` is sufficient.

---

## Existing nprlib Pattern

Four examples establish the range:

| File | Lines | Pattern |
|---|---|---|
| `edge_brighten.mtlx` | 18 | bare `nodegraph`, no nodedef |
| `starfield.mtlx` | 25 | bare `nodegraph`, uses `viewdirection` — closest precedent |
| `gooch_shade.mtlx` | 30 | instantiates externally-defined node |
| `toon_shade.mtlx` | 151 | full `nodedef` + `nodegraph` + material instance, namespace |

Interior mapping targets the `starfield.mtlx` pattern: bare nodegraph, `viewdirection`-based, no `nodedef`.

---

### 1. Godot GLSL (~90 lines)
- Works in **object space** — vertex shader passes vertex and camera positions directly, ray direction = `obj_vertex − obj_cam`, no matrix math
- Ray origin: actual camera position in object space, offset by delta; `ceil()` for tiled room boundaries
- Intersection: per-axis with `ceil()` + `step()`
- Face detection: **if/else branching** — determines which face was hit, picks UV swizzle (`.zy`, `.xy`, `.xz`), maps into 3×2 face atlas
- Depth: scales ceiling/floor distance by `room_depth × 2`

### 2. Pablode (MaterialX XML, 155 lines)
- Works in **tangent space** — builds TBN matrix from surface T/B/N, transposes it, transforms `viewdirection` through it; portable across any mesh orientation, costs ~6 extra nodes
- Ray origin: from fractional UV `(frac(u)×2−1, frac(v)×2−1, +1)` — Z=+1 places origin on front face of `[−1,1]³` cube; `frac()` gives per-tile tiling free
- Intersection: vectorized AABB slab method — `dist = |1/dir| − origin·(1/dir)`, then two `min` nodes; fully branchless
- Face detection: **none needed** — normalizes hit position → spherical coordinates → equirectangular UV via `atan2(x, z)` + `asin(y)`. Continuous, no per-face branching
- Texture: single equirectangular panorama
- Depth: multiply view direction Z by `1/depth`

### 3. Jakethorn MXSL (~50 lines)
- Works in **object space** via `viewdirection("object")`, negated explicitly
- Ray origin: `vec3(uv×2−1, −1)` — same math as pablode, Z=−1 convention (opposite sign)
- Intersection: explicit slab — `dn = (−origin−1)/dir`, `dp = (−origin+1)/dir`, `max(dn, dp)` per component
- Face detection: **if/else branching** — checks which axis of hit position is closest to ±1, samples one of 5 separate `image()` calls
- Texture: 5 separate images, one per visible face

### Comparison Table

|  | Godot | Pablode | Jakethorn |
|---|---|---|---|
| Coordinate space | Object | Tangent | Object |
| Ray origin | Camera pos | Fractional UV | Fractional UV |
| Branching | Yes (6-way) | **No** | Yes (5-way) |
| Texture format | 3×2 face atlas | **Equirectangular** | 5 separate images |
| Seam risk | Face edges | Poles | Face edges |
| Textures needed | 1 | **1** | 5 |

---

## Why It Matters for MaterialX

**MaterialX nodegraphs are pure dataflow — no `if/else`.**

The Godot and jakethorn face-detection approaches require 6-way or 5-way branching. Expressing that branchlessly in MaterialX needs ~60 extra `step()`-based masking nodes to select and blend face UVs — defeats the "simple example" goal entirely.

Pablode's equirectangular approach was designed with this constraint in mind. The pipeline `normalize → atan2/asin → image` is fully branchless and maps cleanly to MaterialX dataflow nodes.

---

### Decision 1: Equirectangular projection (pablode approach)

Chosen because it's the only approach expressible without branching in MaterialX's dataflow model. Single texture, continuous UV mapping, ~30 nodes.

### Decision 2: Tangent space (not object space)

Pablode's TBN approach is more portable. Object-space approaches only work correctly when the mesh is axis-aligned with the world — tangent space works on any oriented mesh surface.

### Decision 3: No explicit negation needed — viewdirection is already inward

MaterialX `viewdirection` is documented as "from shading point toward camera," but `HwViewDirectionNode.cpp` (line 69) emits `normalize(positionWorld − viewPosition)` — surface point minus camera position — which is the **camera → surface** (inward) direction.

In tangent space, `viewDir_tangent.z < 0` for a front-facing surface. The slab formula `|1/dir| − origin·(1/dir)` works correctly with this inward direction:

- `inv_dir.z = 1/dz` where `dz < 0`, so `inv_dir.z < 0`
- `abs_inv_dir.z = |1/dz| > 0`
- `dist.z = |1/dz| − 1·(1/dz) = |1/dz| − (negative) = 2/|dz| > 0`

For a straight-on view (`dz = −1`), `dist.z = 2`, hitting the back wall at `t = 2`. For angled views `min_dist < 2` and the ray hits a side wall. No negation node is needed.

### Decision 4: atan2 seam fix — negate both dx and dz

Vanilla `atan2(dx, dz)` has its ±π discontinuity at dz < 0 (the back wall of the box). That's the most-viewed face — a visible seam artifact.

Fix: `atan2(−dx, −dz)` shifts the discontinuity to dz > 0 (the front face, +Z), which is never visible in interior mapping (the ray origin is on the front face, so the ray never hits it). Back wall, side walls, floor, and ceiling are all seam-free.

### Decision 5: Rotation offset 0.25

After negating dx and dz, atan2=0 → lon = 0.5 (back wall center). Adding 0.25 keeps the same texture mapping as the unnegated version: back wall center → U = 0.75.

### Decision 6: Precise spherical projection constants

| Constant | Value | Source |
|---|---|---|
| Longitude scale | `−0.15915` | `−1/(2π)`, converts radians → [0,1] |
| Latitude scale | `0.31831` | `1/π` |

### Decision 7: Filter mode

`filtertype="closest"` is deliberate — bilinear filtering bleeds across the `0/1` UV wrap boundary, producing a visible horizontal seam artifact at the equirectangular face boundaries.

---

## Node Graph Summary

```
viewdirection ─┐
tangent ────────┤
bitangent ──────┤─► creatematrix(TBN) ─► transpose ─► transformmatrix ─► viewDir_tangent
normal ─────────┘

texcoord ─► floor ─► subtract(frac) ─► ×2 ─► convert(vec3) ─► +(−1,−1,+1) ─► ray_origin

viewDir_tangent ─► 1/dir ─► |·| ─► abs_inv_dir ─┐
ray_origin ──────────────► ×(1/dir) ──────────────┤─► subtract(dist) ─► min3 ─► min_dist

ray_origin + viewDir_tangent × min_dist ─► hit_pos

hit_pos ─► normalize ─► separate3
  ├─ outx → ×(−1) = neg_dx ─┐
  ├─ outy → asin ─► ×(1/π) ─► +0.5 ─► latitude ──────────────────► combine2 ─► map_uv
  └─ outz → ×(−1) = neg_dz ─┴─► atan2(neg_dx, neg_dz) ─► ×(−1/2π) ─► +0.5 ─► +0.25 ─► longitude_rotated ┘

map_uv ─► image(room.png, periodic, closest) ─► out
```

Total: ~35 nodes. No branching. One texture.

---

## What Was Ruled Out

- **Per-face UV + branching (Godot/jakethorn style)**: requires `step()`-based masking, ~60 extra nodes, defeats simplicity goal
- **Object-space approach**: mesh-orientation-dependent, less portable
- **5-texture variant (jakethorn)**: 5 image samples, more complex texture authoring requirements
- **Nodedef + surfacematerial wrapper (toon_shade style)**: deferred — maintainer asked for simple example first; formal node definition is a follow-up PR

</details>

Open questions
1. Should we be worried about copyright of the images?
2. What is the right place for the images?
3. Should we add support for multi-textures variant?

I will create a separate PR for a node-def version for interior-mapping
@jstone-lucasfilm, @jakethorn, @pablode Please review and share feedback if you have a chance!

---

## Changes since PR open

<details>
<summary>Commit <a href="https://github.com/codesavory/MaterialX/commit/40e07cef"><code>40e07cef</code></a></summary>

- **`closest` filter comment** — added inline note that `filtertype="closest"` is deliberate: bilinear filtering bleeds across the `0/1` UV wrap boundary, producing a visible horizontal seam
- **Decision 3 rewrite** — replaced hand-wavy "abs_inv_dir absorbs the sign" with a sourced proof: `HwViewDirectionNode.cpp:69` emits `normalize(positionWorld − viewPosition)` (camera→surface, inward), so `viewDir_tangent.z < 0` for front-facing surfaces and the slab formula gives `dist.z = 2/|dz| > 0` without any negation node needed
- **Reference renders** — see screenshots below (drag-dropped via GitHub web UI)

> **Texture provenance** — `room.png` and `cornell.png` are equirectangular panoramas placed alongside the `.mtlx`. Please advise if source documentation is required for ASWF compliance.

</details>


codesavory:
What is the review process, does this require more changes or should I add someone manually to the reviewers list?

jstone-lucasfilm:
@codesavory No additional changes needed before we can review, and we just need to find the time and focus!

pablode:
Hi @codesavory, thanks for this contribution and some notes from my side:
- I created the cube maps and put the whole repo under public domain, so they are fine to use. The textures seen in room.png are originally from Polyhaven and also licensed under CC0.
- However I'm not sure if it makes sense to include them in this repository, given their large file sizes (20 MB and 10 MB) respectively. Perhaps they should be downsized, or a cubemap with solid colors could be good enough to showcase the effect.
- The original implementation lead to the contribution of two new MaterialX nodes, \<latlongimage\> and \<fract\>. Perhaps it would make sense to use them to optimize the graph?

jstone-lucasfilm:
Thanks for this contribution, @codesavory!  The branchless equirectangular approach is a good fit for MaterialX, and the slab intersection looks correct to me.  Three issues that I'd consider important to address before we merge:

1. Both textures are 8192x4096 PNGs, at 24 MB and 11 MB, while our largest existing asset is about 6 MB.  Since `cornell.png` isn't referenced by the example, I'd suggest dropping it, and downsizing `room.png` to 2K or smaller in `resources/Images`, following the convention of our other examples.  Poly Haven assets are already credited in `THIRD-PARTY.md`, so the CC0 source is fine to include.
2. The `viewdirection` node defaults to world space, while `tangent`, `bitangent`, and `normal` default to object space, so the view vector will be transformed through a mismatched tangent frame.  This renders correctly on our identity-transformed test sphere but breaks on any transformed mesh, as #1710 addressed in the Gooch and toon examples, and setting `space="world"` on the three geometric nodes should resolve it.
3. As @pablode notes, the `latlongimage` (#2168) and `fract` (#2169) nodes were contributed as a result of the original implementation, and I'd suggest using them here.  I'd expect `latlongimage` with a `viewdir` of `hit_dir * (-1, 1, -1)` and a `rotation` of 90 degrees to produce the same coordinates as the current graph, which would bring the node count from 35 down to about 20.

Two smaller notes: `filtertype="closest"` only switches the GL sampler to nearest, and the seam you observed more likely comes from `vaddressmode="periodic"` at the poles, which `latlongimage` avoids with `mirror`.  And since the `file` input carries no color space, the PNG is currently interpreted as linear, so I'd suggest declaring `srgb_texture` if that matches the source.

Otherwise, this looks really promising, and with the changes above it ought to be ready to include in our `nprlib` test suite.

PR #2952, opened 2026-05-26
Migrate JavaScript unit tests to Playwright

jstone-lucasfilm:
This changelist migrates the JavaScript unit tests from Mocha and Chai to Playwright, harmonizing on a single JavaScript test framework for both browser and unit tests.  The following specific changes are included:

- Convert the unit tests from Mocha and Chai to Playwright, removing the `mocha` and `chai` dependencies.
- Run the Node and browser tests as separate Playwright projects from a single configuration.
- Stage the WASM build artifacts through a Playwright `globalSetup` rather than npm `pretest` hooks.
- Restore assertions that had silently become no-ops, so each `expect` again verifies its condition.
- Fix an inverted `readXIncludes` getter on `XmlReadOptions`, uncovered by the restored assertions.

### DIFF
--- a/javascript/MaterialXTest/codeExamples.spec.js
+++ b/javascript/MaterialXTest/codeExamples.spec.js
@@ -122,13 +122,13 @@ test.describe('Code Examples', () =>
         // For each material node, locate its surface shader and traverse the
         // dataflow graph upstream, gathering the nodes that contribute to the
         // shading result.  The marble surface is driven by a procedural noise
-        // network, so the traversal crosses into the bound nodegraph and
-        // reaches a fractal3d noise node.
+        // network bound through a nodegraph, so the traversal crosses from the
+        // top-level document into that nodegraph.
         const materialNodes = doc.getMaterialNodes();
         expect(materialNodes.length).toBe(1);
 
         let nodeCount = 0;
-        let foundNoise = false;
+        let crossedIntoNodeGraph = false;
         for (const materialNode of materialNodes)
         {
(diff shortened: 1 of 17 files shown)


javascript/MaterialXTest/codeExamples.spec.js
```diff
@@ -119,27 +119,43 @@ describe('Code Examples', () =>
...
+            const shaderNodes = mx.getShaderNodes(materialNode);
+            for (const shaderNode of shaderNodes)
+            {
+                for (const edge of shaderNode.traverseGraph())
+                {
+                    const upstreamElem = edge.getUpstreamElement();
+                    if (upstreamElem instanceof mx.Node)
+                    {
+                        nodeCount++;
+                        if (upstreamElem.getCategory() === 'fractal3d')
```
kwokcb: Maybe for robustness, as this seems to be a traversal check, can we not just check if the parent is equal to the document (or if the parent is a nodegraph). If not then it found an upstream node inside a nodegraph ?

kwokcb (commented):
This looks like a good change. I only have a comment about why check for a fractal specifically in one test.

javascript/MaterialXTest/codeExamples.spec.js
jstone-lucasfilm: That's a good idea, @kwokcb, and I've made that change, so that this test isn't so coupled to the specific details of the procedural marble example.

kwokcb (approved):
Thanks for the change. Looks good to go.

PR #2954, opened 2026-05-28
Unrecognized colorspace should throw rather than warn to stderr

doug-walker:
Currently, if an unrecognized colorspace string is encountered in MaterialXGenShader, it prints a warning to std::cerr. This is not ideal for several reasons:
- The render proceeds and will have incorrect colors
- Printing to stderr is difficult for applications to deal with
- Raising an exception would be more in line with how MaterialX usually handles this type of situation

This PR  changes the behavior to raise ExceptionShaderGenError.

This change required omitting the test file ocio_color_management.mtlx from the MaterialXGenMdl tests. The test was previously succeeding (when compiled with MATERIALX_BUILD_OCIO=ON), although it should not have.

In addition, a change was made to improve the name of the color space conversion functions generated by the OcioColorManagementSystem. Previously the names only used the cacheIDs of the OCIO Processors, which was not very friendly to troubleshooting. Now the names are more similar to the "srcColorSpace_to_dstColorSpace" naming used by the DefaultColorManagmentSystem. This avoids a problem when MaterialX is compiled with the current OCIO main branch due to a change in the cacheID formatting.

While testing, I noticed that the MaterialX createValidName function sometimes creates names with double underscores. OCIO's shader generator avoids this since (although some drivers accept it) it is technically undefined behavior. I added a sanitizeName function locally to work around this.

As I worked further, I noticed that the GenShader tests were not actually testing color management, even though MaterialXTest/MaterialXGenShader/GenShaderUtil.cpp explicitly tries to enable color management. The problem was that it was not ensuring that a targetColorSpace was set (as the rendering tests do). If that is empty, the color management shaders are never generated or tested. I ensured that is now being set. As a result, I needed to make sure the ocio_color_management.mtlx test file was omitted from the GenShader tests when OCIO is not enabled.

Finally, I fixed a bug with the DefaultColorManagementSystem where the conversion from "lin_rec709_scene" to "lin_rec709" was not handled properly as a no-op. I added a test case for this in color_management.mtlx.

Claude Code was used during the creation of this PR.

### DIFF
--- a/source/MaterialXTest/MaterialXGenShader/GenShaderUtil.h
+++ b/source/MaterialXTest/MaterialXGenShader/GenShaderUtil.h
@@ -199,7 +199,13 @@ class ShaderGeneratorTester
     virtual void setTestStages() = 0;
 
     // Add files in to not examine
-    virtual void addSkipFiles() { };
+    virtual void addSkipFiles()
+    {
+#ifndef MATERIALX_BUILD_OCIO
+        // ocio_color_management.mtlx uses color spaces which require OCIO.
+        _skipFiles.insert("ocio_color_management.mtlx");
+#endif
+    };
 
     // Add nodedefs to not examine
     virtual void addSkipNodeDefs() { };
--- a/resources/Materials/TestSuite/stdlib/color_management/color_management.mtlx
+++ b/resources/Materials/TestSuite/stdlib/color_management/color_management.mtlx
@@ -188,5 +188,15 @@
       <input name="base_color" type="color3" nodename="color_lin_displayp3_color3" />
     </standard_surface>
     <output name="color_lin_displayp3_out" type="surfaceshader" nodename="color_lin_displayp3_standard_surface5" />
+    <constant name="color_lin_rec709_scene" type="color4">
+      <input name="value" type="color4" value="0.5, 0.0, 0.0, 1.0" colorspace="lin_rec709_scene" />
+    </constant>
+    <convert name="color_lin_rec709_scene_color3" type="color3">
+      <input name="in" type="color4" nodename="color_lin_rec709_scene" />
+    </convert>
+    <standard_surface name="color_lin_rec709_scene_standard_surface" type="surfaceshader">
+      <input name="base_color" type="color3" nodename="color_lin_rec709_scene_color3" />
+    </standard_surface>
+    <output name="color_lin_rec709_scene_out" type="surfaceshader" nodename="color_lin_rec709_scene_standard_surface" />
   </nodegraph>
 </materialx>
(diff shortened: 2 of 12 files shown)


jstone-lucasfilm:
Thanks for putting this together, @doug-walker!

Moving from a `std::cerr` warning to a proper `ExceptionShaderGenError` is clearly the right direction, and I especially appreciate your catch that our GenShader tests weren't actually exercising color management, in contrast to our color-managed render test suite.  The OCIO shader naming and the no-op fix in `DefaultColorManagementSystem` are both welcome as well.

One thing I'd like to work through before we merge: now that an unsupported transform throws, it reaches a spot in the Graph Editor that wasn't quite ready for it.  `RenderView::reloadShaders` currently catches only `mx::ExceptionRenderError`, so a new `ExceptionShaderGenError` would propagate unhandled and take down the editor on shader reload.  This looks like a pre-existing gap rather than anything introduced here (the neighboring `loadMaterialLibrary` and the Viewer's `reloadShaders` both already catch `std::exception`), but since this PR is what makes the path reachable, it seems like a natural place to address it.

For example, you could mirror the existing pattern as follows:

```cpp
catch (mx::ExceptionRenderError& e)
{
    for (const std::string& error : e.errorLog())
    {
        std::cerr << error << std::endl;
    }
}
catch (std::exception& e)   // ExceptionShaderGenError, etc.
{
    std::cerr << e.what() << std::endl;
}
```

Let me know how this sounds to you, and I'm happy to help with the Graph Editor update as needed.

doug-walker:
Thank you @jstone-lucasfilm, I'll add a commit with your proposed changes!

source/MaterialXTest/MaterialXGenShader/GenShaderUtil.h
```diff
@@ -199,7 +199,13 @@ class ShaderGeneratorTester
     virtual void setTestStages() = 0;
 
     // Add files in to not examine
-    virtual void addSkipFiles() { };
+    virtual void addSkipFiles()
+    {
+#ifndef MATERIALX_BUILD_OCIO
+        // ocio_color_management.mtlx uses color spaces which require OCIO.
+        _skipFiles.insert("ocio_color_management.mtlx");
```
jstone-lucasfilm, line 206: Looking through the code in more detail, I noticed one potential issue that we may want to address before merging.  In the commented line, we're adding a common file to the base implementation of `ShaderGeneratorTester::addSkipFiles`, which will require that all subclasses of `ShaderGeneratorTester` call down to this base implementation for consistency.

But I'm noticing that the `OslShaderGeneratorTester` implementation of `addSkipFiles` fails to call down to the base class, opening up the possibility of a new failure when we run OSL tests in our nightly builds:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/cefe4b162fca2d7e76016ae352d70b7c1ea96eeb/source/MaterialXTest/MaterialXRenderOsl/RenderOsl.cpp#L100

For safety, perhaps we should add a call to `ShaderGeneratorTester::addSkipFiles` at the end of `OslShaderGeneratorTester::addSkipFiles` as well, so that it includes your new common file as well?

It's worth noting that this is a pre-existing issue, not introduced by your change, but the current PR has the potential to trigger a new error because of the pre-existing issue.

doug-walker: Good catch, Jonathan!  I've made the change.

jstone-lucasfilm (approved):
This looks good to me, thanks @doug-walker!

PR #2955, opened 2026-05-28
UsdPreviewSurface `opacityMode` string input

### LINKED ISSUE #2896: USDPreviewSurface `opacityMode` port type differs
In MaterialX `ND_UsdPreviewSurface_surfaceshader` specifies the `opacityMode` input as type `integer`
```
<input name="opacityMode" type="integer" enum="transparent,presence" enumvalues="0,1" value="0" uiname="Opacity Mode" />
```

The UsdPreviewSurface specification defines it to be a `token`. I think this most closely matches the MaterialX concept of having a `string` input type with a `enum` attribute defining the supported types.   This is in fact what we're already doing for `wrapS` and `wrapT` on `ND_UsdUVTexture_23`

```
<input name="wrapS" type="string" value="periodic" enum="black,clamp,periodic,mirror" uniform="true" />
<input name="wrapT" type="string" value="periodic" enum="black,clamp,periodic,mirror" uniform="true" />
```

I think we should follow the same pattern for `opacityMode`.

Aligning the types is important for the interchange between the two shading system nodes, but also because the current MaterialX convention using the `enumvalues` attribute it not supported by the `UsdMtlx` translation code.

ld-kerley:
Resolves #2896

Add version of `ND_UsdPreviewSurface_surfaceshader` with `opacityMode` input as a string for compatibility with OpenUSD version of UsdPreviewSurface.

Added unit tests for both modes.

Updated validation logic to allow for mis-matched types if enumeration is present - previously a `NodeGraph` implementation would fail validation if a `string` input with an `enum` attribute was used as its resolved `integer` type.


libraries/bxdf/usd_preview_surface.mtlx
```diff
@@ -25,6 +25,25 @@
...
+    <input name="diffuseColor" type="color3" value="0.18, 0.18, 0.18" uimin="0,0,0" uimax="1,1,1" uiname="Diffuse Color" />
+    <input name="emissiveColor" type="color3" value="0, 0, 0" uimin="0,0,0" uisoftmax="1,1,1" uiname="Emissive Color" />
+    <input name="useSpecularWorkflow" type="integer" value="0" uimin="0" uimax="1" uistep="1" uiname="Use Specular Workflow" />
+    <input name="specularColor" type="color3" value="0, 0, 0" uimin="0,0,0" uimax="1,1,1" uiname="Specular Color" />
+    <input name="metallic" type="float" value="0" uimin="0.0" uimax="1.0" uiname="Metallic" />
+    <input name="roughness" type="float" value="0.5" uimin="0.0" uimax="1.0" uiname="Roughness" />
+    <input name="clearcoat" type="float" value="0" uimin="0.0" uimax="1.0" uiname="Clearcoat" />
+    <input name="clearcoatRoughness" type="float" value="0.01" uimin="0.0" uimax="1.0" uiname="Clearcoat Roughness" />
+    <input name="opacity" type="float" value="1" uimin="0.0" uimax="1.0" uiname="Opacity" />
+    <input name="opacityMode" type="string" enum="transparent,presence" value="transparent" uiname="Opacity Mode" />
```
kwokcb, line 38: Can this be made to have an `enumvalues` attribute. For consistency + to make mapping explicit for hardware. "0, 1" area probably fine.  Thx.

source/MaterialXCore/Element.cpp
```diff
@@ -827,7 +827,18 @@ bool ValueElement::validate(string* message) const
             validateRequire(valueElem != nullptr, res, message, "Interface name not found in referenced declaration");
```
kwokcb, line 827: All string enums should resolve to integers for h/w. Is this what you were seeing ? i.e. There should not be type mismatches -- so perhaps a fix is required elsewhere.

kwokcb (commented):
The enum mismatch seems like a logic issue somewhere else ?

Issue #2900, opened 2026-05-08
ND_UsdPrimvarReader_vector2 fails shader generation for Hardware shading languages


ld-kerley:
USD uses `UsdPrimvarReader` to access primvars from the scene - most commonly texture coordinates to feed `UsdUVTexture`. 

The current MaterialX shader generation for the hardware shading languages (which OpenUSDs Storm uses), currently fails code generation for these nodes.

[test_file.zip](https://github.com/user-attachments/files/27534437/test_file.zip) (updated to include texture file)

The test files zip contains three files representing the same Material.

  ```mermaid                                                                                                                                                                                                                                                                  
  graph LR                                                                                                                                                                                                                                                                    
      Primvar["UsdPrimvarReader"]                                                                                                                                                                                                                               
      Tex["UsdUVTexture"]                                                                                                                                                                                                                                           
      SR["UsdPreviewSurface"]                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                              
      Primvar --> Tex                                                                                                                                                                                                                                                   
      Tex --> SR                                                                                                                                                                                                                                              
  ```                                                                                                                                                                                                                                                                         

### test_uvtexture_uv_primvarreader.usda 
Which is the USDShade version - this renders fine in Strom

<img width="1257" height="1142" alt="Image" src="https://github.com/user-attachments/assets/d86a43c8-d86f-457d-b9b8-941732f7a8b6" />

### test_uvtexture_uv_primvarreader_mx.usda 
Which is the MaterialX version in USD - this does not render in Storm.

<img width="1257" height="1142" alt="Image" src="https://github.com/user-attachments/assets/8f6dab2d-1154-4122-9d50-52ef1cb248d2" />

And emits the following error...
```
ERROR: Usdview encountered an error while rendering.
	Error in 'pxrInternal_v0_26_5__pxrReserved__::HdSt_ApplyMaterialXFilter' at line 1629 in file /Users/leekerley/dev/github_ld_kerley/OpenUSD/pxr/imaging/hdSt/materialXFilter.cpp : 'Unable to create the Glslfx Shader.
MxException: No 'geomprop' parameter found on geompropvalue node 'primvar'. Don't know what property to bind'
Coding Error: in HdSt_ApplyMaterialXFilter at line 1629 of /Users/leekerley/dev/github_ld_kerley/OpenUSD/pxr/imaging/hdSt/materialXFilter.cpp -- Unable to create the Glslfx Shader.
[...]

PR #2957, opened 2026-05-29
Resolve geompropvalue names through nested node graphs in HW codegen

ld-kerley:
Resolves the MaterialX part of #2900 - Note a OpenUSD PR is still necessary to fix things when running inside of OpenUSD.

A geompropvalue node wrapped in a node graph (e.g. UsdPrimvarReader) receives the name of the geometric property through the graph interface rather than as a local value. The hardware generators only inspected the node's local "geomprop" input, so generation threw "No 'geomprop' parameter found" whenever the name was supplied from an enclosing graph.

Track the active compound instance on a stack in GenContext (pushed via the new ScopedSetCompoundInstanceNode RAII guard from CompoundNode during createVariables / emitFunctionDefinition / emitFunctionCall). HwGeomPropValueNode now climbs that stack in tandem with each graph boundary it crosses, so the property name resolves correctly across one or more levels of nested compounds.

Also updated MaterialXRender to allow testing of UsdPrimvarReader for reading texture coordinates "st".

Add Mesh stream aliasing so getStream() can map the bound "i_geomprop_st" attribute onto the existing UV0 texture-coordinate stream; the glTF and OBJ loaders register this alias so the USD "st" primvar binds to mesh UVs.

Add a GenShader test covering two levels of nested geompropvalue interface resolution across GLSL/MSL/Slang.

### DIFF
--- a/source/MaterialXGenHw/Nodes/HwGeomPropValueNode.cpp
+++ b/source/MaterialXGenHw/Nodes/HwGeomPropValueNode.cpp
@@ -13,17 +13,77 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
+static const string GEOMPROP = "geomprop";
+
 ShaderNodeImplPtr HwGeomPropValueNode::create()
 {
     return std::make_shared<HwGeomPropValueNode>();
 }
 
-void HwGeomPropValueNode::createVariables(const ShaderNode& node, GenContext&, Shader& shader) const
+// Find the ShaderInput that contains the actual name of the geomprop to read
+static const ShaderInput* getGeomPropValueInput(const ShaderNode& node, const GenContext& context)
 {
+    // We should always have the "geomprop" input locally on this node
     const ShaderInput* geomPropInput = node.getInput(GEOMPROP);
+    if (!geomPropInput)
+    {
+        throw ExceptionShaderGenError("No 'geomprop' input found on geompropvalue node '" + node.getName() + "'.");
+    }
+
+    // Depth into the compound instance stack, measured from the innermost
+    // instance. Incremented each time we cross a graph boundary so that nested
+    // compounds resolve to the correct enclosing instance.
+    size_t depth = 0;
+
+    while (geomPropInput)
+    {
+        // If it has a local value set, then we've found the ShaderInput and can return that
+        ConstValuePtr value = geomPropInput->getValue();
+        if (value)
+        {
+            return geomPropInput;
+        }
+
+        // If we don't have a local value, then we may have an incoming connection from a containing NodeGraph
+        // Note : the "geomprop" input on the base node is "uniform" so it cannot be driven by the output from
+        // another node.
+        const ShaderOutput* upstreamPort = geomPropInput->getConnection();
+        if (!upstreamPort)
+        {
+            return nullptr;
+        }
+
+        const ShaderNode* upstreamNode = upstreamPort->getNode();
+        if (!upstreamNode || !upstreamNode->isAGraph())
+        {
+            return nullptr;
+        }
+
+        // We've reached a graph input socket. Step out to the matching compound
+        // instance, walking the instance stack outward in tandem with the graph
+        // boundary we just crossed, so that nested compounds resolve to the
+        // correct instance rather than always the innermost one.
+        const ShaderNode* compoundInstance = context.getCompoundInstanceNode(depth++);
+        if (!compoundInstance)
+        {
+            return nullptr;
+        }
+
+        // Consider the input on the compound instance to be the new potential
+        // source for the value of "geomprop". The next iteration either finds a
+        // value here or continues climbing.
+        geomPropInput = compoundInstance->getInput(upstreamPort->getName());
+    }
+
+    return nullptr;
+}
+
+void HwGeomPropValueNode::createVariables(const ShaderNode& node, GenContext& context, Shader& shader) const
+{
+    const ShaderInput* geomPropInput = getGeomPropValueInput(node, context);
     if (!geomPropInput || !geomPropInput->getValue())
     {
-        throw ExceptionShaderGenError("No 'geomprop' parameter found on geompropvalue node '" + node.getName() + "'. Don't know what property to bind");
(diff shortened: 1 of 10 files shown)


source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -452,3 +452,94 @@ TEST_CASE("GenShader: Track Application Variables", "[genshader]")
...
+// A geompropvalue node wrapped inside a nodegraph (e.g. UsdPrimvarReader) receives
+// the name of the geometric property to read through the graph interface rather than
+// as a local value. This verifies that hardware shader generation resolves that name
+// across more than one level of nested compounds, so the geometric stream is bound.
+TEST_CASE("GenShader: Nested geompropvalue interface", "[genshader]")
+{
+    // "outer_primvar" wraps "inner_primvar", which wraps a geompropvalue node. The
+    // "geomprop" name is threaded down through both graph interfaces from the top
+    // level value "st", so resolution must climb two compound instances to find it.
+    std::string testDocumentString =
```
pablode, line 465: Could perhaps a _raw string literal_ be used here to avoid the backslashes?

ld-kerley: I'm not opposed - I was just following the precedent set in the test immediately prior to this. 

Maybe its cleaner to put up a separate PR for this clean up after this one lands - I don't mind doing that - that applies the new raw-string style across all the tests (if there are more than the ones in this file).  It does feel like a good improvement to me.

pablode: Sounds like a reasonable approach.

source/MaterialXGenHw/Nodes/HwGeomPropValueNode.cpp
```diff
@@ -13,17 +13,77 @@
 
 MATERIALX_NAMESPACE_BEGIN
 
+static const string GEOMPROP = "geomprop";
```
kwokcb, line 16: Think you can use `GeomProp::CATEGORY`

source/MaterialXGenHw/Nodes/HwGeomPropValueNode.cpp
```diff
@@ -13,17 +13,77 @@
...
+
 ShaderNodeImplPtr HwGeomPropValueNode::create()
 {
     return std::make_shared<HwGeomPropValueNode>();
 }
 
-void HwGeomPropValueNode::createVariables(const ShaderNode& node, GenContext&, Shader& shader) const
+// Find the ShaderInput that contains the actual name of the geomprop to read
+static const ShaderInput* getGeomPropValueInput(const ShaderNode& node, const GenContext& context)
```
kwokcb, line 24: Niklas can speak better than I on threading but I don't think we should add any static methods in code generation. Maybe add this as a method on `ShaderNode` instead ?

source/MaterialXRender/CgltfLoader.cpp
```diff
@@ -491,6 +491,7 @@ bool CgltfLoader::load(const FilePath& filePath, MeshList& meshList, bool texcoo
                     texcoordStream = mesh->generateTextureCoordinates(positionStream);
                     mesh->addStream(texcoordStream);
                 }
+                mesh->addStreamAlias("i_" + MeshStream::GEOMETRY_PROPERTY_ATTRIBUTE + "_st", texcoordStream->getName());
```
kwokcb, line 494: This assumes a assumption on both the code gen and geometry loader side to "know" about adding "_st" postfix alias. Would it be possible to reflect this mapping within code generation so renderers can pull the appropriate given name or alias name ?

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -452,3 +452,94 @@ TEST_CASE("GenShader: Track Application Variables", "[genshader]")
...
+    "<?xml version=\"1.0\"?> \
```
kwokcb, line 466: Would it be more maintainable to keep this as a test file.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -452,3 +452,94 @@ TEST_CASE("GenShader: Track Application Variables", "[genshader]")
...
+    mx::OutputPtr output = graph->getOutput("out");
+    REQUIRE(output);
+
+    // The geomprop name "st" resolves through both graph interfaces to the bound
+    // vertex attribute "i_geomprop_st" only when nested compounds are climbed in
+    // tandem with the instance stack.
+    const std::string expectedAttribute = mx::HW::IN_GEOMPROP + "_st";
+
+    // Only the hardware backends use HwGeomPropValueNode; OSL and MDL do not.
+#ifdef MATERIALX_BUILD_GEN_GLSL
```
kwokcb, line 515: Is there a more robust check that building GLSL ? I think this is currently required for related HW backends, but would it be better long term to insert a build time `MATERIALX_BUILD_GEN_HW` define ?

source/MaterialXGenHw/Nodes/HwGeomPropValueNode.cpp
```diff
@@ -13,17 +13,77 @@
...
+    {
+        throw ExceptionShaderGenError("No 'geomprop' input found on geompropvalue node '" + node.getName() + "'.");
+    }
+
+    // Depth into the compound instance stack, measured from the innermost
+    // instance. Incremented each time we cross a graph boundary so that nested
+    // compounds resolve to the correct enclosing instance.
+    size_t depth = 0;
+
+    while (geomPropInput)
```
kwokcb, line 38: Checking that this will work if this has to read through a functional nodegraph (i.e. the graph connection is a nodedef implementation). I think it does since shader gen embeds implementation graphs.

kwokcb (commented):
The "_st" mapping feels a bit strange to me a prone to error my omission.
I'm not sure what is better but that's my only main concern with some minor
additional comments.

PR #2958, opened 2026-05-30
Improve Fresnel weighting in GGX energy compensation

jstone-lucasfilm:
This changelist improves the multiple-scattering energy compensation of the GGX microfacet model in hardware shading languages, tinting the recovered energy by the average Fresnel reflectance over the hemisphere rather than a single-direction Fresnel value.  The following specific changes are included:

- Update `mx_ggx_energy_compensation` to take a `FresnelData` argument and compute the cosine-weighted average of the Fresnel reflectance over the hemisphere internally, replacing the prior overloads that accepted a precomputed single-angle value.
- Update the conductor, dielectric, generalized Schlick, and Chiang hair closures to pass `FresnelData` directly, removing the redundant Fresnel evaluations that existed only to feed energy compensation.

As measured by our render test suite in GitHub CI, this change consistently improves the visual parity between MSL and OSL renders, with a few typical examples below (per-pixel RMS error, before -> after):

- `standard_surface_brick_procedural`: 0.00925 -> 0.00916
- `standard_surface_marble_procedural`: 0.00571 -> 0.00558
- `open_pbr_carpaint`: 0.00734 -> 0.00720
- `open_pbr_default`: 0.00627 -> 0.00619

### DIFF
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -499,6 +499,7 @@ vec3 mx_ggx_dir_albedo(float NdotV, float alpha, FresnelData fd)
 }
 
 // Compute the cosine-weighted average of the Fresnel reflectance over the hemisphere.
+// https://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides_v2.pdf
 vec3 mx_fresnel_average(FresnelData fd)
 {
     vec3 F0 = mx_compute_fresnel(1.0, fd);
--- a/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
+++ b/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
@@ -173,19 +173,6 @@ float mx_ggx_dir_albedo(float NdotV, float alpha, float F0, float F90)
     return mx_ggx_dir_albedo(NdotV, alpha, vec3(F0), vec3(F90)).x;
 }
 
-// https://blog.selfshadow.com/publications/turquin/ms_comp_final.pdf
-// Equations 14 and 16
-vec3 mx_ggx_energy_compensation(float NdotV, float alpha, vec3 Fss)
-{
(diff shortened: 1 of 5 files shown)


jstone-lucasfilm:
To provide a visual intuition for the improvements to our render test suite introduced by this PR, I'll attach the first page of our render test output, before and after the PR.

The per-pixel RMS error measurements can be found with the (mostly black) difference images on the right side of each page.

Before this PR:
MaterialX_RenderComparison_05_20_2026

After this PR:
MaterialX_RenderComparison_05_29_2026

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
```diff
@@ -511,6 +498,27 @@ vec3 mx_ggx_dir_albedo(float NdotV, float alpha, FresnelData fd)
...
 
+// Compute the cosine-weighted average of the Fresnel reflectance over the hemisphere.
+vec3 mx_fresnel_average(FresnelData fd)
+{
+    vec3 F0 = mx_compute_fresnel(1.0, fd);
+    vec3 F90 = (fd.model == FRESNEL_MODEL_SCHLICK && !fd.airy) ? fd.F90 : vec3(1.0);
+
+    // The constant 1/21 is exact for a Schlick term with an exponent of 5, while for
+    // a generalized Schlick exponent n it would be 2 / ((n + 1) * (n + 2)).
+    return F0 + (F90 - F0) * (1.0 / 21.0);
```
shill-lucasfilm, line 510: There are a couple of potential references you could add here.

1. The 1/21 result is in the Real-Time Rendering book (Eq. 9.60 on page 347):

Image

2. The general result (albeit using a falloff parameter p =  1/n) is covered by Imageworks' 2017 PBS slides:

Image

(Also included in RTR as  Eq. 9.59.)

jstone-lucasfilm: Thanks for these suggestions, @shill-lucasfilm, and I've added the Imageworks link to the average Fresnel computation.  Although I'd love to add a link to RTR as well, I'm realizing that we don't have a precedent for linking to offline books in MaterialX, so I've omitted this for now.

shill-lucasfilm (approved):
LGTM!

PR #2961, opened 2026-06-03
Fix incorrect math in `chiang_hair` BSDF

### LINKED ISSUE #2960: Incorrect math in GLSL code for `chiang_hair_bsdf`
In `mx_chiang_hair_bsdf()` the `Np` for the case of `i == 3` is supposed to be one over two-pi, but the code ends up computing half-pi due to precedence (two pi needs to be wrapped with parenthesis in the denominator to make it one over two-pi).
The entire ternary expression there is redundant, because `mx_hair_azimuthal_scattering()` already considers this case correctly (using half over pi which is equivalent to one over two pi).

Image

msuzuki-nvidia:
Fix incorrect math in `chiang_hair` BSDF.

This PR closes #2960.


jstone-lucasfilm (approved):
Thanks for the quick turnaround, @msuzuki-nvidia, and this fix looks correct to me!

Issue #2956, opened 2026-05-28
Harmonize MaterialX XML syntax


ashwinbhat:
MaterialX uses XML syntax for transportation of material data. There is desire to ensure that the MaterialX XML syntax supports the following for cleaner data interop:

- [ ] Validation using standard XML validation tools
- [ ] Support Canonical XML Version 1.1 (https://www.w3.org/TR/xml-c14n11/)
- [ ] Support an XML Schema (XSD)

Also related: https://github.com/AcademySoftwareFoundation/MaterialX/issues/2648

kwokcb:
A note from experience for trying to support json / gltf schema that there are an unlimited number of element types since node instances use element category names. This can make validation hard to be deterministic. 

The approach taken for gltf was to instead have 1 node type and category  and data type as mandatory. This matches a similar approach in USD where node instances are Shaders but with exact definition specification by node id. 

This would be a big change for xml so just putting this up for comment.

dgovil:
It would be really great to make the mtlx files be fully xml spec compliant. 
We often get requests to use a more spec compliant XML parser for performance and security reasons, or we just use different readers because we aren't using the MaterialX library directly for other reasons (analysis , processing etc)

Moving away from the PugiXML specific quirks and making sure they can be read independently would be a huge boon for this

kwokcb:
Added [link ](#2730 )to `MaterialX` `namespaces` which break validation since they are being treated as undeclared `XML` `namespaces`.

kwokcb:
Here is a simple checker using Python XML:
```python
"""
XML Compliance Checker for MaterialX (.mtlx) files
"""

import os
import sys
import argparse
from xml.etree import ElementTree as ET
import re

class XMLComplianceChecker:
    """
    @brief Checks XML compliance of .mtlx files and reports errors with line numbers and content.    
    """
    def __init__(self):
        """
        @brief Initializes the XMLComplianceChecker with an empty error list.
        """
        self.errors = []
    
    def check_file(self, filepath):
        """
        @brief Checks the XML compliance of a single file and records any errors found.
        @param filepath The path to the file to check.
        @return True if the file is compliant, False otherwise.
        """
        self.errors = []
        
        if not os.path.exists(filepath):
            self.errors.append({
                'file': filepath,
                'line': None,
                'message': f"File not found"
            })
            return False
        
        try:
            ET.parse(filepath)
            return True
        except ET.ParseError as e:
            error_msg = str(e)
[...]

PR #2963, opened 2026-06-05
Fix XML validation check for mxformat script

kwokcb:
## Change

Update #2956 

Fix so that XML validate returns an error condition. Would previously return nothing so would always pass.

### Libraries

```
python python\Scripts\mxformat.py libraries -x -y
Found 56 MaterialX files in "libraries"
- Check XML syntax
Reformatted 56 documents
```

### Resources

```
python python\Scripts\mxformat.py resources -x -y
Found 205 MaterialX files in "resources"
- Check XML syntax
- Warning: Document resources\Materials\TestSuite\nprlib\toon_shade.mtlx is not well-formed XML: unbound prefix: line 19, column 2
- Warning: Document resources\Materials\TestSuite\stdlib\texture\udim.mtlx is not well-formed XML: not well-formed (invalid token): line 4, column 78
Reformatted 205 documents
```

Note that two XML errors are now found in resources files where none were found before.

### DIFF
--- a/python/Scripts/mxformat.py
+++ b/python/Scripts/mxformat.py
@@ -16,6 +16,7 @@ def is_well_formed(xml_string):
         ET.fromstring(xml_string)  
     except ET.ParseError as e:
         error = str(e)
+    return error
 
 def main():
     parser = argparse.ArgumentParser(description="Reformat a folder of MaterialX documents in place.")


kwokcb:
@ashwinbhat, I had already put in a xml syntax check earlier but had a logic error. This fixes it so you can easily see which files pass mtlx validation but fail xml validation.

ashwinbhat (approved):
Looks good to me. Thanks for this fix

jstone-lucasfilm (approved):
Thanks for the fix, @kwokcb!  There's still a good deal of work needed before we could enable this flag in our GitHub CI, as MaterialX intentionally supports features not allowed in standard XML, but this change seems like a good step in the direction of strict XML alignment.

PR #2964, opened 2026-06-05
Initial scattering equations for the PBR specification

jstone-lucasfilm:
This changelist expands the MaterialX PBR specification with the formal mathematical foundations of its shading nodes, so that the behavior of each BSDF, EDF, and VDF can be reproduced directly from the document rather than from the cited references or shading code implementations.

- Add a Scattering Framework section defining the formal notation, the BSDF and EDF distributions, and the descriptive terminology shared by the per-node equations.
- Add Reflectance Models and Light Transport sections covering the shared microfacet model, energy conservation, and the governing light transport and volume transfer equations.
- Add inline equation sections to each shading node, mapping its artist-facing inputs to the symbols used in its defining equations.
- Add an appendix with the extended derivations for the EON, Zeltner sheen, Chiang hair, and thin-film iridescence models.

This work was developed as a collaboration betweeen the AOUSD Materials Working Group and the ASWF.  The first iteration of the Scattering Framework was written by Anders Langlands, and early review and feedback were provided by André Mazzone, Stephen Hill, and Arnon Marcus.


documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -119,9 +141,9 @@ The surface normal used for shading calculations is supplied as input to each BS
...
+In our volume shader model the scattering of light in a participating medium is controlled by a volume distribution function (VDF), with coefficients controlling the rate of absorption and scattering. The VDF represents what physicists call a _phase function, _describing how the light is distributed from its current direction when it is scattered in the medium. This is analogous to how a BSDF describes scattering at a surface, but with one important difference: a VDF is normalized, summing to 1.0 if all directions are considered. Additionally, the amount of absorption and scattering is controlled by coefficients that gives the rate (probability) per distance traveled in world space. The **absorption coefficient** sets the rate of absorption for light traveling through the medium, and the **scattering coefficient** sets the rate of which the light is scattered from its current direction. The unit for these are $m^{-1}$.
```
dbsmythe: Minor formatting nitpick: the markdown underscores around "phase function" in the second sentence are misplaced.

documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -676,6 +1335,530 @@ The MaterialX PBS Library includes a number of nodegraphs that can be used to ap
...
+The polarized phase shifts are computed with the two-argument arctangent $\operatorname{atan2}(y, x)$, which is required to resolve the correct quadrant:
+
+```math
+\phi_{23}^s = \operatorname{atan2}\left(2\eta_2 V \cos\theta_t,\; U^2 + V^2 - (\eta_2 \cos\theta_t)^2\right)
+```
+<p></p>
+
+```math
+\phi_{23}^p = \operatorname{atan2}\left(2\eta_2 \eta_3^2 \cos\theta_t \left(2\bar{k}U - (1 - \bar{k}^2)V\right),\; \left(\eta_3^2 (1 + \bar{k}^2) \cos\theta_t\right)^2 - \eta_2^2 (U^2 + V^2)\right)
+```
```
dbsmythe: The use of "$\operatorname" is not parsing correctly for me in my browser (Firefox 149.0.2 running in MacOS Sequoia 15.7.1).  Can these expressions be written without using "operatorname"?

dbsmythe (changes requested):
This looks like a fantastic change, much needed.  I will let others proofread the details of the descriptions and equations, I just had a few formatting notes:

documents/Specification/MaterialX.PBRSpec.md
jstone-lucasfilm: Good catch, and fixed!

documents/Specification/MaterialX.PBRSpec.md
jstone-lucasfilm: Good catch, and I've replaced the use of `\operatorname` with `\mathrm` for more consistent rendering across browsers.

portsmouth:
Looks like a lot of work, nice job!

I will note that this gives quite a bit more detail in terms of formulas than the OpenPBR spec does. I assume the point is to make it more clear what the existing MaterialX spec should correspond to in an implementation.

In the OpenPBR spec we could have written similar formulas down, but we opted not to since:

- in most cases, a reference is enough to allow someone to implement it, so long as we make it unambiguous in the text what variant to implement. To give *all* the details, you'd have to write a book length document (and PBRT does exist..).
- the hope was to be fully specific only in the sense that we give unambiguous BSDFs, and layer structure, then require the rest be determined by light transport.
- though in some cases, even in the BSDFs we didn't want to commit to a particular formula , as not all implementations would be able (or want) to adhere to it. An example would be the microfacet multi-scatter, as various reasonable implementations are available. 

For the layering model in particular, one outcome of having a more general description in terms of physical light transport in slabs is that certain physical effects come "for free" in the ground truth, e.g. darkening and roughening effects of the coat and fuzz, and the view-dependence of the coat absorption. Those would have to be put in by hand otherwise, and are missing from the current Material layering model. 

You could argue this makes the OpenPBR spec harder to implement, but it also pushes implementations to be more physically correct and converge on the same underlying physical model, rather than committing to one looser approximation. 

It would be good though to have the conversation about whether OpenPBR should be more specific, possibly to match what was done here in some cases.

anderslanglands:
My view is that a specification should be precise about what the intended result is. Leaving it vague because some implementations have different requirements (eg realtime), just opens the door to divergence. 

If a particular implementation wants to make a different choice about something, then that is their choice to make (eg because I have a hard real-time requirement, or I have a better model for later transport than what the spec proscribes) and a comparative conformance suite will clearly show those differences, but there should never be any confusion about what is the correct, intended behaviour. 

there clearly _is_ a correct, intended behaviour because you have a reference implementation that can be tested against. If the spec doesn’t match that exactly then the spec is of limited value IMO. 

Basically it’s my opinion that it’s OK that an implementation doesn’t match the spec exactly, as long as everyone knows _why_ it doesn’t match and _how much_ and that’s an intentional choice, not just drift created by a vague specification.

portsmouth:
>My view is that a specification should be precise about what the intended result is. Leaving it vague because some implementations have different requirements (eg realtime), just opens the door to divergence ... there should never be any confusion about what is the correct, intended behaviour...  

I think the OpenPBR spec doesn't leave the intended appearance vague, it defines it precisely via physical light transport in a layered slab structure with defined BSDFs. It deliberately omits a fully enumerated set of approximation formulas (for the BSDF or the layering), since those can be worked out or looked up, and prescribing them in the spec would actually narrow the definition unnecessarily.

This was an early design decision. Compared to something like Standard Surface, where the model is a hard-coded mix of (not fully defined) BSDFs with ad-hoc blend weights encoding an approximate layer structure, the slab-based formulation has some advantages:

 - Physical effects like coat darkening, roughening, and view-dependent absorption arise naturally from the light transport rather than having to be hand-coded as separate terms. Implementations that do the physics correctly get these for free.

 - Different implementations can choose different layering approximations to suit their constraints (albedo-scaling for most production cases, more rigorous layering in an offline renderer) and still be working toward the same well-defined physical ground truth. Convergence in the limit is guaranteed by the physics, not just hoped for.

The trade-off is that it requires more engineering work, but in practice various implementations are out in the wild and the results are close, since everyone is working to the same underlying light transport.

>there clearly is a correct, intended behaviour because you have a reference implementation that can be tested against. If the spec doesn’t match that exactly then the spec is of limited value IMO.

The MaterialX graph we supply wasn't intended to represent the ground truth appearance of the model. It's an approximation we designed to be as close as possible to the spec, but the ground truth is defined by the spec, i.e. physical light transport in the slab, not by that graph. 

Declaring the graph to be the reference would be problematic, because the current MaterialX layering model doesn't fully capture some of the physical effects the spec requires. For example the MaterialX layering operation currently ignores the effect of the internal bounces in the slab, or volumetrics within it, so to approximate that in the graph one has to insert some factors. The light transport tells you what those factors have to be, in a fairly straightforward way.

Designating it the reference in its current state would mean requiring all renderers to replicate its specific approximations, including the less accurate ones. I'd like to see work on improving the MaterialX layering model, but that's a separate issue.

>a comparative conformance suite will clearly show those differences, but there should never be any confusion about what is the correct, intended behaviour.

Agreed, and we're working on developing this test suite to try to identify areas of divergence and work to iron those out. A suite that compares implementations against each other and against ground-truth renders would surface deviations clearly, whether intentional or not, and would be valuable regardless of whether the spec prescribes exact formulas.

It would be good to have a live discussion about this sometime, maybe in the OpenPBR meeting.

documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -119,9 +141,9 @@ The surface normal used for shading calculations is supplied as input to each BS
...
+In our volume shader model the scattering of light in a participating medium is controlled by a volume distribution function (VDF), with coefficients controlling the rate of absorption and scattering. The VDF represents what physicists call a _phase function_, describing how the light is distributed from its current direction when it is scattered in the medium. This is analogous to how a BSDF describes scattering at a surface, but with one important difference: a VDF is normalized, summing to 1.0 if all directions are considered. Additionally, the amount of absorption and scattering is controlled by coefficients that gives the rate (probability) per distance traveled in world space. The **absorption coefficient** sets the rate of absorption for light traveling through the medium, and the **scattering coefficient** sets the rate of which the light is scattered from its current direction. The unit for these are $m^{-1}$.
```
portsmouth:
```suggestion
In our volume shader model the scattering of light in a participating medium is controlled by a volume distribution function (VDF), with coefficients controlling the rate of absorption and scattering. The VDF represents what physicists call a _phase function_, describing how the light is distributed from its current direction when it is scattered in the medium. This is analogous to how a BSDF describes scattering at a surface, but with one important difference: a VDF is normalized, summing to 1.0 if all directions are considered. Additionally, the amount of absorption and scattering is controlled by coefficients that gives the rate (probability) per distance traveled in world space. The **absorption coefficient** sets the rate (per unit-length) of absorption of light energy traveling through the medium, and the **scattering coefficient** sets the rate (per unit-length) at which the light is scattered from its current direction. The units for these are $m^{-1}$.
```

documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -119,9 +141,9 @@ The surface normal used for shading calculations is supplied as input to each BS
...
 
-Light can also be emitted from a volume. This is represented by an EDF analog to emission from surfaces, but in this context the emission is given as radiance per distance traveled through the medium. The unit for this is _Wm<sup>−3</sup>sr<sup>−1</sup>_. The emission distribution is oriented along the current direction.
+Light can also be emitted from a volume. This is represented by an EDF analog to emission from surfaces, but in this context the emission is given as radiance per distance traveled through the medium. The unit for this is $`W\,m^{-3}\,sr^{-1}`$. The emission distribution is oriented along the current direction.
```
portsmouth: >The emission distribution is oriented along the current direction.

Isn't it just assumed to be isotropic?

jstone-lucasfilm: Good catch, Jamie, and I've updated this sentence to align with the rest of the specification.

jstone-lucasfilm:
Thanks to @dbsmythe, @portsmouth, @anderslanglands, and @HardCoreCodin for the excellent review notes, and I believe we're close to being ready to merge these initial scattering equations for the PBR specification. 

Feel free to weigh in with additional thoughts on this PR, and we expect this work to be continually refined through future PRs as well.

niklasharrysson (approved):
I’ve only had a chance to take a high level look, but my impression is that this is an excellent and much-needed improvement to the PBR specification.

It makes the work feel much more complete, rigorous, and mathematically formalized. 

Great work! 😊👏

PR #2965, opened 2026-06-08
Restore XML conformant logic for escaping < and > characters 

kwokcb:
## Changes

Update #2956

From a user point of all run-time strings will still show `<` and `>`. The only difference is this enforces writing conformant escaped characters to file. This does **not** disallow users from using unescaped characters by hand editing.

### Implementation

Replace MaterialX specific code for writing unescaped characters `<` and `>` with original `PugiXML` logic. 

### Tests

- Add unit test for writing escaped characters
- Add test to read back escaped output to make sure that unescaping works properly.

### Example

Here is an original UDIM test file:
```xml
<?xml version="1.0"?>
<materialx version="1.38">
  <image name="image_color" type="color3">
    <input name="file" type="filename" value="resources/Images/grid_udim/grid.<UDIM>.png" colorspace="srgb_texture" />
  </image>
  ...
</materialx>

```

Here is the escaped version written out:
```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <image name="image_color" type="color3" xpos="1.724638" ypos="0.000000">
    <input name="file" type="filename" value="resources/Images/grid_udim/grid.&lt;UDIM&gt;.png" colorspace="srgb_texture" />
  </image>
  ...
</materialx>
```

Loading either one in results in the same unescaped user string be stored at run-time and shown the UI. This is using the GraphEditor:

image


kwokcb:
Pinging @ashwinbhat, @ld-kerley, and @jstone-lucasfilm for feedback / review. 

I'm not sure why `<` and `>` was ever modified to not be escaped as from a user perspective the only time you'd notice is if you hand edit a file -- but even here you can write either escaped or non-escaped, and run-time is always unescaped. A standard XML parser will always handle this as does PugiXML before it was modified to have this non-conformant behaviour.

source/MaterialXFormat/External/PugiXML/pugixml.cpp
```diff
@@ -3947,12 +3947,18 @@ PUGI__NS_BEGIN
 					break;
 				case '<':
 					// MaterialX: Allow angle brackets in MTLX serialization.
-					writer.write(*s);
+					if (flags & format_mtlx)
+						writer.write('<');
+					else
+						writer.write('&', 'l', 't', ';');
```
kwokcb: This is the original / unaltered code.

source/MaterialXFormat/External/PugiXML/pugixml.cpp
```diff
@@ -3947,12 +3947,18 @@ PUGI__NS_BEGIN
...
 					++s;
 					break;
 				case '>':
-					// MaterialX: Allow angle brackets in MTLX serialization.
-					writer.write(*s);
+					// MaterialX: Allow angle brackets in MTLX serialization.					
+					if (flags & format_mtlx)
+						writer.write('>');
+					else
+						writer.write('&', 'g', 't', ';');
```
kwokcb: his is the original / unaltered code.

source/MaterialXFormat/XmlIo.cpp
```diff
@@ -347,10 +347,19 @@ void readFromXmlString(DocumentPtr doc, const string& str, const FileSearchPath&
...
+    {
+        // If not strict add MaterialX specific output formatting.
+        flags |= format_mtlx;
+    }
+
     xml_document xmlDoc;
     xml_node xmlRoot = xmlDoc.append_child("materialx");
     documentToXml(doc, xmlRoot, writeOptions);
-    xmlDoc.save(stream, "  ");
+    xmlDoc.save(stream, "  ", flags);
```
kwokcb: Passing this flag allows the <> option to be allowed, but other fixes for conformance can reuse the same flag.

source/MaterialXTest/MaterialXFormat/XmlIo.cpp
```diff
@@ -371,3 +371,44 @@ TEST_CASE("Locale region testing", "[xmlio]")
...
+    mx::DocumentPtr readDoc = mx::createDocument();
+    readDoc->setDataLibrary(stdlib);
+    mx::readFromXmlString(readDoc , output_string);
+    imageNode = readDoc->getNode("testImage");
+    REQUIRE(imageNode != nullptr);
+    fileInput = imageNode->getInput("file");
+    REQUIRE(fileInput != nullptr);
+    std::string unescaped_string = fileInput->getValueString();
+    // There should not be any escaped characters in the value string
+    REQUIRE(unescaped_string.find("&lt;&gt;") == std::string::npos);
```
kwokcb, line 407: This was always the logic as read always unescapes.

ashwinbhat:
HI @kwokcb having the option is helpful but it does create a situtation where mtlx generated by MaterialX will be non-conformant.  As per https://github.com/AcademySoftwareFoundation/MaterialX/blob/74d314d01e79974b678ca7eba0cdd9b60d56677b/documents/Specification/MaterialX.Specification.md?plain=1#L221 `UDIM` tags are have `<` and `>`, should this be updated to mention that these should be encoded for XML?

kwokcb:
I have put this up to allow maintaining "previous" (undesired) behaviour for discussion. 

**I would advocate to remove this option and always require encoding for any SGML derivatives (XML, HTML etc).**
( HTML5 I believe has looser constraints.) 

 Hence the spec would need the clarification you note.

kwokcb:
After discussing with @ashwinbhat offline, I have made this restore PugiXML behaviour for `<` and `>` to write conformant XML all the time.

PR #2966, opened 2026-06-09
Fixes for Slang shader generation

tdavidovicNV:
Fixes multiple slang bugs:
- Correct handling when _program is reset
- Correct dispatch size in makeTexture
- Fixes a typo
- Uses `mx_matrix_mul` instead of `*` to have Slang do matrix rather than component-wise multiplication
- Correctly handle semicolons when replacing macros.
- Reserve both `mix` (to be replaced) and `lerp` (replacing) so the find-replace mix->lerp works correctly
-


jstone-lucasfilm (approved):
These fixes all look good to me, thanks @tdavidovicNV!

PR #2967, opened 2026-06-09
Bind Slang direct light uniforms

tdavidovicNV:
Enable direct light binding for Slang (was env map only until now)


jstone-lucasfilm (approved):
Thanks for this fix, @tdavidovicNV!

PR #2968, opened 2026-06-09
Fix 1.38 interface-bound input upgrade

tdavidovicNV:
MaterialX 1.39 validates node inputs as having exactly one binding. 1.38 nodegraph implementations allowed having a default value and an interfacename on implementation inputs. During the 1.38 to 1.39 upgrade we remove the stale value if connection is present.

(This is needed to use some of our Houdini 19.5 assets with MaterialX 1.39)


jstone-lucasfilm (approved):
This looks like a correct fix to me, thanks @tdavidovicNV!

PR #2970, opened 2026-06-09
Stabilize values in `standard_surface_unit.mtlx` test case

tdavidovicNV:
The old values led to, after conversion, non-physical values (IOR mostly, took rest as well). That is clearly undefined behavior, and GLSL and OSL came to significantly different results.

I picked new arbitrary values that:
- convert to reasonable ranges
- obviously fail if they are not converted

Migrated values:
- base - was: 1.5m -> 1.5m; now: 75cm -> 0.75m
- specular_roughness - was: 0.25mm -> 0.00025m; now: 250mm -> 0.25m
- specular_IOR - was: 2.41799cm -> 0.0241799m; now: 150cm -> 1.5m
- subsurface - was: 0.4mm -> 0.00004m; now: 400mm -> 0.4m

None of those values are in any way meaningful as distances, before or after, but now they land in a detectable ranges.


jstone-lucasfilm (approved):
This is definitely an improvement, thanks @tdavidovicNV, and I'd have no objection to merging this change.

Ideally we should _document_ the behavior of this test case, so that future readers understand its goals and choices, but that's beyond the scope of your fix.

PR #2974, opened 2026-06-10
Fix syntax and logic issues in `tests_to_html.py`

ppenenko:
## Summary
This PR addresses bugs in `tests_to_html.py` that affect the generation of HTML test reports:

1. **Robust Suffix Stripping**: Replaces hardcoded string slicing lengths (`[0:-8]`) with Python's built-in `removesuffix()` method. This correctly handles variable-length language suffixes (e.g., `_glsl.png` which is 9 characters vs `_osl.png` which is 8 characters), preventing filename truncation bugs.
2. **CSS Syntax Error**: Adds a trailing newline (`\n`) to the `@media print` block to prevent it from concatenating directly with the following `td` rule, which produces invalid CSS.
3. **Three-way Comparison Bug**: Fixes a copy-paste error where `path3` was joined with `args.inputdir2` instead of `args.inputdir3`.

This fixes/addresses the following issue on the Autodesk fork:
* [autodesk-forks/MaterialX#1484](https://github.com/autodesk-forks/MaterialX/issues/1484)

## Test plan
- Run `tests_to_html.py` with variable-length language suffixes (like `glsl` and `osl`) and verify that diff image paths are correctly constructed without truncation.
- Verify that three-way comparisons correctly locate files in `inputdir3`.
- Open the generated HTML report and verify that the CSS is valid and loads correctly.

### DIFF
--- a/python/MaterialXTest/tests_to_html.py
+++ b/python/MaterialXTest/tests_to_html.py
@@ -87,7 +87,7 @@ def main(args=None):
     fh.write("<style>\n")
     # Preserve inline background-color when printing to PDF, so transparent renders
     # don't vanish into the white page.
-    fh.write("@media print { * { print-color-adjust: exact; -webkit-print-color-adjust: exact; } }")
+    fh.write("@media print { * { print-color-adjust: exact; -webkit-print-color-adjust: exact; } }\n")
     fh.write("td {")
     fh.write("    padding: " + str(args.cellpadding) + ";")
     fh.write("    border: " + str(args.tableborder) + "px solid black;")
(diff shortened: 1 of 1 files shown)


python/MaterialXTest/tests_to_html.py
```diff
@@ -87,7 +87,7 @@ def main(args=None):
     fh.write("<style>\n")
     # Preserve inline background-color when printing to PDF, so transparent renders
     # don't vanish into the white page.
-    fh.write("@media print { * { print-color-adjust: exact; -webkit-print-color-adjust: exact; } }")
+    fh.write("@media print { * { print-color-adjust: exact; -webkit-print-color-adjust: exact; } }\n")
```
ppenenko, line 90: The CSS fix.

python/MaterialXTest/tests_to_html.py
```diff
@@ -192,13 +192,15 @@ def main(args=None):
             diffRms1 = diffRms2 = diffRms3 = None
 
             if file1 and file2 and DIFF_ENABLED and args.CREATE_DIFF:
-                diffPath1 = fullPath1[0:-8] + "_" + args.lang1 + "-1_vs_" + args.lang2 + "-2_diff.png"
```
ppenenko, line 195: -8 is right for `osl` but not for `glsl`

python/MaterialXTest/tests_to_html.py
```diff
@@ -173,8 +173,8 @@ def main(args=None):
         langPaths2.append(path2)
 
         if useThirdLang:
-            file3 = file1[:-len(postFix)] + args.lang3 + ".png"
-            path3 = os.path.join(args.inputdir2, path1[len(args.inputdir1)+1:])
+            file3 = f"{file1.removesuffix(postFix)}_{args.lang3}.png"
+            path3 = os.path.join(args.inputdir3, path1[len(args.inputdir1)+1:])
```
ppenenko, line 177: Copy-and-paste bug.

jstone-lucasfilm (approved):
Thanks for these fixes, @ppenenko!

PR #2975, opened 2026-06-10
Validate glTF data in `CgltfLoader`

jstone-lucasfilm:
This changelist adds a `cgltf_validate` step to `CgltfLoader`, rejecting invalid glTF files before their contents are used to construct meshes.

The cgltf parser is intentionally permissive, and does not verify that accessor ranges, buffer view extents, and object references are internally consistent.  Validating the parsed data after buffer loading guards downstream mesh construction code against out-of-bounds reads when an invalid asset is loaded.

This changelist additionally fixes a memory leak in `CgltfLoader`, where a failure in buffer loading or validation would previously return without freeing the data allocated by `cgltf_parse_file`.

### DIFF
--- a/source/MaterialXRender/CgltfLoader.cpp
+++ b/source/MaterialXRender/CgltfLoader.cpp
@@ -149,8 +149,10 @@ bool CgltfLoader::load(const FilePath& filePath, MeshList& meshList, bool texcoo
     {
         return false;
     }
-    if (cgltf_load_buffers(&options, data, input_filename.c_str()) != cgltf_result_success)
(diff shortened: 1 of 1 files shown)


kwokcb (approved):
This looks good to me.

BTW, I noticed that `GeometryHandler::loadGeometry() which calls this method  has a return code which is not checked in for GLSL, MSL, Slang (all copied code I assume).

jstone-lucasfilm:
Good point, @kwokcb, and we should indeed add proper error reporting to our render test suite in a future PR.  For now, this means that the render test suite will silently render empty images rather than reporting invalid meshes, which isn't ideal.

PR #2790, opened 2026-02-23
Expose API + Graphing Utilities for Shader Generator Graphs 

kwokcb:
## API Exposure Change

The graphs produced by shader generation are hard to debug / visualize as they are using a different data model.

The proposed change here is to be able to create a graph diagram using Python or Javascript by
exposing more of the internal C++ API.

As an example usage Mermaid graph generation is provided as a utility.

### API Modifications

- Add Python and Javascript wrappers `ShaderNode`, `ShaderInput` and `ShaderOutput`
- Add access to get graph nodes from `Shader` 
- Add graph creation utility in Python inside of `generateshader.py`. 
- Add graph creation utility in Javascript as a JS module utility function

### Tests

- Python graph generation added to CI tests. (see some examples below)
- Javascript graph generation  add to unit tests.
```
Running 1 test using 1 worker

Generating shader for essl...
Shader graph diagram generated.    # New graph gen test
Generating shader for genglsl...
Generating shader for genmsl...
```

### Sample Results

( Markdown wrappers are added as part of `generateshader.py` )

e.g. Velvet (no input expansion)

```mermaid
graph TB
    geomprop_Nworld["geomprop_Nworld (TEXTURE|GEOMETRIC)"]
    style geomprop_Nworld fill:#4B0336,stroke:#222222,stroke-width:1px,color:#ffffff
    geomprop_Tworld["geomprop_Tworld (TEXTURE|GEOMETRIC)"]
    style geomprop_Tworld fill:#4B0336,stroke:#222222,stroke-width:1px,color:#ffffff
    SR_velvet["SR_velvet (SHADER|SURFACE|CLOSURE)"]
    style SR_velvet fill:#0e7865,stroke:#222222,stroke-width:1px,color:#ffffff
    Velvet["Velvet (MATERIAL|SHADER|SURFACE|CLOSURE)"]
    style Velvet fill:#7ccba2,stroke:#222222,stroke-width:1px,color:#ffffff
    geomprop_Nworld --"out --> coat_normal"--> SR_velvet
    geomprop_Nworld --"out --> normal"--> SR_velvet
    geomprop_Tworld --"out --> tangent"--> SR_velvet
    SR_velvet --"out --> surfaceshader"--> Velvet
```

e.g. Tiled Brass (input expansion)

```mermaid
graph LR
    geomprop_Nworld["geomprop_Nworld (TEXTURE|GEOMETRIC)"]
[...]

PR #2976, opened 2026-06-11
Add computeStructuralHash for shader graphs

ashwinbhat:
Introduce computeStructuralHash in MaterialXGenShader, a free function that produces a size_t hash of a finalized ShaderGraph capturing its topology and port types while ignoring instance-specific data (node/port names, parameter values, and texture paths). Two graphs that wire up the same node implementations in the same pattern hash identically even when their constants differ, enabling shader deduplication/caching and distinct-structure reporting across large material libraries.

The function is an external visitor over the public API of ShaderGraph, ShaderNode, and ShaderPort: it never mutates the graph and folds node implementation hashes, classifications, port structure (type, semantic, colorspace, unit, geomprop, uniform/bind-input flags), and connection topology into a Boost/Hydra-style hash combiner.

The hash is deterministic within a build but session-local: it relies on std::hash and implementation hashes, so it must not be persisted or compared across builds/versions, and it is a fast pre-filter rather than a collision-proof equivalence proof.

Also adds:
- Python binding (computeStructuralHash) in PyMaterialXGenShader.
- A --dumpHash flag to generateshader.py.
- A GLSL structural-hash test exercising generation and determinism.

### DIFF
--- a/python/Scripts/generateshader.py
+++ b/python/Scripts/generateshader.py
@@ -53,6 +53,7 @@ def main():
     parser.add_argument('--validatorArgs', dest='validatorArgs', nargs='?', const=' ', type=str, help='Optional arguments for code validator.')
     parser.add_argument('--vulkanGlsl', dest='vulkanCompliantGlsl', default=False, type=bool, help='Set to True to generate Vulkan-compliant GLSL when using the genglsl target.')
     parser.add_argument('--shaderInterfaceType', dest='shaderInterfaceType', default=0, type=int, help='Set the type of shader interface to be generated')
+    parser.add_argument('--dumpHash', dest='dumpHash', action='store_true', default=False, help='Print the structural graph hash for each generated shader.')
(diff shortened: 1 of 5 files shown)


kwokcb (commented):
@jstone-lucasfilm suggested adding enough API exposure for my PR for building graph diagrams from a `Shader` as seen in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2790. 

Basically expose enough to allow you to build your own utilities.

I think this could be an approach to consider here as well vs an opaque non-customizable hash function. I think this could work: 
- Add a "hash" class and provide your current implementation as one case. 
- Expose it in the API so you can create custom ones in C++, Python or JS.
- e.g. in `generateshader.py` you could even add a "hasher" option like the validator option for custom hashes.

This follows the approach for customizable entry points such as generators and loaders.

---
Thoughts ?

PR #2977, opened 2026-06-12
Fix window focus check for `addNodePopup` in Graph Editor

jstone-lucasfilm:
This changelist fixes a Graph Editor bug that appeared on some platforms, where the TAB-triggered add-node popup would open only once per session.

The root cause is that `addNodePopup` used `ImGuiFocusedFlags_RootWindow` to guard the TAB shortcut, which requires the ImGui NavWindow to be exactly the top-level root window.

On application startup, this condition holds because the root window has initial focus, but on dismissing the popup, ImGui restores focus to the popup's source window, causing the check to fail for subsequent TAB presses.

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3878,7 +3878,7 @@ void Graph::showHelp() const
 
 void Graph::addNodePopup(bool cursor)
 {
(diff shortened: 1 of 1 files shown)


lfl-eholthouser (approved):
This looks great! Thanks @jstone-lucasfilm !

PR #2981, opened 2026-06-15
Remove deprecated `Syntax::typeSupported` method

jstone-lucasfilm:
This changelist removes the deprecated `Syntax::typeSupported` method, along with its `GlslSyntax`, `MslSyntax`, and `SlangSyntax` overrides, from the shader generation modules.

The method was marked `[[deprecated]]` during 1.39.0 development and has since been left without any callers in the codebase, making it unused code maintained only to preserve the deprecation cycle.  Removing it resolves warnings on newer compilers, including MSVC in Visual Studio 2026, which flag overrides of a deprecated virtual function as warnings and, under our warnings-as-errors build, promote them to build failures.

### DIFF
--- a/source/MaterialXGenGlsl/GlslSyntax.cpp
+++ b/source/MaterialXGenGlsl/GlslSyntax.cpp
@@ -356,11 +356,6 @@ GlslSyntax::GlslSyntax(TypeSystemPtr typeSystem) :
             "#define material surfaceshader"));
 }
 
-bool GlslSyntax::typeSupported(const TypeDesc* type) const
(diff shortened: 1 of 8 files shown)


kwokcb (approved):
Looks good to me.

BTW, is there an official deprecation removal policy. e.g. after N releases the API can be removed ?

jstone-lucasfilm:
@kwokcb Not yet, though our upcoming move to semantic versioning in v2.0.0 will give us a great opportunity to define those rules clearly.

Issue #2751, opened 2026-01-16
Improve WgslShaderGenerator to generate actual `wgsl` source


ashwinbhat:
WgslShaderGenerator doesn't actually produce wgsl, but just an adapted variant of glsl that is suitable for conversion to wgsl. For real-time web environments this workflow from MaterialX -> GLSL -> SPIRV -> wglsl is not performant. We should investigate in a direct wgsl generator. 

I've logged this issue to get feedback from community about their thoughts and ideas in this area.

ld-kerley:
I think this is a really good idea, and something we should lean in to across all the languages.  Delivering native code tuned to be as performant as possible to the different destination implementations seems like a good mission statement for the project.

thc1006:
I can't agree with u more!!! and I’m also interested in helping with this.

From the docs it looks like `WgslShaderGenerator` is currently a *WGSL-flavored Vulkan GLSL* generator (i.e., intended for conversion), which matches the issue description. ([MaterialX](https://materialx.org/docs/api/class_wgsl_shader_generator.html))

To scope this properly, might need to  clarify some acceptance criteria?

1. Target stages: fragment-only first, or full vertex+fragment (and compute?)
2. Require node coverage for "done": e.g., Minimal set needed to render MaterialX Viewer / Web Viewer sample graphs.
3. Binding layout expectations: should the generator emit fix `@group/@binding` convention (and which ones) or expose reflection metadata for the app to build bind groups? (WebGPU/WGSL uses group+binding as the core model. ([W3C](https://www.w3.org/TR/WGSL/))
4. Preferred compilation path for validation: direct WGSL compile in WebGPU, or still allowing an intermediate toolchain for tests?

aviralgarg05:
I would like to solve the issue, please assign me

thc1006:
> I would like to solve the issue, please assign me

Dear @aviralgarg05 , I think this issue need more discuss, maybe u can share ur fixing way and tag the Moderator ask idea.

ashwinbhat:
The scope would be fragment and vertex shaders, basically implementing a full `class WgSlShaderGenerator : public HwShaderGenerator`

A native wgsl generator would be more performant, easy to read and could be tuned as needed. Similar to `genglsl` we would have `genwgsl` with wgsl implementation for many of the fragments. 

Now this would be another maintenance overhead since for every new node we would need glsl, osl, msl, mdl, slang and wgsl 

[...]

Issue #2558, opened 2025-09-12
Update MaterialX WebViewer to use WebGPU


ashwinbhat:
The MaterialX Web Viewer uses Three.js WebGL renderer. See relevant code [here](https://github.com/AcademySoftwareFoundation/MaterialX/blob/7864cf3f10a54d1ae042bf570fd73f445e981f77/javascript/MaterialXView/source/index.js#L72-L75)

With the recent addition MaterialX WebGPU Code Generator, enhancing the Web Viewer uses Three.js WebGPURenderer would be useful to demonstrate the new code generator to the community.

scotbrew:
This looks feasible.

**Dependencies:**
- [x]  Three.js supports WebGPU renderer
- [x]  MaterialX generates GLSL code compliant with WGSL (as of MaterialX 1.39.4)
- [x]  Conversion from GLSL code to WGSL 

**Details:**
As Three.js now provides a WebGPU renderer (see [WebGPURenderer.js](https://github.com/mrdoob/three.js/blob/master/src/renderers/webgpu/WebGPURenderer.js)), the main thing required to implement in the MaterialX Web Viewer would be to have the WGSL-compliant GLSL (now generated by MaterialX as of version 1.39.4) converted into WGSL code, ideally with Javascript code inside the MaterialX Web Viewer.  After doing some basic tests from the webgpu-test repository, it looks like that should be possible with the glslang.js and twgsl.js modules.

The next step would be to put together an experimental build of the MaterialX Web Viewer that uses these 3 technologies.

kwokcb:
Adding some notes on  status of investigation by Scott, Ashwin and myself:

- [x] Transpiling is straightforward either as a preprocess, or at runtime. For the viewer we would be looking at runtime.
  - Code gen -> glslang to get SPIR-V
  - SPIR-V to WGSL via `tint` or `twsl` which is the Bablyon wrapper for `tint`.  
   - [ ] There are warnings / possible errors that show up that can be investigated / logged as issues against codegen.

- [ ] `RawShaderMaterial` seems to not be supported in Three.WebGPURenderer. This is the code path for WebGLRenderer now.  Codegen produces raw shader so this could be a blocker.
  - [ ] Is using Bablyon a alternative? Note that it accepts both GLSL and WGSL. The former is transpiled on the fly. It also appears to still support raw shaders.
[...]

PR #2983, opened 2026-06-18
WebGPU Viewer 

ashwinbhat:
- Reorganize GLSL based shader generators into sub folders. 

- Update WGSL shader generator to replace GLSL with wgsl. Introduce GlslToWgsl helper that works with binding context and syntax support.

Add a WebGPU rendering path alongside the existing WebGL one.
The viewer generates WGSL WgslShaderGenerator and routes it to Three.js TSL NodeMaterial.

- mxtsladapter.js: mtlx WGSL -> TSL bridge
- wgslmanifest.js: reflection manifest for wgsl shader

image

This work is towards these issues:
https://github.com/AcademySoftwareFoundation/MaterialX/issues/2558
https://github.com/AcademySoftwareFoundation/MaterialX/issues/2751


jstone-lucasfilm:
Hello @ashwinbhat!

I'm really happy to see these steps forward for WebGPU support, which has long been a goal for MaterialX.  I have some questions, though, that I think are worth working through before we commit to a direction, mostly about the scope of the PR and the long-term shape of the WGSL path:

1. Could this be split into smaller PRs?  As it stands it bundles three fairly independent changes: (a) the essl, vk, and wgsl folder reorganization, (b) the new WGSL generator with GLSL -> WGSL converter, and (c) the JS viewer's WebGPU/TSL path.  At roughly 7k lines they're hard to review as a unit, and the converter, which requires the greatest care in review, is the easiest to lose in the noise.  Would it make sense to propose these separately?
2. Can you clarify the goals that the folder reorganization addresses?  All of our sibling generator modules use a flat, prefix-named layout, so we'd need a strong motivation for making changes to the layout of this one.  These are also installed public headers, so moving them is potentially a breaking change for downstream clients.  Is the reorganization needed beyond giving the new converter a home?  If the converter lived flat (e.g. WgslGlslConverter.cpp), would any restructuring be required at all?
3. Could the bespoke transpiler be replaced by a _production compiler_, integrated as an optional dependency for builds that require WebGPU support?  GlslToWgsl.cpp is roughly 3,800 lines of string rewriting over machine-generated GLSL, plus eight whole-shader post-passes.  It's impressive work, but it's a bespoke transpiler that the MaterialX project would own and maintain indefinitely, and it can break silently as our codebase evolves.  Production compilers already do this job, with the added benefit that their output is valid WGSL by construction -- correctness is guaranteed, with no separate validation step to build or maintain.  Two possible options are Tint and Naga, each with its own tradeoffs.  Have you evaluated either, and what led you to the bespoke approach?

ashwinbhat:
> Hello @ashwinbhat!
> 
> I'm really happy to see these steps forward for WebGPU support, which has long been a goal for MaterialX. I have some questions, though, that I think are worth working through before we commit to a direction, mostly about the scope of the PR and the long-term shape of the WGSL path:
> 
> 1. Could this be split into smaller PRs?  As it stands it bundles three fairly independent changes: (a) the essl, vk, and wgsl folder reorganization, (b) the new WGSL generator with GLSL -> WGSL converter, and (c) the JS viewer's WebGPU/TSL path.  At roughly 7k lines they're hard to review as a unit, and the converter, which requires the greatest care in review, is the easiest to lose in the noise.  Would it make sense to propose these separately?
> 2. Can you clarify the goals that the folder reorganization addresses?  All of our sibling generator modules use a flat, prefix-named layout, so we'd need a strong motivation for making changes to the layout of this one.  These are also installed public headers, so moving them is potentially a breaking change for downstream clients.  Is the reorganization needed beyond giving the new converter a home?  If the converter lived flat (e.g. WgslGlslConverter.cpp), would any restructuring be required at all?
> 3. Could the bespoke transpiler be replaced by a _production compiler_, integrated as an optional dependency for builds that require WebGPU support?  GlslToWgsl.cpp is roughly 3,800 lines of string rewriting over machine-generated GLSL, plus eight whole-shader post-passes.  It's impressive work, but it's a bespoke transpiler that the MaterialX project would own and maintain indefinitely, and it can break silently as our codebase evolves.  Production compilers already do this job, with the added benefit that their output is valid WGSL by construction -- correctness is guaranteed, with no separate validation step to build or maintain.  Two possible options are Tint and Naga, each with its own tradeoffs.  Have you evaluated either, and what led you to the bespoke approach?

Hi @jstone-lucasfilm 

Thanks for taking an initial look, it was draft PR to get some early input and possible more interested contributors 😉
It can certainly be split-up. 

The folder reorganization is independent from WebGPU and can be split up into a separate PR. The reorg is mainly for clarity while working with the code.  Unlike other sibling generators, GLSL generator is getting extended to multiple targets. I found this a little confusing while onboarding new developers, and I think reorganizing will be helpful though not required.

 Regarding the bespoke transpiler for wgsl, I have the same concerns as you do and I would like to seek input from community. 
The core issues we need to address are:
- shader readability: current transpiler options such as Tint and Naga, generate source that is very verbose, hard to read and without comments. This cause issues debugging the shader to diagnose issues and further customization.
- compile time: the current pipeline we offer is:
          `mtlx -> glsl -> spirv (IR) -> wgsl -> [browser] -> spirv (or proprietary IR) -> binary -> driver`
 by using a direct wglsl generator (either custom transpile or with a WGSL shadergen) we can reduce it to 
          `mtlx -> wgsl -> [browser] -> spirv (or proprietary IR) -> binary -> driver`

I'll continue to explore other options that might be more acceptable and meet our desired goals.

ashwinbhat:
Closing this in favor of a better approach https://github.com/AcademySoftwareFoundation/MaterialX/pull/2996

PR #2985, opened 2026-06-19
Align OSL image alpha with texture conventions

jstone-lucasfilm:
This changelist fixes the alpha output of the OSL `image` and `hextiledimage` nodes when sampling a file with fewer than four channels.  The OSL `texture` function returns zero for channels not present in the file, so a `color4` or `vector4` `image` reading an RGB file returned a transparent rather than opaque result.  The fix queries the channel count with `gettextureinfo` and sets the output alpha to one when fewer than four channels are present, matching the GLSL and MSL backends.

More broadly, this change aligns the `image` node with the established convention that a missing fourth channel resolves to one, as already followed by the MaterialX `convert` node, the OpenUSD `UsdUVTexture` node, and standard GPU texture sampling.  The specification text for the `image` node, previously ambiguous on this point, has been clarified to note the fourth-channel exception.

### DIFF
--- a/documents/Specification/MaterialX.StandardNodes.md
+++ b/documents/Specification/MaterialX.StandardNodes.md
@@ -80,7 +80,7 @@ Standard Texture nodes:
 
 Samples data from a single image, or from a layer within a multi-layer image.  When used in the context of rendering a geometry, the image is mapped onto the geometry based on geometry UV coordinates, with the lower-left corner of an image mapping to the (0,0) UV coordinate (or to the fractional (0,0) UV coordinate for tiled images).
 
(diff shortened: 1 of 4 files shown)


jstone-lucasfilm:
From our automated render test suite, here are before and after comparisons of the color4 `blur` node, providing some visual intuition for the improved alignment across languages:

Blur Render Comparison (Before):

RenderComparisonOld

Blur Render Comparison (After):

RenderComparisonNew

meshula (commented):
This looks good to me, populating an un-supplied alpha with one will make users happy and matches the convention we use in Hydra!

PR #2988, opened 2026-06-23
Fix MDL codegen for inline sourcecode implementations

ppenenko:
## Summary

Fix three issues in the MDL shader generator that cause `mdlc` compilation failures when processing `Implementation` elements with bare inline `sourcecode` attributes (no `{{` markers, no `file`+`function`):

1. **Dangling expression bug** — `CustomCodeNodeMdl::emitFunctionDefinition` emitted the inline sourcecode as a bare expression statement, which MDL discards. For single-output nodes, the expression is now assigned to the output variable.

2. **Namespace in function names** — When a `NodeDef` uses a MaterialX namespace (e.g. `namespace="adsk"`), `getName()` returns `adsk:ND_foo`, and the colon is not a valid MDL identifier character. Apply `makeValidName()` to sanitize, consistent with other codegen paths.

3. **Unused variable warning suppression** — Add `-W "181=off"` to the `mdlc` test command, consistent with existing suppressions for C183 and C350. The MDL codegen may emit geometry property and image output variables that are unreferenced in the generated shader.

## Context

These bugs are **latent defects in untested code**. `CustomCodeNodeMdl` is instantiated when an implementation has a `sourcecode` attribute with no `{{` markers. Every upstream MDL sourcecode implementation either uses `{{` markers (routed to `SourceCodeNodeMdl`) or has `file`+`function` attributes (early return in `CustomCodeNodeMdl`). The bare-expression path was never exercised until downstream libraries introduced simple placeholder implementations like `sourcecode="0.5"`.

## Test plan

- [x] Verified locally that MDL generation produces correct `mxp_out = <expression>;` assignment
- [x] Verified `mdlc` compilation passes for all affected nodedefs
- [ ] CI should pass the existing MDL generate and compile tests


ppenenko:
Found while investigating https://github.com/autodesk-forks/MaterialX/issues/1488

ppenenko:
Examples of failed builds:

1. https://github.com/autodesk-forks/MaterialX/actions/runs/27631458817
2. https://github.com/autodesk-forks/MaterialX/actions/runs/27639139831
3. https://github.com/autodesk-forks/MaterialX/actions/runs/27762788608

jstone-lucasfilm (approved):
Thanks for these fixes, @ppenenko!

Issue #2984, opened 2026-06-18
XML Proposal: Use a Single `node` Element for MaterialX Node Instances


kwokcb:
## Background

MaterialX currently serializes node categories as XML element names. For example:

```xml
<image name="abc" type="float"/>
<add name="sum" type="float"/>
<fractal3d name="noise" type="float"/>
```

For schema definition, as MaterialX allows custom node definitions and custom categories the set of valid XML element names is effectively unbounded:

```xml
<my_custom_def1 name="n1" type="float"/>
<my_custom_def2 name="n1" type="float"/>
<my_custom_def3 name="n1" type="float"/>
<my_custom_def4 name="n1" type="float"/>
etc
```

This makes it difficult to define a complete XML Schema (XSD), since every new category introduces a new XML element type.

## Proposal

Replace category-specific XML elements with a single generic `node` element and a new `nodetype` attribute.

Instead of:

```xml
<image name="abc" type="float"/>
```

use:

```xml
<node name="abc" type="float" nodetype="image"/>
```

or using a `nodedef` identifier:

```xml
<node name="abc" type="float" nodedef="ND_image_float"/>
```

or both:

```xml
<node  name="abc" type="float" nodetype="image" nodedef="ND_image_float"/>
```

Additional examples:

Current:

```xml
<image name="abc" type="float"/>
<add name="sum" type="float"/>
<fractal3d name="noise" type="float"/>
```

Proposed:

```xml
<node name="abc" type="float" nodetype="image"/>
<node name="sum" type="float" nodetype="add"/>
<node name="noise" type="float" nodetype="fractal3d"/>
```

## Summary

This proposal replaces category-specific XML elements:

```xml
<image/>
<add/>
<fractal3d/>
<my_custom_node/>
```

with a single extensible `node` element:

```xml
<node nodetype="image"/>
<node nodetype="add"/>
<node nodetype="fractal3d"/>
<node nodetype="my_custom_node"/>
```

The primary advantages are:

- A practical and stable XML schema.
- Natural support for unlimited custom node categories.
- Simpler streaming as every element is handled the same way
- Simpler incremental parsing and resolve node type / definition afterward
- Closer alignment with the OpenUSD `Shader + info:id` model.
- Consistent representation across MaterialX XML, OpenUSD, serializations.
- Node identity becomes data (`nodetype` or `nodedef`) rather than XML syntax.

### Simplified XML Schema Definition

[...]

PR #2990, opened 2026-06-25
Proposal : New XML node instance syntax

kwokcb:
## Prototype

This code prototypes an "additive" approach to introducing this new syntax as described in issue #2984.

### Properties

- XML write option allows either current or proposed syntax. Default is current.
- XML read handles both syntax variants
- This allows moving back and forth between either formats.
- On next version "upgrade", all current libraries can be converted to new syntax and write option defaults to new syntax.

### Examples

- Modify graph editor to write new syntax
- Modify `mxformat.py` utility to write either syntax. i.e. supports bidirectional conversion.
  - Tested converting all libraries and resources using this to new syntax and re-ran tests without failure.


jstone-lucasfilm:
@kwokcb This is a very interesting proposal that demands significant thought and discussion from the MaterialX community, but I'll provide one initial suggestion:

Since you are proposing a brand new _syntax_ in MaterialX, perhaps the best starting point is a PR for new text in the [Specification Proposals](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Proposals.md) document, rather than a PR for an implementation in the codebase?

PR #2991, opened 2026-06-27
Connect `thin_walled` to `surface` node in Standard Surface

msuzuki-nvidia:
Connect `thin_walled` to the surface constructor in `standard_surface` nodedef.

### DIFF
--- a/libraries/bxdf/standard_surface.mtlx
+++ b/libraries/bxdf/standard_surface.mtlx
@@ -91,7 +91,7 @@
            doc="The color of the emitted light." />
     <input name="opacity" type="color3" value="1, 1, 1" uimin="0,0,0" uimax="1,1,1" uiname="Opacity" uifolder="Geometry"
            doc="The opacity of the entire material." />
-    <input name="thin_walled" type="boolean" value="false" uiname="Thin Walled" uifolder="Geometry" uiadvanced="true"
+    <input name="thin_walled" type="boolean" value="false" uniform="true" uiname="Thin Walled" uifolder="Geometry" uiadvanced="true"
            doc="If true the surface is double-sided and represents an infinitely thin shell. Suitable for thin objects such as tree leaves or paper." />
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks for putting this together, @msuzuki-nvidia.  The change looks right on both counts: forwarding `thin_walled` into the `surface` constructor, and marking the interface input `uniform="true"` to match it.

@ld-kerley I'm CC'ing you here since you raised the backwards-compatibility question about changing `uniform` on a shipped interface input, which is a fair point to weigh.  A few concrete things make me think the risk here is quite low:

- Established precedent: OpenPBR made the identical change (forward plus `uniform="true"`) in #2759, so this brings `standard_surface` in line with a convention that has already shipped.
- It is the norm rather than a new pattern.  The booleans that gate surface and BSDF construction are already uniform, including `energy_compensation` and `retroreflective` on the BSDF constructors, and the `ND_surface` constructor's own `thin_walled`.  The non-uniform `thin_walled` interface on `standard_surface` is currently the outlier.
- Required for consistency once connected: the `ND_surface` input is `uniform="true"` and the MDL implementation declares it as `uniform bool`, so a non-uniform interface value feeding it would be a qualifier mismatch.
- No content regresses.  Across all bundled materials and our render test suite, `thin_walled` is only ever authored as a literal value, and is never connected to an upstream node.  The only thing affected would be a third-party document driving it from a varying source, which is not a meaningful operation for a structural surface flag of this kind.

Given that, I believe this is the right next step, but I'd like to confirm this with @ld-kerley before proceeding.  If we are aligned, I would suggest we also note the uniform change in our changelog and release notes for discoverability.

jstone-lucasfilm (approved):
Based on discussion on the ASWF Slack, let's move forward with this improvement, and we can follow up with community-facing notes as we approach the release of MaterialX v1.39.6.

PR #2995, opened 2026-07-02
Add uimin/uimax for roughness and coverage in flake nodes

crydalch:
I just recently noticed the awesome new Flake shaders were missing uimin/uimax for roughness and coverage.

Does this look alright @msuzuki-nvidia ? I'm pretty sure it's correct for these to be clamped, but I guess we could use the 'soft' versions if necessary too.

Thanks!


msuzuki-nvidia:
@crydalch LGTM, thanks for the update!

jstone-lucasfilm (approved):
This looks good to me as well, thanks @crydalch!

PR #2997, opened 2026-07-03
Add `mxspec.py` specification script

jstone-lucasfilm:
This changelist adds an `mxspec.py` script for processing MaterialX specification documents, replacing the `comparenodedefs.py` script with a more general tool that supports three subcommands:

- `compare`: Compare node definitions between a specification document and a data library document, reporting differences in node sets, typed signatures, and default values.
- `normalize`: Normalize table formatting and default value notation, aligning columns and converting defaults to canonical placeholders such as `__zero__` and `__one__`.
- `expand`: Expand polymorphic tables into concrete per-signature tables, with an optional `--nodedefs` flag that adds nodedef name subheaders from the standard libraries.

GitHub CI additionally compares the NPR Nodes specification against its data library on each commit, extending the trackable alignment history already maintained for the Standard and PBR Nodes.

This work was developed as a collaboration between the AOUSD Materials Working Group and the ASWF.


ashwinbhat (approved):
LGTM, I was wondering if we should add some notes about markdown layout requirements since we are parsing table cells.

jstone-lucasfilm:
This is a great point, @ashwinbhat, and I've now added clearer reporting of parser warnings across all of the commands, so that any mismatches in conventions are surfaced to the user.

We could add explicit documentation of our Markdown conventions as well, but my concern would be that we're then creating two distinct homes for these conventions (the specification documents and our supporting Python scripts), which would then need to be kept in sync over time.

python/Scripts/mxspec.py
```diff
@@ -0,0 +1,1224 @@
...
+            defaults[(ndi.nodeName, ndi.signature)] = ndi.defaults
+
+    return signatures, defaults
+
+def expandSpecToNodes(nodeSections, typeCtx):
+    '''Expand a parsed spec into signatures and defaults dicts.'''
+    signatures = {}
+    specDefaults = {}
+    for nodeName, tables in nodeSections.items():
+        sigs = signatures.setdefault(nodeName, set())
```
kwokcb, line 799: Not sure but I think you'd want to get defaults based on the per table item's defaults ?

python/Scripts/mxspec.py
```diff
@@ -0,0 +1,1224 @@
...
+        commonInputNames = {name for name, _ in specInputs} & {name for name, _ in libInputs}
+        if commonInputNames:
+            specInputDict = dict(specSig.inputs)
+            libInputDict = dict(libSig.inputs)
+            if not all(specInputDict[n] == libInputDict[n] for n in commonInputNames):
+                continue
+
+        extraInLib = tuple(sorted(libInputs - specInputs))
+        extraInSpec = tuple(sorted(specInputs - libInputs))
+        return libSig, extraInLib, extraInSpec
```
kwokcb, line 873: Doesn't this just return the first one vs comparing all matches ?

kwokcb (commented):
This looks good to me. 
Some minor comments in matching.

---

A general observation:
- I'm curious if instead of markdown formatted tables it would be easier to use
html instead. 
- Then it's much easier to parse (especially if you tag the tables with metadata), though a bit more trouble to write.

python/Scripts/mxspec.py
jstone-lucasfilm: Good catch, and fixed!

python/Scripts/mxspec.py
jstone-lucasfilm: Yes, the exact match pass above handles most cases, and this fallback pass uses fuzzy matching and returns the first one.  So far, I don't believe we have any real-world cases where there's more than one fuzzy match, so this works well, but indeed we could revisit this if such a case appears in the future.

jstone-lucasfilm:
> A general observation:
> 
> * I'm curious if instead of markdown formatted tables it would be easier to use
>   html instead.
> * Then it's much easier to parse (especially if you tag the tables with metadata), though a bit more trouble to write.

Agreed that HTML would be easier to parse, though I'd lean towards Markdown here since the specifications are hand-authored and reviewed in pull requests, where Markdown tables remain readable in raw text and diffs.

The `normalize` subcommand is meant to address the parsing side of this tradeoff by keeping tables in a canonical form, and if we need table metadata down the road, HTML comments within the Markdown could carry it without changing the rendered document.

PR #2998, opened 2026-07-04
Initial shading model definitions for the PBR specification

jstone-lucasfilm:
This changelist expands the MaterialX PBR specification with functional definitions of its shading models, allowing the composition of each model from physically based and standard node primitives to be understood directly from the document, rather than only from its data library nodegraph.  These functional definitions are designed to complement the originating specification for each shading model, serving as the authoritative statement of its MaterialX realization for consistent interchange across applications and renderers.

- Add a Shading Model Notation section defining the compact functional notation used to define each shading model as a composition of physically based and standard node primitives.
- Add a Shared Functions section factoring out the `ior_to_f0` and `coated_emission` sub-expressions that are reused across models.
- Add full definitions of the Autodesk Standard Surface, UsdPreviewSurface, Khronos glTF PBR, and OpenPBR Surface models, each presented as a standard node interface and functional definition.
- Add an Additional Reference Implementations section describing the Disney Principled and MaterialX Lama models, and update the list of shading translation graphs with direct links to their nodegraphs.

This work was developed as a collaboration between the AOUSD Materials Working Group and the ASWF.


documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -1280,49 +1285,808 @@ Converts the hair scattering color to absorption coefficient using the mapping m
...
+    vector3 main_tangent = if (specular_anisotropy > 0.0) { main_rotated } else { tangent };
+
+    // Specular transmission (refraction), blended in by 'transmission'.
+    BSDF transmission_lobe = dielectric_bsdf(
+        tint = transmission_color,
+        ior = specular_IOR,
+        roughness = transmission_roughness,
+        normal = normal,
+        tangent = main_tangent,
+        scatter_mode = "T");
```
niklasharrysson, line 1484: The `transmission_lobe` has no thin-film parameters here. The Standard Surface reference says thin film applies to specular reflection _and_ transmission, but here it's only applied to reflection. The MaterialX graph implementation is also missing those inputs on the transmission lobe. 

If this is a bug we should make a new issue for that. Or otherwise add a note here on why this is different from the shading model reference.

documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -1280,49 +1285,808 @@ Converts the hair scattering color to absorption coefficient using the mapping m
...
+    // Thin-film thickness, converted from micrometers to nanometers for the BSDF inputs.
+    float thin_film_thickness_nm = thin_film_thickness * 1000.0;
+
+    // Specular transmission (refraction) through the interior volume, blended into the opaque base
+    // by 'transmission_weight'. The tint reverts to white once a finite depth carries color through
+    // the volume instead.
+    color3 transmission_tint = if (transmission_depth > 0.0) { color3(1.0) } else { transmission_color };
+    BSDF dielectric_transmission = dielectric_bsdf(
+        tint = transmission_tint, ior = modulated_eta_s, roughness = main_roughness,
+        normal = geometry_normal, tangent = geometry_tangent, scatter_mode = "T");
```
niklasharrysson, line 1997: Same thing here with missing thin-film on the dielectric transmission lobe. The OpenPBR spec says _"In the case of a dielectric base, the thin film should also generate color fringes in the transmission lobe."_

documents/Specification/MaterialX.PBRSpec.md
```diff
@@ -1280,49 +1285,808 @@ Converts the hair scattering color to absorption coefficient using the mapping m
...
+
+    // Coat facing reflectance (F0), reused by the coat-darkening and emission terms below.
+    float coat_F0 = ior_to_f0(coat_ior);
+
+    // Thin-walled subsurface: a translucent reflection/transmission pair scaled by the subsurface
+    // color, with 'subsurface_scatter_anisotropy' shifting weight from reflection toward
+    // transmission, blended in equal measure.
+    color3 ss_color = max(subsurface_color, 0.0);
+    BSDF ss_reflection = oren_nayar_diffuse_bsdf(color = ss_color, roughness = base_diffuse_roughness,
+        normal = geometry_normal) * (subsurface_color * (1.0 - subsurface_scatter_anisotropy));
```
niklasharrysson, line 1944: It seems the `subsurface_color` weight is applied twice in the lobes for thin-walled subsurface scattering here. It's given as the input color weight to the lobe, but also multiplied externally in the anisotropy expression. Same thing below in the transmission lobe.

The MaterialX implementation graph does the same thing. But I wonder if this is correct? For example, with a 0.5 gray color at zero anisotropy, I think the total albedo becomes 0.25 instead of the specified 0.5.

If this is true and there is a bug here, it's probably best to fix in a separate PR, fixing both the functional notation here and the MTLX graph at the same time.

niklasharrysson:
This looks like a great addition to the spec! The functional notation is so much easier to read than reading MaterialX graph implementations :)

While reading I noticed some inconsistencies compared to the shading model references, noted in separate comments above.

jstone-lucasfilm:
Great to hear from you, @niklasharrysson, and thanks for the review notes!  I've confirmed each of the inconsistencies you noted against the originating specifications, and all three turned out to be genuine issues in our nodegraph implementations, now tracked as new GitHub Issues:

- The missing thin-film effect on the transmission lobe of Standard Surface is tracked in #3025, and the missing thin-film effect on the dielectric transmission lobe of OpenPBR Surface is tracked in #3026.  Since `<dielectric_bsdf>` accepts its thin-film inputs independently of `scatter_mode`, both fixes look straightforward, and I've added notes to the two shader definitions, documenting the current behavior of each composition until those fixes are written.
- The double application of `subsurface_color` in the thin-walled subsurface lobes of OpenPBR Surface is tracked in #3027, and I believe your analysis is correct: the OpenPBR specification defines these as unit-albedo diffuse lobes scaled by a single factor of `subsurface_color`, so the current construction squares the total albedo, and a 0.5 gray at zero anisotropy yields 0.25 as you note.  As you suggest, this fix is best made in a separate changelist, updating the nodegraph and its functional definition here in tandem.

Let me know if you have additional ideas for improvements, or if this latest draft looks good to you.

niklasharrysson (approved):
That's great, thanks @jstone-lucasfilm!

This looks good to me.

PR #3001, opened 2026-07-08
Unify texcoord conventions across render targets

jstone-lucasfilm:
This changelist establishes a consistent convention for texture coordinates across the MaterialX codebase, aligning with the MaterialX specification in placing the origin of texture space at the lower left.  Geometry loaders now always return texture coordinates in the MaterialX convention, and each render target compensates for the image origin of its own texture lookups at shader generation or texture upload time, replacing the configurable vertical flips previously scattered across loaders, viewers, and tests.

- Remove the `texcoordVerticalFlip` argument from `GeometryLoader::load` and `GeometryHandler::loadGeometry`, with `CgltfLoader` now always converting texture coordinates from the upper-left origin of glTF to the lower-left origin of MaterialX, and `TinyObjLoader` passing OBJ coordinates through unchanged.
- Assign `fileTextureVerticalFlip` by render target, enabling it for GLSL, MSL, Slang, and OSL to compensate for their upper-left image origins, and disabling it for MDL and for web clients that flip images vertically on upload, with the option's documentation expanded to describe this convention.
- Render test geometry in OSL testrender through a new `%render_geometry%` scene template token and `OslRenderer::setRenderGeometry` method, replacing the built-in sphere primitive whose texture layout could not be modified.
- Replace the `--materialxtest_mode` flag in MDL render tests with `--uv_scale 0.5 1 --uv_repeat`, aligning the texture layout of the built-in MDL sphere with test-suite geometry without inverting its V coordinates.
- Add a `setTargetGenerationOptions` method to `ShaderRenderTester`, along with shared `findRenderGeometry` and `loadRenderGeometry` utilities, removing duplicated logic and filename-based coordinate flips from the render tests for each target.

In addition to its benefits to consistency and maintainability, this changelist also improves the visual parity between MSL and OSL renders in GitHub CI, with a few typical examples below (per-pixel RMS error, before -> after):

- standard_surface_brick_procedural: 0.00916 -> 0.00882
- standard_surface_gold: 0.01496 -> 0.01276
- open_pbr_carpaint: 0.00720 -> 0.00605
- gooch_shade: 0.00939 -> 0.00738


jstone-lucasfilm:
I'm attaching the latest render comparison PDF generated by GitHub CI with these changes:

[MaterialX_RenderComparison_07_08_2026_GitHub.pdf](https://github.com/user-attachments/files/29820049/MaterialX_RenderComparison_07_08_2026_GitHub.pdf)

jstone-lucasfilm:
In addition to the reviewers noted above, I'm CC'ing @krohmerNV and @jreichel-nvidia for their thoughts from the NVIDIA MDL perspective.

javascript/MaterialXView/source/viewer.js
```diff
@@ -167,8 +151,11 @@ export class Scene
 
                     child.geometry.setAttribute('uv', new THREE.BufferAttribute(new Float32Array(uvs), 2));
                 }
-                else if (flipV)
+                else
                 {
+                    // Convert texture coordinates from the glTF convention, with the
```
kwokcb, line 156: As any geometry loader could be used to load geometry this may not be always true.

source/MaterialXGenShader/GenOptions.h
```diff
@@ -112,9 +112,13 @@ class MX_GENSHADER_API GenOptions
...
-    /// file textures need to be flipped vertically to match the target's
-    /// texture space convention. By default this option is false.
+    /// file textures will be flipped before sampling. Set this option when
+    /// the target samples images with their origin at the upper left (e.g.
+    /// textures uploaded to the GPU in scanline order), compensating for
+    /// the lower-left origin of texture space in MaterialX. Leave it false
+    /// when the target samples images with their origin at the lower left
+    /// (e.g. MDL), or when the client compensates by flipping images
+    /// vertically at upload time, as web clients typically do.
     bool fileTextureVerticalFlip;
```
kwokcb, line 122: A suggestion here to be explicit and say what the target orientation is vs "flip". e.g. ORENTATION_LOWER_LEFT, ORIENTATION_UPPER_LEFT. Then a comparison between source (MaterialX) and (renderer) can be made and flip as required.  

It would be nice to codify MaterialX lower left orientation somewhere in any case.

If the plans to support USD input graphs (or other) goes forward the default uv convention could be different.

source/MaterialXRenderOsl/OslRenderer.cpp
```diff
@@ -114,9 +114,12 @@ void OslRenderer::renderOSL(const FilePath& dirPath, const string& shaderName, c
     const string INPUT_SHADER_PARAMETER_OVERRIDES("%input_shader_parameter_overrides%");
     const string INPUT_SHADER_OUTPUT_STRING("%input_shader_output%");
     const string BACKGROUND_COLOR_STRING("%background_color%");
+    const string RENDER_GEOMETRY_STRING("%render_geometry%");
 
     StringMap replacementMap;
     replacementMap[OUTPUT_SHADER_TYPE_STRING] = outputShader;
+    // POSIX-style separators keep the geometry path valid in scene XML on all platforms.
```
kwokcb: Can you clarify what issue this addresses. Thx.. Not sure what POSIX format and XML have in common and where.

source/MaterialXTest/MaterialXRenderMdl/RenderMdl.cpp
```diff
@@ -167,7 +171,13 @@ RenderUtil::RenderProfileResult MdlShaderRenderTester::runRenderer(
...
                 command += " --camera 0 0 3 0 0 0";
                 command += " --fov 45";
-                command += " --materialxtest_mode"; // align texcoord space with OSL
+
+                // Halve the U range of the built-in sphere, whose texture
+                // coordinates span (0, 0) to (2, 1), aligning it with the
+                // sphere geometry used by other renderers in the test suite.
+                // The default V orientation of the renderer matches the
+                // MaterialX convention, with the origin at the lower left.
+                command += " --uv_scale 0.5 1 --uv_repeat";
```
kwokcb: How do you know this is a the default sphere ?

kwokcb (commented):
It's great to clear up this "flip" semantic. 

I would like to be explicit about the MTLX convention and have gen options explicitly codify the convention (e.g. ORIENTATION_LOWER_LEFT, ORIENTATION_UPPER_LEFT)
The input and target conventions would be both specified and flip is determined if they differ. 

One case which I think is worth clarifying is what happens when the renderer's geometry + shader language is based on a upper-left convention -- such as DirectX. The DX geometry loaders I assume would be used (not MTLX ones) + follow the upper-left convention. In the proposal above instead of "flip", you'd have the source geometry orientation and renderer orientation both being  ORIENTATION_UPPER_LEFT. 

This can occur with the web viewer as well where the renderer's geometry loaders cannot assume to match lower left.

--
One final thing is how does this interact with the "flip" option on image loaders for upper-left and lower-left renderer scenarios. Using both of these has been confusing in the past.

javascript/MaterialXView/source/viewer.js
jstone-lucasfilm: That's a good point, and we should remain flexible about this in the future, though as of today the web viewer loads geometry exclusively through the three.js GLTFLoader, where the glTF convention holds.

source/MaterialXGenShader/GenOptions.h
jstone-lucasfilm: Thanks for this suggestion, @kwokcb, and I agree that the MaterialX convention deserves to be codified more visibly, with the expanded documentation of this option being a first step in that direction.

I see the appeal of expressing the target as an orientation rather than a flip, though since `fileTextureVerticalFlip` is public API with bindings in Python and JavaScript, renaming or retyping it would break source compatibility for downstream clients that set it.  So my proposal would be to consider this improvement at a future API-transition point for MaterialX, such as the initial development phase of MaterialX v1.40 or v2.0.

On the source side, my sense is that we should keep the MaterialX convention fixed rather than configurable, and this changelist is intended as a step in that direction: geometry loaders normalize texture coordinates to the MaterialX convention at load time, so that shader generation only needs to account for the image orientation of its target.  I'd anticipate a future OpenUSD importer following the same pattern, converting texture coordinates at import time if needed, though USD conveniently places its texture origin at the lower left as well.

source/MaterialXRenderOsl/OslRenderer.cpp
jstone-lucasfilm: Good point, and I've clarified this comment in the latest version.  The geometry path is written with POSIX-style separators, since `testrender` supports these on all platforms, including Windows.

source/MaterialXTest/MaterialXRenderMdl/RenderMdl.cpp
jstone-lucasfilm: Good question, and I've clarified this in the latest version as well.  Since no scene geometry is specified on the command line, `df_vulkan` renders its built-in sphere, whose texture coordinates span (0, 0) to (2, 1).

jstone-lucasfilm:
Thanks for these thoughts, @kwokcb.  On the suggestion of explicit orientation enumerations, see my reply in the comment thread above, where my proposal would be to consider this improvement at a future API transition point for MaterialX.

For integrations such as DirectX, where the native convention for both geometry and images places the origin at the upper left, my recommendation would follow the same pattern as the geometry loaders in this changelist: the integration converts texture coordinates to the MaterialX convention at import time, and enables `fileTextureVerticalFlip` to compensate for the upper-left origin of its image lookups.  Passing upper-left coordinates through unmodified may initially seem appealing, since the two conventions cancel out for file textures, but this would invert the behavior of procedural nodes that derive patterns from texture coordinates, e.g. ramps and checkerboards, relative to the MaterialX specification.

On the interaction with vertical flips in image loaders, the guiding principle is that exactly one vertical compensation should be active for a given target: either the generated shader flips the V coordinate of file texture lookups, or the client flips images vertically at upload time, but never both.  The web viewer is an example of the latter approach, flipping images on upload and leaving fileTextureVerticalFlip disabled, and I'm hoping the expanded documentation of this option makes these two paths
clearer than they've been in the past.

source/MaterialXTest/MaterialXRenderMdl/RenderMdl.cpp
```diff
@@ -167,7 +171,14 @@ RenderUtil::RenderProfileResult MdlShaderRenderTester::runRenderer(
...
                 command += " --fov 45";
-                command += " --materialxtest_mode"; // align texcoord space with OSL
+
+                // Since no scene geometry is specified, df_vulkan renders its
+                // built-in sphere, whose texture coordinates span (0, 0) to
+                // (2, 1).  Halve its U range to align it with the sphere
+                // geometry used by other renderers in the test suite.  The
+                // default V orientation of df_vulkan matches the MaterialX
+                // convention, with the origin at the lower left.
+                command += " --uv_scale 0.5 1 --uv_repeat";
```
jreichel-nvidia: The df_vulkan example does not have these fine-grained options (only the DXR example). But we would be willing to add them here, too. Is there a need for --uv_repeat (--materialxtest_mode did only flip+stretch)?

jstone-lucasfilm: Thanks @jreichel-nvidia, and that would be much appreciated!

Taking a closer look at the analytic texture coordinates of the built-in `df_vulkan` sphere, I think you're right that `--uv_repeat` is unnecessary.  Since the sphere's U range spans exactly (0, 2), scaling by 0.5 lands it in (0, 1), and no wrapping should be required.

I've removed `--uv_repeat` from the render test command in the latest version, so `--uv_scale` would be the only new option needed in `df_vulkan`, though we'd welcome the full set of texture coordinate options if you find them valuable for other purposes.

Once these options land in an MDL SDK release, we can raise the minimum `df_vulkan` version in the MaterialX build from its current value of 2025.0.3, and I'd be interested in your thoughts on sequencing.  Would it be reasonable for us to include this MaterialX improvement ahead of the update on the MDL side, with the understanding that MDL render tests would require a `df_vulkan` build from your upcoming release, or would you recommend that we wait?

krohmerNV: First, thanks @jstone-lucasfilm for cleaning this up!
We will add all the `--uv` options from our dxr example to df_vulkan as well and I think we would make `--materialxtest_mode`  depend on the MaterialX version so older MaterialX builds would still work.

Do you have an ETA for the next MaterialX releaase? Maybe we ware able to schedule one from our side before that.

jstone-lucasfilm: Thanks @krohmerNV and @jreichel-nvidia!  Although the timing of our next release will be based on community needs, my best guess is that it will occur in August after SIGGRAPH.

jreichel-nvidia:
I wanted to double-check our work on the new df_vulkan options but ran into some unexpected differences. To rule out mistakes in that example I looked at GLSL and also found some differences. Isn't the idea that all changes are purely internal and renderings stay exacly as before?
[tests_threshold_0.03.html.pdf](https://github.com/user-attachments/files/30219701/tests_threshold_0.03.html.pdf)

jstone-lucasfilm:
Thanks for taking a close look at this, @jreichel-nvidia, and these differences are expected, though I should have called them out more explicitly in the PR description.

First, and most importantly, pixel-identical output is still expected for MaterialX integrations such as Houdini/Solaris, Maya/LookdevX, Hydra/Storm, and Omniverse.  For any given set of generator options, this changelist leaves generated shader code unchanged, and integrations provide their own geometry and generator settings.  The same is true of our own example viewers, with the MaterialX Viewer, Graph Editor, and Web Viewer all rendering identically after this change.  The behavioral changes are confined to our render test suite, which previously assigned texture coordinates in ways that deviated from the MaterialX specification.

Within our render tests, output is expected to change in two specific situations.  The first is procedural patterns driven by texture coordinates, e.g. checkerboards, ramps, and noise, which invert vertically, since shaders now receive texture coordinates in the MaterialX convention, with the origin at the lower left.  The previous test setup passed upper-left-origin coordinates to generated shaders, which was an unintended departure from the MaterialX specification.  The second is file texture lookups whose coordinates are transformed before sampling, e.g. through `place2d`, address modes, or unit-driven scaling.  The vertical compensation previously baked into test geometry sat upstream of the node graph, while `fileTextureVerticalFlip` applies it at the sample site, and the two orderings differ whenever the coordinates are modified in between.  The new results are the ones intended by the MaterialX specification.

Untransformed file texture lookups are mathematically unchanged, and I believe this is confirmed by your comparison PDF, where each of the differing tests falls into one of the two categories above.

For validating the new `df_vulkan` options, I'd recommend comparing render targets against each other with this changelist applied, rather than against pre-change renders, which no longer serve as valid baselines for texcoord-dependent content.  One additional sanity check: since MDL render tests no longer invert the V coordinates of the built-in sphere, MDL and GLSL outputs should shift together, and their visual parity should be equal or better than before, as with the MSL/OSL comparisons noted in the PR description.

jreichel-nvidia:
Ok, makes sense. Thanks for detailed explanation!

While comparing MDL/GLSL/OSL with this changelist applied I noticed a new difference for GLSL with umirror/vclamp: 
[umirror_vclamp_new.pdf](https://github.com/user-attachments/files/30311988/umirror_vclamp_new.pdf)
[umirror_vclamp_old.pdf](https://github.com/user-attachments/files/30311989/umirror_vclamp_old.pdf)

jstone-lucasfilm:
Thanks for taking another close look at this, @jreichel-nvidia, and I believe this difference comes from _texture filtering_ rather than texture coordinates.

Measuring your comparison images by region, the divergence is confined to the band where `vaddressmode="clamp"` is actually clamping.  In the upper half of the sphere, GLSL differs from OSL by ~0.20, while MDL differs from OSL by ~0.02, and in the lower half, where coordinates remain in range, all three agree at baseline levels.  The cause is that `grid.png` carries a one-pixel black grid line along its top edge, while its bottom edge is uniform pastel.  Since these tests scale texture coordinates by 2.0, the clamped band has moved to the north pole, where meridians converge and texture LOD becomes extreme, and `GL_CLAMP_TO_EDGE` returns the edge texel of the selected mip level rather than of the base image.

The measurements bracket this precisely.  The gray wedges in the GLSL render fall in the range of ~0.41 to ~0.53, while the mip level one average of that top edge is 0.42 to 0.50, and the base level is pure black.  OSL `testrender` and MDL `tex::lookup_*` both sample the base level, and so return the black edge row as expected.  My sense is that no change is required here, since this is a property of hardware texture filtering that the new coordinate layout simply exercises for the first time.

tdavidovicNV:
I have tackled this recently and it is a can of worms. I have two counter-example shaders, where this change will change the output of the shader.

This shader attached to the standard "plane.obj" produces different images between `pr-3001` and `main` (same goes for the hextiled normal map)
```
<?xml version="1.0"?>
<materialx version="1.39">
  <hextiledimage name="hextiled" type="color3">
    <input name="file" type="filename" value="resources/Images/grid.png" />
    <input name="tiling" type="vector2" value="2.3, 2.3" />
    <input name="rotation" type="float" value="1.0" />
    <input name="rotationrange" type="vector2" value="-120.0, 120.0" />
    <input name="scale" type="float" value="1.0" />
    <input name="scalerange" type="vector2" value="0.65, 1.35" />
    <input name="offset" type="float" value="1.0" />
    <input name="offsetrange" type="vector2" value="-0.35, 0.35" />
    <input name="falloff" type="float" value="0.5" />
    <input name="falloffcontrast" type="float" value="0.0" />
  </hextiledimage>
  <output name="hextiled_output" type="color3" nodename="hextiled" />
</materialx>
```

main:
main
pr-3001:
pr-3001

tdavidovicNV:
The second (sorry for double posting, they are quite different) is in OSL with UDIMs:

This shader:
```
<?xml version="1.0"?>
<materialx version="1.39">
  <image name="udim_image" type="color3">
    <input name="file" type="filename" value="tile.&lt;UDIM&gt;.png" colorspace="lin_rec709" />
    <input name="uaddressmode" type="string" value="periodic" />
    <input name="vaddressmode" type="string" value="periodic" />
    <input name="filtertype" type="string" value="closest" />
  </image>
  <output name="udim_output" type="color3" nodename="udim_image" />
</materialx>
```

where `tile<UDIM>.png` is just a bunch of differently colored solid color images.

Used on this .obj:
```
# Canonical lower-left UVs spanning six UDIMs:
# bottom: 1001, 1002, 1003; top: 1011, 1012, 1013.
v -1.5 -1.0 0.0
v  1.5 -1.0 0.0
v  1.5  1.0 0.0
v -1.5  1.0 0.0

vt 0.0 0.0
vt 3.0 0.0
vt 3.0 2.0
vt 0.0 2.0

vn 0.0 0.0 1.0

f 1/1/1 2/2/1 3/3/1
f 1/1/1 3/3/1 4/4/1
```
(so U from 0-3 and V from 0-2) gets me this:

main:
main
pr-3001:
pr-3001

The main reason it that, previously, MaterialX passed the full UV coordinates unchanged to OSL/OpenImageIO, which resolved the UDIM before interpreting pixels using its top-left image convention. After this PR, MaterialX v = 1-v before the texture lookup, so it also changes the coordinate used for UDIM selection. A two-row range [0,2] becomes [-1,1], causing the second row to resolve back to UDIM row zero. 

To make things work for UDIM, you cannot just do `v=1-v` flip, but have to do `v = floor(v) + 1 - fract(v)`.. but then *that* breaks warp and mirror modes on tiled textures.

I do not think there is a good solution that can support both. What we do in Falcor2 is that we decide one main mode in which Scene works (upper/lower left), that drives the *actual* texture lookup (after UDIM resolution), and convert any non-compliant geometry using `v=1-v`. Mostly because our standard is the upper-left image convention, which is what most VFX does. While the lower-left is mostly for games (e.g., glTF), which rarely-if-ever use UDIMs, so doing the `1-v` is safe on the assumption that breaking UDIM tile IDs doesn't matter.

jstone-lucasfilm:
Thanks for these two counter-examples, @tdavidovicNV, and since they highlight quite different behaviors, as you note, I'll reply to each in turn.

The `hextiledimage` difference falls into the expected categories described earlier in this thread.  The hex-tiling machinery sits between the texture coordinate input and its file lookups, deriving randomized rotations, scales, and offsets from the transformed coordinate, so the tile pattern shifts when the coordinates feeding the shader change conventions.  Our render tests previously passed upper-left-origin coordinates to generated shaders, while shaders now receive coordinates in the MaterialX convention, with the origin at the lower left.  As with the earlier examples, generated code is unchanged for any given set of generator options, so integrations that provide their own geometry and options render identically before and after this change.

One deeper limitation is worth acknowledging here: since the vertical compensation is applied upstream of the hex-grid hashing, targets that flip file texture lookups (GLSL, MSL, Slang, OSL) compute their hex-tile patterns in flipped space, while targets that leave them unflipped (MDL) compute them in MaterialX space, so hex-tile patterns don't align exactly across the two groups.  This limitation predates the changelist, and a natural follow-up would be to move the compensation in the hex-tiling nodes to their individual file lookups, aligning pattern placement across all targets.

The UDIM example, on the other hand, is a great catch, and I agree that a uniform flip breaks UDIM tile resolution in OSL, which is unique among our generation targets in resolving UDIM filenames at lookup time.  This is fixed in the latest version of the changelist, where the image node implementations in OSL flip the V coordinate within each tile when a `<UDIM>` token is present in the filename, following exactly the formulation you propose, and apply a uniform flip otherwise.

I believe this conditional form resolves the tension between the two flips.  Conventional textures retain the uniform flip, preserving the behavior of clamp and mirror address modes outside the unit square, while UDIM filenames preserve the tile indices used in resolution, with the two formulations agreeing inside the unit square.  The filename test is expressed as an OSL `regex_search`, which is resolved at optimization time for constant filenames, adding no per-sample cost in typical renders.

On the broader design, MaterialX takes the same approach that you describe in Falcor: a single canonical convention for texture space, with its origin at the lower left, and geometry converted to that convention at import time.  The per-tile flip is the piece that lets OSL's lookup-time UDIM resolution coexist with this convention.

If you have the chance, I'd welcome a re-run of your six-tile example against the updated branch, since our own render test geometry remains within the unit square, where the two formulations coincide.

libraries/stdlib/genosl/lib/mx_transform_uv_vflip.osl
```diff
@@ -2,3 +2,15 @@ vector2 mx_transform_uv(vector2 texcoord)
 {
     return vector2(texcoord.x, 1.0 - texcoord.y);
 }
+
+vector2 mx_transform_uv(vector2 texcoord, string filename)
+{
+    if (regex_search(filename, "<UDIM>"))
```
ld-kerley, line 8: Is this safe for GPU OSL support? Maybe it all gets constant folded during optimization - but its not always easy to be sure of this without testing. 

Also is this not going to break existing peoples MaterialX materials?  this is a change in behavior.  Feels safer to save for 1.40 or 2.0.0?

source/MaterialXGenOsl/LibsToOso.cpp
```diff
@@ -371,12 +371,13 @@ int main(int argc, char* const argv[])
     // Register types from the libraries on the OSL shader generator.
     oslShaderGen->registerTypeDefs(librariesDoc);
 
-    // Setup the context of the OSL shader generator.
+    // Setup the context of the OSL shader generator. File texture lookups are
+    // generated with a vertical flip, matching the reference OSL shaders and
+    // compensating for the top-left image origin of OSL texture lookups.
     mx::GenContext context(oslShaderGen);
     context.registerSourceCodeSearchPath(argSearchPath);
-    // TODO: It might be good to find a way to not hardcode these options, especially the texture flip.
```
ld-kerley, line 377: I think we need to keep this TODO - and figure out a way to compile the OSOs for flips both ways around at build time.  OSL/OIIO can be configured to use either orientation - so we need the decision to be dynamic during the shader generator - not hard-coded during the build.

source/MaterialXRender/CgltfLoader.h
```diff
@@ -32,7 +32,7 @@ class MX_RENDER_API CgltfLoader : public GeometryLoader
     static CgltfLoaderPtr create() { return std::make_shared<CgltfLoader>(); }
 
     /// Load geometry from file path
-    bool load(const FilePath& filePath, MeshList& meshList, bool texcoordVerticalFlip = false) override;
+    bool load(const FilePath& filePath, MeshList& meshList) override;
```
ld-kerley, line 35: Isn't this an API breaking change - so we should hold it until 1.40 or 2.0.0?

libraries/stdlib/genosl/lib/mx_transform_uv_vflip.osl
jstone-lucasfilm: These are both good questions, @ld-kerley, and the two-argument form of `regex_search` was chosen with GPU environments in mind, since this is the form that OSL constant folds when its subject and pattern are both constant strings (see `constfold_regex_search` in liboslexec).  Because texture filenames are uniform shader parameters whose instance values are known at specialization time, the UDIM branch should fold to a constant before any GPU code is generated, though I agree that testing in GPU implementations of OSL is worthwhile.  If any issues show up in these environments, a straightforward fallback would be to select the UDIM-aware transform at shader generation time instead.

On the question of backwards compatibility, it's worth noting that no documents, node definitions, or generator option defaults are modified in this changelist, so downstream integrations that leave these options untouched will see identical generated code.  The changes in rendered results are confined to our own tests and tools, where they align our renders with the texture space convention that the specification has always defined, and for clients that enable `fileTextureVerticalFlip` in OSL, the new UDIM-aware transform fixes lookups that previously resolved to incorrect tiles.

source/MaterialXGenOsl/LibsToOso.cpp
jstone-lucasfilm: That's a good point, and since OSL and OIIO hosts can be configured with either image orientation, no single baked choice is right for all consumers of the precompiled modules.  In the latest version of this changelist, the orientation of file texture lookups is controlled by a new `--texVerticalFlip` option on the `LibsToOso` tool, defaulting to true to match standard OIIO configurations.  The TODO has been maintained and slightly updated, noting the future possibility of generating modules for both orientations so that the choice could be made dynamically at network generation time.

source/MaterialXRender/CgltfLoader.h
jstone-lucasfilm: That's a fair concern, and my sense is that this change falls within the incremental API evolution that we've historically included in point releases, such as the removal of blur node helpers in v1.39.4 and the NodeGraph dependency of shader translation in v1.39.5.  Since the default behaviors of `GeometryLoader::load` and `GeometryHandler::loadGeometry` are preserved, clients calling these methods without the `texcoordVerticalFlip` argument will receive exactly the same meshes as before, and clients that passed the argument explicitly can now apply their preferred convention downstream of the loader.

Additionally, the removed argument had proven to be a source of developer confusion, as its effect was inverted between our two geometry loaders, flipping OBJ texture coordinates when true but glTF texture coordinates when false.  Removing it establishes a single clear contract, where loaders always return texture coordinates in the MaterialX convention, and we'll make sure this improvement is clearly called out in the release notes for v1.39.6.

libraries/stdlib/genosl/lib/mx_transform_uv_vflip.osl
tdavidovicNV, line 8: MaterialX also allows `<UVTILE>` token (https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#filename-substitutions), so that one should be checked as well.

tdavidovicNV:
> If you have the chance, I'd welcome a re-run of your six-tile example against the updated branch, since our own render test geometry remains within the unit square, where the two formulations coincide.

I reran the six-tile OSL/UDIM example against the current PR head (8656913c) and can confirm the UDIM tile-selection issue is fixed. It now matches the main exactly.
result

jreichel-nvidia:
MDL SDK 2026.0.2 has been released (https://github.com/NVIDIA/MDL-SDK/releases/tag/2026.0.2). This release adds support for the `--uv_scale` flag which is used by this MR. The MR for vcpkg is pending (https://github.com/microsoft/vcpkg/pull/53708).

jstone-lucasfilm:
Thanks for your latest notes, @tdavidovicNV and @jreichel-nvidia, and I've made the suggested updates for the `<UVTILE>` token and the latest version of the MDL SDK.  We don't yet provide MDL render tests in our GitHub CI, but when we ultimately take that step, we'll be using a version of the MDL SDK that supports the new `--uv_scale` flag.

I'm very open to additional requests for improvements, if you see any issues that haven't been addressed, and my own sense is that this PR is in good shape.

libraries/stdlib/genosl/lib/mx_transform_uv_vflip.osl
fpsunflower: Also worth mentioning that OIIO is able to perform this operation "natively" (including doing the right thing on udims without a string search). When an OSL based renderer uses OIIO as its backend, it can opt to change the default orientation of textures. You can try this in testrender by passing:
```--texoptions "flip_t=1"```

This does raise the question of what MaterialX should assume by default though. I assume that Arnold and Prman both use the older "upper-left" convention out of the box?

fpsunflower: Reading a bit more, it looks like you have some provision for a boolean `fileTextureVerticalFlip` control on the code generator, but it doesn't seem to be used by the OSL code generation. Am I misunderstanding the purpose of that boolean?

There is still the problem that if OSL can be configured either way (indeed, its really up to the underlying texture system, the language itself has no visibiltiy into the texture's pixels) then it isn't really possible to write a shader portable across renderers.

@lgritz Do you remember where we landed on this when it was discussed in the past? The OSL spec doesn't say explicitly which way the texture's pixels should go (ie: which corner (s=0,t=0) maps to).

PR #2999, opened 2026-07-04
Add support for nested NodeGraphs : Core Traversal, Validation and ShaderGen

kwokcb:
## Issue

The 1.39 specification notes that nodegraphs may be nested but the the current API and traversal mechanism implementation does not reflect this.

See proposal [here](https://github.com/AcademySoftwareFoundation/MaterialX/issues/1272)

## Core Changes

- Move the nodegraph parenting API from `Document` to `GraphElement` thus allow for parenting from the derived class `NodeGraph`
- Update traversal and connection logic to make connections to nested nodegraphs be supported.
- Updated core API. Should be no API compatibility issues as the interfaces are still available at `Document` level, but are now available at `NodeGraph` level.
- Update validation logic to recognize nested interface connections.

### ShaderGen Changes (None)

- **None** are required as the logic supports flattening via traversal logic. Once traversal
logic is updated nested graphs get flattened as needed. Note this is similar to how functional
graphs for definitions get flattened.

## Graph Editor Changes (Test)
- **Note** these are only changes for testing workflow and visualization and not finalized.
- Generalize logic to support load and save of nested graphs. 
- Fix logic for for currently supported non-nested graph <-> graph /  node <-> graph connections 
- Layer on connection fixes for nesting.

### Example Workflow
- Shows current usage of information hiding by showing nodegraphs at any level as UI nodes.
- These nodes can be expanded to show only their contents in much the same way nodes with functional nodegraphs can be expanded. Traversal in and out of nesting uses the same mechanism.

https://github.com/user-attachments/assets/0e408d90-04d4-4f9b-a233-3f67ca8412a8

## Tests
- New unit tests added to test nested graph creation traversal and validation.

```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <nodegraph name="level1_graph">
    <nodegraph name="level2_graph">
      <nodegraph name="level3_graph">
        <input name="level3_input" type="color3" interfacename="level2_input2" />
        <output name="level3_output" type="color3" nodename="invert_color3" />
        <invert name="invert_color3" type="color3">
          <input name="in" type="color3" interfacename="level3_input" />
        </invert>
      </nodegraph>
      <input name="level2_input" type="color3" interfacename="level1_input" />
      <input name="level2_input2" type="color3" interfacename="level1_input2" />
      <output name="level2_output" type="color3" nodename="color3" />
      <add name="color3" type="color3">
        <input name="in1" type="color3" interfacename="level2_input2" />
[...]

PR #3002, opened 2026-07-08
Add hotkey to group nodes into a node graph

### LINKED ISSUE #2018: Group nodes to create node graph in the Graph Editor

minhpham1810:
## Summary

Moves the Graph Editor's node-grouping logic into a reusable, unit-tested core method, per maintainer feedback. The editor now delegates to core, grouping stays top-level-only, and the Shift+C hotkey remains a thin trigger.

## Changes

- **`source/MaterialXCore/Node.h`** — Declare new `GraphElement::createNodeGraphFromNodes(nodes, name)`, which groups the given nodes into a new child nodegraph while preserving internal and boundary connections.
- **`source/MaterialXCore/Node.cpp`** — Implement `createNodeGraphFromNodes` plus file-local helpers (ported from the editor) that copy the selected nodes into a new nodegraph, turn external upstream links into nodegraph interface inputs, rewire external downstream consumers to nodegraph outputs (with fan-out dedup), and remove the originals. The method is general and does not set layout attributes or enforce a document-only policy.
- **`source/MaterialXGraphEditor/Graph.cpp`** — Remove the in-editor grouping helpers and rewrite `groupSelectedNodesIntoNodeGraph` to collect the selected nodes and delegate to the core method, keeping only editor concerns (top-level guard, average node position, graph rebuild/relink/render).
- **`source/MaterialXTest/MaterialXCore/Node.cpp`** — Add `[nodegraph]` unit tests covering internal links, upstream/downstream boundary rewiring, fan-out output dedup, original removal, document validation, and the empty-selection case.
- **`source/MaterialXCore/Interface.cpp`** — Remove a redundant duplicate `resolveNameReference<NodeGraph>` no-op in `PortElement::validate` left over from the reverted nesting rework.

## Testing

- `git diff --check`
- `MaterialXTest "Graph editor grouping shortcut"`
- `cmake --build build --target MaterialXGraphEditor`

## Related Issue

Fixes #2018


jstone-lucasfilm:
This looks very promising, thanks @minhpham1810, and take a look at the [comment above](https://github.com/AcademySoftwareFoundation/MaterialX/pull/3002#issuecomment-4919969528]) for instructions on resolving the EasyCLA errors.

minhpham1810:
@jstone-lucasfilm I just signed the CLA and changed my commit email, please have a look.

jstone-lucasfilm:
Thanks for the excellent work here, @minhpham1810.  The connection-preservation logic is more thorough than I'd expected, and the hard parts are done well.

This PR has grown into something larger than the summary suggests, though, so before we dive into code review I'd like to align on scope.  Two recommended next steps would be:

1. Update the PR description to reflect the full change.  It currently reads as a Shift+C shortcut fix, but it implements the whole grouping feature along with some MaterialXCore changes.
2. The key decision is nested nodegraphs.  When this was first scoped, @kwokcb suggested the editor guard against nesting; this PR adds support for it instead, which is what motivates the Interface.cpp changes.  I'd suggest we start with grouping at the top document level only, which would cover the original use case, keep us on well-supported ground, and require no MaterialXCore changes.  We can then take up nested nodegraphs (and the resolution work they require) as a deliberate follow-up.

Once we've settled the scope, I'm glad to go deep on the grouping logic itself.  Thanks again for the strong contribution!

minhpham1810:
@jstone-lucasfilm Thank you so much for the detailed feedback! I admitted that I want a bit out of scope in this PR, so I reverted the changes I made in `MaterialXCore`, as well as editing the safeguard for top-level document node grouping. Please let me know if there is any other parts that I should look at.

kwokcb:
I would agree that to only allow top level / document level grouping to occur.

( For reference, I did rework to support graph nesting and fix a other related issues in this [PR ](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2999). )

BTW, there was another [PR ](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2048 ) proposal which took advantage of the cut/paste mechanism. 

From a cursory look and revisit of the previous PR, my suggested approach is to 
1. Put new "grouping" functionality in core so it's a reusable and testable function. 
2. Then expose it in editor.
3. Then add the hotkey as a minor addition.

minhpham1810:
Thank you for the constructive review and feedback. From what we've been discussing so far, I will proceed with what @kwokcb suggested in the comment above. Please let me know if there is any further instructions or directions that you want to highlight @kwokcb @jstone-lucasfilm .

minhpham1810:
@jstone-lucasfilm @kwokcb I just push the changes that relocate grouping logic into core. Please have a look.

jstone-lucasfilm:
Thanks for the revisions, @minhpham1810!

This is looking very promising, but I see two important issues to resolve before we dive into line-level review:

1. Several editor changes from the original nested-nodegraph implementation are still present in the diff, and together they reintroduce the nesting we agreed to defer.  The "Disallow creating nested nodegraphs" guard in `Graph::addNodePopup` is removed, and `Graph::addNode` and `Graph::copyUiNode` now create nodegraphs under `_state.graphElem` rather than the document, so the TAB menu inside a compound nodegraph creates a nested nodegraph even though the Shift+C path is guarded.  The child-nodegraph handling added to `Graph::buildUiNodeGraph` and the nested walk in `Graph::scanNestedGraphDiagnostics` are part of the same earlier work, and they overlap with #2999, where nesting support is being developed as a whole.  I'd suggest reverting these and limiting the editor diff to `groupSelectedNodesIntoNodeGraph`, `rebuildCurrentGraph`, the hotkey, and the help text, which keeps this changelist on well-supported ground.

2. One of the merges from `main` appears to have reverted part of #3012, which introduced the `graphShortcutContext` guard for keyboard input in the graph area.  The copy/paste block retains the guard, but the delete/traverse block in `Graph::drawGraph` is back to a bare `ImGui::IsWindowFocused` check, so Delete, Backspace, the arrow keys, F, and U would once again fire while a popup or file dialog is open, or while a text field has focus.  Restoring `if (graphShortcutContext)` for that block should resolve it.  Since the new `GraphShortcuts.h` predicate duplicates this same guard, I'd also suggest expressing the Shift+C check inline alongside the other hotkeys, which would let us drop the header and the new test directory.

With these two broader items addressed, we should be able to dive into deeper review, and this looks like it will be a welcome addition to the Graph Editor.

PR #3005, opened 2026-07-14
Add XYZ-to-working-space transform for hardware shaders

### LINKED ISSUE #2859: Color-space awareness in thin-film iridescence and blackbody emission
As @mnikelsky pointed out in a recent thread on the ASWF Slack, the [XYZ_TO_RGB](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl#L318) matrix in our GLSL thin-film codebase is taken directly from the original [Belcour paper](https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html), but this matrix transforms colors to _CIE 1931 RGB_ space, which is not one of the linear rendering spaces supported by MaterialX or commonly used in film and game rendering.

The [XYZ_TO_RGB](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genglsl/mx_blackbody.glsl) matrix in our GLSL implementation of blackbody emission is slightly closer to the mark, transforming to the common `lin_rec709` space, but this matrix is still hardcoded and incapable of adapting to the linear working spaces of different rendering environments.

To address both of these issues, we could add a new [shader generation option](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/source/MaterialXGenShader/GenOptions.h#L75) named `xyzToWorkingSpace`, representing the matrix33 that transforms from CIE XYZ to the working color space of the renderer.

By default this option would store the transform from CIE XYZ to `lin_rec709`, but the renderer would have the ability to set another matrix when needed, and because MaterialX shader generators already have access to the working color space, they could store the correct matrix automatically.

We would then swap in this uniform matrix for both of the hardcoded matrices in the GLSL implementations of thin-film iridescence and the blackbody emission node, allowing them to support arbitrary working color spaces in a simple, efficient way.

r800360:
Fixes #2859 

## Summary

* Adds `GenOptions::xyzToWorkingSpace`, defaulting to the CIE XYZ to `lin_rec709` transform.
* Exposes the transform as the private hardware-shader uniform `u_xyzToWorkingSpace`.
* Uses the uniform for blackbody emission and thin-film iridescence.
* Removes the two hardcoded XYZ-to-RGB matrices.
* Adds a focused GLSL shader-generation test covering the default transform, custom overrides, private-uniform propagation, token substitution, and both shader consumers.

## Testing

* `MaterialXTest "GenShader: GLSL XYZ to Working Space"`
* Full `MaterialXTest` suite
* 65 test cases and 14,579 assertions passed


jstone-lucasfilm:
Thanks for this contribution, @r800360, and welcome to the project!  This is a clean implementation of the proposal in #2859, following the same pattern as the existing `u_envMatrix` and `u_shadowMatrix` uniforms in `HwShaderGenerator::createShader`, and it looks like the row-major `Matrix33` round-trips to the same GL and Metal matrix as the original `mat3` literal in `mx_blackbody.glsl`.

One substantive concern is the _white point_ of the thin-film reflectance.  The CIE 1931 RGB matrix from the Belcour paper has an equal-energy white point, and since `mx_eval_sensitivity` is normalized so that a flat reflectance spectrum integrates to XYZ = (1, 1, 1), the original matrix maps a flat spectrum to neutral.  A raw XYZ-to-Rec.709 matrix maps the same XYZ to roughly (1.205, 0.948, 0.909), and because the m=0 terms in `mx_fresnel_airy` are flat-spectrum contributions, this would place a visible warm cast on the base reflectance of every thin-film surface.  Blackbody emission, in contrast, is producing an emitter color, so it seems right for that node to use the unadapted transform.

My current thinking is that the simplest robust fix is to normalize in the shader, so that equal-energy white maps to neutral for any matrix the integrator supplies:

```glsl
// Convert back to RGB reflectance, normalizing so that equal-energy white maps to neutral
I = mx_matrix_mul($xyzToWorkingSpace, I) / mx_matrix_mul($xyzToWorkingSpace, vec3(1.0));
I = clamp(I, 0.0, 1.0);
```

@doug-walker, I'd value your perspective before we commit to an approach, so that we're not relying solely on our own reading of the color science.  Is a per-channel normalization of this kind a reasonable adaptation for reflectance data, or would you recommend a proper chromatic adaptation transform such as Bradford?  And do you agree that blackbody emission should remain unadapted while thin-film reflectance is adapted?

Either way, the `thin_film_bsdf` render references will shift, since the previous code interpreted CIE RGB values as Rec.709, and I'd like to review the before-and-after renders once the white point is settled.

A few smaller notes:

1. `SlangShaderGenerator.cpp` still contains the `const float3x3 XYZ_to_RGB` token replacement, which no longer matches anything and can be removed.

2. The new option should be added to the Python and JavaScript bindings in `PyGenOptions.cpp` and `JsGenOptions.cpp`, alongside the other `GenOptions` fields.

3. The whitespace changes in `HwShaderGenerator::toVec4` and `GenGlsl.cpp` are unrelated to this change, so I'd suggest dropping them to keep the diff focused.

4. In the new test, the checks on the uniform block are the valuable part, while the checks on literal source substrings are tied to variable names in the GLSL and will break on unrelated refactors, and the comparison of `defaultOptions.xyzToWorkingSpace` against `expectedDefault` compares the same literal with itself.  I'd suggest trimming these.

Finally, since this changelist leaves the option as a fixed `lin_rec709` default, an application that already sets `targetColorSpaceOverride` to `acescg` would receive correct `cmlib` transforms everywhere except blackbody and thin-film.  Given that MaterialX supports a small, closed set of linear working spaces, it's worth considering whether the default should derive from the target color space in `createShader`, with `GenOptions::xyzToWorkingSpace` serving as an explicit override.  I'd be interested in your thoughts on this, and it could reasonably be handled as a follow-up.  Overall this is a valuable contribution, and your work on it is much appreciated!

libraries/pbrlib/genglsl/lib/mx_microfacet_specular.glsl
```diff
@@ -302,9 +302,6 @@ vec3 mx_eval_sensitivity(float opd, vec3 shift)
 // https://belcour.github.io/blog/research/publication/2017/05/01/brdf-thin-film.html
 vec3 mx_fresnel_airy(float cosTheta, FresnelData fd)
 {
-    // XYZ to CIE 1931 RGB color space (using neutral E illuminant)
-    const mat3 XYZ_TO_RGB = mat3(2.3706743, -0.5138850, 0.0052982, -0.9000405, 1.4253036, -0.0146949, -0.4706338, 0.0885814, 1.0093968);
-
```
doug-walker, line 307: Replace with:
// Use Bradford chromatic adaptation to convert the white point from illuminant E,
// used by the Belcour paper, to illuminant D65, expected by the xyzToWorkingSpace
// matrix. The matrix converts {1., 1., 1.} to {0.95046, 1., 1.08906}.  (Note that the mat3
// constructor expects a column-major vector.)
const mat3 E_TO_D65 = mat3(0.95314737, -0.03827266,  0.00261539, -0.02661088,  1.02885095, -0.00304736, 0.02391944,  0.00942171,  1.08948971);

Then keep the matrix multiply below, replacing XYZ_TO_RGB with E_TO_D65.

source/MaterialXGenHw/HwConstants.h
```diff
@@ -67,6 +67,8 @@ Uniform variables :
     $lightData[]                        u_lightData[]                       struct     Array of struct LightData holding parameters for active light sources.
                                                                                        The LightData struct is built dynamically depending on requirements for
                                                                                        bound light shaders.
+    $xyzToWorkingSpace                  u_xyzToWorkingSpace                 mat3       Transform from CIE XYZ to the renderer's linear working color space,
+                                                                                       initialized from GenOptions.xyzToWorkingSpace.
```
doug-walker, line 71: Replace "CIE XYZ" with "CIE XYZ D65".

source/MaterialXGenShader/GenOptions.h
```diff
@@ -78,6 +80,10 @@ class MX_GENSHADER_API GenOptions
     GenOptions() :
         shaderInterfaceType(SHADER_INTERFACE_COMPLETE),
         fileTextureVerticalFlip(false),
+        xyzToWorkingSpace(
+            3.2406f, -0.9689f, 0.0557f,
+            -1.5372f, 1.8758f, -0.2040f,
+            -0.4986f, 0.0415f, 1.0570f),
```
doug-walker, line 86: Five digits is not enough to fully initialize a float (needs 6-7). Here are some more digits:
3.240969941905, -0.969243636281, 0.055630079697, -1.53738317757, 1.875967501508, -0.203976958889, -0.498610760293, 0.041555057407, 1.056971514243

source/MaterialXGenShader/GenOptions.h
```diff
@@ -117,6 +123,10 @@ class MX_GENSHADER_API GenOptions
     /// texture space convention. By default this option is false.
     bool fileTextureVerticalFlip;
 
+    /// Transform from CIE XYZ to the renderer's linear working color space.
+    /// Defaults to the CIE XYZ to lin_rec709 transform.
+    Matrix33 xyzToWorkingSpace;
```
doug-walker, line 128: Suggest:
// Transform CIE XYZ values, with a D65 white point, to the renderer's linear working color space.

CIE XYZ does not have a fixed white point. In order for this matrix substitution technique to work, the white point on the XYZ side must be specified.

doug-walker (commented):
A few minor suggestions. 

For the record, the original Slack thread was here:
https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1775741600062129

PR #3009, opened 2026-07-21
Fix missing null input checks in MDL shader generation

jreichel-nvidia:


jstone-lucasfilm (approved):

PR #3012, opened 2026-07-23
Fix graph area keyboard input in Graph Editor

### LINKED ISSUE #3010: MaterialX Graph Editor Node Deletion is Broken
The code checks for:

`if (ImGui::IsWindowFocused(ImGuiFocusedFlags_RootWindow))`

Which is only true after clicking on the enclosing outer pane which can only be done by clicking on the tab bar.
Once clicking anywhere within the actual main graph pane, it is no longer true so all the code in the if block is ignored (which includes the node deletion handling).

Image

kwokcb:
## Issue

Fixes: https://github.com/AcademySoftwareFoundation/MaterialX/issues/3010

- Graph area parenting was added without updating existing event logic.

## Fixes
- Consider child widgets, and add logic for current / active widget focus checking.
- Add firewall checks to make sure that property editor (another child) and menu (at root) events are filtered properly.
- Copy / Paste had state handling issues so also fixed this.

### Testing
- All key input tested manually

### DIFF
--- a/source/MaterialXGraphEditor/Graph.cpp
+++ b/source/MaterialXGraphEditor/Graph.cpp
@@ -3148,7 +3148,11 @@ void Graph::graphButtons()
 
     // Menu keys
     ImGuiIO& guiIO = ImGui::GetIO();
-    if (guiIO.KeyCtrl && !_fileDialogSave.isOpened() && !_fileDialog.isOpened() && !_fileDialogGeom.isOpened())
+    if (guiIO.KeyCtrl &&
+        !guiIO.WantTextInput &&
+        !_fileDialogSave.isOpened() &&
+        !_fileDialog.isOpened() &&
+        !_fileDialogGeom.isOpened())
     {
         if (ImGui::IsKeyReleased(ImGuiKey_O))
         {
@@ -3915,7 +3919,10 @@ void Graph::drawHelpMarker(const char* content)
 
 void Graph::addNodePopup(bool cursor)
 {
-    bool open_AddPopup = (ImGui::IsWindowFocused(ImGuiFocusedFlags_RootAndChildWindows) && ImGui::IsKeyReleased(ImGuiKey_Tab)) ||
+    ImGuiIO& io = ImGui::GetIO();
+    bool open_AddPopup = (ImGui::IsWindowFocused(ImGuiFocusedFlags_RootAndChildWindows) &&
+                          !io.WantTextInput &&
+                          ImGui::IsKeyReleased(ImGuiKey_Tab)) ||
                          (_pinFilterType != mx::EMPTY_STRING && ImGui::IsMouseReleased(0));
     static char input[32]{ "" };
     if (open_AddPopup)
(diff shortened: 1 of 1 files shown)


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3148,7 +3148,11 @@ void Graph::graphButtons()
 
     // Menu keys
     ImGuiIO& guiIO = ImGui::GetIO();
-    if (guiIO.KeyCtrl && !_fileDialogSave.isOpened() && !_fileDialog.isOpened() && !_fileDialogGeom.isOpened())
+    if (guiIO.KeyCtrl &&
+        !guiIO.WantTextInput &&
```
kwokcb, line 3152: `WantTextInput()` added in various places to avoid text input / graph hotkey collisions. Used along with focus tests.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4039,7 +4046,11 @@ void Graph::addNodePopup(bool cursor)
 
 void Graph::searchNodePopup(bool cursor)
 {
-    const bool open_search = ImGui::IsWindowFocused(ImGuiFocusedFlags_RootAndChildWindows) && ImGui::IsKeyDown(ImGuiKey_F) && ImGui::IsKeyDown(ImGuiKey_LeftCtrl);
+    const ImGuiIO& io = ImGui::GetIO();
+    const bool open_search = ImGui::IsWindowFocused(ImGuiFocusedFlags_RootAndChildWindows) &&
```
kwokcb, line 4050: Must consider children since graph editor and property editor are now children not directly at root.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4295,10 +4306,18 @@ void Graph::drawGraph(ImVec2 mousePos)
...
         selectedLinks.resize(linkCount);
-        if (io2.KeyCtrl && io2.MouseDown[0])
-        {
-            _ctrlClick = true;
-        }
+
+        // Keep ctrl-click state frame-local; stale true blocks connection copy in paste.
+        _ctrlClick = io2.KeyCtrl && io2.MouseDown[0];
+
+        const bool graphShortcutContext =
```
kwokcb, line 4313: Robust hotkey filtering checking for graphing area.

HardCoreCodin:
I can confirm this solves my issue #3010

kwokcb:
@lfl-eholthouser Would you have some time to review this? 
It would be good to get hotkeys working again :) Thanks.
( @HardCoreCodin has independently verified the fix. )

--
This does not include automated event testing but  imgui has mechanisms designed for this  kind of thing as noted in [issue](https://github.com/AcademySoftwareFoundation/MaterialX/issues/3013)

jstone-lucasfilm (approved):
This looks like a good improvement to me, thanks @kwokcb!

The consolidated focus check seems like the right direction for keyboard handling in the Graph Editor, and it would be worth considering a follow-up that shares this logic with `graphButtons`, `addNodePopup`, and `searchNodePopup` as well.

PR #3016, opened 2026-07-27
Fix texture flipping for hextiledimage and hextilednormalmap.

jreichel-nvidia:
Add new mxp_flip_v parameter to forward the generator option for texture flipping.


jstone-lucasfilm:
This looks like a great improvement to me, @jreichel-nvidia, and I'd ideally like to merge #3001 before reviewing and merging this complementary work, so that we can leverage the new alignment between specification and codebase that it brings.

PR #3017, opened 2026-07-27
Define layer pass-through semantics

tdavidovicNV:
While reviewing the `layer` equation added in https://github.com/AcademySoftwareFoundation/MaterialX/pull/2964, I found that deriving base attenuation as `1 - E_top` conflates two different outcomes: energy absorbed by the top layer and energy passed toward the base.

This is visible with a colored reflection-only dielectric. For untinted Fresnel albedo `F = 0.04` and tint `c = (1, 0.25, 0.05)`, the complete top albedo is:

```text
E_top = c * F = (0.04, 0.01, 0.002)
```

The current specification therefore gives base attenuation `(0.96, 0.99, 0.998)`. The existing GLSL and OSL `testrender` implementations instead use the untinted unreflected share `(1 - F) = (0.96, 0.96, 0.96)`, treating the tint-removed energy as absorption rather than returning it to the base.

This PR introduces an explicit color-valued pass-through factor that may depend on both incident and outgoing directions:

```math
P_{\mathrm{top}}(\omega_i, \omega_o) \in [0, 1]
```

```math
f(\omega_i, \omega_o) =
f_{\mathrm{top}}(\omega_i, \omega_o) +
P_{\mathrm{top}}(\omega_i, \omega_o)
f_{\mathrm{base}}(\omega_i, \omega_o)
```

This follows the two-leg construction in [Weidlich and Wilkie 2007](https://www.cg.tuwien.ac.at/research/publications/2007/weidlich_2007_almfs/weidlich_2007_almfs-paper.pdf), section 3.2, where the lower-layer BRDF is attenuated by entry transmission, absorption over the incident and outgoing paths, and return transmission.

For nested layers, the pass-through factors multiply recursively. The PR also separates BSDF-over-BSDF composition from a BSDF bound to a VDF: a lossless transmissive interface has complete directional albedo near one, but that should not prevent transmitted light from entering the medium.

The current implementations are not changed:

- **GLSL:** Approximates pass-through using only the outgoing direction, stores it in [`BSDF.throughput`](https://github.com/AcademySoftwareFoundation/MaterialX/blob/ebc8cc8d89c056ba7d7bde0d6ef49bdc417e7de7/libraries/pbrlib/genglsl/mx_dielectric_bsdf.glsl#L13-L68), and multiplies it in [`layer`](https://github.com/AcademySoftwareFoundation/MaterialX/blob/ebc8cc8d89c056ba7d7bde0d6ef49bdc417e7de7/libraries/pbrlib/genglsl/mx_layer_bsdf.glsl#L1-L7). Reflection-only dielectric and sheen keep artistic color out of pass-through, while generalized Schlick reduces its color-valued albedo to an RGB average.
- **OSL (`testrender`):** Also approximates pass-through using only the outgoing direction. It [computes top opacity from `filter_o`](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/f7ab6a7a5074e530d3ef9e107745a193cc811d90/src/testrender/shading.cpp#L1198-L1281) and [scales the base by one minus that value](https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/blob/f7ab6a7a5074e530d3ef9e107745a193cc811d90/src/testrender/shading.cpp#L1645-L1658). Dielectric uses untinted pass-through, sheen uses a scalar estimate based on the maximum color component, and generalized Schlick uses a scalar energy estimate.
- **MDL:** Rewrites a [compatible subset of layer graphs](https://github.com/AcademySoftwareFoundation/MaterialX/blob/ebc8cc8d89c056ba7d7bde0d6ef49bdc417e7de7/source/MaterialXGenMdl/Nodes/ClosureLayerNodeMdl.cpp#L113-L190) into native layerers. Its curve layerers [evaluate both directions and use a symmetric maximum](https://github.com/NVIDIA/MDL-SDK/blob/ebca383a932c94cdc9c965c283658637f251a587/src/mdl/jit/libbsdf/libbsdf.cpp#L6540-L6684). MaterialX sheen uses a scalar weighted-layer approximation based on average color. Unsupported top nodes cannot retain the base.

### DIFF
--- a/documents/Specification/MaterialX.PBRSpec.md
+++ b/documents/Specification/MaterialX.PBRSpec.md
@@ -344,7 +344,7 @@ E_o = \int_{\Omega_i} f(\omega_i, \omega_o) \cos\theta_i \; d\omega_i
 ```
 <p></p>
 
-The directional albedo is the quantity referenced by the [&lt;layer>](#node-layer) node when performing albedo-scaled vertical layering of a top BSDF over a base.
+Directional albedo measures energy scattered by a BSDF, but it does not distinguish energy reflected away from a layer stack, energy transmitted toward the next layer, and energy absorbed by the top layer. The [&lt;layer>](#node-layer) node therefore uses a separate pass-through quantity supplied by each layerable BSDF.
 
 #### Energy Compensation
 
@@ -1098,7 +1098,7 @@ Because $w \in [0, 1]$, the mix of two individually energy-conserving distributi
 <a id="node-layer"> </a>
 
 ### `layer`
-Vertically layer a layerable BSDF such as [&lt;dielectric_bsdf>](#node-dielectric-bsdf), [&lt;generalized_schlick_bsdf>](#node-generalized-schlick-bsdf) or [&lt;sheen_bsdf>](#node-sheen-bsdf) over a BSDF or VDF. The implementation is target specific, but a standard way of handling this is by albedo scaling, using the function "base*(1-reflectance(top)) + top", where the reflectance function calculates the directional albedo of a given BSDF.
+Vertically layer a layerable BSDF such as [&lt;dielectric_bsdf>](#node-dielectric-bsdf), [&lt;generalized_schlick_bsdf>](#node-generalized-schlick-bsdf) or [&lt;sheen_bsdf>](#node-sheen-bsdf) over a BSDF or VDF. The implementation is target specific, but a standard approximation for a BSDF base is albedo scaling, using the function "top + pass_through(top) * base".
 
 |Port  |Description                     |Type       |Default |
 |------|--------------------------------|-----------|--------|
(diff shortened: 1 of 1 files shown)


jstone-lucasfilm:
Thanks for this thoughtful analysis, @tdavidovicNV, and I agree with the core distinction you're drawing here.  Deriving base attenuation from the complete directional albedo of the top conflates energy absorbed by the top layer with energy passed through to the base, and your tinted-dielectric example demonstrates this clearly.  As you note, the existing GLSL and OSL implementations already follow the semantics you propose, keeping artistic tint out of the pass-through term, so this change would align the specification with shipped behavior.

You may have seen that #3022, proposed shortly after this changelist, rewrites the same section of the specification from a complementary angle, defining a *vertical-layering transmittance* for every BSDF in the PBS library, extending layerability to opaque BSDFs through statistical coverage, and adding composition rules for the `mix`, `layer`, `add`, and `multiply` nodes, along with matching GLSL and MDL implementations.  The two proposals share the same architectural move, in which each BSDF supplies its own base-attenuation factor, composed multiplicatively through nested layers, with your pass-through factor and its transmittance playing the same role in the layer equation.

Rather than reconciling the two proposals sequentially, I'd propose that we merge your improvements into the transmittance framework of #3022, which I believe would incorporate:

- Excluding artistic tint and color from the transmittance of interface BSDFs, classifying the removed energy as absorption, so that the specification matches the current GLSL, OSL, and MDL implementations.
- Presenting the bidirectional pass-through factor of Weidlich and Wilkie as the ideal quantity, with the fixed-exitant-direction transmittance of #3022 defined as the reference approximation for targets using transmittance scaling.
- Your clarification that a BSDF layered over a VDF represents a surface boundary bound to an interior medium, so that medium entry is never derived from the complete reflected-plus-transmitted albedo of the top.

As in #3015, I'd suggest omitting the **Implementation notes** section from the specification text itself, since the PBR Specification intentionally avoids target-specific implementation details, and your per-target analysis will remain permanently accessible in the description and discussion of this PR.

If this plan sounds reasonable to you, I'd be glad to integrate these refinements into #3022 with credit to you as a co-author, or alternatively we could merge #3022 first and rebase this changelist on the new transmittance framework, whichever you'd prefer.  Either way, I want to make sure the distinctions you've drawn between reflection, absorption, and pass-through are preserved in the final specification text.

jstone-lucasfilm:
@tdavidovicNV Following up on the plan above, I've now integrated the refinements from this changelist into the transmittance framework of #3022, and I wanted to note how each of the distinctions you've drawn is preserved in the merged text:

- The transmittance of an interface BSDF is now evaluated with its physical Fresnel reflectance alone, with non-physical color inputs such as the `tint` of `dielectric_bsdf` attenuating only the scattered response, and the energy they remove classified as absorption within the interface.  This matches the shipped implementations you documented, and dovetails with the framework's existing rule for shader-semantic `multiply` nodes, which likewise preserve transmittance while attenuating the scattered response.  The Schlick reflectance inputs of `generalized_schlick_bsdf` are explicitly contrasted as parameterizing the physical Fresnel curve itself, contributing to a color-valued transmittance in the reference model, which the current GLSL implementation approximates with its RGB average, as noted in your per-target analysis.
- The bidirectional pass-through factor of Weidlich and Wilkie is presented as the ideal quantity underlying vertical layering, with the fixed-exitant-direction transmittance `T_o` defined as its reference approximation, and bidirectional evaluation noted as a refinement available to targets that directly simulate light transport through the layer stack.
- The `layer` node section is restructured into parallel "Layering over a BSDF" and "Layering over a VDF" subsections, with the latter adopting your framing of a surface boundary bound to an interior medium: entry into the medium is governed by the Fresnel transmittance of the interface, and is never derived from the complete reflected-plus-transmitted albedo of the top BSDF.

Your per-channel definition survives as well, with the transmittance defined as a color quantity taking values in `[0, 1]` per channel, and your observation on the recursive transmittance product is included in the layer equations, making the occlusion of a stack independent of its grouping.

I'd welcome your review of the new specification text in #3022, to validate that it faithfully preserves the distinctions between reflection, absorption, and pass-through that you've drawn here.  If you're happy with the integrated version, I'd propose that we continue the discussion there, and this changelist can be closed in its favor.

tdavidovicNV:
Thanks, @jstone-lucasfilm. I’ve reviewed the #3022, and I am more than happy to close this and continue there!

PR #3023, opened 2026-08-05
Fix prefilter env tangent-frame seam and add mx_isinf wrapper

ashwinbhat:
Fixes a tangent-frame seam in mx_generate_prefilter_env by building the frame from the latlong parameterization instead of calling mx_orthonormal_basis, which flips discontinuously where N.z crosses zero at longitude ±90°.

Also adds an mx_isinf wrapper to the GLSL/MSL/Slang math libraries and updates mx_chiang_hair_bsdf.glsl to use it. This is required because wgsl does not support it.

Validated with before/after GLSL renders using --envMethod 1 (prefiltered environment maps).
[mx_orthonormal_basis_comparison.html.pdf](https://github.com/user-attachments/files/30761472/mx_orthonormal_basis_comparison.html.pdf)


jstone-lucasfilm (approved):
Thanks for these improvements, @ashwinbhat!

Issue #2343, opened 2025-04-11
Document Geometry Resources in a README.md


jstone-lucasfilm:
This issue proposes that we add a README.md to the root of our Geometry Resources folder (https://github.com/AcademySoftwareFoundation/MaterialX/tree/main/resources/Geometry), explaining key details such as the diameter of the sphere, the UV mapping strategy for the cube, and the origins of the Shader Ball and Boombox.

redartengineer:
Hello! This is my first time contributing to open source. I'm interested in this project and would be happy to try putting together a README documenting the geometry resources and the details mentioned here.

jstone-lucasfilm:
Welcome, @redartengineer, and I've assigned this issue to you!

redartengineer:
@jstone-lucasfilm I’ve submitted a PR! I’d be happy to apply any feedback.

PR #982, opened 2022-06-06
Chess set example

crydalch:
Adds the Chess Set geometry and materials from the [Karma: A Beautiful Game](https://www.sidefx.com/tutorials/karma-a-beautiful-game/) tutorial.

Beautiful_Game

Original artwork and assets by Moeen Sayed and Mujtaba Sayed. Thanks to @jstone-lucasfilm for help converting the scene .glb and .mtlx, and finding optimal sizes for contributing.

Here are some test images. MaterialXView:
chess_set_materialxview

And a render from Karma:
chess_set_karma

PR #3029, opened 2026-08-12
Add documentation for geometry resources

redartengineer:
Adds a README to the Geometry resources folder documenting the sphere, cube, Shader Ball, and Boombox as well as Chess Set, Cloth, Plane, and Teapot.
Referencing #2343.


jstone-lucasfilm:
Welcome, @redartengineer, and thanks for this contribution!  From an initial review, most of these details look accurate to me, though I'd note a few areas for improvement:

1. For the Boombox, I'd suggest crediting the original source alongside its contributor.  The model is the BoomBox sample from the Khronos [glTF-Sample-Assets](https://github.com/KhronosGroup/glTF-Sample-Assets) repository, created and donated by Microsoft under a CC0 license, and it was contributed to MaterialX by Ed Mackey in #870.
2. For the Chess Set, the header of `standard_surface_chess_set.mtlx` credits Side Effects for contributing the geometry and materials, with the artwork by Moeen Sayed and Mujtaba Sayed, so I'd suggest carrying the Side Effects credit into the README as well.
3. Since `shaderball_ao.png` is the one file in this folder that the README doesn't yet cover, it may be worth a sentence in the Shader Ball section, noting that this is the pre-baked ambient occlusion map that MaterialXView pairs with the shader ball geometry.

A few smaller notes:

- The section heading "Shaderball" and the prose "Shader Ball" could be unified, and I'd lean towards "Shader Ball" for both.
- It may be worth linking each historical note to its originating pull request (e.g. #187 for the Shader Ball, #819 for the plane update, #982 for the Chess Set), so that readers can trace the full story of each asset.
- Since readers might otherwise assume a single quad, it's worth mentioning that the plane is a subdivided 30x30 grid, making it usable for per-vertex effects such as displacement.

Overall this looks like a great first contribution, and with a few refinements it should be ready to merge.

redartengineer:
@jstone-lucasfilm Thanks for the feedback! I've updated the README with the provided feedback by including the intended credits and contributors. I also added additional information for the other geometries.
Any other feedback I can apply as well!

jstone-lucasfilm:
Thanks for the updates, @redartengineer!  The new credits and PR links look great, and the README is very close to ready.  Two corrections that I'd consider important before we merge:

1. In the Teapot section, `teapot.obj` was actually added in #187 alongside the sphere, plane, and original shader ball.  It was #198 that introduced MaterialXView with the teapot as its default mesh, and #215 that switched that default to the Shader Ball.  Since the rest of the README works at the level of years, I'd suggest dropping the day-level dates here as well.
2. In the Cube section, I'd suggest describing `cube.obj` as "a unit cube centered at the origin", stating its dimensions in units as we do for the sphere and plane.  It may also be worth pointing readers to `resources/Materials/TestSuite/stdlib/texture/udim.mtlx`, which is the UDIM test material that pairs with this geometry.

Two optional refinements, which you're welcome to include or leave for follow-up:

- The Shader Ball GLB contains two meshes, `Preview_Mesh` and `Calibration_Mesh`, which the example looks assign materials to.
- The plane lies in the XY plane facing +Z, which is useful context for anyone testing displacement.

Does that sound reasonable to you?

redartengineer:
@jstone-lucasfilm No problem, I will apply the new feedback!

Issue #3037, opened 2026-08-16
MaterialX Graph Editor Keyboard-less UX


HardCoreCodin:
It is almost to the point of being able to use only the mouse.
The main 2 things missing are:

Adding nodes - maybe the middle mouse button, as that is pretty standard now
Deleting nodes - maybe an x button at the top-right corner of every node
Alternatively: Have a right-click menu (I know it is used for panning, but there's ways of differentiating a quick click vs. a click-and-hold)
In it, there could be menu options for adding a new or deleting all selected nodes and/or connections. It could also contain the graph auto-layout.

Also / in-between-option: Do what LookDevX does for nodes (see attached screenshot) with some hamburger button at the top-right corner of the node that brings up a menu:
This can already support deletion, copy, cut, and also show the docstring there instead of that NodeInfo button on the properties pane.

Also, support node collapse/expand button on the node itself
And have full node documentation of the node and input descriptions - either when hovering the mouse over the node title and over its inputs, or as a separate modal dialog or some such accessible through some button on the node.

<img width="612" height="538" alt="Image" src="https://github.com/user-attachments/assets/0be302fc-7136-4b05-a5c5-fc3beb11a7cf" />

catherineazelbee:
Would I be able to get assigned for this? Hoping to complete over the weekend for Dev Days!

kwokcb:
Basic building blocks + some feature items in #3043.

PR #3043, opened 2026-08-19
Graph Editor : UX Updates Towards "Keyboardless" Workflow

### LINKED ISSUE #3032: Ctrl+F triggers framing of the node graph in MaterialX Graph Editor
It should only bring up the search bar and ignore the fact that the F key happens to also be involved.

kwokcb:
## Updates

Update #3037.
Fixes #3032.

- Add "Add node" : `RMB` on background
- Add "hamburger" control : `LMB` brings up context senstive menu. 
  - Adds in rename (1) and delete node. 
  - Others can easily add more items for menu equivalents or new functionality.
 
- Fix so that focus  (`F`) does not occur when using `CTRL-F` (search)

Notes:

(1)  Rename is in-menu to avoid constant renaming causing "flashing" in UI and slow re-render on every keystroke.
(2) the remapping of pan should make it more consistent with DCCs / Web based editors and can handle 2-button mice / touch interfaces. 

## Implementation

Some refactoring for code reuse as much as possible for shared actions between keyboard vs mouse actions.

## Results

1. New "hamburger" control

image

2. Rename node in context

| | |
|:-:|:-:|
| image | image |

3. Add node (via RMB)
image


source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -27,6 +27,7 @@ const int FILTER_ALPHA = 50;
 const float BASE_UI_FONT_SIZE = 18.0f;
 const float BASE_PIN_ICON_SIZE = 18.0f;
 const float MIN_PIN_ICON_SIZE = 18.0f;
+const float HDR_TEXT_INDENT = 4.0f;
```
kwokcb, line 30: Global header indent

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -145,6 +146,9 @@ Graph::Graph(const std::string& materialFilename,
     _layoutPending(false),
     _needsNavigation(false),
     _delete(false),
+    _nodeMenuToOpen(-1),
```
kwokcb, line 149: Track node menu popup and deletion info if delete chosen

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -1995,6 +1999,55 @@ void Graph::drawPinIcon(const std::string& type, bool connected, int alpha, floa
     ed::PinRect(iconMin, iconMax);
 }
 
+void Graph::drawNodeMenu(UiNodePtr node)
+{
+    const float buttonSize = computeHamburgerSize();
```
kwokcb, line 2004: Node menu icon : "hamburger"

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2062,6 +2116,60 @@ float Graph::computeIconSize()
...
 }
 
+float Graph::computeHamburgerSize()
+{
+    return ImGui::GetTextLineHeight();
+}
+
+float Graph::computeNodeContentWidth(UiNodePtr node)
+{
+    // Compute the node's content width (widest of the title row, shown
```
kwokcb, line 2126: Utilitly to compute max text width for node

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -2292,7 +2409,8 @@ std::vector<int> Graph::createNodes(bool nodegraph)
                     ImGui::GetCursorScreenPos() + ImVec2(ed::GetNodeSize(node->getId()).x - hdrPadL - 2.f * hdrInset, ImGui::GetTextLineHeight() + hdrPadB),
                     ImColor(ImColor(85, 85, 85, 255)), 0.f);
                 ImGui::Indent(hdrTextIndent);
-                ImGui::Text("%s", node->getName().c_str());
+                ImGui::TextUnformatted(node->getName().c_str());
```
kwokcb, line 2412: Extend title with node menu icon.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3007,6 +3127,206 @@ void Graph::deleteNode(UiNodePtr node)
     _state.nodes.erase(_state.nodes.begin() + nodeNum);
 }
 
+bool Graph::deleteNodeById(ed::NodeId nodeId)
+{
```
kwokcb, line 3131: Extract out deletion logic.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3007,6 +3127,206 @@ void Graph::deleteNode(UiNodePtr node)
...
+    if (readOnly())
+    {
+        _popup = true;
+    }
+    return false;
+}
+
+void Graph::renameNode(UiNodePtr node, const std::string& newName)
+{
+    if (!node || newName.empty() || node->getName() == newName)
```
kwokcb, line 3159: Extract out rename logic. Fixed a few cases of up and downstream connection updates that were missing from previous inlined the property editor logic.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3497,130 +3817,25 @@ void Graph::propertyEditor()
         // Set and edit name
         ImGui::Text("Name: ");
         ImGui::SameLine();
+        const bool readOnlyGraph = readOnly();
         std::string original = _currUiNode->getName();
         std::string temp = original;
         float availableWidth = ImGui::GetContentRegionAvail().x;
         ImGui::PushItemWidth(availableWidth);
-        ImGui::InputText("##edit", &temp);
+        // Commit the rename on Enter or when the field loses focus, not on every keystroke.
```
kwokcb, line 3825: There was a lot of UI flickering when renaming during keystroke, especially as rename triggers renders.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -3497,130 +3817,25 @@ void Graph::propertyEditor()
...
+        // The name is not editable in read-only graphs.
+        const bool nameSubmitted = ImGui::InputText("##edit", &temp,
+            ImGuiInputTextFlags_EnterReturnsTrue | (readOnlyGraph ? ImGuiInputTextFlags_ReadOnly : 0));
+        const bool nameEdited = ImGui::IsItemDeactivatedAfterEdit();
         ImGui::PopItemWidth();
 
         std::string docString = "NodeDef Doc String: \n";
-        if (_currUiNode->getNode())
```
kwokcb, line 3508: Extracted and enhanced in shared utility rename method.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4126,9 +4354,10 @@ void Graph::addNodePopup(bool cursor)
 void Graph::searchNodePopup(bool cursor)
 {
     const ImGuiIO& io = ImGui::GetIO();
+    const bool searchModifier = io.KeyCtrl || io.KeySuper;
```
kwokcb, line 4357: Was missing super key for  mac.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4365,14 +4593,10 @@ void Graph::drawGraph(ImVec2 mousePos)
...
 
-        // Set up popups for adding a node when tab is pressed
-        ImGui::PushStyleVar(ImGuiStyleVar_WindowPadding, ImVec2(8.f, 8.f));
-        ImGui::SetNextWindowSizeConstraints(ImVec2(250.0f, 300.0f), ImVec2(-1.0f, 500.0f));
-        addNodePopup(TextCursor);
         searchNodePopup(TextCursor);
         addPinPopup();
         readOnlyPopup();
-        ImGui::PopStyleVar();
+        _nodeMenuRects.clear();
```
kwokcb, line 4599: Fully clear state.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4610,27 +4876,9 @@ void Graph::drawGraph(ImVec2 mousePos)
...
-                                ed::DeselectNode(id);
-                                ed::DeleteNode(id);
-                                _currUiNode = nullptr;
-                            }
-                            else if (readOnly())
-                            {
-                                _popup = true;
-                            }
-                        }
+                        deleteNodeById(id);
```
kwokcb, line 4881: Uses common delete node logic.

source/MaterialXGraphEditor/Graph.cpp
```diff
@@ -4833,7 +5083,15 @@ void Graph::drawGraph(ImVec2 mousePos)
         ed::Resume();
     }
 
+    // Suppress the node editor's built-in F shortcut when Ctrl/Cmd+F is used for node
+    // search. The editor processes its shortcuts inside ed::End(), so disable them for
+    // this frame and restore them immediately after.
+    if ((io2.KeyCtrl || io2.KeySuper) && ImGui::IsKeyPressed(ImGuiKey_F))
+    {
+        ed::EnableShortcuts(false);
```
kwokcb, line 5091: This was very tricky to fix since you need to superscede the imgui node editor logic which always assumes F key (regardless of modifier) is pressed. Reulsint in CMD/CTRL-F always kicking off "focus on node" logic.

source/MaterialXGraphEditor/Graph.h
```diff
@@ -415,6 +432,7 @@ class Graph
 
     // Options
     bool _saveNodePositions;
+    std::vector<ImVec4> _nodeMenuRects;
```
kwokcb, line 432: Used to track node menu location (hamburger) for hover hilighting

source/MaterialXGraphEditor/Main.cpp
```diff
@@ -266,6 +266,9 @@ int main(int argc, char* const argv[])
     // Create editor config and context.
     ed::Config config;
     config.SettingsFile = nullptr;
+    config.DragButtonIndex = 0;
```
kwokcb, line 269: Set mapping of mouse buttons.

kwokcb:
@lfl-eholthouser, @jstone-lucasfilm : if you have some time could you take a look at this PR. It addresses a few issues but naturally not all but maybe useful to allow building on further for upcoming dev days.  Thanks.

jstone-lucasfilm:
Thanks for this proposal, @kwokcb!  The rename consolidation and the Ctrl+F fix both look like clear improvements, and I'd be glad to see them merged soon.  The new `renameNode` corrects the downstream update logic for node renames, committing on Enter rather than every keystroke is the right call, and disabling editor shortcuts around `ed::End` matches the pattern in the ImGui Node Editor widgets example.

On the mouse remap, I'd suggest we reconsider before merging.  The PR moves panning to the middle button to avoid a conflict with a right-click menu, but ImGui Node Editor already resolves that conflict for us: its `ContextMenuAction` distinguishes a right-click from a right-drag, which is why the default config places both navigate and context menu on button 1.  The blueprints example demonstrates the intended pattern with `ed::ShowBackgroundContextMenu` and `ed::ShowNodeContextMenu`.  Building on that API rather than polling `IsMouseReleased(1)` would keep right-drag as pan, open the node menu on a right-click anywhere on the node, and remove the need for `_nodeMenuRects` and its hover tracking.  It would also address a few potential issues in your current proposal:

- Right-clicking a node body, pin, or link opens the Add Node popup, since only the hamburger rectangle is excluded.
- A right-drag that ends on the background opens the Add Node popup as well.
- Panning becomes unavailable on trackpads and two-button mice, which works against the goal of a keyboardless workflow.

If the hamburger feature is included in the current PR, I'd expect it to respond to a left click, matching the convention for that control.

The MacOS scaling change in `Main.cpp` seems separable from this work, and I'd suggest we take it up in its own PR with before-and-after screenshots.  With our current ImGui version, skipping the content scale means fonts are rasterized at 1x and upscaled by the framebuffer scale, so this trades doubled text for blurred text on Retina displays.  The usual approach for ImGui 1.88 is to rasterize at the scaled size and set `FontGlobalScale` to the inverse.

A few smaller notes:

- `documents/DeveloperGuide/GraphEditor.md` still lists right-click as pan, and should follow whatever binding we settle on.
- Moving `addNodePopup` after node creation leaves the search, add-pin, and read-only popups without the window padding style they previously shared.
- The pin offset fix for the pins-on-border option is a good improvement, and worth a mention in the PR description.

Given the above, one option would be to split the rename and Ctrl+F work into a PR we can merge right away, with the context menu work following on the editor's native API.

Does that approach sound good to you?

kwokcb:
Hi @jstone-lucasfilm 

The keyboard and mouse interaction is tricky and quite hard to get correct. Fixing f key at same time is less complicated.

Keeping current main mappings sounds reasonable, if possible.

Mac change can be pulled out. I can remove

I won't have time currently to do work on this. Maybe better to unassigned me. I will post on Slack if anyone wants to complete this pr for dev days.

kwokcb:
Found a few cycles to restore default mapping and make hamburger be on LMB.

kwokcb:
- Documentation has been updated in app and docs.
- Cleaned up padding code so it uses a global padding factor. Make all previous + new node menu have the same
padding.

PR #3059, opened 2026-08-31
Avoid find X11 on WASM

wingfiring:
Emscripten sets UNIX and clears APPLE, but provides no X11. X11 is only needed by the MaterialXRenderHw / MaterialXRenderGlsl modules, which are never built for Emscripten.


ashwinbhat:
@jstone-lucasfilm @meshula this affects USD/MaterialX WASM integration. We can patch it from USD build scripts side but we would prefer that it's addressed here. Thanks.

jstone-lucasfilm:
Thanks for this contribution, @wingfiring!

Your change is correct as written, but I'd suggest we consider a slightly more general form of the fix.  The X11 lookup entered `MaterialXConfig.cmake.in` in #2752, which addressed a genuine issue: the exported MaterialXRenderHw and MaterialXRenderGlsl targets carry `X11::X11` and `X11::Xt` in their link interfaces, and CMake errors if those imported targets don't exist in the consuming project.  That fix, however, made X11 a hard requirement of every Linux install of MaterialX, regardless of whether the hardware render modules were built, and Emscripten is only one of the configurations affected.  Headless Linux builds with `MATERIALX_BUILD_RENDER=OFF`, builds with `MATERIALX_BUILD_RENDER_PLATFORMS` or `MATERIALX_BUILD_GEN_GLSL` disabled, and Android are all subject to the same issue, and each would need its own exception under the current approach.

Since the top-level `CMakeLists.txt` already computes `MATERIALX_BUILD_RENDER_HW` as the condition under which these targets are built, I'd propose gating the lookup on that variable rather than on the platform, so that the installed config mirrors the install itself:

```
if(@MATERIALX_BUILD_RENDER_HW@ AND UNIX AND NOT APPLE)
    find_dependency(X11 REQUIRED COMPONENTS Xt)
    ...
endif()
```

One detail worth noting: `MATERIALX_BUILD_RENDER_HW` is currently defined only within the render block near line 561, so it would need a default of OFF ahead of that block for the substitution to be well-formed when rendering is disabled.

Does that generalized approach sound reasonable to you?

wingfiring:
Agreed — that's a cleaner fix; I've updated the PR to gate on @MATERIALX_BUILD_RENDER_HW@ and moved its OFF default ahead of the render block. Thanks for the suggestion!

jstone-lucasfilm (approved):
Great work, @wingfiring, and this latest version looks ready to go!

Issue #2182, opened 2025-01-13
Triplanar with different colorspaces gets wrong color transform


rasmusbonnedal:
### Description

When using the `<triplanarprojection>` node (which has 3 filename inputs) and different colorspaces on the different inputs only one colorspace transform is generated.

### To reproduce

Create a `<triplanarprojection>` node with different colorspaces on the different input and observe the generated OSL code.

![image](https://github.com/user-attachments/assets/d32888f2-7796-4d43-8192-2f4600f77f20)
[repro-triplanar.zip](https://github.com/user-attachments/files/18396033/repro-triplanar.zip)

### Result
Generated OSL code:
```
    NG_triplanarprojection_color3(triplanarprojection_filex_, triplanarprojection_filey_, triplanarprojection_filez_, triplanarprojection_layerx, triplanarprojection_layery, triplanarprojection_layerz, triplanarprojection_default, geomprop_Pobject_out1, geomprop_Nobject_out1, triplanarprojection_upaxis, triplanarprojection_blend, triplanarprojection_filtertype, triplanarprojection_framerange, triplanarprojection_frameoffset, triplanarprojection_frameendaction, triplanarprojection_out);
    color triplanarprojection_out_cm_out = color(0.0);
    NG_srgb_texture_to_lin_rec709_color3(triplanarprojection_out, triplanarprojection_out_cm_out);
```

### Expected result
I'd expect that three colorspace transforms would be emitted, one for each file input, with the corresponding source colorspace for each file input.

LouRohanNV:
I ran into what looks like another symptom of this issue while testing `ND_triplanarprojection_color3` in MaterialX 1.39.5. If all three filename inputs use `srgb_texture`, shader generation can crash. I reproduced it in USDView and in the MaterialX Playground, which reports a WebAssembly memory error. I reported the crash separately as issue #3060.

Tracing the graph helped explain why the crash is related. The triplanar node samples each axis separately, but MaterialX associates all three filename color spaces with the node's single color output. With identical inputs, it queues the same output transform three times. Removing the duplicate requests prevents the crash.

That crash fix does not solve the color problem in this issue. It leaves one sRGB conversion after the triplanar blend, but each sample needs conversion before the blend. This also affects identical color spaces whenever multiple axes contribute. A blend value of 1 makes the difference easier to see.

[...]

PR #3061, opened 2026-09-02
Prevent a ShaderGraph crash from duplicate color transforms

### LINKED ISSUE #3060: Duplicate color-transform requests from ND_triplanarprojection_color3 can crash shader generation
### Summary

I ran into a crash while testing `ND_triplanarprojection_color3` with ordinary sRGB color textures. It first appeared in USDView, but the problem is not specific to OpenUSD. I reproduced it with the MaterialX Playground 1.39.5 WebAssembly build by marking all three filename inputs as `srgb_texture`.

The same setup works with `raw`, which led me to the color-transform path. From tracing shader generation, each filename input appears to queue the same transform on the triplanar node's single color output. Those duplicate requests can leave an invalid pointer in `ShaderGraph` and cause either a native crash or a WebAssembly memory error.

I have a small patch that keeps only one exact port-and-transform request. It prevents the crash in my tests. The patch is not a complete triplanar color fix. Color conversion still occurs after the three samples are blended. Issue #2182 covers that broader color-order problem.

### Tested configurations

- MaterialX 1.39.5 at tag `v1.39.5`
- [MaterialX Playground Node Graph Editor](https://joaovbs96.github.io/MaterialXPlayground/#!graph) with its bundled MaterialX 1.39.5 WebAssembly build
- OpenUSD 26.08 USDView with Storm
- Ubuntu 22.04.5 LTS on x86-64
- GCC 11.4.0
- Google Chrome 152.0.7977.75
- ESSL and GLSL shader generation
- Default MaterialX color management

### Minimal MaterialX document

```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <triplanarprojection name="triplanarprojection1" type="color3" xpos="0.103" ypos="-1.9615">
    <input name="filex" type="filename" value="" uniform="true" colorspace="srgb_texture" />
    <input name="filey" type="filename" value="" uniform="true" colorspace="srgb_texture" />
    <input name="filez" type="filename" value="" uniform="true" colorspace="srgb_texture" />
  </triplanarprojection>
  <open_pbr_surface name="open_pbr_surface1" type="surfaceshader" xpos="2.1302" ypos="-2.5982">
    <input name="base_color" type="color3" nodename="triplanarprojection1" />
  </open_pbr_surface>
</materialx>
```

No image files are required. Empty filename values are sufficient to reach the failing shader-generation path.

### Minimal OpenUSD stage

The stage contains one sphere, one OpenPBR shader, and one `ND_triplanarprojection_color3` node. It uses empty asset inputs with `srgb_texture` color-space metadata. It requires no textures or USDZ package.

Open the stage with Storm:

```bash
usdview --renderer Storm --clearsettings --select /World/Sphere materialxTriplanarColorTransformCrash.usda
```

OpenUSD 26.08 with unmodified MaterialX 1.39.5 exits on signal 11 during shader generation. The exact-request deduplication patch allows the same stage to complete.

```
#usda 1.0
(
    defaultPrim = "World"
    metersPerUnit = 1
    upAxis = "Y"
)

def Xform "World"
{
    def Material "Material" (
        prepend apiSchemas = ["MaterialXConfigAPI"]
    )
    {
        string config:mtlx:version = "1.39"
        token outputs:mtlx:surface.connect = </World/Material/OpenPBR.outputs:out>

        def Shader "OpenPBR"
        {
            uniform token info:id = "ND_open_pbr_surface_surfaceshader"
            color3f inputs:base_color.connect = </World/Material/Triplanar.outputs:out>
            token outputs:out
        }

        def Shader "Triplanar"
        {
            uniform token info:id = "ND_triplanarprojection_color3"
            asset inputs:filex = @@ (
                colorSpace = "srgb_texture"
            )
            asset inputs:filey = @@ (
                colorSpace = "srgb_texture"
            )
            asset inputs:filez = @@ (
                colorSpace = "srgb_texture"
            )
            color3f outputs:out
        }
    }

    def Sphere "Sphere" (
        prepend apiSchemas = ["MaterialBindingAPI"]
    )
    {
        rel material:binding = </World/Material>
        double radius = 1
    }
}
```

### What I tried

1. Open the [MaterialX Playground Node Graph Editor](https://joaovbs96.github.io/MaterialXPlayground/#!graph).
2. Load the document above or recreate manually with nodes

Image

### What happened

MaterialX Playground can report this error:

```text
RuntimeError: memory access out of bounds
    at ShaderGenerator.generate
```

The native OpenUSD reproduction crashes with this relevant stack:

```text
MaterialX_v1_39_5::ShaderGraph::topologicalSort()
MaterialX_v1_39_5::ShaderGraph::finalize()
MaterialX_v1_39_5::ShaderGraph::create()
MaterialX_v1_39_5::HwShaderGenerator::createShader()
HdStMaterialXShaderGenBaseGlsl::generate()
```

### What I expected

Shader generation completes without invalid memory access. `ShaderGraph` stores one request for an exact shader-port and color-transform pair.

### Cause

`ND_triplanarprojection_color3` has three filename inputs and one color output. When all three inputs use `srgb_texture`, MaterialX adds the same color transform three times to the output. Each transform node gets the same identifier.

The second map entry replaces the first node, but `_nodeOrder` still contains a pointer to the first node. `topologicalSort()` later reads that invalid pointer and crashes.

### Test with unique identifiers

I first gave each transform node a different identifier. This stopped the crash and confirmed the collision. However, MaterialX then applied the same sRGB conversion three times after the triplanar blend. The output became almost black. Unique identifiers protect the graph, but they do not produce the correct shader.

### Proposed crash fix

`populateColorTransformMap()` should ignore an exact duplicate request. An exact duplicate uses the same shader port, source space, target space, and value type.

This change keeps separate requests for different ports or different color transforms. It also reduces this case to one color transform and prevents the crash.

### Remaining color issue

The remaining color transform still occurs after the triplanar blend. Correct sRGB blending requires one transform after each texture sample and before the blend. This difference matters when multiple projection axes contribute. Issue [#2182](https://github.com/AcademySoftwareFoundation/MaterialX/issues/2182) covers that larger change.

### Pull request #2768

Pull request [#2768](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2768) is already included in MaterialX 1.39.5. It fixes collisions between different document nodes. This crash instead comes from repeated transform requests for one shader output. Current `main` still contains this transform-request behavior. Therefore, pull request #2768 does not fix this case.

### Workarounds

Using `raw` avoids the failing path because it requests no color transform. That workaround produces incorrect color for an sRGB color texture.

Using `ND_triplanarprojection_vector3` followed by `ND_convert_vector3_color3` also avoids the crashing `color3` path. That graph bypasses automatic input color conversion. It is not a complete color-management correction.

LouRohanNV:
### Summary

Each filename input on `ND_triplanarprojection_color3` can request a color transform for the same shader output. Those requests create transform nodes with the same identifier. The node map replaces the first node, but the node order and downstream connections still point to it. This can crash shader generation, whether the inputs use identical or different color spaces.

This change keeps the first transform request for each output port and ignores later requests for that port. It also makes `ShaderGraph::addNode()` throw `ExceptionShaderGenError` on a duplicate identifier, instead of replacing a node with active references. Input transform requests need no deduplication because `applyInputTransforms()` visits each shader input once.

Fixes #3060.

### Tests

The regression test generates shaders from triplanar documents with identical and different input color spaces. Both cases check that generation succeeds and that exactly one output transform remains, using the first input's color space. These document tests exercise the `finalize` path where the crash occurs.

The focused test passes with 10 assertions. All 11 tests in `[genshader]` pass, with 251 assertions.

```bash
./bin/MaterialXTest 'GenShader: Duplicate Output Color Transforms'
./bin/MaterialXTest '[genshader]'
```

### Limitation

The retained transform still applies after the triplanar blend. For different input color spaces, the first retained conversion applies to the combined result, so the colors can still be incorrect. Correct conversion of each image sample before the blend remains a separate change under #2182.

### DIFF
--- a/source/MaterialXGenShader/ShaderGraph.cpp
+++ b/source/MaterialXGenShader/ShaderGraph.cpp
@@ -12,6 +12,7 @@
 
 #include <MaterialXTrace/Tracing.h>
 
+#include <algorithm>
 #include <queue>
 
 MATERIALX_NAMESPACE_BEGIN
@@ -1224,13 +1225,25 @@ void ShaderGraph::populateColorTransformMap(ColorManagementSystemPtr colorManage
             ColorSpaceTransform transform(sourceColorSpace, targetColorSpace, shaderPort->getType());
             if (colorManagementSystem->supportsTransform(transform))
             {
+                // A multi-file color node can request the same transform once per filename input.
+                // Retain only unique port and transform pairs to prevent duplicate transform nodes.
                 if (asInput)
                 {
-                    _inputColorTransformMap.emplace_back(static_cast<ShaderInput*>(shaderPort), transform);
+                    const auto entry = std::make_pair(static_cast<ShaderInput*>(shaderPort), transform);
+                    if (std::find(_inputColorTransformMap.begin(), _inputColorTransformMap.end(), entry) ==
+                        _inputColorTransformMap.end())
+                    {
+                        _inputColorTransformMap.push_back(entry);
+                    }
                 }
                 else
                 {
-                    _outputColorTransformMap.emplace_back(static_cast<ShaderOutput*>(shaderPort), transform);
+                    const auto entry = std::make_pair(static_cast<ShaderOutput*>(shaderPort), transform);
+                    if (std::find(_outputColorTransformMap.begin(), _outputColorTransformMap.end(), entry) ==
+                        _outputColorTransformMap.end())
+                    {
+                        _outputColorTransformMap.push_back(entry);
+                    }
                 }
             }
             else
--- a/source/MaterialXTest/MaterialXGenShader/GenShader.cpp
+++ b/source/MaterialXTest/MaterialXGenShader/GenShader.cpp
@@ -17,6 +17,7 @@
 #include <MaterialXGenShader/Exception.h>
 #include <MaterialXGenShader/GenContext.h>
 #include <MaterialXGenShader/OcioColorManagementSystem.h>
+#include <MaterialXGenShader/ShaderGraph.h>
 #include <MaterialXGenShader/ShaderTranslator.h>
 #include <MaterialXGenShader/Util.h>
 
@@ -42,6 +43,21 @@
 #include <set>
 
 namespace mx = MaterialX;
+
+class TestShaderGraph : public mx::ShaderGraph
+{
+  public:
+    using mx::ShaderGraph::ShaderGraph;
+
+    size_t addOutputColorTransform(mx::ColorManagementSystemPtr colorManagementSystem,
+                                   mx::ShaderOutput* output,
+                                   const mx::ColorSpaceTransform& transform)
+    {
+        populateColorTransformMap(colorManagementSystem, output,
+                                  transform.sourceSpace, transform.targetSpace, false);
+        return _outputColorTransformMap.size();
+    }
+};
 
 //
 // Base tests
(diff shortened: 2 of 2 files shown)


jstone-lucasfilm:
Thanks for this proposed fix, @LouRohanNV, as well as for the great overview in #3060.  Your diagnosis matches what I see in the code.  Each filename input on a color-typed node queues its own output transform in `applyInputTransforms`, and since every transform node for a given output is named `<output>_cm`, the second call to `ShaderGraph::addNode` replaces the first node in `_nodeMap` while `_nodeOrder` and the rerouted downstream connections still point to it.

One issue, though, is that the exact-pair check only covers the case where all three inputs share a color space.  If the inputs specify different color spaces, as in the configuration described in #2182, the three requests differ in their source color space and survive the check, so the same collision and crash still occur.  Ideally I'd like this PR to handle the full class of collisions rather than the single-color-space instance, and I believe two changes would get us there:

- In `populateColorTransformMap`, deduplicate the output map by shader port alone, retaining the first request.  A shader output can carry only a single transform node in the current design, so this matches the behavior of single-filename nodes such as `image`, and it makes the input-side check unnecessary, since `applyInputTransforms` visits each shader input exactly once.
- In `ShaderGraph::addNode`, throw an `ExceptionShaderGenError` when a node with the given unique ID already exists, rather than replacing the map entry.  This turns any future collision of this kind into a clear error rather than a use-after-free.

On the test, I'd suggest a document-level regression test in place of the `TestShaderGraph` subclass, following the pattern of the "Color Space No-Op" test in `GenShader.cpp`, since that exercises the `finalize` path where the crash occurs.  Two documents would cover both the identical and differing color space cases above.

More broadly, I agree that the correct long-term behavior for a nodegraph-implemented color node is to apply each transform to the inner `image` node before the blend, and it seems completely reasonable to defer that improvement to a future PR for #2182.

jstone-lucasfilm (approved):
Great work, thanks @LouRohanNV, and this fix looks ready to merge.

PR #3062, opened 2026-09-05
Fix pass-through transforms for legacy color space names

jstone-lucasfilm:
This changelist follows up on the color interop alignment in #3042, resolving a regression for documents that use legacy color space names such as `lin_rec709`.

With shader generation now targeting `lin_rec709_scene`, every color input in a legacy document was seen as requiring a transform, and the default color management system expressed the equivalence of these names as a pass-through `dot` node rather than omitting the transform.  This added a redundant node and renamed uniform for each color input, and in MDL it dropped the uniform qualifier from inputs such as glTF PBR `attenuation_color`, producing shaders that fail to compile.

- Add a virtual `ColorManagementSystem::isNoOpTransform` method, with a `DefaultColorManagementSystem` override that recognizes legacy and color interop names for the same color space, and consult it in `ShaderGraph` so that equivalent color spaces introduce no transform nodes.
- Preserve the uniform flag of an input when inserting a color or unit transform node, so that MDL declares the published value correctly.
- Update `TextureBaker` and the Metal render test to the `lin_rec709_scene` working space, with baked documents now written using color interop names.
- Expose `isNoOpTransform` in Python, with unit tests covering the alias logic and the generated GLSL and MDL shaders.

### DIFF
--- a/source/MaterialXGenShader/ColorManagementSystem.h
+++ b/source/MaterialXGenShader/ColorManagementSystem.h
@@ -74,6 +74,16 @@ class MX_GENSHADER_API ColorManagementSystem
         return isReservedNoOpColorSpace(colorSpace);
     }
 
+    /// Returns true if no color transformation is required between the given source
+    /// and target color spaces, meaning that the two names refer to the same color
+    /// space. The base implementation compares the names directly. Subclasses may
+    /// extend this to recognize aliases of the same color space, such as a legacy
+    /// name and its color interop equivalent.
+    virtual bool isNoOpTransform(const string& sourceColorSpace, const string& targetColorSpace) const
+    {
+        return sourceColorSpace == targetColorSpace;
+    }
+
     /// The colorSpace strings should not be shown directly in a user interface. This function
     /// converts a colorSpace into a user-facing name. For example, "lin_rec709_scene" becomes
     /// "Linear Rec.709 (sRGB)". If the colorSpace is not recognized, it is returned unchanged.
--- a/source/MaterialXGenShader/DefaultColorManagementSystem.cpp
+++ b/source/MaterialXGenShader/DefaultColorManagementSystem.cpp
@@ -67,6 +67,14 @@ const StringMap COLOR_SPACE_USER_FACING_NAMES =
     { "data", "Data" }
 };
 
+// Return the base color space name for the given color space, resolving any
+// legacy or color interop alias through COLOR_SPACE_REMAP.
+const string& remapColorSpace(const string& colorSpace)
+{
+    auto it = COLOR_SPACE_REMAP.find(colorSpace);
+    return it != COLOR_SPACE_REMAP.end() ? it->second : colorSpace;
+}
+
 } // anonymous namespace
 
 //
@@ -90,8 +98,13 @@ const string& DefaultColorManagementSystem::getName() const
 
 string DefaultColorManagementSystem::getUserFacingName(const string& colorSpace) const
 {
-    string baseSpace = COLOR_SPACE_REMAP.count(colorSpace) ? COLOR_SPACE_REMAP.at(colorSpace) : colorSpace;
+    const string& baseSpace = remapColorSpace(colorSpace);
     return COLOR_SPACE_USER_FACING_NAMES.count(baseSpace) ? COLOR_SPACE_USER_FACING_NAMES.at(baseSpace) : colorSpace;
+}
+
+bool DefaultColorManagementSystem::isNoOpTransform(const string& sourceColorSpace, const string& targetColorSpace) const
+{
+    return remapColorSpace(sourceColorSpace) == remapColorSpace(targetColorSpace);
 }
 
 NodeDefPtr DefaultColorManagementSystem::getNodeDef(const ColorSpaceTransform& transform) const
@@ -101,8 +114,8 @@ NodeDefPtr DefaultColorManagementSystem::getNodeDef(const ColorSpaceTransform& t
         throw ExceptionShaderGenError("No library loaded for color management system");
     }
 
-    string sourceSpace = COLOR_SPACE_REMAP.count(transform.sourceSpace) ? COLOR_SPACE_REMAP.at(transform.sourceSpace) : transform.sourceSpace;
-    string targetSpace = COLOR_SPACE_REMAP.count(transform.targetSpace) ? COLOR_SPACE_REMAP.at(transform.targetSpace) : transform.targetSpace;
+    const string& sourceSpace = remapColorSpace(transform.sourceSpace);
+    const string& targetSpace = remapColorSpace(transform.targetSpace);
 
     // After remapping, the transform may be a no-op (e.g. lin_rec709_scene -> lin_rec709).
     if (sourceSpace == targetSpace)
--- a/source/MaterialXTest/MaterialXGenShader/GenShader.cpp
+++ b/source/MaterialXTest/MaterialXGenShader/GenShader.cpp
@@ -552,6 +552,120 @@ TEST_CASE("GenShader: No-op Color Spaces", "[genshader]")
 #endif
 }
 
+TEST_CASE("GenShader: Equivalent Color Spaces", "[genshader]")
+{
+    // A legacy color space name and its color interop equivalent refer to the same
+    // color space, so DefaultColorManagementSystem requires no transform between them.
+    mx::DefaultColorManagementSystemPtr colorManagementSystem =
+        mx::DefaultColorManagementSystem::create("genglsl");
+    CHECK(colorManagementSystem->isNoOpTransform("lin_rec709", "lin_rec709"));
+    CHECK(colorManagementSystem->isNoOpTransform("lin_rec709", "lin_rec709_scene"));
+    CHECK(colorManagementSystem->isNoOpTransform("lin_rec709_scene", "lin_rec709"));
+    CHECK(colorManagementSystem->isNoOpTransform("srgb_texture", "srgb_rec709_scene"));
+    CHECK(colorManagementSystem->isNoOpTransform("acescg", "lin_ap1_scene"));
+    CHECK(!colorManagementSystem->isNoOpTransform("lin_rec709", "acescg"));
+    CHECK(!colorManagementSystem->isNoOpTransform("srgb_texture", "lin_rec709_scene"));
+    CHECK(!colorManagementSystem->isNoOpTransform("lin_rec709", "bogus_colorspace"));
+
+    mx::FileSearchPath searchPath = mx::getDefaultDataSearchPath();
+    mx::DocumentPtr libraries = mx::createDocument();
+    mx::loadLibraries({ "libraries" }, searchPath, libraries);
+
+    // A document authored in the legacy "lin_rec709" working space, generated for the
+    // equivalent "lin_rec709_scene" target, must not acquire pass-through color transform
+    // nodes. The glTF PBR attenuation_color input is uniform, and a pass-through node
+    // inserted here would publish it as a varying value, which fails to compile in MDL.
+    std::string legacyDocString =
+    "<?xml version=\"1.0\"?> \
+      <materialx version=\"1.39\" colorspace=\"lin_rec709\"> \
+        <gltf_pbr name=\"legacy_shader\" type=\"surfaceshader\"> \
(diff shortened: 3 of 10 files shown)


source/MaterialXGenShader/ColorManagementSystem.h
doug-walker, line 43: Should this use isNoOpTransform?

source/MaterialXGenShader/DefaultColorManagementSystem.cpp
doug-walker, line 125: Do we still want to return a dot node? This was added recently in my PR #2954, thinking that some kind of nodedef must be returned, but maybe there is a better approach?

source/MaterialXGenShader/ColorManagementSystem.h
```diff
@@ -74,6 +74,16 @@ class MX_GENSHADER_API ColorManagementSystem
...
 
+    /// Returns true if no color transformation is required between the given source
+    /// and target color spaces, meaning that the two names refer to the same color
+    /// space. The base implementation compares the names directly. Subclasses may
+    /// extend this to recognize aliases of the same color space, such as a legacy
+    /// name and its color interop equivalent.
+    virtual bool isNoOpTransform(const string& sourceColorSpace, const string& targetColorSpace) const
+    {
+        return sourceColorSpace == targetColorSpace;
+    }
```
doug-walker, line 81: This seems to me more like an areColorSpacesEqual function rather than an isNoOpTransform. If it were the latter, I think it would need to return true if isNoOpColorSpace were true for either source or target.

The only place isNoOpColorSpace is currently used is in that same if-statement in populateColorTransformMap in shaderGraph.cpp. I'm thinking that we should build the isNoOpColorSpace functionality into isNoOpTransform and then we could simplify the API by removing isNoOpColorSpace (since it has not been included in an official release yet)?

I'm not sure there is a reason for wanting to know if a color space is a no-op, other than in order to decide if a transform involving it is a no-op, so isNoOpTransform might be a better level of abstraction. But I have no objection if you prefer to keep things more granular and instead have isNoOpColorSpace and areColorSpacesEqual.

Unlike for the DefaultColorManagementSystem, in the OcioColorManagementSystem, it's possible for isNoOpColorSpace to be false and areColorSpacesEqual to be false (if based on alias name resolution) and yet still return a dot node as the result because when OCIO optimizes the mathematical operators it's possible that the result is an identity. But it seems isNoOpColorSpace is probably intended to be a more lightweight function and so perhaps we should just implement it on the OCIO side by just doing alias name resolution and isData on the color space. I think it would depend on how undesirable it is to return a dot node.

doug-walker (commented):
Good catch, Jonathan! Please see my questions below.

I'd like to add the new function to OcioColorManagementSystem too. Would you prefer I make a separate PR, or provide code you could include in your PR?

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -552,6 +552,120 @@ TEST_CASE("GenShader: No-op Color Spaces", "[genshader]")
...
+          <input name=\"surfaceshader\" type=\"surfaceshader\" nodename=\"legacy_shader\" /> \
+        </surfacematerial> \
+      </materialx>";
+    mx::DocumentPtr legacyDoc = mx::createDocument();
+    mx::readFromXmlString(legacyDoc, legacyDocString);
+    legacyDoc->setDataLibrary(libraries);
+    mx::ElementPtr legacyElement = legacyDoc->getChild("legacy_material");
+    REQUIRE(legacyElement);
+
+    // When a genuine transform is required, the inserted color transform node must
```
meshula: can we use a different word than "genuine" here? I think it is being used to mean something like "Since a transform is to be performed, we must ... pass through something something?"

I am struggling to map this comment to what the string does. I think my request is to replace the comment with one that more plainly describes what is going on here.

Side note, any reason not to switch to unescaped multiline C++ string constants at this point? I don't think there is any viable vfxplatform that requires early 21st century cpp compilers any more (which are the last to not have multiline strings)

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -552,6 +552,120 @@ TEST_CASE("GenShader: No-op Color Spaces", "[genshader]")
...
+    CHECK(!colorManagementSystem->isNoOpTransform("lin_rec709", "bogus_colorspace"));
+
+    mx::FileSearchPath searchPath = mx::getDefaultDataSearchPath();
+    mx::DocumentPtr libraries = mx::createDocument();
+    mx::loadLibraries({ "libraries" }, searchPath, libraries);
+
+    // A document authored in the legacy "lin_rec709" working space, generated for the
+    // equivalent "lin_rec709_scene" target, must not acquire pass-through color transform
+    // nodes. The glTF PBR attenuation_color input is uniform, and a pass-through node
+    // inserted here would publish it as a varying value, which fails to compile in MDL.
```
meshula: "which fails to compile in MDL". Is this LLM internal note-keeping? My apologies for the question, but "fails to compile in MDL" is a weird clause. I think what it is trying to say is that 

"the color input value is uniform. inserting a pass-through node here must not republish it as a variying value."

"fails to compile in MDL" is a symptom that was addressed, it is not a contract on the user of the code.

meshula (changes requested):
I don't have anything to add beyond deferring to Doug's comments where he's left them.

Overall it looks good to me, although I'd appreciate slight tweaks to two of the comments which (huge apologies if I'm reading it wrong) appear to be LLM book keeping comments.

source/MaterialXGenShader/ColorManagementSystem.h
jstone-lucasfilm: Alas, I can't access our color management system here, since it's the equality operator on the `ColorSpaceTransform` struct!

source/MaterialXGenShader/DefaultColorManagementSystem.cpp
jstone-lucasfilm: I think we still want the dot node as a fallback.  With this changelist, `ShaderGraph` consults `isNoOpTransform` before requesting a nodedef, so the graph itself never receives a pass-through node for these cases.  The remaining consumers are direct callers of `supportsTransform` and `createNode`, and for those I'd rather report a trivially supported transform than throw "No nodedef found".

source/MaterialXGenShader/ColorManagementSystem.h
jstone-lucasfilm: Agreed on both points, and I've taken your first suggestion in the latest update to this PR.  Since neither method has shipped in a release, `isNoOpColorSpace` is removed from the base class and Python bindings, and `isNoOpTransform` now returns true for identical names or when either name is a reserved no-op space.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
jstone-lucasfilm: Good points, @meshula, and I've rewritten the comment in a way that is hopefully clearer.  On the multi-line C++ string constants, I was following @doug-walker's example in the test code above, but I think your suggestion is a good one, and I've updated all of the color-space examples to unescaped strings.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
jstone-lucasfilm: Agreed, @meshula, the original language wasn't great, and I've rewritten it for clarity.

source/MaterialXGenShader/OcioColorManagementSystem.cpp
doug-walker, line 170: In order to handle aliased names, I suggest replacing this with the following, and then calling it below.

```
bool OcioColorManagementSystemImpl::isNoOpTransform(const string& source, const string& target) const
{
    const char* supportedSource = getSupportedColorSpaceName(source.c_str());
    if (!supportedSource)
    {
        return false;
    }
    const char* supportedTarget = getSupportedColorSpaceName(target.c_str());
    if (!supportedTarget)
    {
        return false;
    }

    OCIO::ConstColorSpaceRcPtr sourceObj = _config->getColorSpace(supportedSource);
    OCIO::ConstColorSpaceRcPtr targetObj = _config->getColorSpace(supportedTarget);
    if (!sourceObj || !targetObj)
    {
        return false;
    }
    return sourceObj->isData() || targetObj->isData()
        // The getName call converts aliases into the canonical names.
        || (sourceObj->getName() == targetObj->getName());
}
```

doug-walker (commented):
Your updates look good Jonathan!  I have one suggestion for the OCIO part.

source/MaterialXGenShader/OcioColorManagementSystem.cpp
jstone-lucasfilm: Thanks, @doug-walker, and I've included your suggestion in the latest update to the PR.  One small adjustment that I made is to compare the canonical names as strings rather than pointers, so that we don't depend on OCIO returning identical pointers in this situation, although it's possible that either approach would be safe in the context of OCIO.

Since our GitHub CI doesn't yet build with OCIO enabled, a validation pass on your side may be warranted before we merge.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
doug-walker: Suggest "Color Spaces" --> "Color Transforms"

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
```diff
@@ -462,33 +462,28 @@ TEST_CASE("GenShader: No-op Color Spaces", "[genshader]")
...
             mx::OcioColorManagementSystem::createFromBuiltinConfig("ocio://cg-config-latest", "genglsl");
-        CHECK(ocioColorManagementSystem->isNoOpColorSpace("none"));
-        CHECK(ocioColorManagementSystem->isNoOpColorSpace("data"));
-        CHECK(ocioColorManagementSystem->isNoOpColorSpace("Raw"));
-        CHECK(!ocioColorManagementSystem->isNoOpColorSpace("lin_rec709_scene"));
-        CHECK(!ocioColorManagementSystem->isNoOpColorSpace("ACEScg"));
+        CHECK(ocioColorManagementSystem->isNoOpTransform("none", "lin_rec709_scene"));
+        CHECK(ocioColorManagementSystem->isNoOpTransform("data", "lin_rec709_scene"));
+        CHECK(ocioColorManagementSystem->isNoOpTransform("Raw", "lin_rec709_scene"));
+        CHECK(ocioColorManagementSystem->isNoOpTransform("lin_rec709", "lin_rec709_scene"));
```
doug-walker, line 487: Please add the following to test that OCIO-based alias resolution is working:
CHECK(ocioColorManagementSystem->isNoOpTransform("ACES - ACES2065-1", "lin_ap0"));

doug-walker (commented):
That looks perfect Jonathan! I proposed a new test below to validate the OCIO-based alias resolution is working. I cloned your branch and built with OCIO enabled and all tests pass.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
meshula: thanks, the new comment reads clearly to me now.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
meshula: thanks, the new comment reads clearly to me now.

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
jstone-lucasfilm: Done!

source/MaterialXTest/MaterialXGenShader/GenShader.cpp
jstone-lucasfilm: Done!

jstone-lucasfilm:
@doug-walker @meshula Let me know if this latest version looks good to you, and ideally I'd like to merge this work soon, as it addresses a failure in our nightly CI builds that was introduced by https://github.com/AcademySoftwareFoundation/MaterialX/pull/3042.  (Multi-hour nightly CI builds FTW!)

doug-walker (approved):
LGTM!

Issue #1749, opened 2024-03-25
Fix behavior of `range` node with inverted output and clamp


makkbru:
when the range node is used to invert the input, activating clamp will make it output the wrong result. its probably clamping to min and max in the input without checking if the min value is smaller than the max value

jstone-lucasfilm:
Adding a link to the current graph definition of the `range` node in MaterialX:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/stdlib/stdlib_ng.mtlx#L3568

sckleung:
Hello, I am a new contributor for ASWF Dev Days this year and would like to look in to this as a first issue. I am unable to log on to the Slack channel currently.

jstone-lucasfilm:
Welcome @sckleung!  Here's a link that we often recommend for creating an ASWF Slack account, and you can find our team there under #materialx.

https://www.aswf.io/get-involved/

ekim-01:
Hello!  This is my first year as a contributor to ASWF Dev Days.  I would like to work on this issue as a first task.  Is it too late to have this task assigned to me?  Thank you!

ld-kerley:
@ekim-01 welcome to the project - I looks like this issue is available - so I'll assign it to you - glad to have you on board to help out - if you have any questions please feel free to reach out on the MaterialX ASWF slack channel. #materialx.

jstone-lucasfilm:
@ekim-01 Since we're approaching our next [ASWF Dev Days event](https://www.aswf.io/dev-days/) in May, I wanted to check in on the status of this project, to see if you'd like to continue working on it.  Let us know!

jstone-lucasfilm:
@ekim-01 I'll unassign you from this issue for now, so that it's available for our upcoming ASWF Dev Days event, but feel free to request it again if you are interested.

baileyperillo:
Hello, I am Bailey and I am a student for Codepath's AI301 course. 
I wanted to try and practice open source work and saw that this was a good first issue. I was wondering if this is still open and if I could work on this issue? 
If so, where would I be able to find the issue, such as where would the range node be located? and can you explain more on the issue, such as what it's supposed to do?

jstone-lucasfilm:
[...]

PR #3064, opened 2026-09-09
Fix range node clamp when output bounds are inverted

seang-ilm:
Fix for issue #1749 

When outlow is greater than outhigh (e.g. for color inversion), clamping  produces incorrect results. This sorts the clamp bounds with min/max before applying `doclamp` across all range nodegraph.

I've got a simple `.mltx` to demonstrate the issue and the fix

checkerboard_two

```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <nodedef name="ND_passthrough" node="passthrough">
    <input name="file" type="filename" value="resources/Images/grid.png" />
  </nodedef>
  <nodegraph name="NG_passthrough" xpos="8.695652" ypos="0.000000">
    <output name="out" type="vector3" xpos="20.992754" ypos="-1.448276" nodename="range_color3" />
    <range name="range_color3" type="vector3" xpos="12.985507" ypos="-3.706897">
      <input name="in" type="vector3" nodename="image_color3" />
      <input name="inlow" type="vector3" value="0.25, 0.25, 0.25" />
      <input name="inhigh" type="vector3" value="0.28, 0.28, 0.28" />
      <input name="outlow" type="vector3" value="1, 1, 1" />
      <input name="outhigh" type="vector3" value="0.6, 0.6, 0.6" />
      <input name="doclamp" type="boolean" value="false" />
      <input name="gamma" type="vector3" value="1, 1, 1" />
    </range>
    <image name="image_color3" type="vector3" xpos="10.666667" ypos="-3.741379">
      <input name="file" type="filename" value=".\checkerboard_two.png" />
    </image>
  </nodegraph>
</materialx>
```

**Before**

image

**After**

image


jstone-lucasfilm:
Thanks @seang-ilm, this looks like a good fix, and I like the graph-based, language-independent approach you've taken here.

One request before we merge: the specification for `range` in our Standard Nodes specification currently describes `doclamp` as follows:

> If true, the output is clamped to [`outlow`, `outhigh`]

With this change, the output is clamped to the range spanned by those two values regardless of their order, so we should update the specification to something more like the following:

> If true, the output is clamped to the range between `outlow` and `outhigh`, regardless of which is larger

Other than that detail, I think this PR looks ready to go.

seang-ilm:
> One request before we merge: the specification for `range` in our Standard Nodes specification currently describes `doclamp` as follows:
> 
> > If true, the output is clamped to [`outlow`, `outhigh`]
> 
> With this change, the output is clamped to the range spanned by those two values regardless of their order, so we should update the specification to something more like the following:
> 
> > If true, the output is clamped to the range between `outlow` and `outhigh`, regardless of which is larger
> 
> Other than that detail, I think this PR looks ready to go.

Good point, updated here https://github.com/AcademySoftwareFoundation/MaterialX/commit/c5f67460abcebe7a72ef9c79299b84075643f8df#diff-50ef1c78b280aa9cf648261c862cf45e53540f243a92910195471595d54defccR1631-R1634

Your wording makes sense, so have used it verbatim.

jstone-lucasfilm (approved):
Thanks for the fix, @seang-ilm!

Issue #3068, opened 2026-09-12
Out of range access on empty swizzle


seang-ilm:
Another edge case on the upgrade path.

If you add this test case to `MaterialXTest/MaterialXCore/Document.cpp` and run with `MATERIALX_DYNAMIC_ANALYSIS=ON`.

``` c++
TEST_CASE("Document upgrade swizzle empty value", "[document]")
{
    mx::DocumentPtr doc = mx::createDocument();
    mx::XmlReadOptions readOptions;
    readOptions.upgradeVersion = false;
    mx::readFromXmlString(doc,
        "<?xml version=\"1.0\"?>"
        "<materialx version=\"1.38\">"
        "  <nodegraph name=\"NG1\">"
        "    <swizzle name=\"swz1\" type=\"color3\" nodedef=\"ND_swizzle\">"
        "      <input name=\"in\" type=\"color3\" value=\"\"/>"
        "      <input name=\"channels\" type=\"string\" value=\"\"/>"
        "    </swizzle>"
        "  </nodegraph>"
        "</materialx>",
        mx::FileSearchPath(), &readOptions);
    REQUIRE_NOTHROW(doc->upgradeVersion());
}
```

```
/usr/include/c++/11/bits/stl_vector.h:1046:34: runtime error: reference binding to null pointer of type 'struct value_type'
```

It's specifically on the upgrade path for 1.38-1.39, that converts old swizzle nodes to constant nodes.

The `in` input has an empty value `value=""`.  This means `inInput->hasValue();` returns true and we go down path

``` c++
convertToConstantNode = true; // because inInput exists and has a value attribute
...
if (convertToConstantNode)
{
    ...
    string valueString = inInput ? inInput->getValueString() : "0"; // <-- ""
    StringVec origValueTokens = splitString(valueString, ARRAY_VALID_SEPARATORS); // <-- empty vector {}
    ...
    newValueTokens.push_back(origValueTokens[0]); // <-- out of range access here
```

https://github.com/AcademySoftwareFoundation/MaterialX/blob/8b4b22247cd92c67b1daf34eb0151a97282453ec/source/MaterialXCore/Version.cpp#L1264-L1288

kshivam4781:
Opened a PR for this: #3069

PR #3069, opened 2026-09-12
Fix null/out-of-range access in the switch and swizzle version upgrade

### LINKED ISSUE #3067: nulldref possible on Switch block
A tidied up fuzz finding. This one's on a highly infeasible path.

You'll only see it triggered if you have `MATERIALX_DYNAMIC_ANALYSIS=ON`.

If you add this test case to `MaterialXTest/MaterialXCore/Document.cpp`

``` c++
TEST_CASE("Document upgrade switch null value", "[document]")
{
    mx::DocumentPtr doc = mx::createDocument();
    mx::XmlReadOptions readOptions;
    readOptions.upgradeVersion = false;
    mx::readFromXmlString(
        doc,
        R"(<?xml version="1.0"?>
            <materialx version="1.38">
            <nodegraph name="NG1">
                <switch name="sw1" type="integer" nodedef="ND_switch">
                    <input name="which" type="integer" value="abc"/>
                </switch>
            </nodegraph>
            </materialx>
            )",
        mx::FileSearchPath(),
        &readOptions
    );
    REQUIRE_NOTHROW(doc->upgradeVersion());
}
```

```
/home/sean/MaterialX/source/MaterialXCore/Version.cpp:1171:45: runtime error: member call on null pointer of type 'struct element_type'
```

It looks like it's specifically in the upgrade path for 1.38-1.39. When `which` can't be parsed it returns `null`.

I think it just needs a null check before accessing it in this block.

``` c++
void Document::upgradeVersion()
...

    // Upgrade from 1.38 to 1.39
    if (majorVersion == 1 && minorVersion == 38)

            else if (nodeCategory == "switch")
            {
                // Upgrade switch nodes from 5 to 10 inputs, handling the fallback behavior for
                // constant "which" values that were previously out of range.
                InputPtr which = node->getInput("which");
                if (which && which->hasValue())
                {
                    auto whichValue = which->getValue(); // <-- returns null
                    if (whichValue->isA<int>() && whichValue->asA<int>() >= 5) // <-- nulldref here
```

https://github.com/AcademySoftwareFoundation/MaterialX/blob/8b4b22247cd92c67b1daf34eb0151a97282453ec/source/MaterialXCore/Version.cpp#L1190-L1207

kshivam4781:
Fixes #3067 and #3068.

Both are fuzz findings from @seang-ilm on the same code path: `Document::upgradeVersion()`'s 1.38-to-1.39 step for `switch` and `swizzle` nodes.

**#3067 (nulldref on switch):** `which->getValue()` returns null when the stored value string can't be parsed as the input's declared type (e.g. `value="abc"` on an `integer` input). The existing code called `whichValue->isA<int>()` without checking for that, so a bad `which` value crashes the upgrade instead of just being left alone. Fix adds a null check before both `isA`/`asA` calls.

**#3068 (out-of-range on swizzle):** when a `swizzle` node's `in` input has an empty value (`value=""`), `origValueTokens` ends up empty, but the fallback branch for an invalid/missing channel name unconditionally did `origValueTokens[0]`. Fix falls back to `"0"` when `origValueTokens` is empty, matching the existing convention in this same function of treating a missing/unparseable value as `0`.

Both issues include a ready-made `TEST_CASE` repro; I added them verbatim to `MaterialXTest/MaterialXCore/Document.cpp`.

### Validation

- Built with `MATERIALX_BUILD_TESTS=ON` and `MATERIALX_DYNAMIC_ANALYSIS=ON` (ASan+UBSan), GCC 13, `MATERIALX_BUILD_GEN_GLSL=ON` (required by `MaterialXTest`'s GenShader test), everything else generation/render-related off.
- Red-before: reverted just the `Version.cpp` fix and reran the two new tests — both reproduce the exact runtime errors quoted in the issues (`member call on null pointer of type 'struct element_type'` for #3067, `reference binding to null pointer of type 'struct value_type'` for #3068).
- Green-after: with the fix restored, both new tests pass under ASan+UBSan.
- Full local test binary (Core/Format/GenShader/GenGlsl/GenHw scope): **14427 assertions in 69 test cases, all passed**, no new sanitizer trips anywhere else in that scope.
- `clang-format --dry-run` clean on the lines this PR touches (pre-existing formatting drift elsewhere in `Version.cpp`/`Document.cpp` left untouched, per the contributing guide's "make unrelated changes in a separate PR").

I didn't build `MATERIALX_BUILD_RENDER`/the other shader-gen backends or the full CI matrix locally, since neither touched file is in that path.


jstone-lucasfilm:
Thanks for this proposal, @kshivam4781!  Can you add a comment on both of the GitHub Issues that you're tackling, so that we can assign them to you?  Additionally, see the instructions at https://github.com/AcademySoftwareFoundation/MaterialX/pull/3069#issuecomment-5648494646 on how to resolve the CLA authorization warnings, so that we can begin reviewing this PR.

jstone-lucasfilm:
Thanks for taking this on, @kshivam4781!  This PR looks very close to the mark as it stands, though I have two suggestions that I believe would strengthen the proposal before we merge:

1. The `switch` fix can be expressed as a single guard.  Since `ValueElement::getValue` already returns a null pointer when no value is present, the existing `hasValue` check becomes redundant once we test the pointer itself, and the repeated `whichValue &&` conditions can be removed:

   ```cpp
   InputPtr which = node->getInput("which");
   ValuePtr whichValue = which ? which->getValue() : nullptr;
   if (whichValue)
   {
       if (whichValue->isA<int>() && whichValue->asA<int>() >= 5)
       {
           which->setValue(0);
       }
       else if (whichValue->isA<float>() && whichValue->asA<float>() >= 5)
       {
           which->setValue(0.0);
       }
   }
   ```

2. Both of these issues are _undefined behavior_ rather than thrown exceptions, so `REQUIRE_NOTHROW` would only catch a regression in our dynamic analysis build, and on other platforms the tests may pass or crash depending on the compiler.  I'd suggest adding a check on the upgraded document in each test, which would give us coverage in every build and also document the fallback behavior that you've chosen.  For the `swizzle` case, that might look like the following:

   ```cpp
   mx::NodePtr node = doc->getNodeGraph("NG1")->getNode("swz1");
   REQUIRE(node->getCategory() == "constant");
   REQUIRE(node->getInput("value")->getValueString() == "0, 0, 0");
   ```

   For the `switch` case, a check that the `which` input retains its original value string would serve the same purpose.

One smaller note:

- Since the two test cases were taken from separate issues, one uses an unescaped string literal for its document and the other uses escaped concatenation.  I'd suggest unescaped strings for both, which are easier to read and match the direction we've taken in other recent PRs to MaterialX.

kshivam4781:
Thanks @jstone-lucasfilm! Pushed both suggestions:

- a071eb8: collapsed the switch fix to the single guard you gave, verbatim.
- 39ff484: both tests now assert real post-upgrade state instead of just REQUIRE_NOTHROW (the swizzle assertions are exactly what you suggested; the switch test checks that `which` keeps its original "abc" value string), and the swizzle test's document literal is now an unescaped raw string, matching the switch test.

Full local run: `bin/MaterialXTest` -- 74 test cases, 17815 assertions, all passed (Debug, `MATERIALX_BUILD_TESTS=ON`, GenGlsl on, other gen backends/render/python off). `clang-format --dry-run` clean on every line both commits touch.

jstone-lucasfilm:
Thanks for the quick turnaround, @kshivam4781.  Both changes look good, and I appreciate the added assertions on the upgraded documents, which will give us coverage of these paths in all builds rather than only in dynamic analysis.

One small refinement before we merge: the new comment in the `switch` block ends by noting that the `hasValue()` check above is now redundant, but that check has been removed in this same change, so the comment now describes the history of the code rather than its current state.  I'd suggest reducing it to a single line that states the property we rely on, along the lines of the following:

```
// getValue() returns null for both a missing and an unparseable value.
```

With that minor comment update, I believe this PR is ready to merge.  Thanks again for taking this on!

jstone-lucasfilm (approved):
Great work, thanks @kshivam4781!

PR #3072, opened 2026-09-15
Reapply target-specific defaults after replacing context options

jreichel-nvidia:
The option distributeLayerOverBsdfMix for genmdl got lost in dc337ca.


jstone-lucasfilm (approved):

PR #3074, opened 2026-09-16
Apply thin-film to the dielectric transmission lobe in OpenPBR Surface

### LINKED ISSUE #3026: Thin film is not applied to the transmission lobe in OpenPBR Surface
As @niklasharrysson notes in his [review of #2998](https://github.com/AcademySoftwareFoundation/MaterialX/pull/2998#discussion_r3744921838), the OpenPBR specification states, in [Thin-film iridescence](https://academysoftwarefoundation.github.io/OpenPBR/index.html#model/thin-filmiridescence):

> In the case of a dielectric base, the thin film should also generate color fringes in the transmission lobe. This is important, for example, when rendering soap bubbles.

In our nodegraph implementation, thin-film variants are provided for the [dielectric reflection](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/open_pbr_surface.mtlx#L385-L394) and [metal](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/open_pbr_surface.mtlx#L418-L427) lobes, blended in by `thin_film_weight`, but the [dielectric transmission lobe](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/bxdf/open_pbr_surface.mtlx#L360-L367) has no thin-film counterpart, so transmission through the surface generates no color fringes.

Since `<dielectric_bsdf>` accepts its thin-film inputs independently of `scatter_mode`, one proposed fix would be to add a thin-film variant of the transmission lobe, blended in by `thin_film_weight` in the same manner as the reflection lobes, updating the functional definition and removing the corresponding note in the PBR specification at the same time.

The reference implementation in the OpenPBR repository shares this construction, and we would propagate any fix made here to that project as well.

hikmetba-bit:
## Summary
Fixes #3026.

The dielectric reflection lobe (`dielectric_reflection` / `dielectric_reflection_tf`) and the metal lobe (`metal_bsdf` / `metal_bsdf_tf`) each have a thin-film variant blended in by `thin_film_weight`, but the dielectric transmission lobe (`dielectric_transmission`) had no thin-film counterpart, so transmission through the surface (e.g. a soap bubble) generated no color fringes, as described in the issue.

## Fix
Added `dielectric_transmission_tf`, identical to `dielectric_transmission` but with `thinfilm_thickness`/`thinfilm_ior` set (mirroring how `dielectric_reflection_tf` relates to `dielectric_reflection`), and a `dielectric_transmission_tf_mix` node that blends the two by `thin_film_weight`, the same pattern used for the reflection lobe. `dielectric_volume_transmission`'s `top` input now points at the mixed node instead of the bare `dielectric_transmission` node. Everything downstream (`dielectric_volume_transmission`, `dielectric_substrate`, `dielectric_base`) is unchanged.

Since MaterialX's `mix` node computes `fg*mix + bg*(1-mix)`, `mix=0` evaluates to exactly `bg`, so this is a no-op whenever `thin_film_weight` is 0 (the default) — existing materials that don't use thin-film render identically.

## Test plan
- Could not build/render in this environment (no CMake/C++ toolchain available), so I could not run MaterialX's shader-generation validation suite directly.
- Validated the edited file is well-formed XML (`xml.etree.ElementTree.parse`).
- Traced every node/interface reference in the new nodes (`if_transmission_tint`, `modulated_eta_s`, `main_roughness`, `geometry_normal`, `geometry_tangent`, `thin_film_thickness_nm`, `thin_film_ior`, `thin_film_weight`) back to their existing definitions in the same file to confirm they exist and have the expected types.
- Found `resources/Materials/Examples/OpenPbr/open_pbr_soapbubble.mtlx`, an existing example that sets `transmission_weight="1.0"` and `thin_film_weight="1.0"` — exactly the soap-bubble scenario the issue names. Confirmed `geometry_thin_walled` (also set `true` there) only affects the separate diffuse subsurface selection (`opaque_base`/`selected_subsurface`), not the `dielectric_transmission` lobe, so this example does exercise the fixed path. Its rendered output is expected to change now that transmission correctly picks up thin-film color fringes — this is the intended fix, not a regression, but I could not re-render it to confirm visually.
- Did not modify the OpenPBR specification document/reference implementation in the separate `AcademySoftwareFoundation/OpenPBR` repo mentioned in the issue; this PR only covers the MaterialX nodegraph implementation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jstone-lucasfilm:
Thanks for this proposal, @hikmetba-bit!  Can you add a comment on the GitHub Issue that you're tackling, so that we can assign it to you?  Additionally, see the instructions at https://github.com/AcademySoftwareFoundation/MaterialX/pull/3074#issuecomment-5701890571 on how to resolve the CLA authorization warnings, so that we can begin reviewing this PR.

Issue #2948, opened 2026-05-23
Windows MinGW shared build broken


brechtsanders:
Since version 1.39.5 building with `-DBUILD_SHARED_LIBS:BOOL=ON` on Windows (in my case with GCC+MinGW-w64) fails.

The reason is that in `source/MaterialXCore/Value.h` there is a template class
```c++
template <class T> class TypedValue : public Value
```
whose functions are defined with `MX_CORE_API`, which is in that case defined as `__declspec(dllexport)`.

This results in the following build errors:
```raw
In file included from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Element.h:16,
                 from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Geom.h:14,
                 from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Interface.h:14,
                 from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Definition.h:14,
                 from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Node.h:14,
                 from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Material.h:14,
                 from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Look.h:14,
                 from R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Document.h:14:
R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Value.h:130:17: error: function 'MaterialX_v1_39_5::TypedValue<T>::TypedValue()' definition is marked dllimport [-Wtemplate-body]
  130 |     MX_CORE_API TypedValue() :
      |                 ^~~~~~~~~~
R:/winlibs16.1.0msvcrt64/MaterialX-1.39.5/source/MaterialXCore/Value.h:134:26: error: function 'MaterialX_v1_39_5::TypedValue<T>::TypedValue(const T&)' definition is marked dllimport [-Wtemplate-body]
  134 |     MX_CORE_API explicit TypedValue(const T& value) :
      |                          ^~~~~~~~~~
[...]

PR #3075, opened 2026-09-16
Fix building shared libraries on Windows using MinGW64

SavaLione:
### Summary
Hi,
There was an issue to building MaterialX on a Windows system using the MinGW64 compiler:
* #2948 

This PR fixes this compilation issue, and the dynamical libraries are now successfully build.

The referenced issue suggests the addition of new GitHub CI entries.
However, this PR does not include them.
I can work on the CI implementation if it is still required and the additional information is provided:
* How is it acceptable get the MinGW binaries (e.g., whether binaries provided by Stephan T. Lavavej are acceptable)
* Which additional tools may a container download and use (is it possible to use bash or ninja?)
* What is the preferred way to cache these binaries?

### Tests
Tests were successfully compiled and executed.
All 94 tests cases passed (39629 assertions).


jstone-lucasfilm:
Thanks for this proposal, @SavaLione!  Can you add a comment on the GitHub Issue that you're tackling, so that we can assign it to you? Additionally, see the instructions at https://github.com/AcademySoftwareFoundation/MaterialX/pull/3075#issuecomment-5705198126 on how to resolve the CLA authorization warnings, so that we can begin reviewing this PR.

SavaLione:
The CLA was signed via EasyCLA, a comment was also added to the issue

jstone-lucasfilm:
Apologies, @SavaLione, another developer requested this one first, but feel free to request any other unclaimed issue for Dev Days.

SavaLione:
Yeah, sure, I should've checked the full issue conversation earlier.
It still may be useful to the person who initially requested this issue first.
And there a new workflow has to be added, so the current PR is not fully covering all requirements from the issue.

jstone-lucasfilm:
Thanks for understanding, @SavaLione, and indeed the other developer may find this work useful for their task.  I've added a few more Good First Issues to our list for Dev Days, if you'd like to take on another task for the event tomorrow.

cod1r:
I think I will add to this PR and finish out the CI stuff if that's ok.

I don't want to throw away mister Sava's good work here :)

jstone-lucasfilm:
Much appreciated, @cod1r, and I think that's a great path forward.

jstone-lucasfilm (approved):
This change looks correct to me, thanks @SavaLione, and let's move forward with the merge, so that the next steps proposed in https://github.com/AcademySoftwareFoundation/MaterialX/pull/3081 can be validated.

Issue #3004, opened 2026-07-12
Standard Surface with colors outside of 0-1 renders incorrectly


joaovbs96:
This Standard Surface material from [physicallybased.info](https://physicallybased.info), the Gold material, has some color3 values beyond the 0-1 range, namely the base color and specular color.

```
<?xml version="1.0"?>
<materialx version="1.38" colorspace="lin_rec709" doc="Downloaded from https://physicallybased.info on 2026-07-10">
  <standard_surface name="SR_Gold" type="surfaceshader">
    <input name="base_color" type="color3" value="1.059, 0.773, 0.307" />
    <input name="specular_color" type="color3" value="1.001, 0.985, 0.523" />
    <input name="specular_roughness" type="float" value="0" />
    <input name="metalness" type="float" value="1" />
  </standard_surface>
  <surfacematerial name="Gold" type="material">
    <input name="surfaceshader" type="surfaceshader" nodename="SR_Gold" />
  </surfacematerial>
</materialx>
```

The material renders with a green circle in the middle. I've observed this on both GLSL backends (MaterialXView, MaterialXEditor and the webviewer):

<img width="300" height="auto" alt="Image" src="https://github.com/user-attachments/assets/397188d0-0e8b-4e03-b865-25e56a2c96ad" />

and MDL:

<img width="300" height="auto" alt="Image" src="https://github.com/user-attachments/assets/1176bfa5-8937-4449-8342-7a9698819188" />

[Per discussion in the Slack channel](https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1783684488714539?thread_ts=1783680653.424079&cid=C0230LWBE2X), Jamie Portsmouth mentioned:

> I think this is because Standard Surface uses the [Gulbrandsen](https://jcgt.org/published/0003/04/03/paper.pdf) model for its metallic Fresnel (i.e. maps base and spec color to n, k), and this gives NaNs if the input colors are outside the [0,1] range. The fix would just be to clamp the channels before computing n, k.
> It won't occur in OpenPBR (even if you supply color channels exceeding 1) since we use a different metallic Fresnel. But also we [said](https://github.com/AcademySoftwareFoundation/OpenPBR/pull/208) the colors are assumed to be clamped anyway.

Note that, while I mentioned GLSL and MDL, it's possible this also happens in other backends as well, but I have no way of testing them.

SavaLione:
Hi,
[...]

PR #3078, opened 2026-09-17
Fix edge color clamping in `artistic_ior`

SavaLione:
### Summary
Hi,

This fixes the issue related to the green circle on some materials:
* #3004

Three shaded backends were changed and tested: GLSL, OSL, and MDL.

I'm somewhat sure about the code for the GLSL and OSL shaders.
With the MDL shader, I'm not sure whether it is considered acceptable practice to add functions directly to the XML file.
Nevertheless, all tests completed successfully, and everything looks good to me (including the MDL backend).

### Tests
I compiled MaterialX Viewer as a standalone application and as a web application using Emscripten.
The MDL backend was tested using the Nvidia MDL SDK (version 2026.0.2-391700.2276) and the supplied dxr.exe application in the examples.
Visually, everything looks OK.

Tests were successfully compiled and executed.
All 94 test cases passed (39629 assertions).

MaterialX Viewer web:
image

MDL before the fix:
image

MDL after the fix:
image


jstone-lucasfilm:
Thanks for taking this on, @SavaLione!  This looks like a good fix, and the GLSL and OSL changes are exactly what I'd expect.  I have one request on the MDL side before we merge, along with a suggestion on test coverage.

On the MDL question you raised, this logic belongs in the MDL module rather than the implementation XML.  The function body for `mx_artistic_ior` lives in `source/MaterialXGenMdl/mdl/materialx/pbrlib_1_6.mdl`, where it already clamps `mxp_reflectivity` in the same manner as the GLSL and OSL code, so the inline `math::clamp` on `reflectivity` in this PR is redundant with that existing line.  I'd suggest leaving `pbrlib_genmdl_impl.mtlx` as it is on `main`, and instead applying the new clamp in the function itself, alongside the existing one:

```
color n = math::lerp(n_max, n_min, math::clamp(mxp_edge_color, color(0.0), color(1.0)));
```

Since each of the later versioned modules (1.7 through 1.11) re-exports this function from `pbrlib_1_6`, that one line covers every MDL version we generate.

On test coverage, none of our current test materials drive `artistic_ior` outside the [0, 1] range, so our render tests wouldn't catch a regression of this fix.  I'd suggest adding a case with an `edge_color` above 1.0 to `resources/Materials/TestSuite/pbrlib/bsdf/conductor.mtlx`, which would exercise this path across all of our shading languages.

Overall this looks like a great second contribution, and with the refinements above, I believe it should be ready to merge.

SavaLione:
Hi @jstone-lucasfilm ,

Thanks for the helpful information about the project's internal structure and how the MDL backend works.
Now I'm finally able to see where the MDL shader generation happens, and it definitely doesn't make sense to edit the XML files to fix the clamp issue (as you pointed out).

I have reverted all changes to the XML file and applied your suggested change to `pbrlib_1_6.mdl`.

Also, I implemented a test case that checks this particular issue:
image

The values for this test case are taken from the material mentioned in the initial issue (#3004), so it is possible to visually inspect the clamping on the gold surface.
If some tests use pixel diff, or if anyone needs to see a more noticeable difference, I can push the values outside the [0, 1] range.

jstone-lucasfilm (approved):
Excellent work, @SavaLione, and this looks ready to merge.

Issue #3039, opened 2026-08-16
MaterialX Graph Editor pin preview to node


HardCoreCodin:
Support pinning the preview render to a specific node in the graph to always preview its result, even as choosing and interacting with other nodes in the graph.

MatthieuBEAUD:
Hello, I'd like to take a look at this for the september 2026 dev days. It that possible ?

jstone-lucasfilm:
Welcome, @MatthieuBEAUD, and I've assigned it to you!

PR #3079, opened 2026-09-17
GraphEditor: Add pin/lock preview to node feature

MatthieuBEAUD:
## Issue

https://github.com/AcademySoftwareFoundation/MaterialX/issues/3039

## Description

Added a feature to be able to lock with node is used for the preview render:

Added:
- Key binding: R (for Render, and to match similar keybinding in Houdini Solaris), but it could be P for Preview or L for Lock
- Help text in menu: under Help/Graph/Viewing
- Color on the UI node header to highlight the currently used render preview node (when locked only)
- function to update the renderNode (check if locked, and if new value different than the current value) and recompile the material (factorized since it was used in 4 places)

Notes:
- I used the "lock" word instead of "pin" in the variable names and comment to avoid confusion with the node pins (inputs).
- I did not add any button, as I thought it would be out of scope. I also figured that it's something that should be added on top of https://github.com/AcademySoftwareFoundation/MaterialX/pull/3043, like adding a button next to, or in, the hamburger menu. Should another issue be added for this ?
- Currently, when diving in a subgraph, the lock/pin remains on the upper level, maybe some UX could indicate that a pin/lock is still active, or maybe we should disable the pin when diving in a subgraph ? Let me know if that should be something to fix/dig into.

## Demo

materialx_preview_render_pin_demo_2


MatthieuBEAUD:
<a href="https://easycla.lfx.linuxfoundation.org/#/?version=2">CLA Signed</a><br >The committers listed above are authorized under a signed CLA.<ul><li>:white_check_mark: login: MatthieuBEAUD / name: Matthieu Beaud (3a128422e8cae56951a88add7f33e8f832b32911, d3d3c1cc6c175e7fbce6f946fd8fe1f05303d97a)</li></ul>

jstone-lucasfilm:
@MatthieuBEAUD For the CLA authorization, take a look at the instructions in the comment on resolving these warnings, and the Linux Foundation is usually very responsive in helping out with this.

kwokcb:
This looks like a nice clean way to add in this functionality. 

I have a  few workflow cases that would be good to check / resolve:

1. What happens if you  lock a node and then delete that node. Does the lock node state get cleared ?
2. What happens if the node you lock cannot be rendered? It is possible to occur.
3. I think this will work for nodes inside a `functional` graph ( node graph used to represent a definition or `nodedef`).
I'm not sure if what the behaviour should/would be there and if it's desired. It seems like a nice feature but would be
good to clarify what is expected here. 

For both compound (non definition graphs) and funcation graphs  for when you dive into the graph, lock a graph node and then go out of it.  I can suggest 2 options:
a. Instead of relying on the node being visible. Add the node name / path to the render panel instead. 
b. Add the visual cue to the parent graph.

Both may be useful even when you scroll the graph so that locked node is no longer visible. a. can be useful longer term as well to allow filtered selection of node / node outputs to render.

Think b. is the simplest to add. 

Thanks.

jstone-lucasfilm:
Thanks for this proposal, @MatthieuBEAUD!  See the instructions at https://github.com/AcademySoftwareFoundation/MaterialX/pull/3079#issuecomment-5717252677 on how to resolve the CLA authorization warnings, so that we can begin reviewing this PR.

MatthieuBEAUD:
> This looks like a nice clean way to add in this functionality.
> 
> I have a few workflow cases that would be good to check / resolve:
> 
> 1. What happens if you  lock a node and then delete that node. Does the lock node state get cleared ?
> 2. What happens if the node you lock cannot be rendered? It is possible to occur.
> 3. I think this will work for nodes inside a `functional` graph ( node graph used to represent a definition or `nodedef`).
>    I'm not sure if what the behaviour should/would be there and if it's desired. It seems like a nice feature but would be
>    good to clarify what is expected here.
> 
> For both compound (non definition graphs) and funcation graphs for when you dive into the graph, lock a graph node and then go out of it. I can suggest 2 options: a. Instead of relying on the node being visible. Add the node name / path to the render panel instead. b. Add the visual cue to the parent graph.
> 
> Both may be useful even when you scroll the graph so that locked node is no longer visible. a. can be useful longer term as well to allow filtered selection of node / node outputs to render.
> 
> Think b. is the simplest to add.
> 
> Thanks.

Thank you very much for this feedback ! 

1. I indeed had forgotten about what happens when you delete the locked preview render node. It now disables the lock.
2. When you enable the lock mode, it stores the currRenderNode, so you cannot lock a non-renderable node (this feature never uses the current node selection). And in that case, if we add a button next to a hamburger menu, this button would only need to be present on renderable nodes.   But maybe you meant a renderable node that can't be rendered again ?
3. Good points, I'll think about those, and what the visual cue could be in specific cases:
- if we select a node in the root graph, and then dive in a child, we could have a cue in the top graph path, to show the preview is currently set in the parent
- if the locked preview node is outside of the frame, we could have an arrow cue of that color pointing to the direction on the edge of the frame
- if we lock a node in a subgraph and then go out of it, the cue could indeed simply transfer to the parent node

All that being said, maybe adding the name / path of the current preview node to the viewport would be the more robust way afterall.

I see when I can find time to work on point 3.

Thanks again 🙏

MatthieuBEAUD:
I added the current render node path display above the viewport

Here are some examples:

image

image

image

image

Let me know what you think of this

kwokcb:
Maybe let's keep it simple.

1. Show the locked node name in the render panel as you have done.
2. Hilight the child or parent nodegraph if locking a child node.

If someone scrolls the node out of view then 1) handles that case.

Disallow functional node graph child node locking for now for this PR. There should be a global state that tells you this -- the graph is marked as read-only.

Not all nodes chosen can have its upstream graph or itself rendered. For this PR let's just leave it as rendering nothing, and handle either prevention or fallback as a separate item.

So I think the only addition is to check for read-only graph state.

MatthieuBEAUD:
Thanks for the feedback

I'll try and find a few hours to highlight the parent/child node of the current lock selection, and check of readonly state

Thank you !

MatthieuBEAUD:
I added the check for read-only graph to not allow for the lock of the render preview node in that case.

And in a separate commit the highlight of the parent if the locked currRenderNode is wihthin its children (I also had to check for the nodegraphs in that case). But it could get a bit confusing as you don't know if that node is currently being viewed, or one of its children, that might not be the same as the preview of the node itself.
Also, I'm not sure what should be highlighted if you're in the children and the locked is somewhere in the parents (since there is no node that references the parent, or maybe use the graph hierarchy on top of the graph ?), I suppose the path over the viewport is the reference in that case, right ? 

Let me know if the highlight of the parent should be kept, or if it's redundant with the path over the viewport and we should only keep that text cue for out of scope/sight locked preview node.

Here is a quick demo of the highlight of the parent of the locked render node:
materialx_preview_render_pin_demo_3

kwokcb:
Suggestion

Parent / child indicators exist and we can use title bar color for now.
So the only small suggestion is that if it's at the parent level to dim or mute the "lock" colour so there is some differentiator. (Actually if you pick the output node on a graph it is the graph that is the target but that's "splitting hairs").

So something like this: (apologies in advance but this is AI generated slop, but shows the basic rule set :))

image

Thought blurb:

1. I tried to look for existing shader graph workflows and the closest I could find for the usage of parent/child indicators was in Houdini and Katana. Nuke also has some parent/child state, but couldn't find anything for Maya, Blender.
2. If a Houdini expand/collapse graph interaction was added vs this editor's complete context switch with single graph display then I think that would be a match.

From what I've seen having both a title color cue and and icon are both useful with a icon addition post #3043 PR.

Does this sound okay?

MatthieuBEAUD:
Yes perfectly clear, I muted the color for when the child is selected then 👌 and we'll wait post https://github.com/AcademySoftwareFoundation/MaterialX/pull/3043 to add the icon

I also changed the color to match your example palette, as the dimmed yellow-ish felt confusing and too brown, and didn't feel like a muted highlight color.

here are the results:

dimmed (the child node is the current lock preview): 
image

main (the node itself is the current lock preview):
image

PR #3080, opened 2026-09-17
Fix drag-and-drop for local files in the Web Viewer

### LINKED ISSUE #3008: Regression: Web Viewer drag and drop no longer works
## Issue

The MaterialX Web Viewer no longer handles drag and drop of materials / geometry anymore.
Local build shows this is still the ase.

## Log

Some snapshots of material d+d from browser console for released :

Image

and local build:

Image

Does not work for geometry either:

Image

mahdavifar2002:
Fixes #3008

When dropping a local material (.mtlx) or geometry (.glb) file into the viewer, the browser attempted to fetch it via a relative URL to the server, resulting in a 404 Not Found error.

This updates the drop callbacks to asynchronously resolve the `FileEntry` into a `File` object, and generates a temporary Blob URL using `URL.createObjectURL()` to successfully load the assets.

**Notes**
* I tested locally and verified that the 404 error is resolved, both when a material or a geometry file is dropped.
* I'm submitting as part of ASWF Dev Days, Fall 2026. This is my first open-source PR. Thanks for reviewing it!


kwokcb:
Adding @hybridherbst to review as he wrote the original code. Felix can you take a look if you have some time?

Still not sure why this broke.

For checks @mahdavifar2002 could you check zip drag and drop as well. Also if the file name is already from a http address does this code still work as you always try to create a url. Not sure if this is general load code or d+d specific so just double checking.

Thanks.

darchangell:
This seems more complex than the change I needed to get drag and drop working again. For me you can see the three locations where THREE.cache.add() is called in my update for image handling. I needed only add file type to the image name like "file:" or "image:".
https://github.com/AcademySoftwareFoundation/MaterialX/pull/3082

mahdavifar2002:
I agree with @darchangell. I didn't pay attention to the `THREE.cache.add()` calls and how they are intended to work. Let's close this PR.

jstone-lucasfilm:
That approach sounds logical to me, @mahdavifar2002 and @darchangell.  I'll close this out in favor of #3082, and we can pick up the conversation there.

PR #3081, opened 2026-09-17
Add a MinGW build to GitHub CI

### LINKED ISSUE #2948: Windows MinGW shared build broken

cod1r:
@SavaLione already had commits to fix the actual build errors, I'm just adding CI stuff to the workflow files.

closes #2948


jstone-lucasfilm:
Thanks for taking this on, @cod1r!  I've looked through the CI failure in the new MinGW job, and it seems unrelated to the symbol export macros from #3075, which are working as intended.

The build stops in `StbImageLoader.cpp` with a set of `unused-function` errors from the stb headers.  We already suppress that warning for MaterialXRender, but the guard in `source/MaterialXRender/CMakeLists.txt` is `if(UNIX)`, so MinGW receives the full GCC warning set without the corresponding suppressions.  Since these flags apply to GCC and Clang on any platform, I'd suggest keying the block on the compiler rather than the OS:

```
if(NOT MSVC)
    target_compile_options(${TARGET_NAME} PRIVATE -Wno-unused-function)
    if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
        target_compile_options(${TARGET_NAME} PRIVATE -Wno-stringop-overflow)
    endif()
endif()
```

Here are a few additional proposed refinements to the workflow, so that the job succeeds once the compile issue is resolved:

1. The Install MinGW step can be removed entirely.  The `windows-2022` image already ships GCC 14.2.0 in `C:\mingw64`, and that's the compiler CMake actually selected in the failed run, since Chocolatey's changes to PATH don't propagate to later steps.  As written, the step also runs on our four MSVC jobs, adding about a minute to each.
2. Drop `upload_shaders: ON` from the new entry.  Our MSVC shared library job already uploads an artifact named `MaterialX_ReferenceShaders`, and upload-artifact refuses a second upload to the same name, so this step would fail once the build passes.
3. The Setup Build Environment (Windows) step sets `CMAKE_BUILD_PARALLEL_LEVEL=1`, which is intended for MSVC's `/MP` but serializes a Ninja build.  The failed run took seven minutes to reach object 96 of 339, so I'd suggest restricting that setting to the Visual Studio generator.
4. Since Visual Studio isn't involved in this job, a name such as `Windows_MinGW_GCC14_SharedLibs` would align better with our existing naming scheme.

Additionally, now that #3075 has landed, a rebase on main should reduce this PR to just the workflow change.

With these refinements, I believe this PR should be ready to merge, giving us the MinGW validation that #2948 requested.  Let me know if you have any questions!

jstone-lucasfilm:
Thanks for the quick turnaround, @cod1r.  The new GitHub CI job is looking good, and it's great to see the MinGW build complete and pass all tests in about nine minutes.

One proposed refinement on the CMake side: my earlier CMake proposal was intended for `source/MaterialXRender/CMakeLists.txt`, replacing the existing `if(UNIX)` block there, rather than the root `CMakeLists.txt`.  As currently written, `-Wno-unused-function` is disabled for every non-MSVC build of every MaterialX target, and `-Wno-stringop-overflow` is disabled project-wide on GCC, whereas both were previously scoped to the targets that wrap third-party code.  The stringop-overflow diagnostic in particular catches real buffer misuse in first-party code, so I'd prefer to keep it enabled outside of MaterialXRender.

The unused-function errors in the failed run were confined to the stb headers included by `StbImageLoader.cpp`, so a MaterialXRender-only suppression should be sufficient.  The complete change would be to revert the root `CMakeLists.txt` addition and update the guard in `source/MaterialXRender/CMakeLists.txt` as follows:

```
if(NOT MSVC)
    target_compile_options(${TARGET_NAME} PRIVATE -Wno-unused-function)
    if(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
        target_compile_options(${TARGET_NAME} PRIVATE -Wno-stringop-overflow)
    endif()
endif()
```

Two smaller suggestions for the workflow:

1. The `using_ninja_with_mingw` matrix key can be dropped, since the condition we want is really "Visual Studio generator", and that can be derived from the existing `cmake_config` entry:

   ```
   if ("${{ contains(matrix.cmake_config, 'Visual Studio') }}" -eq "true") { Add-Content $env:GITHUB_ENV "CMAKE_BUILD_PARALLEL_LEVEL=1" }
   ```

2. Since #3075 has been merged and this branch is already based on it, the PR description could be updated to reflect that it now consists of the new CI job and the warning guard.

With those changes I believe this looks ready to merge.  Thanks again for taking this on!

cod1r:
I basically got carried on this PR 🤣  but thanks for walking me through this Jonathan Stone!

jstone-lucasfilm (approved):
Great work, @cod1r and @SavaLione, and thanks for bringing this to the finish line!

Issue #1640, opened 2024-01-05
Web Viewer: Add support for drag-and-drop of materials with image references


kwokcb:
## Issue

If you drag-drop a MTLX file with image dependencies the image dependencies cannot be found to be bound.
Seems to be the case for 

1. Image files in same folder as MTLX file
2. Image files in sub-folders of MTLX file
3. Image files in sibling or parent folders

Perhaps this is due to the usage of relative file paths, though would think that 1 would work.

## Diagnosis

Found while loading and examining the console log in a browser.

Avgilles:
Can I pick this issue for the dev day 2024 ?

jstone-lucasfilm:
Absolutely, and welcome!

kwokcb:
As an FYI, AFAIK this could just not be feasible to do as you can't access local file paths due to security. Thus drag-and-drop supports zip files load. If there is a way @Avgilles, I'd be very interested in how to set this up :).

Avgilles:
Hello, is there is a way to have resolve path in material X ?
I see here that the path is get in local when it's drag and drop, but the buffer store the blob url, maybe I can get the blob url and resolve it ? @kwokcb 
https://github.com/AcademySoftwareFoundation/MaterialX/blob/e6b96502e965392c8851bab9de847d3135a48396/javascript/MaterialXView/source/helper.js#L117

hybridherbst:
When all the files are dropped togehter (or the folder containing them), then the references can indeed be resolved with a custom loader. (this is, for example, how all glTF viewers work when dropping .gltf + .bin + some textures).

I'll see if I find time to fix this :)

kwokcb:
Hi @hybridherbst,

Thanks for you comments on this.

Actually @Avgilles was trying to do this for ASWF dev days but if Gilles is still interested maybe just some pointers would appreciate.

Also, the case I don't think is possible is drag-and-drop only the MaterialX file (not the folder or other with images files) and have the file references inside the MaterialX file parsed to load in associated images.

darchangell:
Hey, I'm a first time contributor and would like to attempt this one for ASWF Dev days. Could this be assigned to me?

jstone-lucasfilm:
Welcome, @darchangell, and I've assigned it to you!

One caveat about this task is that we're aware of a regression in drag-and-drop for the Web Viewer, and we have a separate Dev Days task to fix this regression:

[...]

PR #3082, opened 2026-09-17
Dev days fix for drag and dropping images with materials

darchangell:
There was an original error pulling in the material files themselves, fixed with adding "file:" in front of the path when loading files and "image:" when loading images.

The rest of the fix for dragging in images deals with making sure that the images are reachable from the material. We resolved that by flattening the paths to be just in the original source.

A next step for this could be to add some support for adding a material name as an upper level directory around all of the files, but because files and images could come from anywhere and dragged/dropped as separate entities it makes resolving this difficult.


jstone-lucasfilm:
Thanks for this proposal, @darchangell!  See the instructions at https://github.com/AcademySoftwareFoundation/MaterialX/pull/3082#issuecomment-5722463593 on how to resolve the CLA authorization warnings, so that we can begin reviewing this PR.

darchangell:
I've started the authorization process and am only waiting for my manager to sign off on it. Thank you for taking a look!

darchangell:
Phew, just needed to update the author on my commits from yesterday. Should be good to go now. Thank you for your patience!!

jstone-lucasfilm:
Thanks for taking this on, @darchangell, and welcome to ASWF Dev Days!

Now that the CLA is resolved, I've had a chance to read through the changes, and I'd like to pick up where the conversation in #3080 left off, since @kwokcb raised the question there of why drag and drop broke in the first place.

This regression traces back to our upgrade of three.js from r152 to r183 in #2858.  In r152, `FileLoader` and `ImageLoader` looked up `Cache.get(url)` directly, while in r183 they key the cache as `file:${url}` and `image:${url}` respectively.  The drop code still seeds the cache with bare paths, so nothing it adds is ever found, and the loaders fall through to a network request that ends in the 404 from #3008.  Your `file:` and `image:` prefixes are exactly the right fix for this, and that part of the PR I'd be happy to merge as it stands.

On the flattening approach, I approved this idea in #1640, but having now read the implementation alongside the existing drop code, I'd like to refine that guidance.  The drop code was already designed to preserve folder structure, so a dropped folder produces entries such as `/Foo/Foo.mtlx` and `/Foo/textures/a.png`, the search path becomes `/Foo`, and `toThreeUniform` in `helper.js` builds `/Foo/textures/a.png`, which matches the cached key once the prefixes are corrected.  Flattening gives up that working case, and same-named textures in different sub-folders now collide in both folder and zip drops, even though the original zip handling kept the archive path on purpose.  I also hadn't anticipated the `clearFilePrefix` pass in `loadMaterials`, which modifies the loaded document to work around a lookup issue that really belongs in the texture loader, and which only covers half of the loose-file case, since a `file` input whose value carries its own directory, such as `textures/a.png` with no fileprefix, still resolves to a path that the flattened cache doesn't contain.

I'd suggest a fallback in the texture lookup instead, along the lines of what @kwokcb proposed in #1640.  Keep the full paths in the cache as before, and have `dropHandling.js` additionally record a map from basename to full path for each dropped image, warning and keeping the first entry on a collision as we discussed.  Then `toThreeUniform` can consult that map when the full texture path misses the cache and the value isn't an `http`, `blob`, or `data` URL:

```js
const droppedPath = findDroppedImage(value);
if (droppedPath && !THREE.Cache.get(`image:${texturePath}`))
    texturePath = droppedPath;
```

That would keep folder and zip drops intact, leave the document untouched, and handle both the `fileprefix` and the sub-directory forms of the reference.  If that turns out to be more work than you have bandwidth for in the context of Dev Days, it would also be completely fine to land the cache key fix first for #3008, and to follow up on #1640 in a second PR.

A few smaller notes:

- `debugFileHandling` is now set to `true`, which enables verbose console logging for all users of the viewer, and should be reverted before we merge.
- The comment above the image cache line says there's no prefix on the key, but the line adds an `image:` prefix.  It's worth describing the actual mechanism here, which is that the direct `THREE.Cache.get(texturePath)` lookup in `helper.js` misses, and the hit occurs inside the `ImageLoader` that `TextureLoader` calls, so the dropped image is correctly wrapped in a `THREE.Texture`.
- Since this PR now supersedes #3080, adding `Fixes #1640` and `Fixes #3008` to the description will close both issues when it lands.

Issue #2344, opened 2025-04-11
Add support for Zeltner sheen in OSL 1.14+


ld-kerley:
The current OSL implementation of `sheen_bsdf` ignores the `mode` input parameter, this parameter is used in the MaterialX nodegraph implementation of OpenPBR. This means that currently OpenPBR is using the `conty-kulla` sheen model instead of `zeltner`.

jstone-lucasfilm:
Good catch, @ld-kerley, and I believe we can safely use the pattern seen here, so that the additional `mode` parameter is only used in OSL 1.14 and beyond:

https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genosl/mx_chiang_hair_bsdf.osl

jstone-lucasfilm:
For reference, here's the OSL PR in which Zeltner sheen was introduced, and I believe this corresponds to OSL v1.14:

https://github.com/AcademySoftwareFoundation/OpenShadingLanguage/pull/1819

chrissanrow:
Hi, I'm Chris Sanrow. I'll be participating in Dev Days 2026 and I'm wondering if I could take on this issue. Thanks!

jstone-lucasfilm:
Welcome, @chrissanrow, and I've assigned it to you!

jstone-lucasfilm:
This has been addressed in #3083, thanks @chrissanrow!

PR #3083, opened 2026-09-18
Add Zeltner Sheen support when OSL 1.14 is present

chrissanrow:
# Issue

#2344 

# Description

Followed https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/libraries/pbrlib/genosl/mx_chiang_hair_bsdf.osl pattern to conditionally pass the mode parameter to the OSL `sheen_bsdf` function, if the OSL version is above 1.14.

Also updated the `implementation` tag for `IM_sheen_bsdf_genosl` to use the new `mx_sheen_bsdf.osl` file and `mx_sheen_bsdf` function instead of the inline source.

# Testing

100% tests passed out of 97
Generated `.osl` using `generateshader.py` with ` resources/Materials/TestSuite/pbrlib/bsdf/sheen_bsdf.mtlx` and succesfully compiled using oslc (see the following generated .osl)
[test_sheen_bsdf.osl](https://github.com/user-attachments/files/32361549/test_sheen_bsdf.txt)


jstone-lucasfilm (approved):
Great work, thanks @chrissanrow, and this looks ready to merge.

Issue #3056, opened 2026-08-29
Uniform inputs connectability


uvr-jra:
The specification states that uniform inputs can only be connected to the output of a `<constant>` node (possibly through one or more no-op `<dot>` nodes) or to any other node whose output is explicitly declared to be "uniform".

However, connections to non-uniform outputs are allowed in the graph editor and seem to be perfectly valid (see the `frameoffset` uniform input of `tiledimage` in the screenshot below):

<img width="1867" height="1107" alt="Image" src="https://github.com/user-attachments/assets/213c2577-1e3c-4433-8469-b742a10c9dab" />

Am I missing something?

Thanks,

Joël

jstone-lucasfilm:
Good catch, @uvr-jra, and I'd propose that our Graph Editor ought to respect this attribute and disallow connections to uniform inputs of nodes.

catherineazelbee:
Hello, I'd like to claim this issue for Dev Days 2026!

jstone-lucasfilm:
Welcome, @catherineazelbee, and done!

PR #3084, opened 2026-09-18
GraphEditor: Prevent invalid connections to uniform inputs

catherineazelbee:
Fixes Issue #3056 

## Summary
The graph editor originally allowed connections to uniform inputs without validation. This PR adds a check in `Graph::checkCanAddLink()` that blocks any connection where the destination input is marked as uniform.

## Testing
- Verified that floor_integer -> frameoffset is correctly blocked
image 

- Verified that floor_integer -> `add.in1`, `constant_integer`, `dot_integer` remains unaffected
image

- All MaterialXTest unit tests pass

## Notes
Constant node outputs are not currently marked `uniform="true"` in
`stdlib_defs.mtlx`, so `constant -> uniform input` connections are
blocked with this change. Per this [ASWF Slack thread](https://academysoftwarefdn.slack.com/archives/C0230LWBE2X/p1789668840126639), this is acceptable behavior for now, as users can still set uniform values directly on the node.


jstone-lucasfilm:
Thanks for this proposal, @catherineazelbee, and welcome to ASWF Dev Days!  Adding the check to `Graph::checkCanAddLink` is the right place for this logic, and the new validation message follows the pattern of the existing type and kind checks.

After our Slack discussion, I went back and reread the related sections of the specification, and I need to revise the guidance I gave there.  The text on [input elements](https://github.com/AcademySoftwareFoundation/MaterialX/blob/main/documents/Specification/MaterialX.Specification.md#inputs) states the following: "Input elements may be defined to only accept uniform values, in which case the input may provide a `value` or a `nodename` connection to the output of a `<constant>` node (possibly through one or more no-op `<dot>` nodes) or any other node whose output is explicitly declared to be "uniform", but may not provide a `nodename` or `nodegraph` connection to any arbitrary node output or to any nodegraph output."  The description of the `uniform` attribute on nodedef inputs repeats this rule, so the specification treats `constant` as a special category rather than depending on a uniform marking on its output, which is why no `ND_constant_*` nodedef carries one.  Your second commit was actually closer to the intended behavior than my Slack reply suggested, and apologies for the detour!

There's one additional case that we didn't discuss in Slack, which is the handling of nodegraph interface inputs.  When a user connects one of the editor's Input nodes to a node inside a nodegraph, `addLink` records this as an `interfacename` connection, and the specification allows a uniform input to reference a uniform nodedef input in this way.  This is exactly how `NG_tiledimage` in `stdlib_ng.mtlx` wires its `file`, `filtertype`, and `frameoffset` inputs, so with the current change, a nodegraph of that form could no longer be authored in the Graph Editor.

Putting these together, I'd suggest a small helper that answers whether the upstream pin is a uniform source, covering the following cases:

- The upstream UiNode is an interface input, i.e. `uiUpNode->getInput()` is non-null.  Since the specification requires the referenced interface input to be uniform as well, it would be reasonable to mark that nodegraph input as `uniform="true"` when the connection is made.
- The upstream node's category is `constant`, as in your second commit.
- The upstream nodedef output has `getIsUniform()` set, also as in your second commit, which covers nodes such as `geompropvalueuniform`.
- The upstream node's category is `dot`, in which case the helper follows its `in` connection and repeats the test.

It's worth noting that the type check earlier in `checkCanAddLink` already protects string and filename inputs in practice, since the libraries contain very few outputs of those types.  The cases this PR really governs are integer, boolean, and color4 uniform inputs such as `frameoffset`, `index`, and `thin_walled`, which is all the more reason to keep constants and interface inputs available as sources for them.

The broader questions raised on Slack remain good candidates for future work, including a shared `isUniform` query in MaterialXCore that handles the implicit uniformity of string and filename types, and a review of uniformity markings across the standard library, and those would make great candidates for GitHub Issues of their own.

Thanks for all of your work on this, and looking forward to the next revision!

catherineazelbee:
@jstone-lucasfilm Got it! I added the helper ` isUniformSource()` to check for those cases. 

These are the new results from testing with the GraphEditor:

## Confirmed Working
- Originally cited issue #3056 prevented (`non_uniform` -> `uniform` prevented):
Screenshot 2026-09-20 175128

- `constant` -> `uniform_input` permitted
Screenshot 2026-09-20 175152

- Recognizes `dot` connection types by tracing back to the original input
Screenshot 2026-09-20 175216
Screenshot 2026-09-20 175257

- Recongizes uniform and non_uniform outputs with `isUniform()`:
Screenshot 2026-09-20 190451

## More Tests - Not 100% on these...
- Handles `nodegraph` interface inputs (not 100% this is what you meant for the test not mentioned on Slack, but let me know):
Screenshot 2026-09-20 185741
Screenshot 2026-09-20 185746

- Going `non_uniform` -> `constant` -> `uniform input` _is_ currently allowed. Not sure if this is what we want, but let me know!
Screenshot 2026-09-20 175549

The last case I'm stuck on is:
> The upstream UiNode is an interface input, i.e. `uiUpNode->getInput()` is non-null. Since the specification requires the referenced interface input to be uniform as well, it would be reasonable to mark that nodegraph input as `uniform="true"` when the connection is made.
The last case I'm working through is the interface input one:
`uiUpNode->getInput()` is non-null. My current implementation doesn't automatically mark it as `uniform="true"` when the connection is made. Should I add that to addLink? Not sure if you were saying this is something we'd want for this specific issue.

Let me know! Can run more tests in the graph editor or update the code as needed. Thank you!

jstone-lucasfilm:
Thanks for the quick turnaround, @catherineazelbee, and this revision looks very close to the mark.

Focusing on your two questions first:

1. Yes, I'd suggest marking the interface input as uniform in `addLink`, at the two places where the new link is recorded through `setConnectedInterfaceName`.  The specification states that a uniform input inside a nodegraph may only reference a uniform nodedef input, so without this step the editor would write documents that violate that rule, and the Graph Editor currently has no other way to set the attribute.  Since the Input node's element is a real nodegraph input, a call to `setIsUniform(true)` will serialize as expected.  Something along these lines, applied to both branches, might work well:

   ```
   if (uiUpNode->getInput())
   {
       nodeInput->setConnectedInterfaceName(uiUpNode->getName());
       mx::NodeDefPtr nodeDef = uiDownNode->getNode() ? uiDownNode->getNode()->getNodeDef() : nullptr;
       mx::InputPtr nodeDefInput = nodeDef ? nodeDef->getActiveInput(nodeInput->getName()) : nullptr;
       if (nodeDefInput && nodeDefInput->getIsUniform())
       {
           uiUpNode->getInput()->setIsUniform(true);
       }
   }
   ```

2. The `non_uniform -> constant -> uniform` case is acceptable as is.  The `constant` nodedef doesn't mark its `value` input as uniform, and the specification places the responsibility on the application authoring the graph to ensure that a uniform output is actually uniform, so I'd consider this out of scope for the current PR.  It would make a good follow-up issue alongside the shared `isUniform` query we discussed.

One issue that's important to address:

3. In `checkCanAddLink`, the downstream input is looked up with `nodeDef->getInput`, which doesn't see inputs inherited from a parent nodedef.  The default `ND_standard_surface_surfaceshader` inherits from its 1.0.0 version and redeclares only `base` and `base_color`, so `thin_walled` is found only on the parent, and the uniform check is silently skipped for that input.  Switching to `getActiveInput` should resolve this, and would match how `setUiNodeInfo` builds pins from `getActiveInputs`.  The same consideration applies to `getOutput` in the helper, where `getActiveOutput` is the equivalent call.

A few more minor notes:

- The `dot` case iterates over all input pins and follows the first connection it finds.  Since `ND_dot_*` also declares a uniform `note` input, looking up the `in` pin by name would be more robust.
- The check currently runs only when the downstream UiNode wraps a node.  Nodegraph inputs can carry `uniform="true"` as well, and `inputPin->getInput()` already holds that element for compound nodegraph nodes, so a fallback to its `getIsUniform` would cover that additional case.
- The Notes section of the PR description still states that constant connections are blocked, so it would be worth refreshing to describe the latest behavior of your change.

Let me know how this sounds to you, and thanks again for the great work on this!

catherineazelbee:
@jstone-lucasfilm Thank you for the helpful feedback! This is what I've been able to implement:

1. Added setIsUniform(true) calls in both `setConnectedInterfaceName` branches in `addLink` using your suggested code.
2. Added `if (inputPin->getName() == "in")` in the dot node case to avoid following the note input
3. Switched to `getActiveInput` and `getActiveOutput` for `checkCanAddLink` and `isUniformSource` respectively
4. Added the `else if (inputPin->getInput() && inputPin->getInput()->getIsUniform())` fallback for nodegraph inputs in checkCanAddLink

Though I am having trouble getting successful tests with updates 1 and 4. I've gone about creating a `nodegraph` with an `input_integer`, and setting the output of the input node to `frameoffset` in `tiledimage`, as an example. However, when I save this file out, I fail to see `uniform="true"` written into the nodegraph input, or anywhere for that matter. This also makes testing a case like 4 hard, since I believe that _would_ be based on a successful test case from 1? I could be mistaken here, and let me know if I am! I couldn't really think of a better way to go about the uniform assigning, but maybe the code is just missing something that's causing the uniform tag to never be written? Can also keep trying on my end.

For 2 and 3, I'm just unclear on how I could go about testing this in general, besides for 2 going back and rerunning my original dot chain tests, which still worked. For 3, in the `thin_walled` example, a connection like `ifgreater_boolean` is permitted, but I'm not sure if that's what we'd want to be testing here (not sure what counts as a uniform/non-uniform boolean output here). But if there is anything I could run in the graph editor, let me know!

jstone-lucasfilm:
Thanks for the quick turnaround, @catherineazelbee, and this revision addresses every point from the last round.

Let me start with the testing questions, as I think your implementation may already be working correctly!  Both of the results you describe are exactly what the previous commit would produce: `getInput` skips the inherited `thin_walled` input, and that commit had no `setIsUniform` call.  Since the current branch has both fixes in place, I'd suggest rebuilding the Graph Editor from scratch and rerunning your tests.

Your instinct on the boolean case is correct, and the same rule applies as for integers: `constant_boolean` is a valid source, while nodes such as `ifgreater_boolean` and `logical_not` are not, since their outputs aren't declared uniform.  For the nodegraph case, the attribute should appear on the nodegraph's interface input rather than on the node input that references it.

On the latest changes in the PR, I have one suggested refinement on structure:

The nodedef lookup now appears twice in `checkCanAddLink` and twice in `addLink`, and the two sites handle different downstream cases.  A small helper alongside `isUniformSource` would collapse all four:

```
bool isUniformInput(UiNodePtr node, const std::string& inputName)
{
    if (node->getNode())
    {
        mx::NodeDefPtr nodeDef = node->getNode()->getNodeDef();
        mx::InputPtr nodeDefInput = nodeDef ? nodeDef->getActiveInput(inputName) : nullptr;
        return nodeDefInput && nodeDefInput->getIsUniform();
    }
    if (node->getNodeGraph())
    {
        mx::InputPtr input = node->getNodeGraph()->getInput(inputName);
        return input && input->getIsUniform();
    }
    return false;
}
```

In `checkCanAddLink`, the two branches can then become a single condition of `isUniformInput(uiDownNode, inputPin->getName()) && !isUniformSource(uiUpNode, outputPin->getName())`, and both blocks in `addLink` become `if (isUniformInput(uiDownNode, nodeInput->getName()))`.  This also resolves an edge case in the current revision, where an Input node connected to a uniform input of a compound nodegraph passes the check but is never marked uniform, since that branch of `addLink` has no nodedef to consult.

A few minor notes:

- The `dot` trace recurses through pin connections without a visited set, and the editor doesn't currently guard against cycles, so a `dot` loop would recurse without bound.  A simple depth limit or `std::set` of visited node ids would be sufficient.
- The label reads as a blanket ban on uniform inputs, when constants and uniform outputs are allowed.  Something like "Uniform inputs accept only constant or uniform sources" would give users a clearer hint.
- Since `isUniformSource` doesn't use any state of the `Graph` class, it could live as a static function in the anonymous namespace at the top of the file, alongside the new helper above.

With those changes I believe this looks ready to merge.  Thanks for your hard work and persistence with this!

catherineazelbee:
@jstone-lucasfilm Thank you for the feedback! Here's what I was able to do with my latest commit:

- Added the `isUniformInput` helper
- Moved `isUniformInput` and `IsUniformSource` to `Graph.cpp` as static functions
- Changed label suggestion to `Uniform inputs accept only constant or uniform sources`
- Added a depth limit for the dot trace loop in `isUniformSource`

After rebuilding, all the original tests that were passing still pass, but I'm still having issues with the uniform inputs test case and `thin_walled`...
1. I'm going about testing the addition to `addLink` for setting uniform inputs when a relevant connection is made like this:
- I make a `nodegraph`, then set a single `input_integer` that is connected to `frameoffset`
Screenshot 2026-09-23 152358
Screenshot 2026-09-23 152353

- When I save out this file, the resulting `.mtlx` file does not tag the input as `uniform="true"`:

```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <nodegraph name="my_nodegraph">
    <input name="input_integer" type="integer" value="0"/>
    <tiledimage name="tiledimage_color3" type="color3">
      <input name="frameoffset" type="integer" interfacename="input_integer" />
    </tiledimage>
  </nodegraph>
</materialx>
```
2. Currently an `ifgreater_boolean` -> `thin_walled` connection is permitted...
Screenshot 2026-09-23 145541

For these two issues, maybe `isUniformInput` isn't recognizing these inputs correctly? Though I'm not sure how to confirm this, as the code makes sense to me and I can't really see another way to go about it. Will keep testing on my end, but let me know what you think!

catherineazelbee:
@jstone-lucasfilm Thank you for the feedback! Here's what I did in my latest commit(s). (Apologies for the messy commits, but hopefully everything is clear in Files Changed):

- Added the `isUniformInput` helper
- Moved `isUniformInput` and `IsUniformSource` to `Graph.cpp` as static functions
- Changed label suggestion to `Uniform inputs accept only constant or uniform sources`
- Added a depth limit for the dot trace loop in `isUniformSource`

After rebuilding, all the original tests that were passing still pass, but I'm still having issues with the uniform inputs test case and `thin_walled`...
1. I'm going about testing the addition to `addLink` for setting uniform inputs when a relevant connection is made like this:
- I make a `nodegraph`, then set a single `input_integer` that is connected to `frameoffset`
Screenshot 2026-09-23 152358
Screenshot 2026-09-23 152353

- When I save this file, the resulting `.mtlx` file does not tag the input as `uniform="true"`:

```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <nodegraph name="my_nodegraph">
    <input name="input_integer" type="integer" value="0"/>
    <tiledimage name="tiledimage_color3" type="color3">
      <input name="frameoffset" type="integer" interfacename="input_integer" />
    </tiledimage>
  </nodegraph>
</materialx>
```
2. Currently an `ifgreater_boolean` -> `thin_walled` connection is permitted...
Screenshot 2026-09-23 145541

For these two issues, maybe `isUniformInput` isn't recognizing these inputs correctly? Though I'm not sure how to confirm this, as the code makes sense to me and I can't really see another way to go about it. Will keep testing on my end, but let me know what you think!

jstone-lucasfilm:
Hey @catherineazelbee -- all of your latest code looks exactly right to me, and I think your work on this PR may already be done!

The results you're seeing probably come from a stale local build of the Graph Editor: the commit you tested called `isUniformSource` with two arguments after the signature gained a third, so the build failed and left the previous executable in place.  Your follow-up commit fixed this, and a clean rebuild should make both tests pass.

A quick way to confirm the binary is current is that the new "Uniform inputs accept only constant or uniform sources" label appears only in the new build.  Checking the saved file for the uniform attribute was exactly the right test, by the way, and it did its job in catching that the build wasn't current.

Let me know if that resolves it, and thanks again for sticking with this!

catherineazelbee:
@jstone-lucasfilm Ah, you're right, that's my bad! I was able to build my current version, and I'm getting better results!

- Now the uniform input tagging works! Rerunning my previous test from my last comment yields the `.mtlx` file, with `uniform="true" successfully being added!:
```xml
<?xml version="1.0"?>
<materialx version="1.39">
  <nodegraph name="my_nodegraph">
    <input name="input_integer" type="integer" value="0" uniform="true"/>
    <tiledimage name="tiledimage_color3" type="color3">
      <input name="frameoffset" type="integer" interfacename="input_integer" />
    </tiledimage>
  </nodegraph>
</materialx>
```
- However, the `ifgreater_boolean -> thin_walled` case is still being permitted:
image

Other `ifgreater` connections are still successfully blocked, so not sure why boolean specifically is being permitted:
image

This is the case when I try this on both `standard_surface_surfaceshader` and `standard_surface_surfaceshader_100` nodes (not sure if that makes any difference, though). `thin_walled` is definitely marked as `uniform="true"` as far as I can tell, so not sure why this specific case is still being permitted. Let me know what you think!

jstone-lucasfilm:
Thanks for the rebuild, @catherineazelbee, and it's great to see `uniform="true"` showing up on the nodegraph input as expected!

On the `ifgreater_boolean -> thin_walled` case, I believe your code is working correctly.  I think the result comes from the data libraries your Graph Editor is loading, not from the logic in this PR.  The blocked `ifgreater_integer -> frameoffset` case and the permitted `ifgreater_boolean -> thin_walled` case go through exactly the same path in `isUniformInput` and `isUniformSource`, so the only thing that differs between them is the library data.  The `uniform="true"` marking on `thin_walled` in Standard Surface is quite recent: it was added in #2991 in July and isn't in any MaterialX release yet.  Your branch was originally created from a version of `main` from before that change, so any copy of the libraries from before your recent merge would still treat `thin_walled` as a varying input.  That would also explain why `standard_surface_surfaceshader` and `standard_surface_surfaceshader_100` behave the same way, since the default version inherits `thin_walled` from the 1.0.0 nodedef.

A quick way to confirm is to try the same connection into `geometry_thin_walled` on `open_pbr_surface`, which has been marked uniform since February and is included in v1.39.5.  If that connection is blocked while `thin_walled` on Standard Surface is still allowed, the editor is loading an older copy of the libraries.  The Graph Editor finds its libraries by searching upward from the executable's folder for the first folder that contains `libraries/targets`.  Checking `libraries/bxdf/standard_surface.mtlx` in that location should tell you whether `thin_walled` carries the uniform marking.  If you're running the editor from an install folder, running the install step again will refresh its copy of the libraries.

Other than that, the only issue that I see is a few stray deviations from our formatting conventions (e.g. the missing space between `if` and an opening parenthesis), and you can address those automatically by running `git clang-format upstream/main` (or whichever remote points to the main MaterialX repository), which formats just the lines that you have changed relative to `main`.

Thanks again for sticking with this, and let me know how the library test goes!

PR #3085, opened 2026-09-18
Fix Document Validation for undefined types on input parameters (#2813)

### LINKED ISSUE #2813: Document Validation doesn't check for undefined types
In our test suite we were seeing that the [Document Validate Method](
https://github.com/AcademySoftwareFoundation/MaterialX/blob/9c478d94ad555bb12b28654c1dbfaa45221d3669/source/MaterialXCore/Document.cpp#L406) doesn't validate types on input parameters as being defined.

For example we have some versions of matrix adds that use 2x2 mats like so

```
  <nodedef name="ND_add_matrix22" node="add" nodegroup="math">
    <input name="in1" type="matrix22" value="1.0,0.0, 0.0,0.0" />
    <input name="in2" type="matrix22" value="0.0,0.0, 0.0,0.0" />
    <output name="out" type="matrix22" defaultinput="in1" />
  </nodedef>
```

This passes validation even though there is no typedef for `matrix22`.  We do have our own separate validation suite that catches this, but I think it would be great if we can have the validate method also validate types as well. It would also help catch typos and the like.

PawanTheGod:
Fixes #2813 

### Description
Currently, `Document::validate()` does not check if the types assigned to input parameters (or other typed elements) are actually defined within the document. As reported, using an undefined type like `matrix22` instead of `matrix33` quietly passes validation.

This PR addresses the issue by adding validation to the `TypedElement` base class.

### Changes Made
- Added a `validate()` override to `TypedElement` in `Element.h` and `Element.cpp`.
- The new validation logic checks if the element's type definition can be successfully resolved via `getTypeDef()`.
- If the type is not found (and is not empty or the special `"multioutput"` type), it cleanly leverages MaterialX's native `validateRequire` to append the error: `"Type '<type>' does not exist"`.
- Because `Input`, `Output`, `GeomPropDef`, and other typed elements all inherit from `TypedElement` (via `ValueElement` / `PortElement`), this single change robustly protects the entire graph against typos and undefined types.


jstone-lucasfilm:
Thanks for this proposal, @PawanTheGod!  We'll hold off on reviewing this for now, as it's still marked as a draft and doesn't yet successfully pass our build and test suite, but let us know if you need any guidance on this.

PR #3087, opened 2026-09-19
Fix double UI scaling on macOS Retina displays

### LINKED ISSUE #3077: MaterialX Graph Editor : Double scaling occurs of text (Minor)
## Issue

Text in property panel, main menu and title area for graph panel are twice too large.

### Workaround

Use the the uiscale command option to scale by 1/2 (0.5).

### Implementation Notes

On Mac, ImGui already tries to compensate for Retina, high DPI displays.
Thus this should not be done when initializing the app (as for Windows and X).

This is one possible way to approach fixing in `Main.cpp`. Needs to be vetted.
```cpp
     // Derive a single effective UI scale from device DPI and optional CLI multiplier.
+    float deviceScale = 1.0f;
+#if !defined(__APPLE__)
+    // On macOS (Retina) the framebuffer tracks the content scale, and ImGui already
+    // converts logical units to pixels using io.DisplayFramebufferScale, so scaling
+    // the UI by the content scale as well doubles the scale.  On Windows and X11
+    // the framebuffer is unscaled, so the content scale is the only way to get the DPI.
     float xscale = 1.0f, yscale = 1.0f;
     glfwGetWindowContentScale(window, &xscale, &yscale);
-    const float deviceScale = (xscale > yscale) ? xscale : yscale;
-    const float effectiveUiScale = deviceScale * ((uiScale > 0.0f) ? uiScale : 1.0f);
-    const float scaledFontSize = std::max(1.0f, fontSize * effectiveUiScale);
+    deviceScale = (xscale > yscale) ? xscale : yscale;
+#endif
+const float effectiveUiScale = deviceScale * ((uiScale > 0.0f) ? uiScale : 1.0f);
+const float scaledFontSize = std::max(1.0f, fontSize * effectiveUiScale);
 ```

La1itchauhan:
## Summary
Fixes #3077

On macOS Retina displays, text in the property panel, main menu, and graph panel title area appears twice as large as intended.

## Root Cause
ImGui already compensates for Retina high-DPI displays internally via `io.DisplayFramebufferScale`. The existing code in `Main.cpp` additionally reads `glfwGetWindowContentScale()` and multiplies the font size and style metrics by it, resulting in double scaling.

## Fix
Wrap the `glfwGetWindowContentScale()` query in a platform guard (`#if !defined(__APPLE__)`) so that it only runs on Windows and X11, where the framebuffer is unscaled and the content scale is the only way to account for DPI. On macOS, `deviceScale` defaults to `1.0f`, allowing ImGui to handle DPI scaling on its own.

## Changes
- **`source/MaterialXGraphEditor/Main.cpp`**: Added `#if !defined(__APPLE__)` guard around the content-scale query block.

## Testing
- **Windows / X11**: No behavioral change. `glfwGetWindowContentScale()` continues to provide the DPI multiplier as before.
- **macOS (Retina)**: `deviceScale` remains `1.0f`, preventing the double scaling. The `--uiScale` CLI option still works as an additional user-specified multiplier on all platforms.


jstone-lucasfilm:
Thanks for this proposal, @La1itchauhan!  Can you add a comment on the GitHub Issue that you're tackling, so that we can assign it to you?  Also, let us know if this contribution is part of ASWF Dev Days, so that we'll know whether to assign the Dev Days 2026 label as well.

La1itchauhan:
Thanks for the heads up! I've commented on the original issue (#3077) . I'd be happy for this to be counted as part of ASWF Dev Days 2026 please go ahead and add the label

jstone-lucasfilm (approved):
This looks correct to me, @La1itchauhan, and thanks for the contribution!

I'd like to give @kwokcb a chance to weigh in before merging the change, just to make sure I'm not missing anything important.

kwokcb:
@jstone-lucasfilm , I think the one concern you had was that additional scaling is occuring which would affect font size. It should be the case now on Mac that this makes it so there is no scaling (scale by 1) of font's by default. 

As such,  I think it's ready for check-in.

kwokcb (approved):

